PR #81, opened 2020-06-08
chore: move to Github Workflow from Gitlab CI

### LINKED ISSUE #82: Replace Gitlab CI for Github Workflow
Last functional piece of what is missing to finish migrating hathor-core to Github.

jansegre:
Opening as a PR on Github instead of a MR on Gitlab to both trigger a Github Workflow build, and mark the final phase of the migration.

Closes #82


.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: |
+        pip install pipenv
+        pipenv install --ignore-pipfile --deploy --system --dev
+    - name: Generate protobuf files
+      run: make protos
+    - name: Install extra dependencies
+      run: |
+        sudo apt-get install -y graphviz librocksdb-dev libsnappy-dev zlib1g-dev libbz2-dev liblz4-dev
+        pip install "cython<0.30"
+        pip install python-rocksdb==0.7.0
```
obiyankenobi, line 23: Are these 2 lines needed after the MR to update the Dockerfile (https://gitlab.com/HathorNetwork/hathor-python/-/merge_requests/411/diffs)?

jansegre: Yes, they are. The Github workflow doesn't use the Dockerfile. These 2 lines will be needed whenever we want to install rocksdb.

It may be able to fix this with Poetry. At the very least in Poetry we can have "extra" dependencies (just like on setup.py) and move python-rocksdb there, still I'm not sure if it iis  possible to force Cython to be installed before python-rocksdb with it.

obiyankenobi: Ok, got it.

PR #86, opened 2020-06-10
refactor(logging): structlog where missing, shorter/cleaner messages

jansegre:
TODO:

- [x] change base to `dev` after `refactor/speed-initialization` is merged

This is what it looks like with this PR:

[asciicast](https://asciinema.org/a/HlJwgYFzbStVRINIj0CWHFHtj?speed=30&theme=tango)

And this is what it looked like before (thumbnail isn't great):

[asciicast](https://asciinema.org/a/kURqhQBfLuGoZPf4PWJ76uN80?speed=30&theme=tango)

### DIFF
--- a/hathor/cli/util.py
+++ b/hathor/cli/util.py
@@ -1,4 +1,6 @@
 from argparse import ArgumentParser
+from collections import OrderedDict
+from datetime import datetime
 from enum import Flag
 from functools import reduce
 from operator import or_
@@ -11,14 +13,23 @@ def create_parser() -> ArgumentParser:
     return configargparse.ArgumentParser(auto_env_var_prefix='hathor_')
 
 
-def setup_logging(debug: bool = False, capture_stdout: bool = True, *, _test_logging: bool = False) -> None:
+def setup_logging(debug: bool = False, capture_stdout: bool = False, *, _test_logging: bool = False) -> None:
+    import logging
     import logging.config
 
     import colorama
     import structlog
     import twisted
+    from twisted.logger import LogLevel
 
-    logger = structlog.get_logger()
+    # Mappings to Python's logging module
+    twisted_to_logging_level = {
+        LogLevel.debug: logging.DEBUG,
+        LogLevel.info: logging.INFO,
+        LogLevel.warn: logging.WARNING,
+        LogLevel.error: logging.ERROR,
+        LogLevel.critical: logging.CRITICAL,
+    }
 
     # common timestamper for structlog loggers and foreign (stdlib and twisted) loggers
     timestamper = structlog.processors.TimeStamper(fmt="%Y-%m-%d %H:%M:%S")
@@ -110,6 +121,13 @@ def setup_logging(debug: bool = False, capture_stdout: bool = True, *, _test_log
 
             return sio.getvalue()
 
+        def _repr(self, val):
+            if isinstance(val, datetime):
+                return str(val)
+            else:
+                return super()._repr(val)
+
+    # See: https://docs.python.org/3/library/logging.config.html#configuration-dictionary-schema
     logging.config.dictConfig({
             'version': 1,
             'disable_existing_loggers': False,
@@ -139,11 +157,15 @@ def setup_logging(debug: bool = False, capture_stdout: bool = True, *, _test_log
                 # },
             },
             'loggers': {
+                # set twisted verbosity one level lower than hathor's
+                'twisted': {
+                    'handlers': ['default'],
+                    'level': 'INFO' if debug else 'WARN',
+                    'propagate': False,
+                },
                 '': {
-                    # 'handlers': ['default', 'file'],
                     'handlers': ['default'],
                     'level': 'DEBUG' if debug else 'INFO',
-                    'propagate': True,
                 },
             }
     })
@@ -164,17 +186,36 @@ def setup_logging(debug: bool = False, capture_stdout: bool = True, *, _test_log
             structlog.processors.StackInfoRenderer(),
             structlog.processors.format_exc_info,
             structlog.stdlib.ProcessorFormatter.wrap_for_formatter,
-            # structlog.twisted.JSONRenderer(),
         ],
-        context_class=dict,
+        context_class=OrderedDict,
         logger_factory=structlog.stdlib.LoggerFactory(),
         wrapper_class=structlog.stdlib.BoundLogger,
         cache_logger_on_first_use=True,
     )
 
-    twisted.python.log.startLoggingWithObserver(twisted.logger.STDLibLogObserver(), setStdout=capture_stdout)
+    twisted_logger = structlog.get_logger('twisted')
+
+    def twisted_structlog_observer(event):
+        try:
+            level = twisted_to_logging_level.get(event.get('log_level'), logging.INFO)
+            kwargs = {}
+            msg = ''
+            if not msg and 'log_format' in event:
+                msg = event['log_format'].format(**event)
+            if not msg and 'format' in event:
+                msg = event['format'] % event
+            failure = event.get('log_failure')
+            if failure is not None:
+                kwargs['exc_info'] = (failure.type, failure.value, failure.getTracebackObject())
+            twisted_logger.log(level, msg, **kwargs)
+        except Exception as e:
+            print('error when logging event', e)
+
+    # start logging to std logger so structlog can catch it
+    twisted.python.log.startLoggingWithObserver(twisted_structlog_observer, setStdout=capture_stdout)
 
     if _test_logging:
+        logger = structlog.get_logger()
         logger.debug('Test: debug.')
         logger.info('Test: info.')
         logger.warning('Test: warning.')
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -15,7 +15,6 @@ limitations under the License.
 """
 
 import datetime
-import json
 import random
 import sys
 import time
@@ -181,8 +180,7 @@ class HathorManager:
     def start(self) -> None:
         """ A factory must be started only once. And it is usually automatically started.
         """
-        self.log.info('Starting HathorManager...')
-        self.log.info('Network: {network}', network=self.network)
+        self.log.info('start manager', network=self.network)
         # If it's a full verification, we save on the storage that we are starting it
         # this is required because if we stop the initilization in the middle, the metadata
         # saved on the storage is not reliable anymore, only if we finish it
@@ -249,7 +247,7 @@ class HathorManager:
     def stop(self) -> Deferred:
         waits = []
 
-        self.log.info('Stopping HathorManager...')
+        self.log.info('stop manager')
         self.tx_storage.stop_running_manager()
         self.connections.stop()
         self.pubsub.publish(HathorEvents.MANAGER_ON_STOP)
@@ -296,7 +294,7 @@ class HathorManager:
 
         This method runs through all transactions, verifying them and updating our wallet.
         """
-        self.log.info('Initializing node...')
+        self.log.info('initialize')
         if self.wallet:
             self.wallet._manually_initialize()
         t0 = time.time()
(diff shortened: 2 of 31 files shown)


hathor/p2p/node_sync.py
```diff
@@ -355,23 +357,20 @@ def sync_from_timestamp(self, next_timestamp: int) -> Iterator[Deferred]:
 
         :param next_timestamp: Timestamp to start the sync
         """
-        self.log.debug('sync-{p} Sync starting at {next_timestamp}', p=self.short_peer_id,
-                       next_timestamp=next_timestamp)
+        self.log.debug('sync start', ts=next_timestamp)
```
obiyankenobi, line 360: Why did you remove the peer id from all the logs in this file?

hathor/stratum/stratum.py
```diff
@@ -523,8 +523,10 @@ def handle_submit(self, params: Dict, msgid: Optional[str]) -> None:
             tx.verify_pow()
         except PowError:
             # Transaction pow was not enough, but the share was succesfully submited
-            self.log.info('Insufficient weight, keep searching', tx=tx)
+            self.log.info('high hash, keep mining', tx=tx)
```
obiyankenobi, line 526: The original message seems better to me ("insufficient weight" x "high hash").

jansegre: "high hash" is the expression that Bitcoin uses, and one problem with "insufficient weight" is that it was often mistaken for an error, that happened a lot at the beginning when new people were mining with their own fullnodes

hathor/p2p/node_sync.py
jansegre: I moved it to `self.log = logger.new(peer_id_short=self.short_peer_id)`, with that every call to `self.log.*` will include `peer_id_short`.

obiyankenobi: Perfect, thanks!

hathor/stratum/stratum.py
obiyankenobi: Got it

hathor/manager.py
```diff
@@ -290,37 +288,29 @@ def _initialize_components(self) -> None:
...
 
         # self.stop_profiler(save_to='profiles/initializing.prof')
         self.state = self.NodeState.READY
-        self.log.info(
-            'Node successfully initialized (total={total}, avg={avg:.2f} tx/s in {dt} seconds).',
-            total=cnt,
-            avg=cnt / (t2 - t0),
-            dt=t2 - t0,
-        )
+        self.log.info('ready', tx_count=cnt, tx_rate=cnt / (t2 - t0), total_dt=t2 - t0)
```
pedroferreira1, line 313: I am still not sure why change 'Node successfully initialized' to 'ready'. Right now we have some things that could be ready, e.g. prometheus, metrics, stratum. I like being verbose in this case

tests/cli/test_multisig_address.py
```diff
@@ -22,16 +22,16 @@ def test_generate_address(self):
         # Last element is always empty string
         output.pop()
 
-        self.assertEqual(len(output), pubkey_count * 10 + 11)
+        self.assertEqual(len(output), pubkey_count * 11 + 11)
```
pedroferreira1, line 25: Why this has changed?

jansegre: This test is kind of weird because it depends on the specific number of lines from the output. Because there is a new log output, the number of lines changed.

hathor/manager.py
jansegre: It's scoped by `hathor.manager`, so in my head this was fine.

tests/cli/test_multisig_address.py
pedroferreira1: Ah, now I understand.

hathor/manager.py
pedroferreira1: Ah, right. I forgot about the scope, then it's fine.

As you said before, I think an image of the new logs would be good.

hathor/merged_mining/coordinator.py
```diff
@@ -719,12 +719,11 @@ async def submit_to_bitcoin(self, job: SingleMinerJob, work: SingleMinerWork) ->
...
                 f.close()
-                self.log.warn('Work rejected from Bitcoin Fullnode. Relevant job+work dumped.',
-                              code=res, filename=f.name)
+                self.log.warn('block rejected from Bitcoin', code=res, dump_file=f.name)
             else:
-                self.log.warn('Work rejected from Bitcoin Fullnode.', code=res)
+                self.log.warn('block rejected from Bitcoin', code=res)
         else:
-            self.log.info('Bitcoin RPC submit response', res=res)
+            self.log.info('other result from Bitcoin', res=res)
```
msbrogli, line 726: Maybe "unexpected result from Bitcoin"?

jansegre:
If I comment `self.manager.start()` on `hathor.cli.run_log` and try to GET `/v1a/status` this is what the logs looks like with this PR:

```
2020-07-31 16:39:02 [info     ] [hathor.cli.run_node] hathor-core v0.36.1            genesis=3fdff62 my_peer_id=7181d232a16ba95001de22d3151b027ed31677c8cd6dc7eb88d538b250186bbf version=0.36.1
2020-07-31 16:39:02 [info     ] [hathor.cli.run_node] with storage                   path=/Users/jan/Projects/hathor-core/datamain storage_class=TransactionRocksDBStorage
2020-07-31 16:39:02 [info     ] [hathor.cli.run_node] with cache                     capacity=100000 interval=5
2020-07-31 16:39:07 [critical ] [twisted]                                
Traceback (most recent call last):
  File "/Users/jan/.local/share/virtualenvs/hathor-core-ygIHHmG7/lib/python3.6/site-packages/twisted/web/server.py", line 199, in process
    self.render(resrc)
  File "/Users/jan/.local/share/virtualenvs/hathor-core-ygIHHmG7/lib/python3.6/site-packages/twisted/web/server.py", line 259, in render
    body = resrc.render(self)
  File "/Users/jan/.local/share/virtualenvs/hathor-core-ygIHHmG7/lib/python3.6/site-packages/twisted/web/resource.py", line 250, in render
    return m(request)
  File "/Users/jan/Projects/hathor-core/hathor/p2p/resources/status.py", line 74, in render_GET
    'state': self.manager.state.value,
AttributeError: 'NoneType' object has no attribute 'value'
```

And then before it looked like this:

```
2020-07-31 16:45:06 [info     ] [twisted] Log opened.
2020-07-31 16:45:07 [info     ] [twisted] Hathor v0.36.1 (genesis 3fdff62)
2020-07-31 16:45:07 [info     ] [twisted] My peer id is 7181d232a16ba95001de22d3151b027ed31677c8cd6dc7eb88d538b250186bbf
2020-07-31 16:45:07 [info     ] [twisted] Using Wallet at /Users/jan/Projects/hathor-core/datamain
2020-07-31 16:45:07 [info     ] [twisted] Using TransactionRocksDBStorage at /Users/jan/Projects/hathor-core/datamain
2020-07-31 16:45:07 [info     ] [twisted] Using TransactionCacheStorage, capacity 100000, interval 5s
2020-07-31 16:45:07 [info     ] [twisted] Site starting on 9080
2020-07-31 16:45:07 [info     ] [twisted] Starting factory <twisted.web.server.Site object at 0x10fba7080>
2020-07-31 16:45:10 [critical ] [twisted]                                exc_info=(<class 'AttributeError'>, AttributeError("'NoneType' object has no attribute 'value'",), <traceback object at 0x10fba14c8>)
2020-07-31 16:45:10 [info     ] [twisted] "127.0.0.1" - - [31/Jul/2020:16:45:10 +0000] "GET /v1a/status HTTP/1.1" 500 4771 "-" "curl/7.64.1"
```

hathor/merged_mining/coordinator.py
jansegre: I'm not sure if it's unexpected. If it's not `None` (success) or `'high-hash'` they could all be called `unexpected` because it would signal an implementation error, actually even `'high-hash'` would be unexpected because of the `if block_hash.to_u256() > block_target.to_u256()` condition above. Maybe I could refactor these since I understand these errors better now, and because we'll validate the template before sending the work to the miner (which is on another branch).

msbrogli: That would be great! You could also include a list of the known responses from the Bitcoin full node.

hathor/cli/util.py
```diff
@@ -139,11 +156,15 @@ def __call__(self, _, __, event_dict):
...
+                    'handlers': ['default'],
+                    'level': 'INFO' if debug else 'WARN',
+                    'propagate': False,
+                },
                 '': {
-                    # 'handlers': ['default', 'file'],
                     'handlers': ['default'],
                     'level': 'DEBUG' if debug else 'INFO',
-                    'propagate': True,
+                    # 'propagate': True,
```
msbrogli, line 167: Should we remove this line? Or add a comment explaining when it should be uncommented?

hathor/cli/util.py
```diff
@@ -164,17 +185,37 @@ def kwargs_formatter(_, __, event_dict):
...
+            failure = event.get('log_failure')
+            if failure is not None:
+                kwargs['exc_info'] = (failure.type, failure.value, failure.getTracebackObject())
+            twisted_logger.log(level, msg, **kwargs)
+        except Exception as e:
+            print('error when logging event', e)
+
+    # start logging to std logger so structlog can catch it
+    twisted.python.log.startLoggingWithObserver(twisted_structlog_observer, setStdout=capture_stdout)
+    # twisted.python.log.startLoggingWithObserver(twisted_structlog_observer, setStdout=False)
```
msbrogli, line 215: Remove this line?

hathor/cli/util.py
```diff
@@ -164,17 +185,37 @@ def kwargs_formatter(_, __, event_dict):
...
+            if not msg and 'log_format' in event:
+                msg = event['log_format'].format(**event)
+            if not msg and 'format' in event:
+                msg = event['format'] % event
+            failure = event.get('log_failure')
+            if failure is not None:
+                kwargs['exc_info'] = (failure.type, failure.value, failure.getTracebackObject())
+            twisted_logger.log(level, msg, **kwargs)
+        except Exception as e:
+            print('error when logging event', e)
```
msbrogli, line 211: Should we exit the process?

hathor/cli/util.py
jansegre: I'll link to python's logging docs, which explain what all theses settings mean and remove that line.

hathor/cli/util.py
jansegre: Yeah, I'll remove this one too.

hathor/cli/util.py
jansegre: I'd rather not. I have on idea what could cause this. And if it turn out to be "frequent" we're going to have a bad time managing our nodes that don't restart automatically and even if they did it would take a good while to load/sync and be up and running again.

Let's first see how it goes.

hathor/merged_mining/coordinator.py
jansegre: OK, I'll do this on `feat/merged-mining-better-api` where I'm already touching this part. Also, although it might be helpful to list known responses and symptoms, it's probably not that useful. It should always return `None`, otherwise there is an implementation error. At least that's what I get from all the errors I've encountered. But I'll add a brief list anyway so I can remember it better in case there is a regression.

hathor/cli/util.py
jansegre: Done.

hathor/cli/util.py
jansegre: Done. What do you think?

hathor/merged_mining/coordinator.py
jansegre: Here: https://github.com/HathorNetwork/hathor-core/pull/114/files#diff-eb36e58f1ad6ea122750ed239d9cd974L709-L729

PR #88, opened 2020-06-10
feat(storage): Save genesis and its metadata in storage

msbrogli:
Migrated from [MR!385](https://gitlab.com/HathorNetwork/hathor-python/-/merge_requests/385)

### DIFF
--- a/hathor/cli/main.py
+++ b/hathor/cli/main.py
@@ -127,8 +127,16 @@ class CliManager:
             capture_stdout = False
         else:
             capture_stdout = getattr(module, 'LOGGING_CAPTURE_STDOUT', False)
-        setup_logging(debug, capture_stdout)
 
+        # Enable pudb to stop on an unhandled exception.
+        pudb = '--pudb' in sys.argv
+        if pudb:
+            sys.argv.remove('--pudb')
+            import pudb
+            pudb.set_trace(paused=False)
+            capture_stdout = False
+
+        setup_logging(debug, capture_stdout)
         module.main()
 
 
--- a/hathor/transaction/genesis.py
+++ b/hathor/transaction/genesis.py
@@ -62,10 +62,19 @@ def _get_genesis_hash() -> bytes:
 GENESIS_HASH = _get_genesis_hash()
 
 
-def get_genesis_transactions(tx_storage: Optional[TransactionStorage]) -> List[BaseTransaction]:
+def _get_genesis_transactions_unsafe(tx_storage: Optional[TransactionStorage]) -> List[BaseTransaction]:
+    """You shouldn't get genesis directly. Please, get it from your storage instead."""
     genesis = []
     for tx in GENESIS:
         tx2 = tx.clone()
         tx2.storage = tx_storage
         genesis.append(tx2)
     return genesis
+
+
+def is_genesis(hash_bytes: bytes) -> bool:
+    """Check whether hash is from a genesis transaction."""
+    for tx in GENESIS:
+        if hash_bytes == tx.hash:
+            return True
+    return False
--- a/hathor/transaction/storage/remote_storage.py
+++ b/hathor/transaction/storage/remote_storage.py
@@ -1,5 +1,5 @@
 from math import inf
-from typing import TYPE_CHECKING, Any, Callable, Dict, Generator, Iterator, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Generator, Iterator, List, Optional, Set, Tuple, Union
 
 import grpc
 from grpc._server import _Context
@@ -101,23 +101,21 @@ class TransactionRemoteStorage(TransactionStorage):
     def __init__(self, with_index=None):
         super().__init__()
         self._channel = None
-        self._genesis_cache: Dict[bytes, BaseTransaction] = None
         self.with_index = with_index
         # Set initial value for _best_block_tips cache.
-        self._best_block_tips = [x.hash for x in self.get_all_genesis() if x.is_block]
-
-    def _create_genesis_cache(self) -> None:
-        from hathor.transaction.genesis import get_genesis_transactions
-        self._genesis_cache = {}
-        for genesis in get_genesis_transactions(self):
-            assert genesis.hash is not None
-            self._genesis_cache[genesis.hash] = genesis
+        self._best_block_tips = []
 
     def connect_to(self, port: int) -> None:
         if self._channel:
             self._channel.close()
         self._channel = grpc.insecure_channel('127.0.0.1:{}'.format(port))
         self._stub = protos.TransactionStorageStub(self._channel)
+
+        # Initialize genesis.
+        self._save_or_verify_genesis()
+
+        # Set initial value for _best_block_tips cache.
+        self._best_block_tips = [x.hash for x in self.get_all_genesis() if x.is_block]
 
     def _check_connection(self) -> None:
         """raise error if not connected"""
@@ -142,10 +140,6 @@ class TransactionRemoteStorage(TransactionStorage):
     @convert_grpc_exceptions
     def save_transaction(self, tx: 'BaseTransaction', *, only_metadata: bool = False) -> None:
         self._check_connection()
-
-        # genesis txs and metadata are kept in memory
-        if tx.is_genesis and not only_metadata:
-            return
 
         tx_proto = tx.to_proto()
         request = protos.SaveRequest(transaction=tx_proto, only_metadata=only_metadata)
@@ -507,7 +501,15 @@ class TransactionRemoteStorage(TransactionStorage):
             if not list_item.HasField('transaction'):
                 break
             tx_proto = list_item.transaction
-            yield tx_or_block_from_proto(tx_proto, storage=self)
+            tx = tx_or_block_from_proto(tx_proto, storage=self)
+            lock = self._get_lock(tx.hash)
+            with lock:
+                tx2 = self.get_transaction_from_weakref(tx.hash)
+                if tx2:
+                    tx = tx2
+                else:
+                    self._save_to_weakref(tx)
+            yield tx
 
     @convert_grpc_exceptions_generator
     def _topological_sort(self):
@@ -545,13 +547,11 @@ class TransactionRemoteStorage(TransactionStorage):
         return result.count
 
     def get_genesis(self, hash_bytes: bytes) -> Optional['BaseTransaction']:
-        if not self._genesis_cache:
-            self._create_genesis_cache()
+        assert self._genesis_cache is not None
         return self._genesis_cache.get(hash_bytes, None)
 
     def get_all_genesis(self) -> Set['BaseTransaction']:
-        if not self._genesis_cache:
-            self._create_genesis_cache()
+        assert self._genesis_cache is not None
         return set(self._genesis_cache.values())
 
     @convert_grpc_exceptions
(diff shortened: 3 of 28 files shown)


hathor/transaction/storage/remote_storage.py
```diff
@@ -534,13 +533,11 @@ def get_tx_count(self) -> int:
...
     def get_genesis(self, hash_bytes: bytes) -> Optional['BaseTransaction']:
-        if not self._genesis_cache:
-            self._create_genesis_cache()
+        assert self._genesis_cache is not None
         return self._genesis_cache.get(hash_bytes, None)
 
     def get_all_genesis(self) -> Set['BaseTransaction']:
-        if not self._genesis_cache:
-            self._create_genesis_cache()
+        assert self._genesis_cache is not None
```
obiyankenobi, line 540: Isn't `self._genesis_cache` removed?

pedroferreira1: The genesis are being added to `_genesis_cache ` on `_save_or_verify_genesis`.

tests/tx/test_genesis.py
```diff
@@ -20,39 +20,42 @@ def get_genesis_output():
...
 
 class GenesisTest(unittest.TestCase):
-    def test_pow(self):
-        genesis = get_genesis_transactions(None)
+    def setUp(self):
+        super().setUp()
+        self.storage = TransactionMemoryStorage()
 
+    def test_pow(self):
+        genesis = self.storage.get_all_genesis()
```
pedroferreira1, line 28: You could have this line on the `setUp` method, so you don't repeat it in all the tests.

tests/tx/test_tx.py
```diff
@@ -517,7 +517,6 @@ def test_tx_methods(self):
 
         # Validate is_genesis without storage
         tx_equal.storage = None
-        self.assertFalse(tx_equal.is_genesis)
```
pedroferreira1, line 520: Why did you remove it? Maybe we should just check if this raises `ValueError`

hathor/transaction/block.py
```diff
@@ -123,7 +123,7 @@ def create_from_struct(cls, struct_bytes: bytes,
 
     def calculate_height(self) -> int:
         """Return the height of the block, i.e., the number of blocks since genesis"""
-        if self.is_genesis:
+        if not self.parents:
```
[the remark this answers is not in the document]
msbrogli: The `is_genesis` is a method, so I had to use another verification to skip the genesis. This is one of the dangers of using `@property` decorator.

hathor/transaction/genesis.py
```diff
@@ -62,7 +62,9 @@ def _get_genesis_hash() -> bytes:
 GENESIS_HASH = _get_genesis_hash()
 
 
-def get_genesis_transactions(tx_storage: Optional[TransactionStorage]) -> List[BaseTransaction]:
+def get_genesis_transactions(tx_storage: Optional[TransactionStorage], safe: bool = False) -> List[BaseTransaction]:
+    if not safe:
+        raise RuntimeError('You shouldn\'t get genesis directly. Please, get it from your storage instead.')
```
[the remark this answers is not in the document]
msbrogli: The idea behind it is that you should not use it unless you know what you're doing. I don't think the `get_genesis_transactions_safe` is clear that you probably shouldn't use it.

tests/tx/test_tx.py
msbrogli: The `is_genesis` is a method marked as `@property`. It does not work anymore without a storage.

tests/tx/test_genesis.py
msbrogli: They are independent test but it happened that all of them need the genesis. We have plenty of things in the `setUp` of other tests because all tests used those things but there are some new tests that doesn't. But I can move there if we agree that's the best place to be.

hathor/cli/main.py
```diff
@@ -138,8 +138,8 @@ def main():
     except KeyboardInterrupt:
         logger.warn('Aborting and exiting...')
         sys.exit(1)
-    except Exception:
-        logger.exception('Uncaught exception:')
+    except Exception as e:
+        logger.exception('Uncaught exception:', e)
```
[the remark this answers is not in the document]
msbrogli: I agree. I just rolled back my change. Thanks!

hathor/transaction/genesis.py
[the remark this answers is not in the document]
obiyankenobi: In that sense, I don't think a parameter `safe=True` helps. Maybe starting the method with `_` to indicate it's "internal".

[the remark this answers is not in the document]
msbrogli: I agree that `safe` may not be the best parameter. Maybe `i_know_what_im_doing=True` parameter?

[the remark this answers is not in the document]
obiyankenobi: Still seems weird to me. We have several methods that should be used with extreme caution - e.g.`_add_to_cache()`. In none of them we have such a parameter.

tests/tx/test_genesis.py
pedroferreira1: I am okay keeping it the way it is. The idea was just to prevent code duplication.

tests/tx/test_tx.py
pedroferreira1: Yes, that's why I said we could check for a `ValueError` because that's what you raise when there is no storage.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -575,7 +600,7 @@ def _save_transaction(self, tx: BaseTransaction, *, only_metadata: bool = False)
     def _reset_cache(self) -> None:
         """Reset all caches. This function should not be called unless you know what you are doing."""
         if not self.with_index:
-            raise NotImplementedError
+            raise RuntimeError('Cannot reset cache because it has not been enabled.')
```
jansegre, line 603: Why not just `assert self.with_index, 'Cannot reset cache because it has not been enabled.'`? Is this a situation we can actually reach except from a miss-configured code setup?

hathor/transaction/genesis.py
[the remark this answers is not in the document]
jansegre: `safe=True` actually gives me the wrong idea, I understood you could only do that if `safe` is turned on, as in there is no way to turn a manual unsafe mode that would skip something. For me a parameter does not make sense. I'd rather it be called `_get_genesis_transactions`, `_get_genesis_transactions_unsafe` or `get_genesis_transactions_unsafe` as in "it is unsafe to call this method if you don't know what you're doing"

I'm pretty sure we have many other methods that can only be called in specific situations and they're all either implicitly checked (like having to call `.start()` on managers/listeners/...) or prefixed with an underscore (like `TransactionStorage._manually_initialize`). I think adding another way to mark a method that is unsafe to call unless you know the context will just make it harder to understand what's actually going on. An underscore would be enough for me, but if we want to be extra careful and `_unsafe` suffix makes a lot of sense to me.

tests/tx/test_genesis.py
jansegre: I mean, `setUp` is just to make it practical, we could have another `TestCase`, but it's just so few tests that it's maybe not worth the line count increase.

tests/tx/test_tx.py
jansegre: Why does it need a storage? We should be able to tell from looking at the settings only.

hathor/transaction/block.py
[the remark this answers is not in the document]
jansegre: Being a property shouldn't matter much, but from a comment below I get that it's because it uses the storage, but I think we could do away with depending on the storage, since using the storage is being more trouble than being useful.

hathor/transaction/genesis.py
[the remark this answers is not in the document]
msbrogli: I agree. Can you review it again, please? I kept the changes in a different commit.

hathor/transaction/block.py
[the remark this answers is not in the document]
msbrogli: I didn't understand your suggestion. I agree that it would be better to use `if self.is_genesis:`, and I'll check what part breaks when I do it.

tests/tx/test_tx.py
msbrogli: One of the ideas of this PR is to use the storage as the point of truth regarding genesis. Maybe we can create a `is_genesis(tx_id: bytes)` method in `hathor.transaction.genesis`, and use it directly.

hathor/transaction/storage/transaction_storage.py
msbrogli: Exactly. That's why I chose RuntimeError, but we can use an assert statement as well.

msbrogli: Changed! Can you check again, please?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -549,17 +568,23 @@ class BaseTransactionStorage(TransactionStorage):
     def __init__(self, with_index: bool = True, pubsub: Optional[Any] = None) -> None:
         super().__init__()
 
+        # Pubsub is used to publish tx voided and winner but it's optional
+        self.pubsub = pubsub
```
obiyankenobi, line 572: Already done on line 580.

hathor/transaction/block.py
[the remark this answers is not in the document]
jansegre: What I suggested was:

```python
if not self.parents:
    assert self.is_genesis
    ...
```

But now I understand that the change of the `if` condition from `self.is_genesis` to `not self.parents` is precisely because `self.is_genesis` needs a storage. If that's the case, fixing `self.is_genesis` seems like it would make it possible to use that for the `if` condition. Which makes a little more sense to me, but it's probably subjective anyway. Looking back it doesn't seem that important.

hathor/transaction/genesis.py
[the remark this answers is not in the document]
jansegre: Looks great now.

tests/tx/test_tx.py
jansegre: While I think it's generally useful to contain the concerns, even the storage will have to rely on the settings to know what are the genesis transactions. So I don't think it's useful to "prevent" these settings from being useful outside the storage.

I'd say that the storage can definitely be the gatekeeper of all metadata, but determining if a transaction is part of the genesis isn't inherently something that should depend on the metadata.

 And because there actually are use cases (like this assert or `calculate_height`'s `if`) that are prevented from using `is_genesis` because the lack of a storage would break it. I think it's more useful not to use the storage when it makes sense.

hathor/transaction/storage/transaction_storage.py
jansegre: Looks good.

hathor/transaction/block.py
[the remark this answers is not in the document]
msbrogli: I agree. I just reverted the change. Would you check again, please?

tests/tx/test_tx.py
msbrogli: I agree. I reverted this change. Would you review it again, please?

hathor/transaction/storage/transaction_storage.py
msbrogli: Fixed. Thanks!

tests/tx/test_tx.py
pedroferreira1: Good to me

jansegre: Great!

hathor/transaction/block.py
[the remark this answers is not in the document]
jansegre: Great!

PR #89, opened 2020-06-10
refactor: speed initialization

pedroferreira1:
1. Remove lock for `get_transaction` while initializing;
2. Add `--fast-init` parameter to skip tx validation and metadata/consensus update when initializing (we trust the data on the storage);
3. Remove `@deprecated` decorators;
4. Remove calculus of total tx/block weight on metrics (it was not used anywhere).


hathor/cli/run_node.py
```diff
@@ -49,6 +49,8 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--recursion-limit', type=int, help='Set python recursion limit')
         parser.add_argument('--allow-mining-without-peers', action='store_true', help='Allow mining without peers')
         parser.add_argument('--min-block-weight', type=int, help='Minimum weight for blocks')
+        parser.add_argument('--fast-init', action='store_true',
+                            help='Skip some transaction validations when initializing the node')
```
jansegre, line 53: I don't see the benefit of this not being the default. I'd rather have a `--full-verification` flag, with reversed meaning. Specially if the speedup from this is really significant. Particularly, if the concern is to protect ourselves from maliciously altered storage data, that wouldn't matter, because then they could as well alter the node code so it's skipped anyway. Setups where this would matter (external storage) could take this into account and add the `--full-verification` flag.

To add to this, defaulting to faster startup seem to be the norm (like Bitcoin only verifying the last 6 blocks by feault).

Since @msbrogli suggested this, what do you think?

hathor/manager.py
```diff
@@ -482,30 +494,43 @@ def on_new_tx(self, tx: BaseTransaction, *, conn: Optional[HathorProtocol] = Non
...
-        try:
-            tx.update_initial_metadata()
-            self.consensus_algorithm.update(tx)
-        except Exception:
-            pretty_json = json.dumps(tx.to_json(), indent=4)
-            self.log.error('An unexpected error occurred when processing {tx.hash_hex}\n'
-                           '{pretty_json}', tx=tx, pretty_json=pretty_json)
-            self.tx_storage.remove_transaction(tx)
-            raise
+            if self._fast_init:
```
obiyankenobi, line 515: I think we could merge the code a bit:
```
self.tx_storage._add_to_cache(tx)
if not self._fast_init:
    tx.reset_metadata()
else:
    # the code below
```

From what I've seen, there's no harm in adding to the index before resetting the metadata.

hathor/transaction/base_transaction.py
```diff
@@ -879,6 +881,7 @@ def clone(self) -> 'BaseTransaction':
         """
         new_tx = self.create_from_struct(self.get_struct())
         if hasattr(self, '_metadata'):
+            assert self._metadata is not None  # FIXME: is this actually true or do we have to check if not None
```
obiyankenobi, line 884: I think if we have `_metadata` set, it will never be `None`, but I don't see harm in having this assert.

jansegre: If this is the case we could annotate it with `_metadata: TransactionMetadata`, this would allow us to confidently remove this assertion, but I'd guess it would need some work to get around all the typing issues that would be revealed.

But I'm OK leaving this for a greater full mypy-strict refactor.

hathor/manager.py
pedroferreira1: Yeah, it's the same. Did the merge you suggested.

hathor/transaction/base_transaction.py
pedroferreira1: @obiyankenobi  The assert is required for mypy.

And I agree with @jansegre, we could do it in other ways, but this requires fixing the typing issues, which I agree that could be done in another refactor.

hathor/consensus.py
```diff
@@ -1,5 +1,5 @@
 from itertools import chain
-from typing import TYPE_CHECKING, Iterable, List, Optional, Set
+from typing import TYPE_CHECKING, Iterable, List, Optional, Set, cast
```
pedroferreira1, line 2: The problem I am facing now is about the metadata saved on storage.

When we load without any specific parameters, we do a fast init, where we trust that the data on the storage is correct, so we just load it. When we do a full verification, we reset the metadata of each transaction and calculate every step of the consensus for each transaction.

The problem is when I start loading my storage with a full verification and stop before it's fully initialized, so the txs that I had already loaded would have partial metadata informations because the txs not loaded could change its consensus. Then if I decide to do a fast load, I can't trust the metadata informations that are on the storage and were updated with partial information.

I am still thinking about the best solution but my first thought is to mark the metadata I reset with a full verification as 'unreliable'. It will be reliable only after a full initialization, then I can trust it. While doing a fast init, if the metadata is unreliable, I have to do the full check again.

Another option would be to have a 'temporary' storage while loading the transactions that is only updated after it's completely initialized. However this would require the double of storage to run an initialization.

Any thoughts on that? @jansegre @obiyankenobi @msbrogli 

(I created this comment here on this line to start a discussion thread, not only a single comment)

jansegre: This makes me dislike the "full verification" even more (besides being slow it can mess up your storage).

I don't think a temporary storage would be really good, it would either need not so trivial implementations for all backends or have a really expensive workaround (like when loading storage A, redirecting all writes to a memory storage and after it's loaded flushing all the writes back to storage A).

I think some sort of state, or lock would be simpler and would solve the problem. Even a lock file would be enough, this is probably going to have to be storage specific, but we lock the storage when starting a full verification, and unlock when its finished. If it's already locked when we start to load (full verification or not) we have to assume the storage is in a broken state and either a) refuse to continue, b) wipe the storage, or c) make a full verification; which are all bad and will need a long full sync to recover (unless we make periodic storage backups, which would be good for many reasons, but probably need backend specific support).

msbrogli: I agree that a temporary storage is not a good solution.

Following @jansegre's idea, we already have to improve our storage to support general items (key/value). In this sense, we can set the key `runningFullVerification` to true when it starts and to false when it ends. So, if it is interrupted, we can detect in the next run. In this case, a full verification will be required and the fast init cannot be used. What do you think?

We should also create a new design to discuss how we can run a full verification without changing the metadata. I think it shouldn't be complicated, but we don't have to do it here.

jansegre: I also think we don't have to do this on this PR. Also, a way to run non-destructive verification would enable verifying only the last few txs/blocks (like the last 100 blocks, or everything in the last 1h from the last timestamp in storage). Which is much more reasonable to me than a full verification.

pedroferreira1: I was thinking about a way to keep the unmodified storage data as clean, so we would need to calculate again only the ones modified by the full verification. But I agree that there are no good solutions for that.

With @jansegre suggestion to change the parameter to --full-verification, it's harder for someone to run this by mistake, so I guess it's fine to assume that if you stop in the middle you can't run a fast init again until you finish a full init.

jansegre: @pedroferreira1 I think it's fine either way, if you always run verifications you won't have problems or if you always run without verifications you won't have problems. You'll only have problems if you run verifications, stop in before it finishes, and run without verifications.

Maybe adding a comment on the CLI's help for `--full-verification` (like "This will break your storage if you stop midway") is enough for now.

I just don't want to delay this PR much longer.

pedroferreira1: The changes for this are here: https://github.com/HathorNetwork/hathor-core/pull/89/commits/fad36e6a6fd02e77a89ddc825fc2320acaf063fd

@msbrogli @jansegre @obiyankenobi

hathor/cli/run_node.py
msbrogli: @jansegre But it is not so easy to temper with blocks/transactions because you'd have to solve the PoW of the whole DAG after it. On the other side, it is very easy to temper with the metadata, and it would affect the consensus.

Anyway, I agree that I myself would run with `--fast-init` most of the times. So, I agree that we can have it by default. I'm only concerned people would never run a full verification, and it can be dangerous.

Maybe we should add a counter of "number of initializations without full verification" and warn users from time to time to run a full verification, just like linux partitions do.

hathor/consensus.py
msbrogli: If the full node is killed for any reasons, and it can't shutdown properly, we would need to run a full verification as well. So, we need to be careful about it.

hathor/transaction/storage/binary_storage.py
```diff
@@ -153,6 +163,25 @@ def get_tx(hash_bytes, path):
                     yield tx
 
     def get_count_tx_blocks(self):
-        files = os.listdir(self.path)
+        files = os.listdir(self.tx_path)
         assert len(files) % 2 == 0
         return len(files) // 2
+
+    def enable_full_verification(self) -> None:
```
msbrogli, line 170: I think we shouldn't do anything so specific. Why don't we just create a generic `Registry` that has a given number of parameters? So, we can store other things there in the future.

hathor/manager.py
```diff
@@ -191,6 +208,8 @@ def start(self) -> None:
         self.tx_storage.disable_lock()
         # Initialize manager's components.
         self._initialize_components()
+        if self._full_verification:
+            self.tx_storage.disable_full_verification()
```
msbrogli, line 212: Why do we need to disable it?

hathor/manager.py
```diff
@@ -182,6 +183,22 @@ def start(self) -> None:
...
+        # this is required because if we stop the initilization in the middle, the metadata
+        # saved on the storage is not reliable anymore, only if we finish it
+        if self._full_verification:
+            self.tx_storage.enable_full_verification()
+        else:
+            # If it's a fast initialization and the last time a full initialization stopped in the middle
+            # we can't allow the full node to continue, so we need to remove the storage and do a full sync
+            # or execute an initialization with full verification
+            if self.tx_storage.running_full_verification_active():
+                self.log.error(
```
msbrogli, line 196: We should also mark as `full_verification_required` if the full node is not properly shutdown.

obiyankenobi: How is that possible? If it's not properly shut down, we can't detect and take any action before it's shut down, right?

hathor/cli/run_node.py
jansegre: > @jansegre But it is not so easy to temper with blocks/transactions because you'd have to solve the PoW of the whole DAG after it. On the other side, it is very easy to temper with the metadata, and it would affect the consensus.

My point was, that even easier than tempering with the metadata, is tempering with the code. Assuming one has disk access.

> Maybe we should add a counter of "number of initializations without full verification" and warn users from time to time to run a full verification, just like linux partitions do.

That sounds good too. We should discuss this in a separate issue.

hathor/consensus.py
jansegre: > If the full node is killed for any reasons, and it can't shutdown properly, we would need to run a full verification as well. So, we need to be careful about it.

Then we should review our storages, it seems perfectly reasonable for our storages to be resilient against sudden crashes. RocksDB supports transactions, and we should take advantage of that so that we never leave the database in an inconsistent state.

hathor/cli/run_node.py
pedroferreira1: @msbrogli do you agree to create another issue to discuss this counter of "number of initializations without full verification"?

hathor/manager.py
pedroferreira1: Before calling `self._initialize_components()` I enter 'full verification mode' and after that I disable it. It's just to know if I've stopped a full initialization in the middle

hathor/transaction/storage/binary_storage.py
pedroferreira1: My idea was to create a way to add any key/value to our storages, that's why I created another folder 'attributes' for it. I just left everything in the same method because we are using it only here but maybe I should create another generic method, e.g.

```
def add_key_value(self, key, value):
    filepath = os.path.join(self.attributes_path, key)
    self.save_to_json(filepath, value)

def enable_full_verification(self):
    self.add_key_value(settings.RUNNING_FULL_VERIFICATION_ATTRIBUTE, True)

```

hathor/consensus.py
pedroferreira1: I think today we also have this problem with transactions (not only metatadas) and a full verification after a sudden crash wouldn't fix it. To prevent this from happen in both metadata and tx data, we would need to refactor our storage, which I think we should do, but I am just wondering if we should do here, given that this is a problem we always had.

jansegre: Definitely not on this PR, which I think has stretched too far from its original purpose already. I'm okay with some things being a little worse before we can make them a lot better, as long is it's nothing too serious.

hathor/transaction/storage/binary_storage.py
msbrogli: I really prefer this way. And I like creating a new subfolder (or just a new file for it) because it allows us to put all the code in `BaseTransactionStorage` without specifics for each storage type.

pedroferreira1: Did this refactor in this commit: https://github.com/HathorNetwork/hathor-core/pull/89/commits/44d1197720ee7eb84554207baa8f57cb398a32dc

Can you see if it's good?

hathor/cli/run_node.py
```diff
@@ -49,6 +49,10 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--recursion-limit', type=int, help='Set python recursion limit')
         parser.add_argument('--allow-mining-without-peers', action='store_true', help='Allow mining without peers')
         parser.add_argument('--min-block-weight', type=int, help='Minimum weight for blocks')
+        parser.add_argument('--x-fast-init-beta', action='store_true',
+                            help='Execute a fast initialization, which skips some transaction verifications. '
+                            'This is still a beta feature because there are some issues when using both '
+                            'initializations.')
```
obiyankenobi, line 55: What do you mean by "both initializations"?

hathor/conf/settings.py
```diff
@@ -247,3 +247,6 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
 
     # Amount in which tx min weight reaches the middle point between the minimum and maximum weight
     MIN_TX_WEIGHT_K: int = 100
+
+    # Key storage attribute to save if the full node is running a full verification
+    RUNNING_FULL_VERIFICATION_ATTRIBUTE: str = 'running_full_verification'
```
obiyankenobi, line 252: Doesn't seem like a setting to me. Why not place it in `transaction_storage.py`?

pedroferreira1: To be honest when I first it was on `transaction_storage.py` but then I see `STORAGE_SUBFOLDERS` on settings and this not only storage specific but it's also `compact_storage.py` specific and it was on settings. Then I decided to move.

I can move it back to `transaction_storage.py`, not a problem. It was my first thought also

hathor/cli/run_node.py
pedroferreira1: Fast initialization and full verification. But actually the problem is with sudden crashes, I will improve this help text.

pedroferreira1: Updated the help text, can you see if it's good?

hathor/manager.py
pedroferreira1: I've changed to `start_full_verification` and `finish_full_verification`. I think the names enable/disable were not the best options, as Jan said

hathor/cli/run_node.py
pedroferreira1: Created the issue as we've discussed. https://github.com/HathorNetwork/hathor-core/issues/122

hathor/consensus.py
pedroferreira1: Issue created. https://github.com/HathorNetwork/hathor-core/issues/123

hathor/conf/settings.py
obiyankenobi: I think it should be in `transaction_storage.py` and we should move `STORAGE_SUBFOLDERS` also in the future.

hathor/cli/run_node.py
```diff
@@ -49,6 +49,10 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--recursion-limit', type=int, help='Set python recursion limit')
         parser.add_argument('--allow-mining-without-peers', action='store_true', help='Allow mining without peers')
         parser.add_argument('--min-block-weight', type=int, help='Minimum weight for blocks')
+        parser.add_argument('--x-fast-init-beta', action='store_true',
+                            help='Execute a fast initialization, which skips some transaction verifications. '
+                            'This is still a beta feature because there might be some issues if there is a sudden'
+                            'crash on the full node.')
```
obiyankenobi, line 55: This is still a beta feature as it may cause issues when restarting the full node after a crash .

hathor/cli/run_node.py
obiyankenobi: Proposed a new text.

hathor/manager.py
msbrogli: We can have a flag `not-properly-shutdown` set to true. During the shutdown, after the finish our tasks, we mark it as false.

msbrogli: Another very common way to do it is to create an empty file `running` that is deleted during the shutdown.

hathor/conf/settings.py
msbrogli: I agree with @obiyankenobi .

hathor/transaction/storage/transaction_storage.py
```diff
@@ -541,6 +558,41 @@ def get_all_sorted_txs(self, timestamp: int, count: int, offset: int) -> Transac
         """
         raise NotImplementedError
 
+    def add_value(self, key: str, value: str) -> None:
```
msbrogli, line 561: It is missing a unittest for `add_value` and `remove_value`.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -541,6 +558,41 @@ def get_all_sorted_txs(self, timestamp: int, count: int, offset: int) -> Transac
...
+        """ Save value on storage
+            Need to be a string to support all storages, including rocksdb, that needs bytes
+        """
+        # XXX Some storages (e.g. memory storage) don't need this attribute
```
msbrogli, line 565: I think we should raise a `NotImplementedError` here, and have an empty method in the storages that don't need this attribute. By the way, why don't we need it in the memory storage?

hathor/transaction/storage/transaction_storage.py
msbrogli, line 561: Why must the value be an `str`? Can't we save using json or pickle?

pedroferreira1: The idea was to simplify the first implementation. I think we should support `Any` type but rocksdb for instance requires only bytes, so depending on the type it's not just a direct conversion. 

That's why I started with `str` only because it covers what we need for now and many cases, then we can expand later.

pedroferreira1: Actually there is a test (`test_full_verification_attribute` on `tests/tx/test_tx_storage.py`) that tests the methods using `add_value` and `remove_value`.

Do you think it would be better to have another one that directly calls `add_value` and `remove_value`?

hathor/conf/settings.py
pedroferreira1: Done

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Added to memory and remote storage. It was my mistake because the previous method were not needed there. Can you see if it's good?

hathor/transaction/storage/remote_storage.py
```diff
@@ -8,6 +8,8 @@
 from twisted.logger import Logger
 
 from hathor import protos
+print('#### ')
+print(protos)
```
jansegre: I think these are leftovers from testing.

pedroferreira1: Yeah, this was remove in the last commit, thanks!

hathor/manager.py
pedroferreira1: @msbrogli would something like this work for you? https://github.com/HathorNetwork/hathor-core/pull/89/commits/bdb68cd711b783cfbafb08f4d20848c2e66b9dcf

msbrogli: It seems good! In the explanation, I'd add that the metadata is the only part that may be wrong (not the blocks / transactions themselves).

hathor/transaction/storage/transaction_storage.py
msbrogli: Ok. We can improve it later since it is not a requirement right now.

hathor/transaction/storage/transaction_storage.py
msbrogli: It is. Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: I think it is ok if you've tested trying to delete a non-existing key and also getting a non-existing key.

hathor/manager.py
msbrogli: I got it. Ok! Can you drop a comment explaining it, please?

hathor/transaction/storage/binary_storage.py
msbrogli: It is. Thanks!

hathor/manager.py
pedroferreira1: Done

hathor/manager.py
pedroferreira1: Done

hathor/transaction/storage/transaction_storage.py
pedroferreira1: I've added another test to test the add/remove/get methods. Can you see if it's good?
https://github.com/HathorNetwork/hathor-core/pull/89/commits/afafc19673aa288fd4ed24ce0c114e12efa9975a

msbrogli: Great! Thanks!

PR #92, opened 2020-06-12
fix(manager): Use cache on generate_mining_block

msbrogli:


hathor/manager.py
```diff
@@ -344,11 +344,13 @@ def generate_mining_block(self, timestamp: Optional[float] = None,
...
 
-        if not timestamp:
+        original_timestamp = timestamp
+
+        if timestamp is None:
             timestamp = max(self.tx_storage.latest_timestamp, self.reactor.seconds())
 
         if parent_block_hash is None:
-            tip_blocks = self.tx_storage.get_best_block_tips(timestamp)
+            tip_blocks = self.tx_storage.get_best_block_tips(original_timestamp)
```
pedroferreira1, line 353: I think it's important to add a comment to explain why we need to use `original_timestamp` (to keep the None and get the cache), so we don't get confused in the future

msbrogli: Fixed. Thanks! Can you review the comment please?

hathor/manager.py
```diff
@@ -344,11 +344,15 @@ def generate_mining_block(self, timestamp: Optional[float] = None,
         """
         from hathor.transaction.scripts import create_output_script
 
-        if not timestamp:
+        # We keep the original timestamp to use it in the `get_best_block_tips()` method,
+        # because this method has cache when `timestamp` is None.
+        original_timestamp = timestamp
```
obiyankenobi, line 349: Isn't it easier to do this after we get the tip_blocks? This way there's no need to save this and the code is more readable.

hathor/manager.py
```diff
@@ -344,11 +344,15 @@ def generate_mining_block(self, timestamp: Optional[float] = None,
...
+
+        if timestamp is None:
             timestamp = max(self.tx_storage.latest_timestamp, self.reactor.seconds())
```
obiyankenobi, line 352: Isn't it easier to do this after we get the tip_blocks? This way there's no need to save `original_timestamp ` and the code is more readable.

msbrogli: Great one! Fixed. Thanks!

PR #93, opened 2020-06-13
refactor: add all tips cache and merkle tree cache

pedroferreira1:


hathor/transaction/storage/transaction_storage.py
```diff
@@ -333,6 +344,36 @@ def get_height_best_block(self) -> int:
 
         return highest_height
 
+    def get_merkle_tree(self, timestamp: int) -> Tuple[bytes, List[bytes]]:
```
msbrogli, line 347: Even though I got the point, I'm not sure we should move both methods to the storage because they are exclusively used by the sync algorithm. Can't we do the cache inside the sync code?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -584,7 +625,15 @@ def get_all_tips(self, timestamp: Optional[float] = None) -> Set[Interval]:
         assert self.all_index is not None
         if timestamp is None:
             timestamp = self.latest_timestamp
+
+        if self._all_tips_cache is not None and timestamp >= self._all_tips_cache.timestamp:
```
msbrogli, line 629: It feels wrong to me. I think we should compare to `self.latest_timestamp`.

obiyankenobi: Every time a new tx arrives, the cache is invalidated. So the cache will always be up to date with `self.latest_timestamp`, ie `self._all_tips_cache.timestamp == self.latest_timestamp`.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -45,6 +53,9 @@ def __init__(self):
         # Cache for the best block tips
         self._best_block_tips = []
 
+        # Cache for the latest timestamp of all tips with merkle tree precalculated to be used on the sync algorithm
```
obiyankenobi, line 56: I think a better explanation would be good. As far as I understand, this cache is always kept up to date with `self._latest_timestamp` and is invalidated all time a new txs arrives.

hathor/transaction/storage/transaction_storage.py
obiyankenobi: We probably have to cache on p2p manager, as we have one node_sync instance per peer.

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Yes, that's the idea. `self.latest_timestamp` is supposed to be equal to `self._all_tips_cache.timestamp` if there is a `self._all_tips_cache`

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Yes, my first implementation was on the p2p manager. However it was not easy to invalidate the cache and I had to use the pubsub to get the events and invalidate the cache.

We already know that we have a bug where we emit repeated pubsub events, so I was invalidating the cache in some moments I shouldn't, what could make the node waste CPU.

That's why I moved to the storage to use the same method to invalidate the cache as the one from the tips.

If you have any thoughts about how to improve this, I can refactor the code.

obiyankenobi: Why was it needed to use pubsub? `NodeSync` has access to `HathorManager`, which in turn can access `ConnectionsManager`. But that seems like hack. 

A better approach could be adding `ConnectionsManager` to `NodeSync`. When we create `NodeSync`, we have access to the protocol, which stores this object. We could add it also there.
```
    def __init__(self, protocol: 'HathorProtocol', reactor: Clock = None) -> None:
        ...
        self.protocol = protocol
        self.manager = protocol.node
        # something like this
        self.connections = protocol.connections
```

pedroferreira1: Why do I need `ConnectionsManager`? I need every moment a tx is added to the cache.

obiyankenobi: `ConnectionsManager` is the p2p manager. I thought this would be the best place to have the cache. It can be accessed by the `NodeSync` code (if you make the changes above). It can also be accessed on the storage through `HathorManager`.

It may be easier to have the cache in `HathorManager`, but since it's related to sync, having it inside `hathor/p2p/` seems more logical.

pedroferreira1: Ah, okay, now I understand. So add the cache to the `ConnectionsManager` and make the storage call a `reset_cache` method on it every time a tx is added to the cache?

To be honest I prefer to keep all caches on the storage but I can do this refactor if you feel is better.

@msbrogli @obiyankenobi

obiyankenobi: Yeah, that's the idea.

Regarding the location, I was looking at all the caches in our storage and it looks a bit confusing. We'll probably need to refactor it at some point. So for this index, I'd vote for adding it under the sync code, but I'm also fine if we add it to the storage with all the other indexes.

hathor/transaction/storage/transaction_storage.py
msbrogli: In this case, we should add an assert to double check it.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -584,7 +625,15 @@ def get_all_tips(self, timestamp: Optional[float] = None) -> Set[Interval]:
         assert self.all_index is not None
         if timestamp is None:
             timestamp = self.latest_timestamp
+
+        if self._all_tips_cache is not None and timestamp >= self._all_tips_cache.timestamp:
+            return self._all_tips_cache.tips
+
         tips = self.all_index.tips_index[timestamp]
+        if timestamp >= self.latest_timestamp:
```
msbrogli, line 633: This cache invalidation mechanism seems to fail if a new transaction arrives with the same timestamp as the previous one (or with an older timestamp). In this case, I think we will wrongly return an outdated tips list.

obiyankenobi: Not really. In `_add_to_cache`, the tips cache is always invalidated, regardless of the received transaction's timestamp. That's where the cache is invalidated.

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Added the assert

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Added an explanation about the cache being invalidated on the add cache and about the latest_timestamp.

Can you see if it's good?

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Yes, exactly. If a new transaction arrives in an previous (or older) timestamp, the cache will be invalidated when adding it to the cache and the tips will be recalculated when this method is executed again with a `timestamp >= latest_timestamp`

hathor/transaction/storage/transaction_storage.py
msbrogli: I finally understood the difficulty to put the cache out of the storage. I agree that it can be in the storage for now. Even though I agree with @obiyankenobi that we'll need to refactor it later, I don't think we should put it in the `ConnectionsManager`.

hathor/transaction/storage/transaction_storage.py
msbrogli: I finally got it. I haven't seen the change in the `_add_to_cache` method. You're right! Thanks!

hathor/transaction/storage/transaction_storage.py
pedroferreira1: Ok, so will keep on the storage for now.

PR #94, opened 2020-06-14
fix(mining_api): improve can_start_mining condition

jansegre:
Currently an outdated peer will make `can_start_mining` return `True` before it makes sense to allow mining. This PR improves this condition to filter-out peers too into the future or past.


hathor/p2p/manager.py
```diff
@@ -107,6 +110,31 @@ def has_synced_peer(self) -> bool:
                 return True
         return False
 
+    def is_practically_synced(self) -> bool:
+        """ Returns whether we can consider for practical resons that we are synced to the network.
+
+        This method name is intentionally a bit vague to give room for some imprecisions and subjective protection
+        agains malicious or rougue nodes. This method is primarily intended for use by APIs that need to wait for the
```
obiyankenobi, line 117: **against** malicious **rogue**

hathor/p2p/manager.py
```diff
@@ -107,6 +110,31 @@ def has_synced_peer(self) -> bool:
...
+
+        For example: `get_block_template` API can work at anytime, but before the node has synced, it will lead to
+        blocks way into the past with too low difficulty (because of decay), which is bad, but not catastrophic.
+        """
+        # How it works: filter out the peers that seem to have a reasonable timestamp (not way into the future, and not
+        # way into the past). Test if we're synced with any of them.
+        timestamp = int(self.reactor.seconds())
+        # XXX: reusing settings parameter, should we have another param? does it really matter?
+        sync_threshold = settings.P2P_SYNC_THRESHOLD
+        reasonable_timestamp_range = range(timestamp - sync_threshold, timestamp + sync_threshold)
```
obiyankenobi, line 128: This might give false negatives, I think. We update a peer's `synced_timestamp` with the latest block/tx we get from it. `settings.P2P_SYNC_THRESHOLD` is currently 1 minute, but we might have a period when it takes more than one minute to find a new block. In that case, even if a peer is synced, this function will return false.

jansegre: Yeah, I've tested and it actually happens quite often. I've made a specific param for this and set it to 600s (10m), which should be plenty and safe.

hathor/p2p/manager.py
jansegre: Thanks! Fixed.

hathor/p2p/manager.py
obiyankenobi: We can have it as `n * AVG_TIME_BETWEEN_BLOCKS` and we can calculate the probability that we have n periods without any blocks.

jansegre: Another way to do this would be taking the highest timestamp that is not much into too in the future from now and using that as a reference. That way it doesn't matter if there are no blocks for 1h (in case of an extreme hashrate drop).

obiyankenobi: Not sure I understand well your suggestion.

hathor/p2p/manager.py
```diff
@@ -107,6 +110,30 @@ def has_synced_peer(self) -> bool:
...
+        """
+        # How it works: filter out the peers that seem to have a reasonable timestamp (not way into the future, and not
+        # way into the past). Test if we're synced with any of them.
+        timestamp = int(self.reactor.seconds())
+        sync_threshold = settings.PRACTICALLY_SYNC_THRESHOLD
+        reasonable_timestamp_range = range(timestamp - sync_threshold, timestamp + sync_threshold)
+        for conn in self.get_ready_connections():
+            assert conn.state is not None
+            assert isinstance(conn.state, ReadyState)
+            if conn.state.get_sync_plugin().synced_timestamp not in reasonable_timestamp_range:
```
msbrogli, line 131: I think we should use an interval comparison instead of a list. We can do it in O(1) using two comparisons instead of in O(n) using a list.

hathor/p2p/manager.py
msbrogli: Even using `n * AVG_TIME_BETWEEN_BLOCKS`, it may happen to find a long period without any blocks because the hash rate has plummet. Maybe we should use `MAX_DISTANCE_BETWEEN_BLOCKS`. What do you think?

hathor/p2p/manager.py
jansegre: I'm not using a list, `range` is its own type that is efficient for this.

hathor/p2p/manager.py
jansegre: `MAX_DISTANCE_BETWEEN_BLOCKS` won't matter, there is an `n` that makes `n * AVG_TIME_BETWEEN_BLOCKS` value's higher than that, it could be better if the meaning of the constant made more sense to use here, but I think it doesn't

My proposal use the highest `synced_timestamp` (excluding nodes that are more than some `FUTURE_THRESHOLD` more than `datetime.now`), and use that as a reference. **Instead** of using `datetime.now` as a base for these cut-offs. Now `P2P_SYNC_THRESHOLD` could be reused, because it won't matter how long the network stays without blocks, because we're relative to whichever peer has highest timestamp. It's still possible for a peer to fake it's timestamp when the network does not have blocks, but we could eventually remove that peer from the `ready_connections` when it fails to provide a block with said timestamp, so these can still be independent.

What do you think? Does it make sense now? @msbrogli @obiyankenobi

hathor/p2p/manager.py
msbrogli: I didn't know the `range` type. I just read Python's documentation. Thanks!

hathor/p2p/manager.py
jansegre: I had a wrong understanding of what `synced_timestamp` was. I've made an update to use `peer_timestamp`. Also done this to `is_synced`. These changes make much more sense to me know, but I still have to try them out.

hathor/p2p/node_sync.py
```diff
@@ -266,7 +266,7 @@ def stop(self) -> None:
     def is_synced(self) -> bool:
         """ Return True if we are synced.
         """
-        return self.manager.tx_storage.latest_timestamp - self.synced_timestamp <= self.sync_threshold
+        return self.manager.tx_storage.latest_timestamp - self.peer_timestamp <= self.sync_threshold
```
obiyankenobi, line 270: We should add a comment on the declaration of `self.peer_timestamp` explaining what that is. Also, maybe move its declaration right before `self.synced_timestamp` so it's easy to see that both metrics are available and the difference between them.

hathor/p2p/manager.py
obiyankenobi: It makes more sense now.

I'm wondering how easy it is to make tests to check this heuristics. It'd be good to have them, as it's easy for us to come up with some different scenarios and the expected outcome. But I'm not sure about creating "fake" connections.

hathor/p2p/node_sync.py
jansegre: Done. What do you think?

hathor/p2p/node_sync.py
```diff
@@ -264,7 +265,7 @@ def stop(self) -> None:
...
```
msbrogli, line 268: Why are we changing from `synced_timestamp` to `peer_timestamp`? Even if we have a transaction/block near the peer_timestamp, we may be way behind syncing with the peer.

hathor/p2p/manager.py
msbrogli: I agree that it would be good to test the scenarios. We can just create the peers and fill the variables to generate the scenarios.

hathor/p2p/node_sync.py
obiyankenobi: > Even if we have a transaction/block near the peer_timestamp, we may be way behind syncing with the peer.

I'm not sure we can have a tx way ahead the other blocks and transactions.

As I understand, `peer_timestamp` is the peer's latest block/tx timestamp. So if we are around that timestamp, we are synced with it.

But it actually should be the absolute of `self.manager.tx_storage.latest_timestamp - self.peer_timestamp`, as this will be a negative number if we're behind the peer.

obiyankenobi: > `is_synced` isn't supposed to be symmetrical (that is, indicate whether the other peer is in sync with us) otherwise it will get out of sync every time we get too ahead of the rest of the network (which when you're mining is not unlikely)

But we should invert it, then. Let's analyze `self.manager.tx_storage.latest_timestamp - self.peer_timestamp`. Suppose I just started my node and I'm syncing with a peer that's all synced up. `self.peer_timestamp` will be the latest timestamp, close to `datetime.now()`, while `self.manager.tx_storage.latest_timestamp` (the latest transaction I have) is smaller than that, so the difference will be negative. If it's negative, then it will be smaller than `self.sync_threshold` and the function will return `True`.

jansegre: Wait, that totally makes sense, so now I think it should not be working, it should return `True` from the start. Is `peer_timestamp` what I think it is then? I'll have to debug this better, because this shouldn't be working but it is.

obiyankenobi: Yeah, the definitions are a bit confusing here. We should clear understand it better.

As far as I've read the code, `peer_timestamp` should be what we've discussed: the peer's latest timestamp, as reported by it.

hathor/p2p/manager.py
```diff
@@ -96,13 +99,36 @@ def stop(self) -> None:
...
+        for conn in connections:
+            assert conn.state is not None
+            assert isinstance(conn.state, ReadyState)
+            best_timestamp = max(best_timestamp, conn.state.get_sync_plugin().peer_timestamp)
+        # check if any peer within a reasonable synced timestamp is synced with us
         for conn in connections:
             assert conn.state is not None
             assert isinstance(conn.state, ReadyState)
+            peer = conn.state.get_sync_plugin()
+            reasonable_timestamp = range(best_timestamp - peer.sync_threshold, best_timestamp + peer.sync_threshold)
```
obiyankenobi, line 129: Now that I think of it, no peer should be higher than `best_timestamp`, since it's already the highest one. Maybe we should use `reasonable_timestamp = range(best_timestamp - peer.sync_threshold, best_timestamp)`

jansegre: Makes sense. I'll do this.

jansegre:
I'd like to propose the following change: initially return `False` and once the "start condition" is reached always return `True` (even if the condition is lost).

Why? Because the alternative means that connected peers can influence your ability to mine, a rouge peer could potentially make you think you're out of sync and you'll just stop mining sequentially, but miners won't stop, so they'll just keep mining in parallel and make a bunch of obviously orphan blocks. If that's the case it's better to just use `--allow-mining-without-peers` and the "start mining" condition would effectively not be used by the ones that it was made for.

What would be the overall effect of this proposal? Once you can start mining you'll start and won't ever go back to "not mining" unless the node is restarted. If the node goes out of sync for network reasons it will start producing an orphan chain (assuming those network reasons don't disconnect it from the miners) (and instead of producing a bunch of orphan blocks of the same height). When it is back online, it might be that the chain it produced outweighs the best public chain (which would also mean that this node has >51% of the network's hashrate, which is already a bad scenario) and that would imply in a re-org, otherwise it will just be an orphan chain, the same way it would produce orphan blocks.

Is there a superior alternative to all of this? I think there is, it would probably be better if the condition was merely informative, and not "blocking". Meaning that any API that uses `can_start_mining` for gating a result, could instead include a "is_synced" field, and the users of that API decide what to do with that info. This would mean that a mining proxy that connects to multiple nodes could exclude results from nodes out of sync unless all nodes are out of sync, and this apparently would lead to the best of both worlds. But because this affects more APIs, it should move slower and with feedback from the users of those APIs. And thus I don't believe we have to take this "is_synced field" route right now.

What do you think? @msbrogli @obiyankenobi @pedroferreira1

obiyankenobi:
> a rouge peer could potentially make you think you're out of sync and you'll just stop mining sequentially

I'm not sure that would happen. As far as I understand, you just need one peer in sync so that it returns `True`. Unless he's connected only to rogue peers, in which case there's nothing we can do.

jansegre:
> > a rouge peer could potentially make you think you're out of sync and you'll just stop mining sequentially
> 
> I'm not sure that would happen. As far as I understand, you just need one peer in sync so that it returns `True`. Unless he's connected only to rogue peers, in which case there's nothing we can do.

A rouge peer can tell others it has blocks in the future that they don't have. Wouldn't everyone now think they're out of sync?

msbrogli:
> A rouge peer can tell others it has blocks in the future that they don't have. Wouldn't everyone now think they're out of sync?

In the future? It could have new blocks up to `X` minutes in the future only.

jansegre:
> In the future? It could have new blocks up to `X` minutes in the future only.

Right now that `X` is smaller than our sync threshold, so the issue currently exists (on `dev` and on this PR). But that can be accounted for. Should we account it on the "start mining condition"? Probably not, I think we should at the very least temporarily ban those peers. It doesn't looks like we do this today.

OK, but that's just one of the motivations. Regardless of that there is still the side effect of blindly mining on a peer out of sync vs blindly mining on the same height: the first produces a single chain at a predictable rate (after a while of not finding blocks decay will adjust difficulty ...) and the second produces a lot (by very much) of chains of height 1 (after a while it's only a few chains, but if it takes too long decay will make it so there are way too many chains, and on each decay the rate of orphan chains being produced increases a lot). Today, with or without this proposal (and regardless of a sync attack being prevented), we are on the second situation and I'd much rather be on the first. Using a state with a single transition (False -> True) for the mining condition seems like the simplest way to achieve this.

What do you think?

obiyankenobi:
> but miners won't stop, so they'll just keep mining in parallel and make a bunch of obviously orphan blocks

> and the second produces a lot (by very much) of chains of height 1 

I still didn't understand why this happens. But I do agree that, given it's true, your suggestion seems much better.

jansegre:
The new main (and only) purpose of the _start mining condition_ would be to prevent mining on the past, which would produce a large amount of new orphan small chains as the node syncs, because the difficulty is generally much smaller on the past, specially on the first few months of the chain.

obiyankenobi:
Sounds good. We just need to make sure the start criteria is on point. If we just require that it's in sync with one peer, for instance, there's an attack where an attacking peer might always say that it's synced right after the genesis, so all nodes would return True very early.

jansegre:
> Sounds good. We just need to make sure the start criteria is on point. If we just require that it's in sync with one peer, for instance, there's an attack where an attacking peer might always say that it's synced right after the genesis, so all nodes would return True very early.

Yes. I'm not proposing to change the criteria. Which as it stands on this PR the case you mentioned wouldn't happen, because we choose from the connected peers the ones that are most on the future (but not after a future limit). Where this might break is if after peer discovery there are only outdated or rogue nodes.

jansegre:
I'll link the issues and PRs that supersede this one when I open them. But for now it's sufficient to say these changes have some fundamental problems that will require more design and discussion to solve properly. This was made clear after a few hours-long discussion with @msbrogli

PR #97, opened 2020-06-21
fix(api): use HTTP error codes on errors

jansegre:


jansegre:
This fixes a problem that can be seen on the merged mining server logs that says key "parents" does not exist. Which happens because the error returned by the fullnode has a status code 200 and is mistaken with a successful response.

PR #99, opened 2020-06-25
feat: websocket subscribe channel wallet service

pedroferreira1:


hathor/websocket/factory.py
```diff
@@ -54,6 +54,14 @@
     HathorEvents.WALLET_ELEMENT_VOIDED.value
 ]
 
+# Possible channels that can be subscribed on the websocket
+# Each channel has a list of events that emits.
+CHANNELS = {
+    'wallet-service': [
```
obiyankenobi, line 60: I'm not sure about using channels. The full node shouldn't have to know this information. It shouldn't need to be aware of the wallet-service, for instance. In the future, if we decide we need an extra message on this service, we'll have to update both the service and the full node.

Ideally, all websocket clients should subscribe to the messages they require. It's their responsibility.

pedroferreira1: The idea for the channels is just to make it easier for people to integrate with the full node ws. If someone wants to make another explorer, doesn't need to know all messages the explorer uses. It can just subscribe to the explorer and check all the messages (maybe this is not useful, since the dev needs to handle the messages in the frontend).

But I agree that we should have the possibility to subscribe to messages. I am just wondering if it's useful to have the channels also, so some common services can integrate easier.

I will add the message subscribe and wait for other inputs to discuss if it's useful to keep the channel subscribe also.

obiyankenobi: > If someone wants to make another explorer, doesn't need to know all messages the explorer uses.

This doesn't make sense to me. If I'm developing a new explorer, I need to know which messages it receives. Also, I might want to have different messages. For instance, I might want to create an explorer that doesn't have the statistics information. If I subscribe to the explorer channel, I'll end up getting them.

> I am just wondering if it's useful to have the channels also, so some common services can integrate easier.

We shouldn't do that, as the full node would need to have knowledge of the other services. This goes agains of what we've been doing, moving logic out of the full node (tx-mining, wallet-service) and leaving the full node only for "essential" tasks.

pedroferreira1: I've updated the code and it's now subscribing to messages. Instead of the channels.

Can you see if it's good?

msbrogli: Even though I understand @pedroferreira1's idea, I agree with @obiyankenobi. Anyway, we should have a list of all possible events to subscribe in the documentation of hathor-core.

hathor/websocket/factory.py
```diff
@@ -314,14 +332,42 @@ def _remove_connection_from_address_dict(self, connection: HathorAdminWebsocketP
         if len(self.address_connections[address]) == 0:
             del self.address_connections[address]
 
+    def _handle_subscribe(self, connection: HathorAdminWebsocketProtocol, message: Dict[Any, Any]) -> None:
+        """ Handler for subscription to a ws message."""
+        ws_message: Optional[str] = message.get('message')
```
msbrogli, line 337: Isn't it easier to receive a list of strings?

hathor/websocket/factory.py
```diff
@@ -120,7 +146,7 @@ def _schedule_and_send_metric(self):
             'type': 'dashboard:metrics',
             'time': reactor.seconds(),
         }
-        self.broadcast_message(data)
+        self.send_or_enqueue(data)
```
msbrogli, line 149: Is it compatible with the current clients of the websocket? I think we can keep compatibility adding a new message `clear-subscriptions` that will be called by new clients. In the future, after every client has been updated, we can start with no subscriptions from the beginning.

hathor/websocket/factory.py
```diff
@@ -226,6 +239,9 @@ def send_or_enqueue(self, data):
         else:
             self.send_message(data)
 
+        # Send message to each connection that subscribed to it
+        self.execute_send(data, self.subscribed_connections[data['type']])
```
obiyankenobi, line 243: There's an important difference here because when we subscribe to a specific message, there's no rate limit. I don't remember exactly why we implemented these limits, but should it also be considered here?

I'm more inclined to let the client deal with it. If they execute an expensive operation when receiving a ws message, they should do some sorte of rate limit on their side. E.g. an explorer should do batch updates on the screen in case there are too many txs arriving so the browser doesn't become unresponsive.

hathor/websocket/factory.py
```diff
@@ -314,14 +332,42 @@ def _remove_connection_from_address_dict(self, connection: HathorAdminWebsocketP
...
+            # With connections subscribing to an specific channel, we need to do this check
+            # In the future all connections should subscribe to channels
+            self.connections.remove(connection)
+
         for address in connection.subscribed_to:
             self._remove_connection_from_address_dict(connection, address)
 
+        # Removing from subsbribed channels set
+        # Should we have an easier way to check in each channels the connections is subscribed to?
+        # For now we have only one channel, so this shouldn't be a problem
```
obiyankenobi, line 366: Comment is out of date (channels).

Given that we don't have many message types and use a set for storing the connections, shouldn't be a very expensive operation.

pedroferreira1: Updated the comments there and added your comment also, I agree it's not expensive to remove from all sets.

hathor/websocket/factory.py
pedroferreira1: We implemented the limits because the explorer and wallet UI was freezing when receiving many messages on the ws.

One of the reasons to do this refactor on the websocket was exactly because we don't want any throttle when receiving messages to update the wallet service. I think we should keep simple for now, i.e. all subscribed messages have no rate limit and we implement it in the future.

We already have the throttle implemented on the ws, so I believe we should allow the client to 'configure' the message subscription, like choosing if wants any throttle and each limit parameter. But we should discuss this in the future when decide to implement it.

hathor/websocket/factory.py
pedroferreira1: Yes, it's fully compatible. The method `send_or_enqueue` will call `broadcast_message` for this message type, however it will also handle sending this message to the subscribed connections.

hathor/websocket/factory.py
pedroferreira1: Yes, updated it! Can you see now?

hathor/websocket/factory.py
pedroferreira1: Yes, I agree. Where do you think we should have it? In the README?

hathor/websocket/factory.py
obiyankenobi: Agreed.

hathor/websocket/factory.py
obiyankenobi: This behavior is already implemented as the websocket is removed from the broadcast group if it subscribes to any message type.

hathor/websocket/factory.py
```diff
@@ -271,6 +287,8 @@ def handle_message(self, connection: HathorAdminWebsocketProtocol, data: bytes)
             self._handle_subscribe_address(connection, message)
         elif message['type'] == 'unsubscribe_address':
             self._handle_unsubscribe_address(connection, message)
+        elif message['type'] == 'subscribe':
```
obiyankenobi, line 290: Should we also have an unsubscribe message? Doesn't seem we'll use it now, but might be useful in the future.

hathor/websocket/factory.py
pedroferreira1: Yes, I forgot to comment about the `clear-subscriptions`. Every time you subscribe to messages I remove the connection to the broadcast set.

hathor/websocket/factory.py
pedroferreira1: Done

PR #110, opened 2020-06-30
feat(cli): run_miner: Add --address parameter

msbrogli:


jansegre:
This seems harmless, but why is it even needed? I'd assume the next `run_miner` change would be removing it.

msbrogli:
> This seems harmless, but why is it even needed? I'd assume the next `run_miner` change would be removing it.

I still use it when running a local full node not connected to any public network. It is better than using `cpuminer` because I can use `--sleep` to keep CPU usage low and still have a flow of new blocks.

PR #116, opened 2020-07-09
feat(merged_mining): detect NiceHash and initial diff from user-agent

jansegre:
Also, thanks to Discord user EuroLine#5553 for the tip on the NiceHash user agent.

### DIFF
--- a/hathor/merged_mining/coordinator.py
+++ b/hathor/merged_mining/coordinator.py
@@ -265,6 +265,7 @@ class MergedMiningStratumProtocol(asyncio.Protocol):
         self.miner_id: Optional[str] = None
         self.miner_address: Optional[bytes] = None
         self.min_difficulty = min_difficulty if min_difficulty is not None else self.MIN_DIFFICULTY
+        self.initial_difficulty = PDiff(self.INITIAL_DIFFICULTY)
         self._current_difficulty = PDiff(1)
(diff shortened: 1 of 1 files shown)


hathor/merged_mining/coordinator.py
```diff
@@ -530,21 +534,37 @@ def handle_configure(self, params: List, msgid: Optional[str]) -> None:
...
+        if '/' in self.user_agent:
+            try:
+                # example: bmminer/2.0.0/Antminer S9j/14500
+                # the last part will often contain an indication of the hashrate (in GH/s)
+                hashrate_ghs = int(self.user_agent.split('/')[-1])
+                self.log.debug('detected hashrate', hashrate_ghs=hashrate_ghs)
+            except ValueError:
+                pass
+            else:
+                self.initial_difficulty = Weight(log2(hashrate_ghs * self.TARGET_JOB_TIME) + log2(1e9)).to_pdiff()
```
obiyankenobi, line 565: Do we need the `else`, can't this just run inside the `try` block?

jansegre: I'd rather not accidentally catch any exceptions that may come from this, because any exceptions here would actually be an error.

obiyankenobi: Got it.

PR #119, opened 2020-07-21
fix(mergedmining): bad-txnmrklroot and other errors on bitcoin.submitblock

jansegre:
This PR preserves the transaction data given by Bitcoin's
getblocktemplate as much as possible. The only transaction we build from
scracth now is the coinbase.

Transactions are temporarily removed (along with the fees) and other
errors on bitcoin.submitblock were also fixed, like bad-cb-height and unexpected-witness

Additionally a proposed merged block is now validated against
bitcoin.getblocktemplate which only affects logging, but should give a
preview of whether bitcoin.submitblock will fail.

### DIFF
--- a/hathor/merged_mining/bitcoin.py
+++ b/hathor/merged_mining/bitcoin.py
@@ -1,5 +1,19 @@
 import struct
-from typing import List, NamedTuple, Sequence, SupportsBytes, Tuple
+from typing import Dict, List, NamedTuple, Sequence, SupportsBytes, Tuple, Union
+
+
+@SupportsBytes.register
+class BitcoinRawTransaction(NamedTuple):
+    hash: bytes
+    txid: bytes
+    data: bytes
+
+    @classmethod
+    def from_dict(cls, data: Dict) -> 'BitcoinRawTransaction':
+        return cls(bytes.fromhex(data['hash']), bytes.fromhex(data['txid']), bytes.fromhex(data['data']))
+
+    def __bytes__(self) -> bytes:
+        return self.data
 
 
 @SupportsBytes.register
@@ -38,7 +52,7 @@ class BitcoinBlockHeader(NamedTuple):
         ])
 
     @classmethod
-    def from_dict(cls, params: dict) -> 'BitcoinBlockHeader':
+    def from_dict(cls, params: Dict) -> 'BitcoinBlockHeader':
         r""" Convert from dict of the properties returned from Bitcoin RPC.
 
         Examples:
@@ -101,11 +115,11 @@ def _merkle_concat(left: bytes, right: bytes) -> bytes:
     return bytes(reversed(left)) + bytes(reversed(right))
 
 
-def build_merkle_path_for_coinbase(merkle_leaves: List[bytes], _partial_path: List[bytes] = []) -> List[bytes]:
-    """ Return the merkle path (unidirectional since it's a list) to the coinbase (first tx) from hash leaves.
+def build_merkle_path_for_coinbase(merkle_leaves: List[bytes]) -> List[bytes]:
+    """ Return the merkle path (unidirectional since it's a list) to the coinbase (not included) from hash leaves.
 
-    >>> [i.hex() for i in build_merkle_path_for_coinbase([bytes.fromhex(tx) for tx in [
-    ...     '32a31fb3f8596e5de0a40a53748839d15e0a1a1d264da5b7dacec9209a59fd2a',
+    >>> tx_list = [bytes.fromhex(tx) for tx in [
+    ...     #'32a31fb3f8596e5de0a40a53748839d15e0a1a1d264da5b7dacec9209a59fd2a',
     ...     '45c5dcbe62075d366b87fa375fb919c7a8ede24eba0a3a094df491aef55184ca',
     ...     '170938fa8cd0d26e796d0b407eaa2d40db8a8c0cb660f68bc0df2cc65ffc3990',
     ...     'd0d6241f43a27980da30ce66250bea94d852b77f5413a8e031a5bb545e4be80e',
@@ -122,12 +136,19 @@ def build_merkle_path_for_coinbase(merkle_leaves: List[bytes], _partial_path: Li
     ...     'b244488bd9bd57fb33ea1c3872d510b64abdce2245e17a6607653e0b17628c7a',
     ...     'efb48aaeb133cca84af2d2fb9c13399db1f9f1fdb85b4210a936c545cd3124e7',
     ...     '232aaff768a2653db16fbe84504ff7d59396eb08663d2f5dda1d1587ce297df8'
-    ... ]])]
+    ... ]]
+    >>> [i.hex() for i in build_merkle_path_for_coinbase(tx_list)]
     ['45c5dcbe62075d366b87fa375fb919c7a8ede24eba0a3a094df491aef55184ca', \
 '6caec8ea3732c953fa195320bb26d2e9f630be5edf384a48e42e26ae7198f844', \
 '188c78ef10ce002f2fc3cf8f445d4d1aa12d5f3ce32420e565c9d3cc4d64d8a2', \
 '67ce1464dc89e67dd30acf8adf74c7ec37fa9f14040b7ecd9127391af1b25f2a', \
 'ee017b11d10898f3b19194f43d9b5b9cf443b8e992797e49f4edd603fee060c7']
+    """
+    return _build_merkle_path_for_coinbase([b''] + merkle_leaves)
+
+
+def _build_merkle_path_for_coinbase(merkle_leaves: List[bytes], _partial_path: List[bytes] = []) -> List[bytes]:
+    """ Internal implementation of `build_merkle_path_for_coinbase`, assumes first `merkle_leave` is the coinbase.
     """
     merkle_leaves = merkle_leaves[:]  # copy to preserve original
     _partial_path = _partial_path[:]  # copy to preserve original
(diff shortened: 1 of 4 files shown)


hathor/merged_mining/coordinator.py
```diff
@@ -741,15 +754,10 @@ def job_request(self) -> None:
             self.send_request('mining.notify', job.to_stratum_params())
 
             # for debugging only:
-            bitcoin_block = job.build_bitcoin_block_header(self.dummy_work(job))
+            bitcoin_block = job.build_bitcoin_block_header(job.dummy_work())
```
pedroferreira1, line 757: I know this line is not from this PR but should we do this calculus always? Even if it's just for debug?

This end up calling `build_merkle_root_from_path`, could this be a costly method? Maybe with the `dummy_work` is really fast but if it's not we could call this line only if we really are in debug

hathor/merged_mining/bitcoin.py
```diff
@@ -38,7 +52,7 @@ def __bytes__(self) -> bytes:
         ])
 
     @classmethod
-    def from_dict(cls, params: dict) -> 'BitcoinBlockHeader':
+    def from_dict(cls, params: Dict) -> 'BitcoinBlockHeader':
```
msbrogli, line 55: `Dict[str, Any]`?

hathor/merged_mining/bitcoin.py
```diff
@@ -297,7 +311,7 @@ def has_witness(self) -> bool:
         return bool(self.script_witness)
 
     @classmethod
-    def from_dict(cls, params: dict) -> 'BitcoinTransactionInput':
+    def from_dict(cls, params: Dict) -> 'BitcoinTransactionInput':
```
msbrogli, line 314: `Dict[str, Any]`?

hathor/merged_mining/bitcoin.py
```diff
@@ -357,7 +371,7 @@ def __bytes__(self) -> bytes:
         return struct.pack('<Q', self.value) + encode_bytearray(self.script_pubkey)
 
     @classmethod
-    def from_dict(cls, params: dict) -> 'BitcoinTransactionOutput':
+    def from_dict(cls, params: Dict) -> 'BitcoinTransactionOutput':
```
msbrogli, line 374: `Dict[str, Any]`?

hathor/merged_mining/bitcoin.py
```diff
@@ -437,6 +451,9 @@ def _to_bytes(self, skip_segwit: bool = False) -> bytes:
         data.extend(struct.pack('<I', self.lock_time))
         return bytes(data)
 
+    def to_raw(self) -> BitcoinRawTransaction:
+        return BitcoinRawTransaction(self.hash, self.txid, bytes(self))
```
msbrogli, line 455: Missing docstring.

hathor/merged_mining/bitcoin.py
```diff
@@ -456,7 +473,7 @@ def txid(self) -> bytes:
         return sha256d_hash(self._to_bytes(skip_segwit=True))
 
     @classmethod
-    def from_dict(cls, params: dict) -> 'BitcoinTransaction':
+    def from_dict(cls, params: Dict) -> 'BitcoinTransaction':
```
msbrogli, line 476: `Dict[str, Any]`?

hathor/merged_mining/coordinator.py
msbrogli, line 757: Shouldn't this line be a comment?

hathor/merged_mining/coordinator.py
```diff
@@ -978,11 +999,32 @@ class MergedJob(NamedTuple):
...
+        return self._new_single_miner_job(
+            protocol.next_job_id(),
+            protocol.xnonce1,
+            protocol.xnonce2_size,
+            protocol.payback_script_hathor,
+            payback_script_bitcoin,
+        )
+
+    def _new_single_miner_job(self, job_id: str, xnonce1: bytes, xnonce2_size: int, payback_script_hathor: bytes,
+                              payback_script_bitcoin: bytes) -> SingleMinerJob:
```
msbrogli, line 1026: Missing docstring. Why did you split the code in a separate method if it is called in only one place?

tests/tx/test_merged_mining.py
```diff
@@ -1403,6 +1403,9 @@ async def get_block_template(self, *, rules: List[str] = ['segwit'],
         await asyncio.sleep(self.response_delay)
         return stub
 
+    async def verify_block_proposal(self, *, block: bytes) -> Optional[str]:
+        return None
```
msbrogli, line 1407: TODO, right?

hathor/merged_mining/bitcoin_rpc.py
```diff
@@ -23,6 +23,12 @@ async def get_block_template(self, *, rules: List[str] = ['segwit'],
         """
         raise NotImplementedError
 
+    @abstractmethod
+    async def verify_block_proposal(self, *, block: bytes) -> Optional[str]:
+        """ Method for the [GetBlockTemplate call](https://bitcoin.org/en/developer-reference#getblocktemplate).
```
obiyankenobi, line 28: Couldn't find this link, just https://developer.bitcoin.org/reference/rpc/getblocktemplate.html

hathor/merged_mining/coordinator.py
```diff
@@ -1234,12 +1281,18 @@ def update_merged_block(self) -> None:
...
             self.hathor_coord_job,
             self.bitcoin_coord_job,
             output_script,
             self.is_next_job_clean(),
         )
+        # validate built job
+        block_proposal = merged_job.build_sample_block_proposal()
+        data = await self.bitcoin_rpc.verify_block_proposal(block=block_proposal)
+        self.log.debug('Validation result', data=data)
+        # job is valid
```
obiyankenobi, line 1294: Why is job valid here? If that's the case, the validation result will always confirm it?

hathor/merged_mining/coordinator.py
jansegre: `build_merkle_root_from_path` makes `N=log(len(transactions))` (typically  hashes of `N` txid's, those txid's are already in memory, so it's just calling `sha256d` `N` times in a row. Which is really really cheap. This also already being done (i.e. not new), and from the monitoring of the server it's pretty noticeable that resource usage is impressively low. So I'm worried with this right now, even if it's only for debugging.

hathor/merged_mining/bitcoin.py
jansegre: I honestly can't be bothered using `Dict[str, Any]`, it isn't that much better than `Dict[Any, Any]` which is just `Dict` which is much less annoying to type. I wish there as an easy way to describe a `Dict` that can be json de/serialized. [But this isn't quite possible yet.](https://github.com/python/typing/issues/182). But we could use Guido's suggestion of doing `JsonDict = Dict[str, Any]` add that to `utils.py` and use where it makes sense (most uses of `Dict` probably).

hathor/merged_mining/coordinator.py
jansegre: Maybe, but it isn't hurting performance and I think is very useful to be able to turn it on just by using `--debug`, which I think can't be verified right now just by looking at the `log` object. And I don't think it's worth propagating a `debug` param, since it would only be used here without clear benefit.

hathor/merged_mining/coordinator.py
jansegre: No, it's called right above on build_sample_block_proposal too.

tests/tx/test_merged_mining.py
jansegre: Not really, I don't really want to port Bitcoin's verification algorithm. But it would be great if we had it. `None` means `valid`, the signature method could use a better docstring though.

hathor/merged_mining/bitcoin_rpc.py
jansegre: I copied the link from `get_block_template`, it was working before, but I think I'll have to fix the other links too.

hathor/merged_mining/coordinator.py
jansegre: Not sure what I thought when I added that comment. It'll be valid if `data is None`. I'll improve this.

jansegre: Fixed. What do you think now?

hathor/merged_mining/bitcoin_rpc.py
jansegre: Fixed.

hathor/merged_mining/coordinator.py
jansegre: Added docstring.

hathor/merged_mining/bitcoin.py
jansegre: Added docstring.

hathor/merged_mining/bitcoin.py
msbrogli: I agree that they are quite the same. But using `Dict[str, Any]` is a step towards strict mode verification by mypy. I like the idea of using a `JsonDict`.

hathor/merged_mining/coordinator.py
msbrogli: Ok! It makes sense. It is a shame we can't check the log level at the `log` object, but I also understand why it is done this way.

hathor/merged_mining/coordinator.py
msbrogli: Ok. Thanks!

tests/tx/test_merged_mining.py
msbrogli: Ok! Is it just a mock for the tests?

jansegre: Yes.

hathor/merged_mining/bitcoin.py
jansegre: I'll do this on a new PR.

jansegre: I've made some progress on `refactor/typing-json-dict`. Trying to be a little more specific than `Dict[str, Any]` has been taking far more work than I'd like to put in now. I'll come back to it later because I may have spotted a few issues because of this. But I won't include any of it on this PR, since there isn't any benefit to the issue it solves.

jansegre: The draft PR is here: https://github.com/HathorNetwork/hathor-core/pull/126 I won't be putting more effort on this for now.

PR #120, opened 2020-07-21
feat(mergedmining): use longpoll on bitcoin.getblocktemplate

jansegre:
- [x] change base to `dev` after `feat/merged-mining-better-api` is merged

### DIFF
--- a/hathor/merged_mining/coordinator.py
+++ b/hathor/merged_mining/coordinator.py
@@ -1133,7 +1133,7 @@ class MergedMiningCoordinator:
         self.sequential_xnonce1 = sequential_xnonce1
         self._next_xnonce1 = 0
         self.job_count = 0
-        self.update_bitcoin_block_task: Optional[Periodic] = None
+        self.update_bitcoin_block_task: Optional[asyncio.Task] = None
         self.update_hathor_block_task: Optional[asyncio.Task] = None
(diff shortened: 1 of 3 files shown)


jansegre:
In draft because it depends on #114

hathor/merged_mining/coordinator.py
```diff
@@ -1177,49 +1177,56 @@ def update_jobs(self) -> None:
...
-        # self.log.debug('bitcoin.getblocktemplate response', res=data_log)
-        self.bitcoin_coord_job = BitcoinCoordJob.from_dict(data)
-        self.log.debug('new Bitcoin block template', height=self.bitcoin_coord_job.height)
-        await self.update_merged_block()
+        backoff = 1
+        longpoll_id = None
+        while True:
+            self.log.debug('get Bitcoin block template')
+            try:
+                backoff = 1
```
msbrogli, line 1210: Should it be after `self.bitcoin_rpc.get_block_template(...)`? It seems `await asyncio.sleep(backoff)` will always wait one second.

jansegre: You're right, my mistake.

jansegre: Fixed.

msbrogli (approved):

PR #124, opened 2020-07-24
feat: skipping block difficulty validation while initializing the node

pedroferreira1:
In my notebook:

`dev`: between 19 - 24 tx/s
`feat/skip-block-weight-validation`: between 80 - 120 tx/s

### DIFF
--- a/hathor/conf/settings.py
+++ b/hathor/conf/settings.py
@@ -248,5 +248,6 @@ class HathorSettings(NamedTuple):
     # Amount in which tx min weight reaches the middle point between the minimum and maximum weight
     MIN_TX_WEIGHT_K: int = 100
 
-    # Quantity of blocks without a weight verification when doing a full verification
-    QUANTITY_BLOCKS_SKIP_WEIGHT_VERIFICATION: int = 1000
+    # When the node is being initialized (with a full verification) we don't verify
+    # the difficulty of all blocks, we execute the validation every N blocks only
+    VERIFY_WEIGHT_EVERY_N_BLOCKS: int = 1000
--- a/hathor/conf/settings.py
+++ b/hathor/conf/settings.py
@@ -247,3 +247,6 @@ class HathorSettings(NamedTuple):
 
     # Amount in which tx min weight reaches the middle point between the minimum and maximum weight
     MIN_TX_WEIGHT_K: int = 100
+
+    # Quantity of blocks without a weight verification when doing a full verification
+    QUANTITY_BLOCKS_SKIP_WEIGHT_VERIFICATION: int = 1000
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -260,6 +260,8 @@ class HathorManager:
         cnt = 0
         cnt2 = 0
 
+        block_count = 0
+
         # self.start_profiler()
         for tx in self.tx_storage._topological_sort():
             assert tx.hash is not None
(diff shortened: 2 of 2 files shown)


hathor/manager.py
```diff
@@ -423,14 +423,15 @@ def validate_new_tx(self, tx: BaseTransaction) -> bool:
             tx = cast(Block, tx)
             assert tx.hash is not None  # XXX: it appears that after casting this assert "casting" is lost
 
-            # Validate minimum block difficulty
-            block_weight = self.calculate_block_difficulty(tx)
-            if tx.weight < block_weight - settings.WEIGHT_TOL:
-                raise InvalidNewTransaction(
-                    'Invalid new block {}: weight ({}) is smaller than the minimum weight ({})'.format(
-                        tx.hash.hex(), tx.weight, block_weight
+            if self.state != self.NodeState.INITIALIZING:
```
obiyankenobi, line 426: I'd just add a comment explaining that we can skip this check during startup as we'll trust the difficulty stored in metadata.

msbrogli, line 426: What if we check one every 120 blocks, i.e., once per hour? To remove this logic from here, we can add a `skip_block_weight_verification` parameter to this method.

pedroferreira1: Done

pedroferreira1: Can you check this commit and see if it's good? https://github.com/HathorNetwork/hathor-core/pull/124/commits/e83a2052ca868ac932b848d93ffd3a6e5435912d

I would just discuss about this number, I added a config on settings with this and first I set as 1000. I am not sure the benefit of having 6k verifications instead of 700. But it's easy to change

hathor/conf/settings.py
```diff
@@ -247,3 +247,6 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
 
     # Amount in which tx min weight reaches the middle point between the minimum and maximum weight
     MIN_TX_WEIGHT_K: int = 100
+
+    # Quantity of blocks without a weight verification when doing a full verification
+    QUANTITY_BLOCKS_SKIP_WEIGHT_VERIFICATION: int = 1000
```
jansegre, line 252: This name is a bit long, that would be OK if the meaning was more clear. The comment can be interpreted as "don't verify 1000 blocks, then verify the rest", but I understood it as "don't verify any blocks, unless the last 1000". Turns out was neither.

I think including "every n" in some form would be enough (at least for my understanding). Some suggestions:

- `SKIP_VERIFY_WEIGHT_EXCEPT_EVERY_N_BLOCKS`: maybe a bit long but in my head this is literally what it does;
- `VERIFY_WEIGHT_EVERY_N_BLOCKS`: avoids double negation (skip+except);
- `BLOCK_VERIFY_WEIGHT_EVERY_N`: ordered more "hierarchical" (`block.verify_weight(every_n)`) but not grammatical;

What do you think?

hathor/manager.py
msbrogli: I think it is great. Thanks!

hathor/conf/settings.py
pedroferreira1: I agree with you that the name is not good. I do like the second option (`VERIFY_WEIGHT_EVERY_N_BLOCKS`). Changed for that one.

PR #126, opened 2020-07-30
refactor: add type JsonDict and others for better dicts

jansegre:
dev


jansegre:
Some comments appeared because I initially targeted `master` instead of `dev`.

PR #128, opened 2020-07-31
fix: stop adding tx to some wrong addresses on wallet index

pedroferreira1:


hathor/indexes.py
```diff
@@ -376,21 +377,22 @@ def _get_addresses(self, tx: BaseTransaction) -> Set[str]:
...
-                    assert tx.hash is not None
-                    address = script_type_out.address
-                    addresses.add(address)
 
-        for txout in tx.outputs:
-            script_type_out = parse_address_script(txout.script)
+        def add_address_from_output(output: 'TxOutput') -> None:
+            script_type_out = parse_address_script(output.script)
             if script_type_out:
                 assert tx.hash is not None
```
msbrogli, line 384: It seems we should remove this assert.

pedroferreira1: Removed

PR #129, opened 2020-07-31
chore: add token info to token creation tx extended json

obiyankenobi:


hathor/transaction/token_creation_tx.py
```diff
@@ -232,6 +232,13 @@ def to_json(self, decode_script: bool = False, include_metadata: bool = False) -
         json['tokens'] = []
         return json
 
+    def to_json_extended(self) -> Dict[str, Any]:
```
pedroferreira1, line 235: Where do we need this?

obiyankenobi: When we send the `network:new_tx_accepted` event, we use `to_json_extended`. Without this, we can't get the token info from a token creation tx.

PR #136, opened 2020-08-31
chore: improve wallet_index access

obiyankenobi:


hathor/indexes.py
```diff
@@ -445,6 +445,9 @@ def get_sorted_from_address(self, address: str) -> List[bytes]:
         """
         return sorted(self.index[address])
 
+    def is_address_empty(seld, address: str) -> bool:
```
pedroferreira1, line 448: self*

obiyankenobi: Thanks

PR #138, opened 2020-09-02
feat: creating POST resource for push tx endpoint

pedroferreira1:
GET resource was kept for compatibility but will be deprecated


tests/resources/transaction/test_pushtx.py
```diff
@@ -85,7 +85,7 @@ def test_get(self):
         self.assertFalse(data['success'])
 
         # force
-        response = yield self.web.get('push_tx', {b'hex_tx': bytes(tx.get_struct().hex(), 'utf-8'), b'force': b'true'})
+        response = yield self.web.post('push_tx', {'hex_tx': tx.get_struct().hex(), 'force': True})
```
msbrogli, line 88: Shouldn't we keep the GET test until the support for GET is removed?

msbrogli (changes requested):

obiyankenobi (approved):

tests/resources/transaction/test_pushtx.py
pedroferreira1: @msbrogli as discussed in privately, I've done a refactor in the tests to have all tests for the GET and the POST. Can you see if it's good?

tests/resources/transaction/test_pushtx.py
```diff
@@ -56,7 +56,14 @@ def test_get(self):
         tx.parents = self.manager.get_new_tx_parents(tx.timestamp)
         tx.resolve()
 
-        response = yield self.web.get('push_tx', {b'hex_tx': bytes(tx.get_struct().hex(), 'utf-8')})
+        push_tx_fn = self.web.post if is_post else self.web.get
+        hex_param = 'hex_tx' if is_post else b'hex_tx'
```
msbrogli, line 60: Why don't we call a function that converts all keys and values to binary if needed?

tests/resources/transaction/test_pushtx.py
```diff
@@ -56,7 +56,14 @@ def test_get(self):
...
+        force_param = 'force' if is_post else b'force'
+
+        tx_hex = tx.get_struct().hex()
+        hex_data = tx_hex if is_post else bytes(tx_hex, 'utf-8')
```
msbrogli, line 64: Why do we have different things? Are we sending binaries in the HTTP POST's body?

msbrogli (approved):

tests/resources/transaction/test_pushtx.py
pedroferreira1: No, the GET requires bytes and the POST doesn't

tests/resources/transaction/test_pushtx.py
pedroferreira1: As discussed privately we decided to keep like this

PR #150, opened 2020-09-18
feat(mergedmining): support DGB as parent-chain

jansegre:


hathor/cli/merged_mining.py
```diff
@@ -28,22 +31,34 @@ def execute(args: Namespace) -> None:
...
-    bitcoin_rpc = BitcoinRPC(args.bitcoin_rpc)
+    kwargs: Dict[str, Any] = {}
+    if args.bitcoin_rpc:
+        kwargs['bitcoin_rpc'] = bitcoin_rpc = BitcoinRPC(args.bitcoin_rpc)
+        kwargs['payback_address_bitcoin'] = args.bitcoin_address
+    elif args.digibyte_rpc:
+        kwargs['digibyte_rpc'] = digibyte_rpc = DigibyteRPC(args.digibyte_rpc)
+        kwargs['payback_address_digibyte'] = args.digibyte_address
+    else:
+        raise ValueError('should have at least one parent chain')
```
msbrogli, line 46: I think we should protect from using more than one parent chain.

hathor/merged_mining/coordinator.py
```diff
@@ -502,6 +512,7 @@ def handle_error(self, error: Dict, data: Any, msgid: Optional[str]) -> None:
     def handle_authorize(self, params: List, msgid: Optional[str]) -> None:
         """ Handles authorize request by always authorizing even if the request is invalid.
         """
+        # XXX
```
msbrogli, line 515: Missing comment.

hathor/merged_mining/coordinator.py
```diff
@@ -714,6 +732,40 @@ async def submit_to_bitcoin(self, job: SingleMinerJob, work: SingleMinerWork) ->
...
+        self.log.debug('digibyte_rpc.submit_block', res=res)
+        if res is None:
+            self.log.info('new Bitcoin block found!!!', hash=bitcoin_block.header.hash.hex())
+            await self.coordinator.update_bitcoin_block()
+        else:
+            # Known reasons:
+            # - high-hash: PoW not enough, shouldn't happen because we check the difficulty before sending
+            # - bad-*: invalid block data
+            # - unexpected-witness: transaction has inputs with witness but isn't marked as containing witnesses
+            self.log.error('block rejected from Bitcoin', reason=res)
```
msbrogli, line 767: digibyte?

hathor/merged_mining/coordinator.py
```diff
@@ -714,6 +732,40 @@ async def submit_to_bitcoin(self, job: SingleMinerJob, work: SingleMinerWork) ->
...
+        except Exception:
+            self.log.warn('submit to Bitcoin failed', exc_info=True)
+            return
+        self.coordinator.update_parent_submitted(job.parent_height, ParentChain.DIGIBYTE)
+        self.log.debug('digibyte_rpc.submit_block', res=res)
+        if res is None:
+            self.log.info('new Bitcoin block found!!!', hash=bitcoin_block.header.hash.hex())
+            await self.coordinator.update_bitcoin_block()
+        else:
+            # Known reasons:
```
msbrogli, line 763: Are these reasons valid for Digibyte? It seems it was copied from Bitcoin.

hathor/merged_mining/coordinator.py
```diff
@@ -1100,20 +1167,27 @@ class MergedMiningCoordinator:
     """
 
     BITCOIN_UPDATE_INTERVAL = 10.0
+    DIGIBYTE_UPDATE_INTERVAL = 3.0
     # very arbitrary, max times since last submit that we consider sending a new winning share
-    BITCOIN_MAX_FUTURE_SUBMIT_SECONDS = 30.0
+    MAX_FUTURE_SUBMIT_SECONDS = {
```
msbrogli, line 1172: Instead of `if` statements, isn't it easier to create a class `ParentSpecificMethods`? I don't know if it is easy to generalize but I feel it is better (and easier to add other parent chains in the future).

hathor/cli/merged_mining.py
msbrogli: It took me sometime to understand that we support both at the same time. Maybe add a comment about it somewhere.

hathor/merged_mining/coordinator.py
```diff
@@ -1222,29 +1302,60 @@ async def update_bitcoin_block(self) -> None:
...
+                assert self.bitcoin_coord_job is not None
                 self.log.debug('new Bitcoin block template', height=self.bitcoin_coord_job.height)
-                await self.update_merged_block()
+                await self.update_merged_block(ParentChain.BITCOIN)
                 if longpoll_id is None:
                     self.log.warn('no longpoll_id received, sleep instead')
                     await asyncio.sleep(self.BITCOIN_UPDATE_INTERVAL)
 
-    def update_bitcoin_submitted(self, height: int) -> None:
+    async def update_digibyte_block(self) -> None:
```
msbrogli, line 1312: I feel we should refactor and move all these specific codes to separate classes, one for each parent chain.

msbrogli (changes requested):

hathor/cli/merged_mining.py
obiyankenobi: From what I understood we only support one at a time. That's why we get only one of the arguments from the parser: `parent_chain = parser.add_mutually_exclusive_group()`

hathor/merged_mining/coordinator.py
```diff
@@ -215,7 +224,7 @@ def _make_bitcoin_block_and_coinbase(self, work: SingleMinerWork) -> Tuple[Bitco
         )
         return bitcoin_header, coinbase_tx
 
-    def build_bitcoin_block_header(self, work: SingleMinerWork) -> BitcoinBlockHeader:
+    def build_parent_block_header(self, work: SingleMinerWork) -> BitcoinBlockHeader:
```
obiyankenobi, line 227: This builds both BTC and DGB block headers?

hathor/merged_mining/coordinator.py
```diff
@@ -714,6 +732,40 @@ async def submit_to_bitcoin(self, job: SingleMinerJob, work: SingleMinerWork) ->
...
+        bitcoin_block_header = job.build_parent_block_header(work)
+        block_hash = Hash(bitcoin_block_header.hash)
+        block_target = Target(int.from_bytes(bitcoin_block_header.bits, 'big'))
+        if block_hash.to_u256() > block_target.to_u256():
+            self.log.debug('high hash for DigiByte, keep mining')
+            return
+        if self.coordinator.should_skip_parent_submit(job.parent_height, ParentChain.DIGIBYTE):
+            self.log.debug('late winning share, skipping DigiByte submit')
+            return
+        bitcoin_block = job.build_bitcoin_block(work)  # XXX: far fewer cases, so it's OK now
```
obiyankenobi, line 750: I'm a bit confused. Does DGB use the same structure as BTC?

hathor/merged_mining/coordinator.py
```diff
@@ -714,6 +732,40 @@ async def submit_to_bitcoin(self, job: SingleMinerJob, work: SingleMinerWork) ->
...
+        data = bytes(bitcoin_block)
+        try:
+            res = await digibyte_rpc.submit_block(data)
+        except Exception:
+            self.log.warn('submit to Bitcoin failed', exc_info=True)
+            return
+        self.coordinator.update_parent_submitted(job.parent_height, ParentChain.DIGIBYTE)
+        self.log.debug('digibyte_rpc.submit_block', res=res)
+        if res is None:
+            self.log.info('new Bitcoin block found!!!', hash=bitcoin_block.header.hash.hex())
```
obiyankenobi, line 760: Bitcoin

tests/tx/test_merged_mining.py
```diff
@@ -1410,3 +1412,7 @@ async def submit_block(self, block: bytes) -> str:
         stub = 'high-hash'
         await asyncio.sleep(self.response_delay)
         return stub
+
+    async def validate_address(self, address: str) -> Dict:
```
obiyankenobi, line 1416: Seems to be from the other PR.

hathor/merged_mining/digibyte_rpc.py
```diff
@@ -0,0 +1,151 @@
...
+from itertools import count
+from typing import Any, Callable, Dict, Iterator, List, Optional, Union, cast
+
+from aiohttp import BasicAuth, ClientSession
+from structlog import get_logger
+
+logger = get_logger()
+
+
+class RPCFailure(Exception):
```
pedroferreira1, line 11: You already have one of this in the `bitcoin_rpc.py`, can we use the same?

hathor/merged_mining/digibyte_rpc.py
```diff
@@ -0,0 +1,151 @@
...
+    def __init__(self, message: str, code: Optional[int] = None):
+        super().__init__(message)
+        self.code = code
+
+
+class IDigibyteRPC(ABC):
```
pedroferreira1, line 17: About this class, is it possible to create a parent class with this? For what I see the only difference between this one and the `IBitcoinRPC` are the default capabilities on the `get_block_template`. I might be wrong, but I'm thinking about the other coins that we might add in the future, this could make it easier.

hathor/merged_mining/digibyte_rpc.py
```diff
@@ -0,0 +1,151 @@
...
+
+class DigibyteRPC(IDigibyteRPC):
+    """ Class for making calls to Digibyte's RPC.
+
+    References:
+
+    - https://bitcoin.org/en/developer-reference#remote-procedure-calls-rpcs
+    """
+
+    USER_AGENT = 'hathor-merged-mining'
```
pedroferreira1, line 53: I think this could go in the conf file? Or maybe a constants file inside the merged_mining module. Just we don't duplicate it in all coins we have

hathor/cli/merged_mining.py
jansegre: Yes, it's one at a time at the command line level. Internally it almost works with both of them, but there are still some unresolved questions when using multiple parent chains, so I opted to block it at the command line level.

hathor/merged_mining/coordinator.py
jansegre: I was using this as a placeholder. Removed.

hathor/merged_mining/coordinator.py
jansegre: Fixed.

hathor/merged_mining/coordinator.py
jansegre: I updated these with feedback from testing.

hathor/merged_mining/coordinator.py
jansegre: I agree, but I used structures like `Dict[ParentChain, X]` precisely to avoid `if`s.

hathor/merged_mining/coordinator.py
jansegre: I agree it could use a good refactoring. Or even be re-written from scratch. I had very little idea what I would need to generalize while I was adding this support, and even now it isn't too clear. However I'm not sure if it'd be best to do this refactoring right now, at the very least I'd avoid doing this on this PR as not to stretch it for too long.

hathor/merged_mining/coordinator.py
jansegre: Yes. The header is exactly the same format.

tests/tx/test_merged_mining.py
jansegre: I've proposed a similar API, but this is the one from Bitcoin that inspired ours. I'm now using it (on both Bitcoin and DigiByte), to simplify address validation, support more address types, and not have to build the output script manually.

hathor/merged_mining/digibyte_rpc.py
jansegre: I guess I could. I can make an `rpc_commons` module for this.

hathor/merged_mining/digibyte_rpc.py
jansegre: I think it makes sense to unify jsonrpc related abstractions, and I'll do that on an `rpc_commons` or even `jsonrpc` module. But I prefer not making a class for common methods, precisely because the little differences are very important, and I also can't foresee what new parent chains (like Bitcoin Cash or SV) will need to be different, so it's safer to assume they all have distinct method sets and implement them independently. But unifying jsonrpc is definitely a plus.

hathor/merged_mining/digibyte_rpc.py
jansegre: I'll move this to a `jsonrpc` module (I'm settling on that now). But I don't think this should normally be customizable. Also I want to move this away from `hathor-core` so that's one more reason not to create a conf that will be removed.

jansegre: Done.

hathor/merged_mining/digibyte_rpc.py
jansegre: Done.

hathor/merged_mining/digibyte_rpc.py
jansegre: Done, made a `jsonrpc` module instead.

hathor/merged_mining/coordinator.py
jansegre: Fixed.

hathor/merged_mining/coordinator.py
jansegre: The block header is identical, there are some differences on the coinbase though, but that is built elsewhere. Also there might be differences with transactions, but it doesn't matter because they're also forwarded transparently.

tests/tx/test_merged_mining.py
obiyankenobi: Excellent

hathor/cli/merged_mining.py
```diff
@@ -28,22 +33,38 @@ def execute(args: Namespace) -> None:
...
         min_difficulty=args.min_diff,
+        constant_difficulty=args.const_diff,
+        **kwargs,
     )
-    logger.info('start Bitcoin RPC', url=args.bitcoin_rpc)
-    loop.run_until_complete(bitcoin_rpc.start())
+    if args.bitcoin_rpc:
+        logger.info('start Bitcoin RPC', url=args.bitcoin_rpc)
+        loop.run_until_complete(bitcoin_rpc.start())
+    elif args.digibyte_rpc:
```
obiyankenobi, line 65: We have the same `if...else` before. Could we start the RPC services there or it needs to be after the coordinator?

hathor/merged_mining/coordinator.py
obiyankenobi: Thanks. Could you just add that to the docstring please? Or change the method name to use bitcoin again and we can add a comment when we call it from the DGB code to say the header is the same as BTC.

hathor/merged_mining/coordinator.py
obiyankenobi: Got it.

pedroferreira1 (approved):

hathor/merged_mining/bitcoin.py
```diff
@@ -282,7 +283,7 @@ class BitcoinTransactionInput(NamedTuple):
     # Transaction version as defined by the sender. Intended for "replacement" of transactions when information is
     # updated before inclusion into a block.
     sequence: int = SEQUENCE_FINAL  # default value disables nLockTime
-    script_witness: List[bytes] = []
+    script_witness: Tuple[bytes, ...] = ()
```
msbrogli, line 286: I'd say `tuple()` is easier to understand.

hathor/merged_mining/coordinator.py
```diff
@@ -259,7 +373,7 @@ class MergedMiningStratumProtocol(asyncio.Protocol):
     DEFAULT_XNONCE2_SIZE = 8  # size in bytes to reserve for extra nonce 2 (which is concatenated with extra nonce 1)
     ESTIMATOR_LOOP_INTERVAL = 30  # in seconds, "frequency" that the function that updates the estimator will be called
     ESTIMATOR_WINDOW_INTERVAL = 60 * 15  # in seconds, size of the window to use for estimating miner's hashrate
-    MIN_DIFFICULTY = 128  # minimum "bitcoin difficulty" to assign to jobs
+    MIN_DIFFICULTY = 50000  # minimum "bitcoin difficulty" to assign to jobs
```
msbrogli, line 376: It's such a huge increase. Why?

hathor/merged_mining/coordinator.py
msbrogli: I've seen `if` statements along the code. I understand that we can refactor it to generalize when we add the next parent chain. That's the idea, right?

msbrogli (approved):

hathor/merged_mining/coordinator.py
jansegre: Accidentally committed this, I increased here for testing but then noticed I could use the `--min-diff 50000` param.

hathor/merged_mining/bitcoin.py
jansegre: Oh! I guess it's subjective, I've never seen `tuple()` without arguments. But it's fine either way for me.

hathor/cli/merged_mining.py
jansegre: I thought there would be some problem, but it's actually fine. I'll do this refactor, thanks.

hathor/merged_mining/coordinator.py
jansegre: Since I'm using "bitcoin like" terminology (in `BitcoinLikeAddress` that is used for DigiByte too), what do you think about `build_bitcoin_like_header`?

obiyankenobi: Sounds good

obiyankenobi (approved):

hathor/merged_mining/coordinator.py
jansegre: Yes. And IMO we would leave this for the next parent chain integration.

hathor/merged_mining/coordinator.py
jansegre: Done.

hathor/cli/merged_mining.py
jansegre: Done.

hathor/merged_mining/coordinator.py
jansegre: Reversed.

hathor/merged_mining/bitcoin.py
jansegre: My best argument pro `()` is that it's the canonical representation (`repr(tuple())` -> `'()'`).

PR #157, opened 2020-10-11
chore: add support for Python 3.7 and 3.8

jansegre:


obiyankenobi (approved):

hathor/merged_mining/coordinator.py
```diff
@@ -69,6 +58,22 @@
 MAGIC_NUMBER = b'Hath'  # bytes.fromhex('48617468') or 0x68746148.to_bytes(4, 'little')
 NICEHASH_MIN_DIFF = 500_000  # as per https://www.nicehash.com/pool-operators
 
+UNRECOVERABLE_ERROR_CODE_MAX = -32600
```
pedroferreira1, line 61: Why are you adding this here? Isn't it the same as in the `hathor/stratum/stratum.py`?

jansegre: The simple answer is that it's complicated.

The module `hathor.merged_mining.coordinator` is in the path of `hathor.transaction` (through `hathor.transaction.aux_pow`, which is imported very early on. So it ends up being very crucial that it's "importable". However `hathor.stratum.stratum` isn't crucial to be "importable", and its use of `multiprocessing.sharedctypes` right at the class definition level (`MinerJob`) does not work on certain platforms (Windows and or PyPy or recent versions of Python, can't remember right now). I chose to refactor this by moving the shared values to `hathor.merged_mining` and not importing it from `hathor.stratum`.

In the short/mid term I think we'll end up deleting `hathor.stratum` and splitting `hathor.merged_mining` into a separate project. So I'm not too worried about these duplicates, furthermore these aren't values that will ever change, so I think it's not terrible for them to be in multiple places for a little while.

pedroferreira1: Makes sense, thanks for the explanation!

pedroferreira1 (approved):

PR #158, opened 2020-10-11
fix(tests): retry tests/p2p/test_split_brain once if it fails

jansegre:
That test is known to randomly fail. We will always re-run it manually when that happens. This PR adds the [flaky](https://github.com/box/flaky) plugin that will do just that automatically for tests marked as flaky (`@pytest.mark.flaky` like this one is now) with the following strategy: out of 2 runs it has to pass one.

This is what it looks like when it passes the first time:

```
===Flaky Test Report===

test_split_brain passed 1 out of the required 1 times. Success!

===End Flaky Test Report===
```

And this is what it looks like when it fails once and has to run again (which took several tries):

```
===Flaky Test Report===

test_split_brain failed (1 runs remaining out of 2).
	<class 'twisted.trial.unittest.FailTest'>
	Items in the first set but not the second:
Interval(3177661205, inf, b'\x07B\x96\x9dx\x01Av\x80\xf0\xb4E\xd3\x8d\xc2\xc5\xc1\x14uDsl\xe7\x80\xd4\x82\x95\xbdZm\xdd;')
	[<TracebackEntry /Users/jan/Projects/hathor-core/tests/p2p/test_split_brain.py:116>, <TracebackEntry /Users/jan/Projects/hathor-core/tests/unittest.py:85>, <TracebackEntry /Users/jan/Library/Caches/pypoetry/virtualenvs/hathor-JIQwCn9w-py3.6/lib/python3.6/site-packages/twisted/trial/_synctest.py:432>, <TracebackEntry /Users/jan/Library/Caches/pypoetry/virtualenvs/hathor-JIQwCn9w-py3.6/lib/python3.6/site-packages/twisted/trial/_synctest.py:375>]
test_split_brain passed 1 out of the required 1 times. Success!

===End Flaky Test Report===
```

I did not have the patience to run try and make it fail twice in a row.

### DIFF
--- a/tests/p2p/test_split_brain.py
+++ b/tests/p2p/test_split_brain.py
@@ -1,5 +1,6 @@
 import random
 
+import pytest
 from mnemonic import Mnemonic
 
 from hathor.graphviz import GraphvizVisualizer
@@ -46,6 +47,8 @@ class HathorSyncMethodsTestCase(unittest.TestCase):
         wallet.unlock(words=words, tx_storage=manager.tx_storage)
         return manager
 
+    # retry the test once if it fails, see: https://github.com/box/flaky
+    @pytest.mark.flaky
     def test_split_brain(self):
         debug_pdf = False
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -225,6 +225,14 @@ pycodestyle = ">=2.5.0,<2.6.0"
 pyflakes = ">=2.1.0,<2.2.0"
 
 [[package]]
+name = "flaky"
+version = "3.7.0"
+description = "Plugin for nose or pytest that automatically reruns flaky tests."
+category = "dev"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
+
+[[package]]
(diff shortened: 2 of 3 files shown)


obiyankenobi (approved):

tests/p2p/test_split_brain.py
```diff
@@ -46,6 +47,8 @@ def create_peer(self, network, unlock_wallet=True):
         wallet.unlock(words=words, tx_storage=manager.tx_storage)
         return manager
 
+    # retry the test once if it fails, see: https://github.com/box/flaky
+    @pytest.mark.flaky
```
pedroferreira1, line 51: How do you configure how many attempts/failures you want? Is it always one out of two?

pedroferreira1 (approved):

tests/p2p/test_split_brain.py
jansegre: The Flaky GitHub says it's `@flaky(max_runs=3, min_passes=2)`, I suppose it also works as `@pytest.mark.flaky(max_runs=3, min_passes=2)`, but I haven't tested it.

PR #164, opened 2020-12-18
feat: create API to return all addresses of a wallet

pedroferreira1:


hathor/wallet/resources/addresses.py
```diff
@@ -0,0 +1,69 @@
...
+            # We have this method only for HD wallet
+            request.setResponseCode(503)
+            return json.dumps({'success': False}, indent=4).encode('utf-8')
+
+        addresses = self.manager.wallet.get_all_addresses()
+
+        data = {
+            'addresses': addresses,
+        }
+        return json.dumps(data, indent=4).encode('utf-8')
```
obiyankenobi, line 39: Shouldn't we add `success: True` to maintain our standard?

obiyankenobi (approved):

hathor/wallet/resources/addresses.py
pedroferreira1: Actually I kept the same behaviour of the `AddressResource`, where we return only the address. But I guess it's and old resource, I will add the `success: True`

jansegre (approved):

hathor/wallet/resources/addresses.py
```diff
@@ -0,0 +1,69 @@
...
+
+        if not self.manager.wallet or self.manager.wallet.type != self.manager.wallet.WalletType.HD:
+            # We have this method only for HD wallet
+            request.setResponseCode(503)
+            return json.dumps({'success': False}, indent=4).encode('utf-8')
+
+        addresses = self.manager.wallet.get_all_addresses()
+
+        data = {
+            'addresses': addresses,
```
msbrogli, line 37: We must paginate. It feels very unsafe to return all addresses.

msbrogli (changes requested):

pedroferreira1:
This was created at first to help some miners that were using our built in wallet in the full node but it's not needed anymore. They are mining without this API

PR #165, opened 2020-12-24
feat: add POST request for address history

pedroferreira1:


hathor/wallet/resources/thin_wallet/address_history.py
```diff
@@ -24,6 +24,27 @@ class AddressHistoryResource(resource.Resource):
     def __init__(self, manager):
         self.manager = manager
 
+    def render_POST(self, request: Request) -> bytes:
```
jansegre, line 27: We also need to add the OpenAPI spec for this method. I don't mind if we only add a `TODO` comment though.

jansegre (approved):

hathor/wallet/resources/thin_wallet/address_history.py
pedroferreira1: Added the TODO

obiyankenobi (approved):

PR #170, opened 2021-01-05
fix(api): validate_address returns 500 on valid address

jansegre:
This is a regression that passes the tests but not on actual usage. I'm figuring out how to test this, it can potentially affect other endpoints that rely on a request parameter to be passed as `str`, but so far this is the only known case (maybe because it's the only usage of `getChild`).

### DIFF
--- a/hathor/transaction/resources/validate_address.py
+++ b/hathor/transaction/resources/validate_address.py
@@ -1,9 +1,11 @@
 import base64
+from typing import Union
 
 from twisted.web import resource
 
 from hathor.api_util import set_cors
 from hathor.cli.openapi_files.register import register_resource
+from hathor.manager import HathorManager
 from hathor.transaction.scripts import create_base_script
 from hathor.util import api_catch_exceptions, json_dumpb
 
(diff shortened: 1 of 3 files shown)


pedroferreira1 (approved):

hathor/transaction/resources/validate_address.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, manager, address):
         super().__init__()
         # Important to have the manager so we can know the tx_storage
         self.manager = manager
+        if isinstance(address, bytes):
+            address = address.decode('ascii')
+        assert isinstance(address, str)
         self.address = address
```
msbrogli, line 41: Missing typing.

msbrogli (approved):

hathor/transaction/resources/validate_address.py
jansegre: Done.

jansegre:
Tests added. These new tests fail when the fix is not included, and pass when it is. The endpoint now properly works when tested in practice.

PR #171, opened 2021-01-06
fix(p2p): re-run discovery when no peers are connected

jansegre:
This PR was created to fix an issue, but it's actually a feature.

To fix the problem of rejecting every peer before getting the whitelist I added the feature of re-running the discovery process periodically if there are no connected peers.

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -245,8 +245,7 @@ class HathorManager:
         for description in self.listen_addresses:
             self.listen(description, ssl=self.ssl)
 
-        for peer_discovery in self.peer_discoveries:
-            peer_discovery.discover_and_connect(self.connections.connect_to)
+        self.do_discovery()
 
         self.start_time = time.time()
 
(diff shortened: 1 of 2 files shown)


pedroferreira1 (approved):

hathor/p2p/manager.py
```diff
@@ -225,8 +225,11 @@ def reconnect_to_all(self) -> None:
         """ It is called by the `lc_reconnect` timer and tries to connect to all known
         peers.
 
-        TODO(epnichols): Should we always conect to *all*? Should there be a max #?
+        TODO(epnichols): Should we always connect to *all*? Should there be a max #?
         """
+        if len(self.connected_peers) < 1:
```
msbrogli, line 230: I think it is a good behavior to add. Please, add some comments explaining what it is doing and why.

msbrogli (changes requested):

hathor/manager.py
```diff
@@ -282,6 +281,10 @@ def stop(self) -> Deferred:
 
         return defer.DeferredList(waits)
 
+    def do_discovery(self) -> None:
```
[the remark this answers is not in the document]
jansegre: Done.

hathor/p2p/manager.py
jansegre: Added. What do you think?

msbrogli (approved):

PR #180, opened 2021-02-02
chore(ci): remove poetry cache, which was causing tests to crash

jansegre:
I thought about just disabling the caches or just hijacking the cache key to force it to be rebuilt (since there isn't a way to force cache invalidation for Github Actions), but after reviewing all builds, not using this cache will add about 3min to build time, which is not much (it usually takes about 30min for the whole matrix to finish).

Given that we don't really know what caused the cache to become a problem, and thus we don't know how to incorporate that into the key to make it "imcompatible" (for instance we already incorporate the OS and Python version, so one cache can't be used on a different os or Python version), we don't know when something like this will happen again. And given that we can't "quickly" solve this by force-invalidating a cache (I think it takes 7 days of no use for a cache to be dropped), it seems to me that just not using the cache is the best current approach.

What do you think?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,23 +51,11 @@ jobs:
       run: brew install -q graphviz rocksdb
     - name: Install Poetry
       run: pip -q --no-input install poetry
-    - name: Get Poetry cache-dir
-      id: poetry-install
-      shell: bash
-      run: echo ::set-output name=cache-dir::$(poetry config cache-dir)
-    - name: Cache Poetry dependencies
-      uses: actions/cache@v2
-      id: poetry-cache
-      # skip on Windows because `Compile protos` hangs
-      if: matrix.os != 'windows-latest'
-      with:
-        path: ${{ steps.poetry-install.outputs.cache-dir }}
-        key: ${{ runner.os }}-py${{ matrix.python }}-pypoetry-${{ hashFiles('poetry.lock') }}
     - name: Install Poetry dependencies (with rocksdb)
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):
For me it's okay to disable the cache. When you said that we add 3 minutes to the build time, it's 3 minutes for each of the 7 different platforms, right? So we add 21 minutes to the full check. (to run the 7 checks in this PR it took 55 minutes)

I was thinking about having all 7 checks only before merging from dev to master and inside PRs we could have 3 or 4, not sure.

But anyway I am okay with moving forward.

jansegre:
> For me it's okay to disable the cache. When you said that we add 3 minutes to the build time, it's 3 minutes for each of the 7 different platforms, right? So we add 21 minutes to the full check. (to run the 7 checks in this PR it took 55 minutes)

Maybe there is a limit, but all of the runs on the build matrix should run in parallel. So it should be +3min to the total duration and +21min to the "build minutes used" (which I'm not sure is limited for open source projects).

> I was thinking about having all 7 checks only before merging from dev to master and inside PRs we could have 3 or 4, not sure.

We only require 2 of those to pass (python-3.7-linux and python-3.7-macos). But if any other fails before they will be cancelled and therefore fail too. Maybe we could change this to let all of them run until the end, so that requirement of only 2 actually makes sense.

msbrogli (approved):

msbrogli:
I agree that we can remove the cache. At the end of the day, it doesn't save considerable time.

PR #181, opened 2021-02-05
fix: when the input spends a HTR utxo, the token_data should always be 0

pedroferreira1:


jansegre (approved):

hathor/transaction/resources/transaction.py
```diff
@@ -67,17 +67,21 @@ def get_tx_extra_data(tx: BaseTransaction) -> Dict[str, Any]:
...
+                for out in tx.outputs:
+                    out_token_uid = tx.get_token_uid(out.get_token_index())
+                    if out_token_uid == token_uid:
+                        output['decoded']['token_data'] = out.token_data
+                        break
+                else:
+                    # This is the case when the token from the input does not appear in the outputs
+                    # This case can happen when we have a full melt, so all tokens from the inputs are destroyed
+                    # So we manually add this token to the array and set the token_data properly
+                    serialized['tokens'].append(token_uid.hex())
```
msbrogli, line 83: What if the token from inputs 0 and 1 do not appear in the outputs? I feel we will duplicate the token in the `serialized['tokens']` for no reason, and it is not supposed to happen.

Why do we search for the token index in the outputs? Isn't it better to search in the `serialized['tokens']` instead?

hathor/transaction/resources/transaction.py
```diff
@@ -67,17 +67,21 @@ def get_tx_extra_data(tx: BaseTransaction) -> Dict[str, Any]:
...
+            if token_uid == settings.HATHOR_TOKEN_UID:
+                # If it's an HTR output, the token_data is 0
+                output['decoded']['token_data'] = 0
             else:
-                # This is the case when the token from the input does not appear in the outputs
-                # This case can happen when we have a full melt, so all tokens from the inputs are destroyed
-                # So we manually add this token to the array and set the token_data properly
-                serialized['tokens'].append(token_uid.hex())
-                output['decoded']['token_data'] = len(serialized['tokens'])
+                for out in tx.outputs:
```
msbrogli, line 74: I don't like searching for each input when we can easily index in a preprocessor. Why don't we create a `Dict[bytes, int]` that maps `token_uid -> token_index`?

msbrogli (changes requested):

hathor/transaction/resources/transaction.py
```diff
@@ -67,17 +67,21 @@ def get_tx_extra_data(tx: BaseTransaction) -> Dict[str, Any]:
 
             # We need to get the token_data from the current tx, and not the tx being spent
             token_uid = tx2.get_token_uid(tx2_out.get_token_index())
```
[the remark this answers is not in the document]
pedroferreira1: I don't really understand the situation you described. The `token_data` is useful to know the token this output references.

If `token_data = 129` => `idx = 0b01111111 & 129 = 1`
If `token_data = 1` => `idx = 0b01111111 & 1 = 1`

So if I understood correctly, both cases reference the same token in the tokens array, so it's correct. But as I said previously, I might have misunderstood.

hathor/transaction/resources/transaction.py
pedroferreira1: It makes sense! I've fixed this and refactored the code as you suggested in the other comment below in this commit: https://github.com/HathorNetwork/hathor-core/pull/181/commits/7ca07bb1f8d4729989e286d4b6d045ea5d2667c3

What do you think?

hathor/transaction/resources/transaction.py
```diff
@@ -54,6 +54,9 @@ def get_tx_extra_data(tx: BaseTransaction) -> Dict[str, Any]:
                     spent_outputs[index] = spent_tx.hash_hex
                     break
 
+    # Maps the token uid to the token_data value
+    token_uid_map: Dict[bytes, int] = {settings.HATHOR_TOKEN_UID: 0}
```
msbrogli, line 58: I feel you're not updating this map in your loop.

msbrogli (changes requested):

hathor/transaction/resources/transaction.py
pedroferreira1: Done, thanks

hathor/transaction/resources/transaction.py
msbrogli: Great! The new code can be a bit simplified (removing duplicate code) if you first check `if token_uid not in token_uid_map`, then you update the token_uid_map if necessary, and finally you do `output['decoded']['token_data'] = token_uid_map[token_uid]`.

hathor/transaction/resources/transaction.py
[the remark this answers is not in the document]
msbrogli: The new algorithm fixes it. Thanks!

msbrogli (approved):

hathor/transaction/resources/transaction.py
pedroferreira1: Refactor done

PR #192, opened 2021-04-21
feat(tx): Add verbose mode to tx.verify()

msbrogli:


msbrogli:
@jansegre @pedroferreira1 Should we print or log the outputs? We could also collect each log in a list of strings passed to the first call (e.g. `lines = []; tx.verify(debug=lines); print(lines)`).

hathor/transaction/block.py
```diff
@@ -143,7 +143,7 @@ def get_block_parent(self) -> 'Block':
         assert isinstance(block_parent, Block)
         return block_parent
 
-    def get_funds_fields_from_struct(self, buf: bytes) -> bytes:
+    def get_funds_fields_from_struct(self, buf: bytes, *, verbose: bool = False) -> bytes:
```
pedroferreira1, line 146: We should use this verbose in `TxOutput.create_from_bytes`, right?

And also log `self.version` and `outputs_len`

hathor/transaction/block.py
```diff
@@ -162,7 +162,7 @@ def get_funds_fields_from_struct(self, buf: bytes) -> bytes:
 
         return buf
 
-    def get_graph_fields_from_struct(self, buf: bytes) -> bytes:
+    def get_graph_fields_from_struct(self, buf: bytes, *, verbose: bool = False) -> bytes:
```
pedroferreira1, line 165: Here also we should use `verbose` in `get_graph_fields_from_struct`, right?

And also log `self.data`.

hathor/transaction/transaction.py
```diff
@@ -128,10 +128,10 @@ def create_from_proto(cls, tx_proto: protos.BaseTransaction,
         return tx
 
     @classmethod
-    def create_from_struct(cls, struct_bytes: bytes,
-                           storage: Optional['TransactionStorage'] = None) -> 'Transaction':
+    def create_from_struct(cls, struct_bytes: bytes, storage: Optional['TransactionStorage'] = None,
+                           *, verbose: bool = False) -> 'Transaction':
```
pedroferreira1, line 132: Should we also log `nonce` here when `verbose=True`?

pedroferreira1 (changes requested):

pedroferreira1:
> @jansegre @pedroferreira1 Should we print or log the outputs? We could also collect each log in a list of strings passed to the first call (e.g. `lines = []; tx.verify(debug=lines); print(lines)`).

As far as I understand, the idea of this PR is to debug print when we are handling transactions in shell, right? If that's the case I think keeping the print is enough.

However, I think we could add a parameter in the cli `--verbose` to set the default of the methods, in order to see the verbosity in the logs (in that case, the logs are better). If we do this, it would be also great to have an API (or another way) to temporarily change this verbosity configuration but I guess this could be done in the future.

hathor/transaction/block.py
msbrogli: Fixed. Thanks!

hathor/transaction/block.py
msbrogli: Fixed. Thanks!

hathor/transaction/transaction.py
msbrogli: Fixed. Thanks!

msbrogli:
> > @jansegre @pedroferreira1 Should we print or log the outputs? We could also collect each log in a list of strings passed to the first call (e.g. `lines = []; tx.verify(debug=lines); print(lines)`).
> 
> As far as I understand, the idea of this PR is to debug print when we are handling transactions in shell, right? If that's the case I think keeping the print is enough.
> 
> However, I think we could add a parameter in the cli `--verbose` to set the default of the methods, in order to see the verbosity in the logs (in that case, the logs are better). If we do this, it would be also great to have an API (or another way) to temporarily change this verbosity configuration but I guess this could be done in the future.

I got it. But, on the other hand, we could send it as part of the `/decode-tx` API. In this case, appending to a list is easier than capturing stdout for a while.

jansegre (approved):

pedroferreira1 (approved):

PR #193, opened 2021-04-21
chore(p2p): Improve p2p logging

msbrogli:


pedroferreira1 (approved):

hathor/p2p/protocol.py
```diff
@@ -182,11 +182,40 @@ def change_state(self, state_enum: PeerState) -> None:
             if self.state:
                 self.state.on_enter()
 
+    def is_state(self, state_enum: PeerState) -> bool:
+        return isinstance(self.state, state_enum.value)
+
+    def get_short_remote(self) -> str:
```
jansegre, line 188: I initially liked this strategy, but after understanding that it needs `update_log_context` to be called every time any information used to generate this short description changes, I think it's weird.

I'd lean more to using separate properties even if it gets a little more verbose, our logs are structured enough that we can start forwarding them to a central logger, and being able to filter by `short_peer_id` without having to worry that the param is mixed with `host/port` or the connection state would be a big plus. In this case we would just keep binding these properties as soon as they're available:

```
self.log = self.log.bind(remote_id=self.get_short_peer_id())
# ...
self.log = self.log.bind(remote_host=f'{host}:{port}')
# ...
self.log = self.log.bind(connected=self.connected)  # either 1/0, or True/False
```

What do you think?

hathor/p2p/states/hello.py
```diff
@@ -10,12 +12,15 @@
...
 
+logger = get_logger()
+
 settings = HathorSettings()
 
 
 class HelloState(BaseState):
     def __init__(self, protocol: 'HathorProtocol') -> None:
         super().__init__(protocol)
+        self.log = logger.new(remote=protocol.get_short_remote())
```
jansegre, line 23: Shouldn't it be enough on `BaseState`?

hathor/p2p/states/peer_id.py
```diff
@@ -11,12 +12,15 @@
...
 
+logger = get_logger()
+
 settings = HathorSettings()
 
 
 class PeerIdState(BaseState):
     def __init__(self, protocol: 'HathorProtocol') -> None:
         super().__init__(protocol)
+        self.log = logger.new(remote=protocol.get_short_remote())
```
jansegre, line 23: Same here.

hathor/p2p/states/ready.py
```diff
@@ -21,6 +24,7 @@ class ReadyState(BaseState):
 
     def __init__(self, protocol: 'HathorProtocol') -> None:
         super().__init__(protocol)
+        self.log = logger.new(remote=protocol.get_short_remote())
```
jansegre, line 27: And also here.

hathor/p2p/protocol.py
msbrogli: I agree. I'll do that. Thanks!

hathor/p2p/states/hello.py
msbrogli: I did it but it was logging as `BaseState` instead of each class. Maybe I did something wrong.

hathor/p2p/protocol.py
msbrogli: Fixed. Can you review again, please?

jansegre: Much better!

jansegre (approved):

PR #194, opened 2021-04-21
feat(p2p): Add netfilter structure for advanced p2p firewall

msbrogli:


hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,133 @@
...
+        raise NotImplementedError
+
+
+class NetfilterMatchAll(NetfilterMatch):
+    def match(self, context: 'NetfilterContext') -> bool:
+        return True
+
+
+class NetfilterMatchAnd(NetfilterMatch):
+    def __init__(self, a, b):
```
pedroferreira1, line 19: For what I understood in the `match` method, `a` and `b` are `NetfilterMatch`, right?

jansegre: Yes, I think it's just missing type annotations.

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,136 @@
...
+    def match(self, context: 'NetfilterContext') -> bool:
+        if not self.a.match(context):
+            return False
+        if not self.b.match(context):
+            return False
+        return True
+
+
+class NetfilterMatchOr(NetfilterMatch):
+    def __init__(self, a, b):
```
jansegre, line 35: Missing type annotations.

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,136 @@
...
+
+        host = ip_network(context.addr.host)
+        if not self.host.supernet_of(host):
+            return False
+
+        return True
+
+
+class NetfilterMatchPeerId(NetfilterMatch):
+    def __init__(self, peer_id):
```
jansegre, line 66: Missing type annotation.

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,136 @@
...
+            return False
+
+        if context.protocol.peer.id != self.peer_id:
+            return False
+
+        return True
+
+
+class NetfilterMatchPeerIdRemoteURL(NetfilterMatch):
+    def __init__(self, name, manager, url, update_interval=30) -> None:
```
jansegre, line 83: Missing type annotations.

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,136 @@
...
+        self.url = url
+        self.update_interval = update_interval
+        self.lc_update = LoopingCall(self.update)
+        self.items = set()
+        self.method = 'GET'
+        self.headers = {
+            'User-Agent': ['hathor-core'],
+        }
+
+    def start(self):
```
jansegre, line 96: These should also be annotated with `-> None`.

hathor/p2p/states/hello.py
```diff
@@ -102,4 +102,17 @@ def handle_hello(self, payload: str) -> None:
...
+        from hathor.p2p.netfilter import get_table
+        from hathor.p2p.netfilter.context import NetfilterContext
+        context = NetfilterContext(
+            protocol=self.protocol,
+            connections=self.protocol.connections,
+            addr=self.protocol.transport.getPeer(),
+        )
+        verdict = get_table('filter').get_chain('post_hello').process(context)
+        if not bool(verdict):
+            self.protocol.close_connection()
```
jansegre, line 115: Should we have a `reason` on `close_connection` so we can better debug why the connection was closed?

hathor/p2p/netfilter/__init__.py
```diff
@@ -0,0 +1,5 @@
+from hathor.p2p.netfilter.instances import get_table
+
+__all__ = [
+    'get_table',
+]
```
jansegre, line 5: Should we add some simple tests? (This has nothing to do with this code location)

hathor/p2p/netfilter/table.py
```diff
@@ -0,0 +1,20 @@
...
+    from hathor.p2p.netfilter.chain import NetfilterChain
+
+
+class NetfilterTable:
+    def __init__(self, name: str):
+        self.name = name
+        self.chains: Dict[str, 'NetfilterChain'] = {}
+
+    def add_chain(self, chain: 'NetfilterChain') -> 'NetfilterChain':
+        self.chains[chain.name] = chain
```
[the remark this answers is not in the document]
msbrogli: Done. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: You're correct, @pedroferreira1. Just added the type annotations. Thanks!

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,136 @@
...
+
+from twisted.internet.defer import Deferred
+from twisted.internet.task import LoopingCall
+
+if TYPE_CHECKING:
+    from hathor.p2p.netfilter.context import NetfilterContext
+
+
+class NetfilterMatch:
+    def match(self, context: 'NetfilterContext') -> bool:
```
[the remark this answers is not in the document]
msbrogli: Done. Thanks!

hathor/p2p/netfilter/chain.py
```diff
@@ -0,0 +1,31 @@
+
```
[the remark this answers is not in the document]
msbrogli: Done. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: Done. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: Done. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: Done. Thanks!

hathor/p2p/netfilter/__init__.py
msbrogli: Surely we should. I'll do it. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: Done. Thanks!

hathor/p2p/netfilter/targets.py
```diff
@@ -0,0 +1,40 @@
...
+
+class NetfilterTarget:
+    terminate: bool
+
+    def execute(self, rule: 'NetfilterRule', context: 'NetfilterContext') -> None:
+        pass
+
+
+class NetfilterAccept(NetfilterTarget):
+    terminate = True
```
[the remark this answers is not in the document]
msbrogli: Terminate means we should finish evaluating and the final verdict is to accept.

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+class NetfilterMatchAll(NetfilterMatch):
+    """Always match."""
+    def match(self, context: 'NetfilterContext') -> bool:
+        return True
+
+
+class NetfilterMatchAnd(NetfilterMatch):
+    """Logic AND operation for two matches. When the first fails, the second is not executed."""
```
giovanecosta, line 43: Didn't understand this part. Doesn't Python Short-Circuit Evaluation enough to prevent second statement from be executed?

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,171 @@
...
+        if context.protocol.peer is None:
+            return False
+
+        if context.protocol.peer.id != self.peer_id:
+            return False
+
+        return True
+
+
+class NetfilterMatchPeerIdRemoteURL(NetfilterMatch):
```
giovanecosta, line 115: No `match` method for this one?

hathor/p2p/netfilter/matches.py
```diff
@@ -0,0 +1,177 @@
...
+        try:
+            host = ip_network(context.addr.host)
+        except ValueError:
+            return False
+
+        if self.host.version != host.version:
+            return False
+
+        if not self.host.supernet_of(host):
+            return False
```
jansegre, line 99: Apparently this is Python 3.7+ only. I don't think we should necessarily accelerate dropping support of Python 3.6 just for this, but at the very least we could leave this commented for a simpler way do this once we do drop 3.6

hathor/p2p/netfilter/__init__.py
msbrogli: Done!

hathor/p2p/netfilter/matches.py
msbrogli: I'm just explaining how the match behaves in this case. It has exactly the same behavior as the python short-circuit evaluation.

hathor/p2p/netfilter/matches.py
msbrogli: Fixed. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: Nice catch! Thanks!

I put an if-else block that run this code for 3.7+ and another code for 3.6. Can you review it agian?

pedroferreira1 (approved):

hathor/p2p/netfilter/matches.py
msbrogli: Even though `overlaps` and `supernet_of` have different semantics, they work the same for us because one host is always `/32`.

I just implemented your suggestion. I changed to `host = ip_address(...)` and used `host not in self.host`.

hathor/p2p/states/hello.py
msbrogli: Done! Can you review it again, please?

jansegre: Looks great!

setup.cfg
```diff
@@ -16,6 +16,9 @@ exclude_lines =
     # Have to re-enable the standard pragma
     pragma: no cover
 
+    # Don't complain about TYPE_CHECKING:
+    if TYPE_CHECKING:
+
```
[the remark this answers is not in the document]
msbrogli: I agree. I'm looking forward to dropping support for 3.6.

PR #198, opened 2021-04-30
fix(websocket): dashboard metrics can stop until node is restarted

jansegre:


hathor/websocket/factory.py
```diff
@@ -105,23 +116,18 @@ def _setup_rate_limit(self):
             self.rate_limiter.set_limit(control_type, config['max_hits'], config['hits_window_seconds'])
             self.buffer_deques[control_type] = deque(maxlen=config['buffer_size'])
 
-    def _schedule_and_send_metric(self):
+    def _send_metrics(self):
         """ Send dashboard metric to websocket and schedule next message
```
pedroferreira1, line 120: Update the docstring. We are not scheduling the next message anymore.

pedroferreira1 (approved):

hathor/websocket/factory.py
jansegre: Done.

hathor/mining/ws.py
```diff
@@ -129,7 +129,12 @@ def send_notification(self, *, method: str, params: Union[List, Dict]) -> None:
         }
         if params:
             request['params'] = params
-        self.sendMessage(json_dumpb(request))
+        try:
+            self.sendMessage(json_dumpb(request))
```
msbrogli, line 133: As we need to capture `Exception`, I think we should move the serialization to outside the try/except block.

hathor/websocket/factory.py
```diff
@@ -86,16 +89,24 @@ def __init__(self, metrics: Optional[Metrics] = None, wallet_index: Optional[Wal
...
+
     def start(self):
         self.is_running = True
 
         # Start limiter
         self._setup_rate_limit()
 
         # Start metric sender
-        self._schedule_and_send_metric()
+        self._lc_send_metrics.start(1, now=False)
```
msbrogli, line 105: Move `1` to a constant (e.g. `LC_SEND_METRICS_INTERVAL`).

msbrogli (changes requested):

hathor/mining/ws.py
jansegre: Actually, I think we might want to capture that too, besides if the serialization fails it should be an implementation error because it does not rely on data that isn't already in a format we can serialize. The point is that if `send_notifications` fails we will miss a send for the rest of the connections on the `broadcast_notification` loop.

hathor/websocket/factory.py
jansegre: Done.

msbrogli (approved):

hathor/mining/ws.py
msbrogli: I agree. Thanks!

PR #199, opened 2021-04-30
chore(ci): use GITHUB_TOKEN for ghcr login

jansegre:
Test run: https://github.com/jansegre/hathor-core/actions/runs/801167847
Some of the generated images:
- `docker run --rm jansegre/hathor-core:experimental-docker-tweaks-python3.8`
- `docker run --rm ghcr.io/jansegre/hathor-core:experimental-docker-tweaks-python3.8`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ aiohttp = "~3.7.3"
 idna = "~2.10"
 # move the following to "build-system.requires" when this poetry pr is merged and released: https://github.com/python-poetry/poetry/pull/2794
 cython = {version = "<0.30", optional = true}
+typing-extensions = "^3.7.4"  # indirect dependency, should not be needed on Python >3.7 but some packages import it anyway
 
 [tool.poetry.extras]
 rocksdb = ["cython", "python-rocksdb"]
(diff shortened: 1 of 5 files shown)


pedroferreira1 (approved):

pyproject.toml
```diff
@@ -75,6 +75,7 @@ aiohttp = "~3.7.3"
 idna = "~2.10"
 # move the following to "build-system.requires" when this poetry pr is merged and released: https://github.com/python-poetry/poetry/pull/2794
 cython = {version = "<0.30", optional = true}
+typing-extensions = "^3.7.4"  # indirect dependency, should not be needed on Python >3.7 but some packages import it anyway
```
msbrogli, line 78: Why do we need this package?

msbrogli (approved):

pyproject.toml
jansegre: We don't directly need that package. That package almost like a polyfill for typing related stuff of built-in modules that were only available on Python >3.7. However some of our dependencies (at least aiohttp, only used by the MM code, but still there might be others, googling leads to the problem appearing on several packages) still try to incorrectly import `typing_extensions` despite it being correctly gated to only be a dependency if Python <= 3.7. Making this dependency explicit here makes it so it is required regardless of the Python version, there is no harm, but eventually we should remove this when all our dependencies have this sorted out.

PR #204, opened 2021-05-01
feat(cli): Set process name

msbrogli:


hathor/cli/run_node.py
```diff
@@ -77,6 +78,9 @@ def prepare(self, args: Namespace) -> None:
         settings = HathorSettings()
         log = logger.new()
 
+        from setproctitle import setproctitle
+        setproctitle('{}hathor-core'.format(args.procname_prefix))
```
pedroferreira1, line 82: This is useful only for debugging or there is another usage?

pedroferreira1 (approved):

jansegre (approved):

hathor/cli/run_node.py
msbrogli: I feel it is easier to find the process when we need. It currently shows a weird cmdline.

PR #206, opened 2021-05-07
feat(simulator): Add a network simulator

msbrogli:


pedroferreira1 (approved):

hathor/simulator/simulator.py
```diff
@@ -0,0 +1,118 @@
...
+            if self.clock.seconds() - latest_time >= status_interval:
+                t1 = time.time()
+                print('[{:8.2f}][rate={:8.2f}] t={:15.2f}    dt={:8.2f}    toBeRun={:8.2f}    delayedCall={}'.format(
+                    t1 - t0,
+                    (self.clock.seconds() - initial) / (t1 - t0),
+                    self.clock.seconds(),
+                    self.clock.seconds() - initial,
+                    interval - self.clock.seconds() + initial,
+                    len(self.clock.getDelayedCalls()),
+                ))
```
jansegre, line 116: When a .format gets this big it's really hard to understand what goes where. An alternative is using variables and an f-string like

```python
foo = ...
print(f'... {foo:fmt} ...')
```
What do you think?

msbrogli: Really better. I also added comments explaining each variable. Can you review again, please?

jansegre (approved):

hathor/simulator/simulator.py
jansegre: Looks great!

PR #217, opened 2021-05-25
feat(sync-v2): base classes and methods, shouldn't break current code

jansegre:


hathor/manager.py
```diff
@@ -120,6 +121,13 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
 
         self.cpu = cpu
 
+        # XXX: first checkpoint must be genesis (height=0)
+        self.checkpoints: List[Checkpoint] = checkpoints or []
+        self.checkpoints_ready: List[bool] = [False] * len(self.checkpoints)
+        if not self.checkpoints or self.checkpoints[0].height > 0:
+            self.checkpoints.insert(0, Checkpoint(0, settings.GENESIS_BLOCK_HASH))
+            self.checkpoints_ready.insert(0, True)
```
msbrogli, line 129: I feel we should either remove `checkpoints_ready` or add:

```python
if ...:
    :
    :
else:
    self.checkpoints_ready[0] = True
```

hathor/manager.py
```diff
@@ -613,37 +624,18 @@ def validate_new_tx(self, tx: BaseTransaction, skip_block_weight_verification: b
...
-                        allowed=tokens_issued_per_block,
-                    )
-                )
+                tx.verify_weight()
+
+            tx.verify_reward()
         else:
+            tx = cast(Transaction, tx)
             assert tx.hash is not None  # XXX: it appears that after casting this assert "casting" is lost
-            assert isinstance(tx, Transaction)
```
msbrogli, line 637: Do the `cast(...)` check whether `isinstance(tx, Transaction)`? I guess it doesn't and we should keep this assert.

hathor/manager.py
```diff
@@ -776,3 +768,17 @@ class ParentTxs(NamedTuple):
     max_timestamp: int
     can_include: List[bytes]
     must_include: List[bytes]
+
+    def get_random_parents(self) -> Tuple[bytes, bytes]:
+        """ Get parents from self.parents plus a random choice from self.parents_any to make it 3 in total.
+
+        Return type is tuple just to make it clear that the length is always 3.
```
msbrogli, line 775: 3 or 2? The code seems to return always two.

hathor/manager.py
jansegre: I'd rather add the `else`. Otherwise the same goes `self.checkpoints`, not being used.

hathor/manager.py
jansegre: No, `cast` bypasses any type checks.

hathor/transaction/base_transaction.py
```diff
@@ -424,8 +425,81 @@ def get_struct(self) -> bytes:
         struct_bytes += self.get_struct_nonce()
         return struct_bytes
 
+    def get_all_dependencies(self) -> Set[bytes]:
+        """Set of all tx-hashes needed to fully validate this tx, including parent blocks/txs and inputs."""
+        from hathor.transaction.genesis import GENESIS_HASHES
```
msbrogli, line 430: Why do we remove the genesis? I feel we should return them just like any other dependency.

hathor/transaction/transaction.py
```diff
@@ -260,6 +263,33 @@ def to_json(self, decode_script: bool = False, include_metadata: bool = False) -
         json['tokens'] = [h.hex() for h in self.tokens]
         return json
 
+    def verify_basic(self, skip_block_weight_verification: bool = False) -> None:
```
msbrogli, line 266: I guess we should add a docstring explaining what a basic validation means for a transaction.

hathor/transaction/transaction.py
```diff
@@ -260,6 +263,33 @@ def to_json(self, decode_script: bool = False, include_metadata: bool = False) -
...
+        if self.is_genesis:
+            # TODO do genesis validation?
+            return
+        self.verify_without_storage()
+        self.verify_parents_basic()
+        self.verify_weight()
+
+    def verify_parents_basic(self) -> None:
```
msbrogli, line 274: Shouldn't it be part of the `verify_without_storage()`?

hathor/transaction/transaction.py
```diff
@@ -260,6 +263,33 @@ def to_json(self, decode_script: bool = False, include_metadata: bool = False) -
...
+
+        # check if parents are duplicated
+        parents_set = set(self.parents)
+        if len(self.parents) > len(parents_set):
+            raise DuplicatedParents('Tx has duplicated parents: {}', [tx_hash.hex() for tx_hash in self.parents])
+
+        if len(self.parents) != 2:
+            raise IncorrectParents(f'wrong number of parents (tx type): {len(self.parents)}, expecting 2')
+
+    def verify_weight(self) -> None:
```
msbrogli, line 286: Same here: should it be part of the `verify_without_storage()`?

hathor/transaction/resources/tips.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+from typing import Optional
```
msbrogli, line 16: I feel we can remove this resource.

tests/resources/transaction/test_mining.py
```diff
@@ -30,6 +30,7 @@ def test_get_block_template_with_address(self):
                 'conflict_with': [],
                 'voided_by': [],
                 'twins': [],
+                'validation': 'initial',  # FIXME: change to 'full' when validations are enabled
```
msbrogli, line 33: Will this test fail when validations are enabled?

tests/utils.py
```diff
@@ -160,7 +160,7 @@ def add_new_block(manager, advance_clock=None, *, parent_block_hash=None, data=b
     if weight is not None:
         block.weight = weight
     block.resolve()
-    block.verify()
+    block.validate_full()
```
msbrogli, line 163: Isn't `verify()` and `validate_full()` the same? If not, where do they differ?

hathor/transaction/transaction_metadata.py
```diff
@@ -21,7 +22,73 @@
...
+    blocks the block parent has to exist and be at least BASIC) have been run. For example, if it's `BASIC` the weight
+    of a tx has been validated and is correct, but it may be spending a tx that has already been spent, we will not run
+    this validation until _all_ the dependencies have reached `FULL` or any of them `INVALID` (which should
+    automatically invalidate this tx). In theory it should be possible to have even more granular validation (if one of
+    the inputs exists, validate that we can spend it), but the complexity for that is too high.
+
+    """
+    INITIAL = 0  # aka, not validated
+    BASIC = 1  # only graph info has been validated
+    CHECKPOINT = 2  # validation can be safely assumed because it traces up to a known checkpoint
```
msbrogli, line 57: As long as the hash matches, right?

hathor/transaction/base_transaction.py
```diff
@@ -424,8 +425,81 @@ def get_struct(self) -> bytes:
...
+        """Set of parent tx hashes, typically used for syncing transactions."""
+        from hathor.transaction.genesis import GENESIS_HASHES
+        return set(self.parents[1:] if self.is_block else self.parents) - set(GENESIS_HASHES)
+
+    def can_validate_full(self) -> bool:
+        """ Check if this transaction is ready to be fully validated, either all deps are full-valid or one is invalid.
+        """
+        assert self.storage is not None
+        deps = self.get_all_dependencies()
+        return (
```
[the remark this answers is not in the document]
jansegre: Using for loops is annoying because it adds a lot of breaks to the flow, I could pull an updated version of this implementation that breaks this up into more methods which makes it much better to use for loops.

hathor/transaction/transaction_metadata.py
```diff
@@ -253,6 +341,9 @@ def create_from_proto(cls, hash_bytes: bytes, metadata_proto: protos.Metadata) -
         metadata.score = metadata_proto.score
         metadata.first_block = metadata_proto.first_block or None
         metadata.height = metadata_proto.height
+        metadata.validation = ValidationState(metadata_proto.validation)
+        if is_genesis(hash_bytes):
+            metadata.validation = ValidationState.FULL
```
msbrogli, line 346: Isn't it already saved in the storage?

hathor/transaction/transaction_metadata.py
```diff
@@ -36,7 +103,9 @@ class TransactionMetadata:
     accumulated_weight: float
     score: float
     first_block: Optional[bytes]
+    first_block_parent: Optional[bytes]
```
msbrogli, line 106: What is it for? I feel we should remove it now because it's not being used anywhere in the PR.

hathor/manager.py
msbrogli: Ok then! Even though I don't feel this information will be stored the manager, we can keep it here and refactor in the next PR.

hathor/transaction/base_transaction.py
```diff
@@ -424,8 +425,81 @@ def get_struct(self) -> bytes:
...
+    def get_tx_dependencies(self) -> Set[bytes]:
+        """Set of all tx-hashes needed to fully validate this, except for block parent, i.e. only tx parents/inputs."""
+        from hathor.transaction.genesis import GENESIS_HASHES
+        parents = self.parents[1:] if self.is_block else self.parents
+        return set(chain(parents, (i.tx_id for i in self.inputs))) - set(GENESIS_HASHES)
+
+    def get_tx_parents(self) -> Set[bytes]:
+        """Set of parent tx hashes, typically used for syncing transactions."""
+        from hathor.transaction.genesis import GENESIS_HASHES
+        return set(self.parents[1:] if self.is_block else self.parents) - set(GENESIS_HASHES)
```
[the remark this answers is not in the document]
jansegre: I'll look into it on the other branches. It's something like where these are being used the genesis would have to be removed, so to remove this from all callers I pulled it into these methods. But I want to confirm this and it might be possible the callers won't have a problem of being returned a genesis tx, or would be better to handle it themselves.

hathor/transaction/block.py
```diff
@@ -235,6 +273,38 @@ def to_json_extended(self) -> Dict[str, Any]:
...
+        """Whether all block parent is in storage and is at least basic-valid."""
+        assert self.storage is not None
+        parent_block_hash = self.parents[0]
+        if not self.storage.transaction_exists(parent_block_hash):
+            return False
+        metadata = self.storage.get_metadata(parent_block_hash)
+        assert metadata is not None
+        return metadata.validation.is_at_least_basic()
+
+    def verify_basic(self, skip_block_weight_verification: bool = False) -> None:
```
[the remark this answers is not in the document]
jansegre: I agree.

hathor/transaction/transaction.py
jansegre: Maybe, these verifications already exist in other forms outside of the usual `verify` chain. They are integrated into the usual `verify` chain on the next branches, but I'm not sure if I can pull that integration yet (there are some restrictions on when some data is available, or on the order that I can verify transacions). I have to confirm if I can pull these changes.

tests/utils.py
jansegre: `verify` does not write on the database, `validate_full` will update (and save?) the `metadata.validation` state

hathor/transaction/transaction_metadata.py
jansegre: Yes, I tried to keep the comment short, is so long already.

hathor/transaction/transaction_metadata.py
jansegre: Not on an outdated storage.

hathor/transaction/transaction_metadata.py
jansegre: Yes, I'll remove this for now.

hathor/transaction/transaction_metadata.py
msbrogli: Just double checking. Thanks!

hathor/manager.py
jansegre: Done.

hathor/manager.py
jansegre: Done.

hathor/transaction/base_transaction.py
```diff
@@ -424,8 +425,81 @@ def get_struct(self) -> bytes:
...
         return struct_bytes
 
+    def get_all_dependencies(self) -> Set[bytes]:
+        """Set of all tx-hashes needed to fully validate this tx, including parent blocks/txs and inputs."""
+        from hathor.transaction.genesis import GENESIS_HASHES
+        return set(chain(self.parents, (i.tx_id for i in self.inputs))) - set(GENESIS_HASHES)
+
+    def get_tx_dependencies(self) -> Set[bytes]:
+        """Set of all tx-hashes needed to fully validate this, except for block parent, i.e. only tx parents/inputs."""
+        from hathor.transaction.genesis import GENESIS_HASHES
```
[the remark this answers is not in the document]
jansegre: Done.

hathor/transaction/base_transaction.py
jansegre: Done.

msbrogli (approved):

hathor/transaction/transaction.py
jansegre: Done.

hathor/transaction/resources/tips.py
jansegre: Done.

hathor/transaction/transaction_metadata.py
jansegre: Done.

hathor/transaction/transaction.py
jansegre: For blocks it actually needs storage.

jansegre: Makes sense.

hathor/transaction/transaction.py
jansegre: I checked it out, and it makes sense.

jansegre: I tried moving it and a bunch of tests broke. While I think the move makes sense and it might actually indicate something wrong on the tests, at this point I rather do this on a new PR.

hathor/transaction/transaction.py
jansegre: I tried moving it and a bunch of tests broke. While I think the move makes sense and it might actually indicate something wrong on the tests, at this point I rather do this on a new PR.

tests/resources/transaction/test_mining.py
jansegre: No. And actually I think 'full' wouldn't make any sense for a block that wasn't even mined yet. I'll remove that comment.

tests/utils.py
jansegre: Also, `validate_full` is _supposed_ to skip some verifications if `validate_basic` has been run before, and also `validate_full` can `skip_block_weight_verification`, while `verify` cannot. ("_supposed_" but it calls `.verify()` internally instead of only running a `verify_post_basic` or something like that)

I considered renaming the `verify` family of methods like we talked but it's too widespread and will result in a very large (although simple) refactor.

I also tried unifying `verify` and `validate_full` but because `validate_full` makes some assumptions about the metadata and also runs `verify_basic` before `verify`.

Either way, I'm moving this to an issue and so I can merge this PR. I completely agree we should improve how this is being handled to minimize any confusion.

PR #218, opened 2021-05-26
fix(ws): Fix exception raised when a connection is closed before finishing handshake

msbrogli:


hathor/websocket/factory.py
```diff
@@ -208,6 +210,9 @@ def execute_send(self, data: Dict[str, Any], connections: Set[HathorAdminWebsock
         for c in connections:
             try:
                 c.sendMessage(payload, False)
+            except Disconnected:
```
pedroferreira1, line 213: When this can happen? Race condition?

In the PR title is written 'Fix exception raised when a connection is closed before finishing handshake' but the `on_client_open` says that the connection is added to the set only after finishing handshake ('Called when a ws connection is opened (after handshaking).').

I just want to understand if this except block 'should never happen'. In that case I think we should add at least a debug/warn log.

pedroferreira1 (approved):

hathor/websocket/factory.py
msbrogli: The fix is really to change from `self.connections.remove(connection)` to `self.connections.discard(connection)`. The exception was caused by trying to remove from `self.connections` a connection that was closed before finishing handshaking. https://github.com/HathorNetwork/hathor-core/pull/218/files#diff-ed970ec39d116754ce09130dcc2370c3f41d93620a0234d588d0639687e96d64R351

The remaining of the changes is a minor refactor because the protocol was directly changing `factory.connections` and I moved this code a method in the factory itself.

I'm not sure when the `Disconnect` situation might happen. I feel that it should be a race condition. I made this change because we don't want to log if the problem was in a connection that was closed.

PR #219, opened 2021-05-26
fix(p2p): Stop calling sync_manager.stop() twice

msbrogli:


hathor/p2p/protocol.py
```diff
@@ -82,7 +82,7 @@ class WarningFlags(str, Enum):
     connection_string: Optional[str]
     expected_peer_id: Optional[str]
     warning_flags: Set[str]
-    connected: bool
+    is_connected: bool
```
jansegre, line 85: We have discussed this parameter before and it was renamed to `aborting` (flipping the boolean value) on one of the sync-v2 branches. The main difference being that `aborting` starts with `False` which is not the inverse of `connected`, which also starts with `False`, so this property is only changed when we decide to close a connection, and not when it is connected. Which from what I remember from what we discussed it is best to leave the state of the connection to twisted.

jansegre:
I'll take a better look later and fuller review later today. There is some refactoring mixed with the actual fix, which isn't exactly a problem but will clash with changes on sync-v2. My first thought would be that there _shouldn't_ be any harm calling stop twice, so instead of raising an exception we should log an error and then independently we would actually fix the issue that was causing it to be called twice. Maybe this happens on other classes that have start/stop methods, should we always raise an exception?

msbrogli:
> I'll take a better look later and fuller review later today. There is some refactoring mixed with the actual fix, which isn't exactly a problem but will clash with changes on sync-v2. My first thought would be that there _shouldn't_ be any harm calling stop twice, so instead of raising an exception we should log an error and then independently we would actually fix the issue that was causing it to be called twice. Maybe this happens on other classes that have start/stop methods, should we always raise an exception?

Letting start/stop execute twice is not safe because of initializations, cleanups, and other actions that are not idempotent. So, we should at least protect it from happening.

I'm not sure what's the best behavior when start/stop calls are unbalanced. I feel that raising an exception is the best way to signal that something is wrong. In some cases, we should allow only one call to start/stop methods.

hathor/p2p/protocol.py
msbrogli: I had to rename it because `connected` is used by the parent class and we do not have full control over it. But I agree that `aborting` is a better way to deal with it.

jansegre:
> Letting start/stop execute twice is not safe because of initializations, cleanups, and other actions that are not idempotent. So, we should at least protect it from happening.
> 
> I'm not sure what's the best behavior when start/stop calls are unbalanced. I feel that raising an exception is the best way to signal that something is wrong. In some cases, we should allow only one call to start/stop methods.

I understand, I'm only worried about `stop` though. In case there are bugs that call stop twice, should it blow up in other parts of the code? Should we try to make idempotent `stop` methods (simply logging an error "already stopped" and not executing anything is an idempotent function) or should we try to move some of these actions to `__del__` which is guaranteed to be called only once (at least on CPython).

hathor/p2p/protocol.py
jansegre: These are the changes: https://github.com/HathorNetwork/hathor-core/pull/217/files#diff-e0be12551dd55b1003c020ef234d6a44a5126754c055b1a96fcdb3900b5be622 if you want to move them to this PR it will avoid renaming it twice in a row.

hathor/p2p/states/ready.py
```diff
@@ -88,24 +78,23 @@ def on_enter(self) -> None:
...
 
-        for plugin in self.plugins.values():
-            plugin.stop()
+        if self.sync_manager.is_started:
+            self.sync_manager.stop()
 
-    def get_sync_plugin(self) -> NodeSyncTimestamp:
-        return cast(NodeSyncTimestamp, self.plugins[self.SYNC_PLUGIN_NAME])
+    def prepare_to_disconnect(self) -> None:
+        self.sync_manager.stop()
```
pedroferreira1, line 91: Is it safe enough to assume `sync_manager` will be started here? Or should we add the same `if` as in the `on_exit` method? `if self.sync_manager.is_started:`

pedroferreira1 (approved):

msbrogli:
> > Letting start/stop execute twice is not safe because of initializations, cleanups, and other actions that are not idempotent. So, we should at least protect it from happening.
> > I'm not sure what's the best behavior when start/stop calls are unbalanced. I feel that raising an exception is the best way to signal that something is wrong. In some cases, we should allow only one call to start/stop methods.
> 
> I understand, I'm only worried about `stop` though. In case there are bugs that call stop twice, should it blow up in other parts of the code? Should we try to make idempotent `stop` methods (simply logging an error "already stopped" and not executing anything is an idempotent function) or should we try to move some of these actions to `__del__` which is guaranteed to be called only once (at least on CPython).

I don't feel we should use the `__del__` method. But I agree we can easily make these methods idempotent just doing nothing. Something like "if you'd like to stop it and it is already stopped, that's good". I guess I like this behavior except by the idea that raising an exception might help us to identify flow/logic problems easily. What do you think?

I understand the downside of raising an exception because we can break things that were working nicely before. But I feel we should put more and more protections to avoid logic mistakes.

jansegre (approved):

PR #220, opened 2021-05-26
feat(sentry): Add support to publish exception events in Sentry

msbrogli:


jansegre (approved):

hathor/cli/run_node.py
```diff
@@ -69,6 +69,7 @@ def create_parser(self) -> ArgumentParser:
                             'This is still a beta feature as it may cause issues when restarting the full node '
                             'after a crash.')
         parser.add_argument('--procname-prefix', help='Add a prefix to the process name', default='')
+        parser.add_argument('--sentry', help='Publish exception events to Sentry.')
```
pedroferreira1, line 72: I prefer being more explicit in the help text saying that we 'Expect the Sentry DSN as parameter.'.

hathor/cli/run_node.py
```diff
@@ -232,10 +233,38 @@ def create_wallet():
...
+            return
+        self.log.info('Starting Sentry', dsn=args.sentry)
+        try:
+            import sentry_sdk
+        except ModuleNotFoundError:
+            self.log.error('ERROR: You must install the package \'sentry-sdk\' to enable sentry.')
+            sys.exit(-3)
+
+        import hathor
+        sentry_sdk.init(
```
pedroferreira1, line 251: Can we have more information about the hostname/entrypoint here? It would be good to know which node generated the exception.

pedroferreira1 (changes requested):

hathor/cli/run_node.py
jansegre: I'm not too familiar with sentry, so maybe we should call it `--sentry-dsn` then.

hathor/cli/run_node.py
luislhl: Maybe this is the way: https://github.com/kiwicom/structlog-sentry#sentry-tags

hathor/cli/run_node.py
```diff
@@ -232,10 +233,41 @@ def create_wallet():
...
+        )
+
+        import logging
+
+        import structlog
+        from structlog_sentry import SentryProcessor
+        processors = structlog.get_config()['processors']
+        processors.append(SentryProcessor(level=logging.ERROR))
+        structlog.configure(
+            processors=processors
```
luislhl, line 266: Are we able to send the exception StackTrace to Sentry with this setup?

By reading this I was left with the impression that maybe we won't be sending important information to Sentry, like the StackTrace, since we are putting SentryProcessor after the `format_exc_info` processor.
> When you want to use structlog's built-in format_exc_info processor, make that the SentryProcessor comes before format_exc_info! Otherwise, the SentryProcessor won't have an exc_info to work with, because it's removed from the event by format_exc_info.

https://github.com/kiwicom/structlog-sentry#usage

hathor/cli/run_node.py
msbrogli: Done!

hathor/cli/run_node.py
msbrogli: Nice catch. It's fixed but it can't capture the stack trace in some cases (e.g. exception inside a deferred).

hathor/cli/run_node.py
msbrogli: I can set the `server_name` to the peer-id. Otherwise, it will send the actual hostname. I feel that the actual hostname might be more useful if we configure it properly during deploy. What do you think?

hathor/cli/run_node.py
```diff
@@ -254,10 +255,31 @@ def create_wallet():
...
-    def start_manager(self) -> None:
+    def start_sentry_if_possible(self, args: Namespace) -> None:
+        """Start Sentry integration if possible."""
+        if not args.sentry_dsn:
+            return
+        self.log.info('Starting Sentry', dsn=args.sentry_dsn)
+        try:
+            import sentry_sdk
+        except ModuleNotFoundError:
+            self.log.error('ERROR: You must install the package \'sentry-sdk\' to enable sentry.')
```
jansegre, line 269: I think this might be a bit confusing because `structlog-sentry` is also needed. But rather than making those dependencies explicit, I prefer to point to the `-E sentry` extras. And also I'm trying to avoid "ERROR" / "WARN" inside log messages.

I suggest something like:

```python
self.log.error('Please use `poetry install -E sentry` for enabling Sentry.')
```

hathor/cli/util.py
```diff
@@ -212,18 +213,30 @@ def kwargs_formatter(_, __, event_dict):
...
+        structlog.stdlib.add_logger_name,
+        structlog.stdlib.add_log_level,
+    ]
+
+    try:
+        from structlog_sentry import SentryProcessor
+    except ModuleNotFoundError:
+        pass
+    else:
+        processors.append(SentryProcessor(level=logging.ERROR))
```
jansegre, line 227: I was going to suggest that we add a `enable_sentry` param and gate the whole `try` block on it. But I understand how this can get complicated. I have a branch where I refactor the argparse structure that would make this possible, I'll add this to a TODO. I think it would really be better to not try to load this at all when we don't need it, and error when we need it but fail to load it.

The way it currently is it might fail silently if the user only installs `sentry_sdk` but not `structlog_sentry`.

hathor/cli/run_node.py
msbrogli: Nice suggestion. Done.

hathor/cli/util.py
msbrogli: You're right. I fixed the silent failure now. Can you review it again, please?

jansegre: Makes sense, I think it's good now.

jansegre (approved):

hathor/cli/run_node.py
luislhl: We can change the instance hostname during deploys, if needed.

We could do something like this to extract the Name of the instance and set it as the hostname inside it, without having to rely on passing input from aws-deploy-tools scripts: https://stackoverflow.com/a/24549602/3477266

But I think that it would be better to do this without having to set the hostname, just setting an ENV VAR maybe, and a custom tag in Sentry. I don't know if there are other bad consequences of changing the hostname.

luislhl (approved):

pedroferreira1 (approved):

hathor/cli/run_node.py
msbrogli: @luislhl I'll add a custom tag in Sentry in a separate PR. Let's see what other pieces of information are useful after we start using it for real.

PR #221, opened 2021-06-02
chore: bump version

jansegre:


pedroferreira1:
@luislhl maybe we could create a CI check just like it was added in the wallets (desktop and mobile) and explorer.

pedroferreira1 (approved):

luislhl:
> @luislhl maybe we could create a CI check just like it was added in the wallets (desktop and mobile) and explorer.

That's true. Noted.

luislhl (approved):

PR #229, opened 2021-06-11
feat(cli): param to enable non-standard transaction and higher max-size

jansegre:


jansegre:
This should make it easy to setup a private node that accepts non-standard or large transactions. What do you think? @msbrogli @pedroferreira1 @luislhl

pedroferreira1 (approved):

msbrogli (approved):

PR #230, opened 2021-06-11
feat(p2p): only say the peer is "blocked" when explicitly configured

jansegre:
This change makes it so we have to explicitly enable "peer blocked" warnings. Which should make it so "internal nodes" that don't use the default whitelist don't annoy everyone.

### DIFF
--- a/hathor/conf/settings.py
+++ b/hathor/conf/settings.py
@@ -154,6 +154,9 @@ class HathorSettings(NamedTuple):
     # that the peer is synced (in seconds).
     P2P_SYNC_THRESHOLD: int = 60
 
+    # Whether to warn the other peer of the reason for closing the connection
+    WHITELIST_WARN_BLOCKED_PEERS: bool = False
+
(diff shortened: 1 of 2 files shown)


hathor/conf/settings.py
```diff
@@ -154,6 +154,9 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
     # that the peer is synced (in seconds).
     P2P_SYNC_THRESHOLD: int = 60
 
+    # Whether too warn the other peer of the reason for closing the connection
+    P2P_WARN_BLOCKED: bool = False
```
pedroferreira1, line 158: This warning is only in the case of a whitelist rejection, right? Maybe we should make this explicit in the comment or in the var name

jansegre: What about `WHITELIST_WARN_BLOCKED_PEERS`?

msbrogli (approved):

PR #234, opened 2021-06-16
chore(ci): use codecov for minimum coverage check

jansegre:
This moves the minimum coverage check to codecov. The pipeline will still fail if coverage is not reached, but the tests checks will still pass, this should make it easier to see coverage specific failures.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,11 +1,19 @@
 codecov:
+  branch: dev
+
+coverage:
+  # https://docs.codecov.com/docs/coverage-configuration
+  range: "80...90"
   # https://docs.codecov.io/docs/commit-status
   status:
+    # TODO: re-enable patch in the future
+    patch: off
     project:
       default:
         # minimum coverage ratio that the commit must meet to be considered a success
-        target: 82%
+        target: 83%
         if_ci_failed: error
         only_pulls: true
+
 github_checks:
-  annotations: false
+  annotations: true
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ tests-doctests:
 
 .PHONY: tests-lib
 tests-lib:
-	pytest --durations=10 $(pytest_flags) --doctest-modules hathor --cov-fail-under=83 $(tests_lib)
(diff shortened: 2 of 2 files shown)


jansegre:
This is what it will look like when a PR fails to reach coverage: https://github.com/jansegre/hathor-core/pull/6

.codecov.yml
```diff
@@ -1,11 +1,19 @@
 codecov:
+  branch: dev
```
pedroferreira1, line 2: This means that we are going to test the coverage only when we have a PR to dev?

.codecov.yml
```diff
@@ -1,11 +1,19 @@
...
+
+coverage:
+  # https://docs.codecov.com/docs/coverage-configuration
+  range: "80...90"
   # https://docs.codecov.io/docs/commit-status
   status:
+    # TODO: re-enable patch in the future
+    patch: off
```
pedroferreira1, line 10: What's this used for?

pedroferreira1 (approved):

.codecov.yml
jansegre: No. This means that the `dev` branch is used for the main configuration. Some options can be loaded right on the branch of a PR, but others need a "global" default that will use that branch.

.codecov.yml
jansegre: When enabling `github_checks.annotations`, there are 2 checks by default:

- `codecov/project`
- `codecov/patch`

I'm not totally sure about the differences, but for the total coverage check is made on `codecov/project`, whereas on `codecov/patch` it checks, for instance, that the diff of the branch reaches the target coverage (for instance, if the PR only adds 1 line, that does not change the total coverage by much `codecov/project` will pass but `codecov/patch` will fail because it will have a "patch coverage" of 0%).

I'm proposing that we start with `codecov/patch` disabled, _should_ have the exact same behavior as we do by checking the coverage when running tests (it doesn't check the patch coverage). And only after we're comfortable with dealing with codecov we would enable `codecov/patch`, which should be much more useful in practice than only looking at the total coverage.

luislhl (approved):

PR #236, opened 2021-06-17
refactor(sync-v2): introduce all structures needed without any sync-v2

jansegre:


hathor/transaction/block.py
```diff
@@ -141,6 +141,8 @@ def get_next_block_best_chain_hash(self) -> Optional[bytes]:
         """
         assert self.storage is not None
         meta = self.get_metadata()
+        if meta.voided_by:
+            return None
```
msbrogli, line 145: I'm not sure what's the best approach here. Should we raise an exception? I feel this method should never be called in voided blocks because there's no such thing as next block in the best chain.

hathor/consensus.py
```diff
@@ -253,7 +253,7 @@ def update_voided_info(self, block: Block) -> None:
                     assert best_block.hash is not None
                     best_meta = best_block.get_metadata()
                     self.log.debug('index previous best block', height=best_meta.height, block=best_block.hash_hex)
-                    storage.add_new_to_block_height_index(best_meta.height, best_block.hash)
```
msbrogli, line 256: Is it a bug fix?

tests/p2p/netfilter/test_factory.py
```diff
@@ -18,7 +18,8 @@ def test_factory(self):
         rule = NetfilterRule(match, NetfilterReject())
         pre_conn.add_rule(rule)
 
-        wrapped_factory = HathorServerFactory('testnet', PeerId(), node=None, use_ssl=False)
+        wrapped_factory = HathorServerFactory('testnet', PeerId(), node=None, use_ssl=False,
+                                              enable_sync_v1=True, enable_sync_v2=False)
```
msbrogli, line 22: Should we test it with sync-v2 only and the bridge? If it doesn't matter, I feel we should add a comment explaining why it doesn't matter.

tests/p2p/netfilter/test_match.py
```diff
@@ -191,3 +195,22 @@ def test_match_peer_id(self):
         # Fail because proto1 is connected to proto2, and the peer id cannot match.
         context = NetfilterContext(protocol=conn.proto1)
         self.assertFalse(matcher.match(context))
+
+
+class SyncV1NetfilterMatchTest(BaseNetfilterMatchTest):
```
msbrogli, line 200: They are all the same across files. Can't we create a method that receives a `TestCase` and return these three variations registered as valid tests?

Another way to go would be to create a `__sync_variations__` and make the scanner find them and generate all three variations automatically. What do you think?

tests/p2p/test_peer_id.py
```diff
@@ -149,14 +149,20 @@ def test_retry_connection(self):
...
     def test_validate_entrypoint(self):
         manager = self.create_peer('testnet', unlock_wallet=False)
         peer_id = manager.my_peer
         peer_id.entrypoints = ['tcp://127.0.0.1:40403']
 
         # we consider that we are starting the connection to the peer
-        protocol = HathorProtocol('testnet', peer_id, None, node=manager, use_ssl=True, inbound=False)
+        # TODO: also test with sync-v2 enabled
+        protocol = HathorProtocol('testnet', peer_id, None, node=manager, use_ssl=True, inbound=False,
+                                  enable_sync_v1=True, enable_sync_v2=False)
```
msbrogli, line 165: Should we use `self._enable_sync_v1` and `self._enable_sync_v2`?

tests/p2p/test_peer_id.py
```diff
@@ -188,7 +194,9 @@ def getPeer(self):
 
     def test_validate_certificate(self):
         peer = PeerId('testnet')
-        protocol = HathorProtocol('testnet', peer, None, node=None, use_ssl=True, inbound=True)
+        # TODO: also test with sync-v2 enabled
+        protocol = HathorProtocol('testnet', peer, None, node=None, use_ssl=True, inbound=True,
+                                  enable_sync_v1=True, enable_sync_v2=False)
```
msbrogli, line 199: Should we use `self._enable_sync_v1` and `self._enable_sync_v2`?

tests/resources/base_resource.py
```diff
@@ -16,12 +16,23 @@ def _manager_kwargs(self):
...
             wallet = self._create_test_wallet()
             tx_storage = getattr(self, 'tx_storage', None)
+            assert (
+                hasattr(self, '_enable_sync_v1') and
+                hasattr(self, '_enable_sync_v2') and
+                (self._enable_sync_v1 or self._enable_sync_v2)
+            ), (
+                'Please set both `_enable_sync_v1` and `_enable_sync_v2` on the class. '
+                'Also they can\'t both be False. '
+                'This is by design so it\'s we don\'t forget to test for multiple sync versions.'
```
msbrogli, line 26: typo "so it's we don't"

msbrogli (changes requested):

hathor/transaction/storage/transaction_storage.py
```diff
@@ -916,14 +921,18 @@ def get_from_block_height_index(self, height: int) -> bytes:
 
 
 class BaseTransactionStorage(TransactionStorage):
-    def __init__(self, with_index: bool = True, pubsub: Optional[Any] = None) -> None:
+    def __init__(self, with_index: bool = True, with_all_index: bool = False, pubsub: Optional[Any] = None) -> None:
```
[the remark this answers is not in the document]
jansegre: `with_all_index` is supposed to include the interval-tree indexes, which will not be included when starting a node with only sync-v2 support

hathor/manager.py
```diff
@@ -133,7 +138,7 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
 
         # XXX Should we use a singleton or a new PeerStorage? [msbrogli 2018-08-29]
         self.pubsub = pubsub or PubSubManager(self.reactor)
-        self.tx_storage = tx_storage or TransactionMemoryStorage()
+        self.tx_storage = tx_storage or TransactionMemoryStorage(with_all_index=enable_sync_v1)
```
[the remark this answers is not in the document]
jansegre: The memory-storage? I think it's still has some use, it is not the default anymore though. I think CompactStorage (json files) is less useful since it can't even hold the current mainnet unless using a filesystem tuned to handled the number of inodes (or a filesystem that doesn't use inodes, like btrfs). The memory storage is useful for tests and platforms where rocksdb isn't supported (currently Windows, but maybe it won't work on M1 Macs for example, this has never been tested).

So at the very least, I'd prefer to have the rocksdb and memory storage, I completely agree we could remove all the other storages.

hathor/manager.py
```diff
@@ -188,8 +195,10 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
         # List of capabilities of the peer
         if capabilities is not None:
             self.capabilities = capabilities
-        else:
+        elif enable_sync_v2:
             self.capabilities = [settings.CAPABILITY_WHITELIST, settings.CAPABILITY_SYNC_V2]
```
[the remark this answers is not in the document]
jansegre: Maybe but I didn't want to interfere with setting the capabilities list directly (I'm not sure we actually need to, and also I think that the manager isn't a good place for this, I think we shouldn't even need to have list of capabilities, we could build one directly on the hello-state based on what is enabled).

hathor/manager.py
```diff
@@ -647,6 +662,17 @@ def _make_block_template(self, parent_block: Block, parent_txs: 'ParentTxs', cur
             score=sum_weights(parent_block_metadata.score, weight),
         )
 
+    def get_current_score(self, parent_block: Optional[Block] = None, timestamp: Optional[int] = None) -> float:
```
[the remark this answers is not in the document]
jansegre: I don't know how that would work. We either need the current score (and thus we have no block instance, so it would be a classmethod), or we need the score from a given parent block, and we have an instance (and thus it can't be a classmethod anymore). Do you have any suggestion? Maybe the signature isn't good, I'm not sure how this is being used it could be refactored in a different way, but I wouldn't make it goal to put it on the block class but rather make it clearer why it exists.

hathor/p2p/manager.py
```diff
@@ -104,6 +107,12 @@ def __init__(self, reactor: ReactorBase, my_peer: PeerId, server_factory: 'Hatho
 
         self.ssl = ssl
 
+        # Object that handles the sync until the last checkpoint for all peers
+        self.sync_checkpoints = SyncCheckpoint(manager)
```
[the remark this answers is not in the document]
jansegre: I don't understand that suggestion. `ConnectionsManager` isn't a factory right? It's more of a factory holder. But the main point is that we don't know if we have to use sync-v1 or sync-v2 until after the handshake, so we can't easily have different protocol factories (currently we have different "plugins" for sync-v1 and sync-v2, they are not actually plugins anymore after a recent refactor). But either way the current `SyncCheckpoint` class works as a singleton and the most obvious place to put that singleton in is the `ConnectionsManager`. We could just not instantiate it when sync-v2 isn't enabled, but it doesn't make much difference, it will only have any runtime impact if it is started, an only sync-v2 starts it.

tests/p2p/netfilter/test_factory.py
jansegre: Nope, I just forgot to fix this test case. I think I did this before making the base class with specialization for each case structure and forgot to update it. Anyway I fixed it to run in all 3 cases.

tests/p2p/netfilter/test_match.py
jansegre: I think there might be a lot of ways go about this. I just didn't want to spend too much time on this.

A decorator wouldn't work unless it modifies the methods (which I'm not a big fan of, because it has to duplicate the logic of detecting which methods are tests and which are not, and then decorate them individually).

I think using something outside of the tests would be the best solution (like running all the tests with different configs), and some tests will be skipped when sync-v1 is enabled and others when sync-v2 is enabled. But this looks like it would be a lot more work and would run multiple times tests that don't have anything to do with sync.

Anyway your suggestion aren't too clear in my mind. We could have a talk to see if I understand or if you have any example we could discuss on top of them.

hathor/p2p/messages.py
```diff
@@ -90,10 +90,26 @@ class ProtocolMessages(Enum):
...
-    HASHES = 'HASHES'
+    GET_BEST_BLOCK = 'GET-BEST-BLOCK'  # Request the best block of the peer
+    BEST_BLOCK = 'BEST-BLOCK'  # Send the best block to your peer
+
+    GET_BLOCK_TXS = 'GET-BLOCK-TXS'  # TODO: rename, maybe GET-TX-RANGE or repurpose GET-TRANSACTIONS above
+    TRANSACTION = 'TRANSACTION'
+
+    GET_MEMPOOL = 'GET-MEMPOOL'  # TODO: rename, maybe GET-TX-RANGE or repurpose GET-TRANSACTIONS above
+    MEMPOOL_END = 'MEMPOOL-END'  # End of mempool sync
+    # MEMPOOL = 'MEMPOOL'
```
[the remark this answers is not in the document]
jansegre: Removed.

hathor/p2p/states/hello.py
```diff
@@ -73,6 +73,8 @@ def handle_hello(self, payload: str) -> None:
         """ Executed when a HELLO message is received. It basically
         checks the application compatibility.
         """
+        from hathor.p2p.protocol import HathorProtocol
```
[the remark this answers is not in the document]
jansegre: Makes sense. Moved it to `hathor.p2p.protocol_version`.

hathor/p2p/states/hello.py
```diff
@@ -92,6 +94,40 @@ def handle_hello(self, payload: str) -> None:
...
+            else:
+                # no compatible sync version to use, this is fine though we just can't connect to this peer
+                self.log.info('no compatible sync version to use')
+                protocol.send_error_and_close_connection('no compatible sync version to use')
+                return
+        elif protocol.enable_sync_v1:
+            # we don't accept sync-v2, so it doesn't matter much whether the remote supports but we'll check anyway
+            assert settings.CAPABILITY_SYNC_V2 not in protocol.node.capabilities
+            if settings.CAPABILITY_SYNC_V2 in data['capabilities']:
+                # they do support so we should it should fallback because of us
```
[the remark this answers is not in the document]
jansegre: Fixed.

tests/p2p/test_peer_id.py
jansegre: Yes. I missed this. Fixed.

tests/p2p/test_peer_id.py
jansegre: Fixed.

tests/resources/base_resource.py
jansegre: Fixed.

hathor/transaction/block.py
jansegre: I might have added this for testing, there is an assert immediately after, which I remember might have failed. I'm removing this and running all tests to see if this caused any problem.

hathor/consensus.py
jansegre: I don't know why I removed this line. I'll review this more carefully.

jansegre: Oh, this isn't a remove, it's a replace. Github line comments aren't great so see what's bellow the diff. I usually try to comment on the end of a block of changes so it's easier to see. Anyway I'm running the tests again with the old method call, but I think this had to do with the loading transactions, where this call can be a reorg sometimes.

jansegre: I think this should actually me removed, instead of replaced. But it needs some more thought.

hathor/transaction/block.py
jansegre: I'm not sure why, but the simulation tests fail too often when I remove this. This branch is may be introducing a bug that is causing this. I have to review it in depth.

jansegre: I think I got it. I'm adding this fix:

```diff
diff --git a/hathor/p2p/node_sync_v2.py b/hathor/p2p/node_sync_v2.py
index d26cd073..c8a5fc54 100644
--- a/hathor/p2p/node_sync_v2.py
+++ b/hathor/p2p/node_sync_v2.py
@@ -56,6 +56,7 @@ class StreamEnd(IntFlag):
     END_HASH_REACHED = 0
     NO_MORE_BLOCKS = 1
     LIMIT_EXCEEDED = 2
+    STREAM_BECAME_VOIDED = 3  # this will happen when the current chain becomes voided while it is being sent
 
 
 class NodeBlockSync(SyncManager):
@@ -924,6 +925,12 @@ class BlockchainStreaming(_StreamingBase):
             self.stop()
             self.node_sync.send_blocks_end(StreamEnd.LIMIT_EXCEEDED)
             return
+
+        if cur.get_metadata().voided_by:
+            self.stop()
+            self.node_sync.send_blocks_end(StreamEnd.STREAM_BECAME_VOIDED)
+            return
+
         self.counter += 1
 
         self.current_block = cur.get_next_block_best_chain()

```

What could happen is that while a peer is sending a stream of blocks to second peer, it is also receiving blocks from a third peer, such that a re-org happens and the chain that it was sending becomes voided. So the block we assumed to be part of the best chain, is now voided and would fail that assert. I'm restoring the assert and adding the check above on the sync-v2 class.

The simulator test that will give the most consistent failures on this case is `tests/simulation/test_simulator.py::SyncV2HathorSimulatorTestCase::test_new_syncing_peer`. I'm running it with:

```
pytest tests/simulation/test_simulator.py -k 'test_new_syncing_peer and v2' -s
```

With the fixes I mentioned it fails much less often, which I'm still debugging. Either way, I'm considering using flaky on some of these tests to actually require it to pass multiple times (we used flaky before to require a test to pass 1 out of 2 tries, but I can require it to pass 3 out of 3, for example).

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: I mean to remove as default choice in `HathorManager` in case a tx_storage is not provided. Maybe we should make this a required parameter with no default choice.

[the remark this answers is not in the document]
jansegre: I think we can remove it as a default choice on `HathorManager`. But on `rocksdb` being required, I agree that for any real deploy that makes sense, but it might still be useful to be able to have the option to use a memory backend for every case where we use rocksdb, even if just for mocking, or testing on platforms that are tricky to install rocksdb into.

tests/p2p/netfilter/test_match.py
jansegre: I think this is fixed now.

hathor/manager.py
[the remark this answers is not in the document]
jansegre: I removed it as a implicit default on `HathorManager`. What do you think?

jansegre:
I've hijacked this commit with a refactor. Mostly because this refactor is mostly of the changes of the previous PR, except is does not include any sync-v2 specific classes (like `SyncCheckpoint` and `NodeBlockSync`). The current reviews mostly regard changes about this refactor anyway, and it should make the review easier to reason about.

hathor/transaction/storage/transaction_storage.py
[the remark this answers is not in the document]
jansegre: I've removed all `with_all_index` changes from this PR and the next. Currently I only have them in a backup branch and my stash, since this doesn't add anything I'll leave it to a later PR or drop it.

hathor/manager.py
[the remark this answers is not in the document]
jansegre: I still think the behavior implemented is the best change in this case and I don't see much benefit in trying to not duplicate `settings.CAPABILITY_WHITELIST`.

hathor/p2p/manager.py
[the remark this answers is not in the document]
jansegre: I've removed `SyncCheckpoint` from this PR, we can discuss this on #275

hathor/cli/run_node.py
```diff
@@ -80,6 +80,8 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--max-output-script-size', type=int, default=None, help='Custom max accepted script size '
                             'on /push-tx API')
         parser.add_argument('--sentry-dsn', help='Sentry DSN')
+        parser.add_argument('--x-sync-bridge', action='store_true',
```
msbrogli, line 83: Should we add it to the list of unsafe parameters for now?

hathor/consensus.py
```diff
@@ -154,14 +154,14 @@ def update_voided_info(self, block: Block) -> None:
 
         # Update accumulated weight of the transactions voiding us.
         assert block.hash not in voided_by
-        for h in voided_by:
+        for h in sorted(voided_by):
```
msbrogli, line 157: Why do we need that?

hathor/consensus.py
```diff
@@ -154,14 +154,14 @@ def update_voided_info(self, block: Block) -> None:
...
             tx = storage.get_transaction(h)
             tx_meta = tx.get_metadata()
             tx_meta.accumulated_weight = sum_weights(tx_meta.accumulated_weight, block.weight)
             storage.save_transaction(tx, only_metadata=True)
 
         # Check conflicts of the transactions voiding us.
-        for h in voided_by:
+        for h in sorted(voided_by):
```
msbrogli, line 164: Why do we need that?

hathor/p2p/node_sync.py
```diff
@@ -248,22 +249,26 @@ def get_cmd_dict(self) -> Dict[ProtocolMessages, Callable[[str], None]]:
             ProtocolMessages.NOT_FOUND: self.handle_not_found,
         }
 
+    @property
```
msbrogli, line 252: I don't like to use properties. Is it ok to remove it?

hathor/p2p/manager.py
```diff
@@ -106,6 +107,9 @@ def __init__(self, reactor: ReactorBase, my_peer: PeerId, server_factory: 'Hatho
 
         self.ssl = ssl
 
+        # Parameter to explicitly enable whitelist-only mode, when False it will still check the whitelist for sync-v1
+        self.whitelist_only = whitelist_only
```
msbrogli, line 111: Shouldn't we use the netfilter to do it?

hathor/p2p/peer_id.py
```diff
@@ -83,6 +83,9 @@ def __str__(self):
         return ('PeerId(id=%s, entrypoints=%s, retry_timestamp=%d, retry_interval=%d)' % (self.id, self.entrypoints,
                 self.retry_timestamp, self.retry_interval))
 
+    def __hash__(self):
```
msbrogli, line 86: Maybe add a comment explaining that two different objects with same id will be considered the same?

hathor/p2p/states/ready.py
```diff
@@ -68,7 +71,16 @@ def __init__(self, protocol: 'HathorProtocol') -> None:
...
 
         # Initialize sync manager and add its commands to the list of available commands.
-        self.sync_manager = NodeSyncTimestamp(self.protocol, reactor=self.reactor)
+        self.sync_manager: SyncManager
+        if protocol.protocol_version is ProtocolVersion.V1:
+            self.log.debug('loading sync-v1')
+            self.sync_manager = NodeSyncTimestamp(self.protocol, reactor=self.reactor)
+        elif protocol.protocol_version is ProtocolVersion.V2:
+            self.log.debug('loading sync-v2 (pretend)')
+            # self.sync_manager = NodeBlockSync(self.protocol, reactor=self.reactor)
```
msbrogli, line 80: Can we remove this old code?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -123,11 +123,11 @@ def __init__(self):
         self._needed_txs_index: Dict[bytes, Tuple[int, bytes]] = {}
 
         # Hold txs that have not been confirmed
-        self._tx_tips_index: Set[bytes] = set()
+        self._tx_tips_index: List[bytes] = []
```
msbrogli, line 126: Why are we changing to a list?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -123,11 +123,11 @@ def __init__(self):
...
 
         # Hold blocks that can be used as the next parent block
         # XXX: if there is more than one they must all have the same score, must always have at least one hash
-        self._parent_blocks_index: Set[bytes] = {BLOCK_GENESIS.hash}
+        self._parent_blocks_index: List[bytes] = [BLOCK_GENESIS.hash]
```
msbrogli, line 130: Why are we changing to a list?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -180,7 +180,7 @@ def add_to_deps_index(self, tx: bytes, deps: Iterable[bytes]) -> None:
     def del_from_deps_index(self, tx: bytes) -> None:
         """Call to remove tx from all reverse dependencies, for example when validation is complete."""
         _deps = self._dir_dep_index.pop(tx, _DirDepValue())
-        for rev_dep in _deps.keys():
+        for rev_dep in sorted(_deps.keys()):
```
msbrogli, line 183: Why do we have to sort it?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -144,7 +144,7 @@ def _get_validation_state(self, tx: bytes) -> ValidationState:
 
     def _update_deps(self, deps: _DirDepValue) -> None:
         """Propagate the new validation state of the given deps."""
-        for tx, validation in deps.items():
+        for tx, validation in sorted(deps.items()):
```
msbrogli, line 147: Why do we have to sort it?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -205,15 +205,15 @@ def next_ready_for_validation(self, *, dry_run: bool = False) -> Iterator[bytes]
         else:
             cur_ready, self._txs_with_deps_ready = self._txs_with_deps_ready, set()
         while cur_ready:
-            yield from iter(cur_ready)
+            yield from iter(sorted(cur_ready))
```
msbrogli, line 208: Why do we have to sort it?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -205,15 +205,15 @@ def next_ready_for_validation(self, *, dry_run: bool = False) -> Iterator[bytes]
...
             if dry_run:
                 cur_ready = self._txs_with_deps_ready - cur_ready
             else:
                 cur_ready, self._txs_with_deps_ready = self._txs_with_deps_ready, set()
 
     def iter_deps_index(self) -> Iterator[bytes]:
         """Iterate through all hashes depended by any tx or block."""
-        yield from self._rev_dep_index.keys()
+        yield from sorted(self._rev_dep_index.keys())
```
msbrogli, line 216: Why do we have to sort it?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -242,7 +242,7 @@ def get_next_needed_tx(self) -> bytes:
         # This strategy maximizes the chance to download multiple txs on the same stream
         # find the tx with highest "height"
         # XXX: we could cache this onto `needed_txs` so we don't have to fetch txs every time
-        height, start_hash, tx = max((h, s, t) for t, (h, s) in self._needed_txs_index.items())
+        height, start_hash, tx = max((h, s, t) for t, (h, s) in sorted(self._needed_txs_index.items()))
```
msbrogli, line 245: Why do we have to sort it?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -280,12 +280,13 @@ def add_to_parent_blocks_index(self, block: bytes) -> None:
         cur_score = not_none(self.get_metadata(next(iter(self._parent_blocks_index)))).score
         if isclose(new_score, cur_score):
             self.log.debug('same score: new competing parent block', block=block.hex())
-            self._parent_blocks_index.add(block)
+            if block not in self._parent_blocks_index:
```
msbrogli, line 283: O(n) search? I feel we should revert to a set.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -280,12 +280,13 @@ def add_to_parent_blocks_index(self, block: bytes) -> None:
...
+                self._parent_blocks_index.append(block)
         elif new_score > cur_score and not meta.voided_by:
             # If it's a high score, then I can't add one that is voided
             self.log.debug('high score: new best parent block', block=block.hex())
             self._parent_blocks_index.clear()
-            self._parent_blocks_index.add(block)
+            self._parent_blocks_index.append(block)
```
msbrogli, line 289: Risk of duplicate entries.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -300,11 +301,13 @@ def iter_tx_tips(self, max_timestamp: Optional[float] = None) -> Iterator[Transa
...
         yield from cast(Iterator[Transaction], it)
 
-    def remove_from_tx_tips_index(self, remove_txs: Set[bytes]) -> None:
+    def remove_from_tx_tips_index(self, remove_txs: Iterable[bytes]) -> None:
         """
         This should be called to remove a transaction from the "mempool", usually when it is confirmed by a block.
         """
-        self._tx_tips_index -= remove_txs
+        for tx in iter(remove_txs):
+            if tx in self._tx_tips_index:
```
msbrogli, line 309: It was O(n+m) and now it is O(n*m). I feel we should revert to set.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -327,8 +330,8 @@ def update_tx_tips(self, tx: BaseTransaction) -> None:
         # confirming tx2 will make it valid while tx1 becomes voided, so it has to be removed
         # from the tips.
         assert tx.hash is not None
-        to_remove: Set[bytes] = set()
-        to_remove_parents: Set[bytes] = set()
+        to_remove: List[bytes] = []
```
msbrogli, line 333: Why use a list here?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -327,8 +330,8 @@ def update_tx_tips(self, tx: BaseTransaction) -> None:
...
+        to_remove_parents: List[bytes] = []
```
msbrogli, line 334: Why use a list here?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -338,8 +341,11 @@ def update_tx_tips(self, tx: BaseTransaction) -> None:
             # themselves become tips (hence we use to_remove_parents)
             meta = tip_tx.get_metadata()
             if meta.voided_by:
-                to_remove.add(tip_tx.hash)
-                to_remove_parents.update(tip_tx.parents)
+                if tip_tx.hash not in to_remove:
```
msbrogli, line 344: The complexity in time increase considerably with these changes. It was O(n+m) and became O(n*m) in some cases.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -350,7 +356,8 @@ def update_tx_tips(self, tx: BaseTransaction) -> None:
                     confirmed = True
                     break
             if confirmed:
-                to_remove.add(tip_tx.hash)
+                if tip_tx.hash not in to_remove:
```
msbrogli, line 359: O(1) became O(n).

hathor/transaction/storage/transaction_storage.py
```diff
@@ -375,7 +382,9 @@ def update_tx_tips(self, tx: BaseTransaction) -> None:
                 to_add.add(tx_hash)
 
         if to_add:
-            self._tx_tips_index.update(to_add)
+            for tx_hash in to_add:
```
msbrogli, line 385: O(n+m) became O(n*m).

hathor/transaction/storage/transaction_storage.py
```diff
@@ -386,7 +395,9 @@ def update_tx_tips(self, tx: BaseTransaction) -> None:
         self.remove_from_tx_tips_index(set(tx.parents))
 
         if tx.is_transaction and tx.get_metadata().first_block is None:
-            self._tx_tips_index.add(tx.hash)
+            assert tx.hash is not None
+            if tx.hash not in self._tx_tips_index:
```
msbrogli, line 399: O(1) became O(n).

hathor/manager.py
```diff
@@ -155,10 +161,12 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
...
         self.ssl = ssl
-        self.server_factory = HathorServerFactory(self.network, self.my_peer, node=self, use_ssl=ssl)
-        self.client_factory = HathorClientFactory(self.network, self.my_peer, node=self, use_ssl=ssl)
+        self.server_factory = HathorServerFactory(self.network, self.my_peer, node=self, use_ssl=ssl,
+                                                  enable_sync_v1=enable_sync_v1, enable_sync_v2=enable_sync_v2)
+        self.client_factory = HathorClientFactory(self.network, self.my_peer, node=self, use_ssl=ssl,
+                                                  enable_sync_v1=enable_sync_v1, enable_sync_v2=enable_sync_v2)
         self.connections = ConnectionsManager(self.reactor, self.my_peer, self.server_factory, self.client_factory,
-                                              self.pubsub, self, ssl, rng=self.rng)
+                                              self.pubsub, self, ssl, whitelist_only=False, rng=self.rng)
```
msbrogli, line 169: Do we need the whitelist_only parameter since it is not being used?

msbrogli (changes requested):

hathor/cli/run_node.py
jansegre: I think we should, will do when that is merged.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: Ok. I agree.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: I feel we could move it to the next PR since it is not being used in this PR.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: Great. Thanks!

hathor/p2p/manager.py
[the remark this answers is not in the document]
msbrogli: Ok. Great!

hathor/consensus.py
msbrogli: I guess you moved this change to the next PR, right?

hathor/manager.py
```diff
@@ -506,19 +516,33 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
...
-        assert len(can_include_intervals) > 0, f'invalid timestamp "{timestamp}", no tips found"'
-        if len(can_include_intervals) < 2:
-            # If there is only one tip, let's randomly choose one of its parents.
-            must_include_interval = can_include_intervals[0]
-            must_include = [must_include_interval.data]
-            can_include_intervals = sorted(self.tx_storage.get_tx_tips(must_include_interval.begin - 1))
-        can_include = [i.data for i in can_include_intervals]
+        # get all that are before "timestamp"
+        # XXX: maybe add some tolerance?
+        tx_tips = list(self.tx_storage.iter_tx_tips(timestamp))
```
msbrogli, line 521: Maybe we should rename to `iter_mempool_tips(...)` since it has a different behavior of `get_tx_tips(...)`?

msbrogli, line 521: Maybe add `tx_tips: List[Transaction]` which is counterintuitive to me because most `tx_tips` are `List[bytes]`.

msbrogli, line 521: This algorithm seems good but I feel we haven't tested this new index enough. Maybe we should do this change in the next PR so we can test this index more before using it in such critical method.

tests/p2p/test_capabilities.py
```diff
@@ -5,10 +5,7 @@
 settings = HathorSettings()
 
 
-class HathorCapabilitiesTestCase(unittest.TestCase):
-    def setUp(self):
-        super().setUp()
-
+class SyncV1HathorCapabilitiesTestCase(unittest.SyncV1Params, unittest.TestCase):
```
msbrogli, line 8: Why is it different from the other files?

tests/p2p/test_double_spending.py
```diff
@@ -81,8 +83,8 @@ def test_simple_double_spending(self):
             spent_meta = spent_tx.get_metadata()
             self.assertEqual([tx1.hash, tx2.hash], spent_meta.spent_outputs[txin.index])
 
-        self.assertNotIn(tx1.hash, [x.data for x in self.manager1.tx_storage.get_tx_tips()])
-        self.assertNotIn(tx2.hash, [x.data for x in self.manager1.tx_storage.get_tx_tips()])
+        self.assertNotIn(tx1.hash, self.manager1.tx_storage._tx_tips_index)
```
msbrogli, line 86: Should we check both indices?

tests/p2p/test_double_spending.py
```diff
@@ -105,9 +107,9 @@ def test_simple_double_spending(self):
             spent_meta = spent_tx.get_metadata()
             self.assertEqual([tx1.hash, tx2.hash, tx3.hash], spent_meta.spent_outputs[txin.index])
 
-        self.assertNotIn(tx1.hash, [x.data for x in self.manager1.tx_storage.get_tx_tips()])
-        self.assertNotIn(tx2.hash, [x.data for x in self.manager1.tx_storage.get_tx_tips()])
-        self.assertIn(tx3.hash, [x.data for x in self.manager1.tx_storage.get_tx_tips()])
```
msbrogli, line 110: Should we check both indices?

hathor/manager.py
jansegre: Agreed, renamed all new `tx_tips` methods/members to `mempool_tips`.

tests/p2p/test_protocol.py
```diff
@@ -176,7 +178,8 @@ def test_invalid_peer_id(self):
         # So we need to stop this call manually, otherwise the reactor would be unclean with a pending call
         # TODO We should use a fake DNS resolver for tests otherwise we would need internet connection to run it
         resolver = twisted.names.client.getResolver().resolvers[2]
-        resolver._parseCall.cancel()
+        if not resolver._parseCall.cancelled:
```
msbrogli, line 181: This is very old. I think we already have a method that properly cleans up the pending calls.

hathor/manager.py
jansegre: I'll move it to the next PR.

tests/p2p/test_twin_tx.py
```diff
@@ -69,6 +71,9 @@ def test_twin_tx(self):
         self.assertEqual(meta2.voided_by, {tx2.hash})
         self.assertEqual(meta2.twins, [tx1.hash])
 
+        # The same as tx1 but with one output different, so it's not a twin
+        tx3 = add_new_double_spending(self.manager, tx=tx1)
```
msbrogli, line 75: Does it already propagate the transaction? Maybe the `self.manager.propagate_tx(tx3)` is not necessary.

hathor/manager.py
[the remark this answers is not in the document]
jansegre: Moved to the next PR.

hathor/manager.py

hathor/p2p/manager.py
jansegre: I think it would be great. But maybe on a separate PR? I don't think it will be a trivial change.

tests/tx/test_tx_storage.py
```diff
@@ -146,11 +146,11 @@ def test_storage_basic(self):
             self.assertEqual(2, self.tx_storage.get_tx_count())
             self.assertEqual(3, self.tx_storage.get_count_tx_blocks())
 
-            block_parents_hash = [x.data for x in self.tx_storage.get_block_tips()]
+            block_parents_hash = self.tx_storage.get_best_block_tips()
```
msbrogli, line 149: I feel we should check all block tips here (not only the ones in the best chain).

hathor/p2p/node_sync.py
jansegre: Leave it as a method? OK

tests/tx/test_tx_storage.py
```diff
@@ -146,11 +146,11 @@ def test_storage_basic(self):
...
             self.assertEqual(1, len(block_parents_hash))
             self.assertEqual(block_parents_hash, [self.genesis_blocks[0].hash])
 
-            tx_parents_hash = [x.data for x in self.tx_storage.get_tx_tips()]
+            tx_parents_hash = self.manager.get_new_tx_parents()
```
msbrogli, line 153: I feel this change is testing a different thing. Maybe we should test both `get_tx_tips()` and `get_new_tx_parents()`?

tests/simulation/test_simulator.py
```diff
@@ -151,3 +166,22 @@ def test_new_syncing_peer(self):
...
             self.assertConsensusValid(node)
             self.assertConsensusEqual(node, late_manager)
+
+
+class SyncV1RandomSimulatorTestCase(unittest.SyncV1Params, BaseRandomSimulatorTestCase):
+    __test__ = True
+
+    # XXX: I'm not sure if marking this as flaky is ideal, but it can fail randomly if it lands on a bad seed,
+    #      known bad seeds are saved bellow, failing twice in a row is still possible but much less likely
+    @pytest.mark.flaky(max_runs=2, min_passes=1)
```
msbrogli, line 176: I think we should remove this and let them fail. This is a simple test and I feel we can debug it quickly.

tests/tx/test_tips.py
```diff
@@ -0,0 +1,182 @@
...
+    def setUp(self):
+        super().setUp()
+
+        self.network = 'testnet'
+        self.manager = self.create_peer(self.network, unlock_wallet=True)
+
+    def test_tips_back(self):
+        add_new_block(self.manager, advance_clock=1)
+        add_blocks_unlock_reward(self.manager)
+        self.assertEqual(len(self.manager.tx_storage._tx_tips_index), 0)
```
msbrogli, line 24: I feel all these `_tx_tips_index` should be renamed to `_tx_mempool_index `, right?

tests/tx/test_tips.py
```diff
@@ -0,0 +1,182 @@
...
+from tests.utils import (
+    add_blocks_unlock_reward,
+    add_new_block,
+    add_new_blocks,
+    add_new_double_spending,
+    add_new_transactions,
+)
+
+
+class BaseTipsTestCase(unittest.TestCase):
```
msbrogli, line 12: These unit tests feel good to me. But I feel we should avoid accessing `_tx_tips_index` and use the public method instead. So we can easily extend them to other indexes in the future.

tests/simulation/test_simulator_itself.py
```diff
@@ -0,0 +1,325 @@
...
+import pytest
+
+from hathor.simulator import FakeConnection, Simulator
+from tests import unittest
+
+# these tests will basically setup two independent simulator instances and assert that they generate the same
+# blocks and transactions to the hash throughout the execution
+
+
+@pytest.mark.skipif(sys.platform == 'win32', reason='set_seed fails on Windows')
```
msbrogli, line 14: Maybe we can remove this skip.

msbrogli (changes requested):

hathor/p2p/manager.py
msbrogli: Sure. And we are not using it here, so it seems safe to just remove it. What do you think?

hathor/p2p/peer_id.py
jansegre: I need to test this better, it's certainly more complex than that, `__hash__` doesn't cause object equality. Also this change might be wrong, it seems `__hash__` _requires_ object equality in the first place.

hathor/p2p/states/ready.py
jansegre: I can remove this commented line, but this is actually new code that will be "uncommented" on the next PR.

hathor/p2p/peer_id.py
msbrogli: So I feel we should take it out of this PR. What do you think?

hathor/manager.py
jansegre: This is being used on `PeerIdState._should_block_peer`, basically so that sync-v2 does not use a "whitelist-only" mode by default (which we can enable for some nodes) and still use whitelist-only on sync-v1.

If we migrate to netfilter this might not be needed, but currently (on this PR) it is in use.

hathor/p2p/peer_id.py
jansegre: I'll remove this along with `sorted/Set` changes.

hathor/transaction/storage/transaction_storage.py
jansegre: Reverted.

hathor/manager.py
msbrogli: I got it. I agree that we should keep it until we migrate to netfilter later.

hathor/p2p/peer_id.py
jansegre: Removed.

hathor/p2p/manager.py
msbrogli: I agree. Thanks!

tests/simulation/test_simulator_itself.py
jansegre: I think we can now, I'll do it.

hathor/transaction/storage/transaction_storage.py
msbrogli: Reverted.

hathor/transaction/storage/transaction_storage.py
msbrogli: Reverted.

hathor/consensus.py
msbrogli: Reverted.

hathor/transaction/storage/transaction_storage.py
msbrogli: Reverted.

hathor/manager.py
msbrogli: Moved to the next PR. Thanks!

tests/tx/test_tips.py
jansegre: Changing to use public methods only.

tests/tx/test_tips.py
jansegre: Renamed.

tests/simulation/test_simulator.py
jansegre: Done.

tests/tx/test_tx_storage.py
jansegre: Makes sense.

tests/tx/test_tips.py
msbrogli: Great! Thanks!

tests/simulation/test_simulator.py
msbrogli: Great. Thanks!

tests/tx/test_tx_storage.py
jansegre: I'm splitting this into two tests: the same as it was before and one with the new indexes. I don't think we have "all block tips" in the new indexes, only the candidates to be block parents for the next block.

tests/p2p/test_twin_tx.py
jansegre: Yes `add_new_double_spending` has that call, I'll remove the redundant call.

tests/p2p/test_protocol.py
jansegre: OK, I'll remove this along the `skipif`

tests/p2p/test_double_spending.py
jansegre: Yes.

tests/p2p/test_double_spending.py
jansegre: Yes.

tests/p2p/test_capabilities.py
jansegre: This is test too specific to sync-v1, the next PR should introduce the equivalent test for sync-v2.

msbrogli: Ok. Thanks!

tests/p2p/test_double_spending.py
msbrogli: Great! Thanks!

tests/p2p/test_double_spending.py
msbrogli: Great! Thanks!

tests/p2p/test_protocol.py
msbrogli: Great!

tests/p2p/test_twin_tx.py
msbrogli: Great! Thanks!

tests/simulation/test_simulator_itself.py
msbrogli: Great!

tests/tx/test_tips.py
msbrogli: Great! Thanks!

tests/tx/test_tx_storage.py
msbrogli: Good one. Thanks!

tests/tx/test_tx_storage.py
msbrogli: So the old test is using `get_tx_tips()` while the new one is using `get_new_tx_parents()`. Great! Thanks!

msbrogli (approved):
Great PR!

hathor/manager.py
```diff
@@ -98,7 +99,12 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
         from hathor.metrics import Metrics
         from hathor.p2p.factory import HathorClientFactory, HathorServerFactory
         from hathor.p2p.manager import ConnectionsManager
-        from hathor.transaction.storage.memory_storage import TransactionMemoryStorage
+
+        if not (enable_sync_v1 or enable_sync_v2):
+            raise TypeError(f'{type(self).__name__}() at least one sync version is required')
+
+        if tx_storage is None:
+            raise TypeError(f'{type(self).__name__}() missing 1 required positional argument: \'tx_storage\'')
```
pedroferreira1, line 107: We are doing this just to prevent changing the parameter orders if we change to a required parameter in the constructor?

hathor/p2p/messages.py
```diff
@@ -90,10 +90,25 @@ class ProtocolMessages(Enum):
     GET_NEXT = 'GET-NEXT'
     NEXT = 'NEXT'
 
-    GET_BLOCKS = 'GET-BLOCKS'  # Request a list of hashes for blocks. Payload is the current latest block.
```
pedroferreira1, line 93: The removed messages are not used in sync-v1?

hathor/p2p/states/ready.py
```diff
@@ -68,7 +71,15 @@ def __init__(self, protocol: 'HathorProtocol') -> None:
         })
 
         # Initialize sync manager and add its commands to the list of available commands.
-        self.sync_manager = NodeSyncTimestamp(self.protocol, reactor=self.reactor)
+        self.sync_manager: SyncManager
+        if protocol.protocol_version is ProtocolVersion.V1:
+            self.log.debug('loading sync-v1')
+            self.sync_manager = NodeSyncTimestamp(self.protocol, reactor=self.reactor)
```
pedroferreira1, line 77: There is not difference between `protocol` and `self.protocol`, right? I saw super being called in the first line of the constructor (`super().__init__(protocol)`). But should we use only `self.protocol` here just for consistency?

tests/p2p/test_protocol.py
```diff
@@ -171,12 +169,6 @@ def test_invalid_peer_id(self):
         yield self._send_cmd(self.conn.proto1, 'PEER-ID', json.dumps(invalid_payload))
         self._check_result_only_cmd(self.conn.peek_tr1_value(), b'ERROR')
         self.assertTrue(self.conn.tr1.disconnecting)
-        # When a DNS request is made to twisted client, it starts a callLater to check the resolv file every minute
```
pedroferreira1, line 174: No need to do this anymore?

tests/tx/test_tips.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertTrue(set(b2.parents).issubset(set([tx3.hash] + [reward_blocks[-1].hash] + tx3.parents)))
+
+        tx4 = add_new_transactions(self.manager, 1, advance_clock=1)[0]
+        # tx4 had no tip, so the parents will be the last block parents
+        self.assertCountEqual(set(tx4.parents), set(b2.parents[1:]))
+        # Then tx4 will become a tip
+        self.assertCountEqual(self.manager.tx_storage.get_mempool_tips_index(), set([tx4.hash]))
+
+    def test_tips_twin(self):
+        # b = add_new_blocks(self.manager, 6, advance_clock=1)
```
pedroferreira1, line 127: Should we remove it and the line below?

pedroferreira1 (approved):

hathor/cli/run_node.py
```diff
@@ -89,6 +90,8 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--sentry-dsn', help='Sentry DSN')
         parser.add_argument('--enable-debug-api', action='store_true', help='Enable _debug/* endpoints')
         parser.add_argument('--enable-crash-api', action='store_true', help='Enable _crash/* endpoints')
+        parser.add_argument('--x-sync-bridge', action='store_true',
+                            help='Enable support for running both sync protocols. DO NOT ENABLE, IT WILL BREAK.')
```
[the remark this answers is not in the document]
msbrogli: The full-node will not work because the sync-v2 is not implemented in this PR.

hathor/manager.py
msbrogli: Exactly. If we remove the default value, we must change the order.

hathor/p2p/messages.py
msbrogli: They aren't.

hathor/p2p/states/ready.py
msbrogli: I think we should use `self.protocol` all along the code (except for the `super()`). In this case, we need to update the `self.log` as well.

tests/p2p/test_protocol.py
msbrogli: This code is really old code. Later we added a proper clean up in the `tearDown()` of the parent class.

tests/tx/test_tips.py
msbrogli: They are commented out in the `test_tips_winner()` method as well. I think the variables are not used anymore and we can remove it. What do you think?

pedroferreira1: Agree

hathor/cli/run_node.py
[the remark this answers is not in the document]
pedroferreira1: But I see that we instance the same `NodeSyncManager` if we use sync-v1 or sync-v2 in the protocol, so it will work?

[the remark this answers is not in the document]
jansegre: It instantiates a sync-v1 manager where there would be a sync-v2. It will _work_ as long as we don't change that between different versions. But we precisely want to change that and two nodes using `--x-sync-bridge --unsafe-mode=mainnet` with different versions of the code _will not_ be able sync because they would be using different protocols thinking they are both using sync-v2.

hathor/p2p/states/ready.py
jansegre: I agree. I'll make that change.

tests/tx/test_tips.py
jansegre: OK, removing...

jansegre: Done.

hathor/p2p/states/ready.py
jansegre: Done.

hathor/cli/run_node.py
```diff
@@ -89,6 +90,8 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--sentry-dsn', help='Sentry DSN')
         parser.add_argument('--enable-debug-api', action='store_true', help='Enable _debug/* endpoints')
         parser.add_argument('--enable-crash-api', action='store_true', help='Enable _crash/* endpoints')
+        parser.add_argument('--x-sync-bridge', action='store_true',
```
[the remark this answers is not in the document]
jansegre: I think it's fine without beta. This should be temporary either way.

PR #240, opened 2021-06-21
feat: add block at height API

jansegre:


hathor/transaction/resources/block_at_height.py
```diff
@@ -0,0 +1,173 @@
...
+                    'burst': 10,
+                    'delay': 3
+                }
+            ]
+        },
+        'get': {
+            'tags': ['block'],
+            'operationId': 'block',
+            'summary': 'Get block at height',
+            'description': 'Returns the data of a block at specific height.',
```
msbrogli, line 106: I'd say "in the best blockchain".

msbrogli (approved):

hathor/transaction/resources/block_at_height.py
jansegre: I just realized how "best blockchain" can sound like "Hathor is the best blockchain". I'll use "best chain" since there are much more uses of "best chain" than "best blockchain" in code.

PR #241, opened 2021-06-21
feat(api): add API for mempool

jansegre:


hathor/transaction/resources/mempool.py
```diff
@@ -0,0 +1,94 @@
...
+                '200': {
+                    'description': 'Success',
+                    'content': {
+                        'application/json': {}
+                    }
+                }
+            }
+        }
+    }
+}
```
pedroferreira1, line 94: We should add tests for this new API.

hathor/transaction/resources/mempool.py
```diff
@@ -0,0 +1,94 @@
...
+        self.manager = manager
+
+    def render_GET(self, request: 'Request') -> bytes:
+        """ Get request /mempool/ that returns transactions on the mempool
+
+            :rtype: string (json list)
+        """
+        request.setHeader(b'content-type', b'application/json; charset=utf-8')
+        set_cors(request, 'GET')
+
```
pedroferreira1, line 48: We should add a limit here. We know the mempool is not supposed to be huge but we can already add this protection. Maybe limit to 100? In this case we can return in the response also `truncated: True`. And leave a TODO comment to add pagination in the future.

And I am not sure only the tx_id is enough for some cases, maybe we can have a parameter to return the full tx?

hathor/transaction/resources/mempool.py
```diff
@@ -0,0 +1,94 @@
...
+            'global': [
+                {
+                    'rate': '50r/s',
+                    'burst': 100,
+                    'delay': 50
+                }
+            ],
+            'per-ip': [
+                {
+                    'rate': '3r/s',
```
pedroferreira1, line 73: Maybe reduce to 1r/s? I don't know why someone would like to request this API 3 times per second. We should also reduce the burst

hathor/transaction/resources/mempool.py
```diff
@@ -0,0 +1,94 @@
...
+        'get': {
+            'tags': ['mempool'],
+            'operationId': 'mempool',
+            'summary': 'List of mempool transactions',
+            'description': 'Returns a list of all transactions currently on the mempool',
+            'responses': {
+                '200': {
+                    'description': 'Success',
+                    'content': {
+                        'application/json': {}
```
pedroferreira1, line 88: Missing the return example

hathor/websocket/factory.py
```diff
@@ -196,6 +196,8 @@ def serialize_message_data(self, event, args):
             tx = data['tx']
             data = tx.to_json_extended()
             data['is_block'] = tx.is_block
+            # XXX: needed to check if the transaction is from the mempool
```
pedroferreira1, line 199: I didn't understand this.

pedroferreira1 (changes requested):

hathor/websocket/factory.py
r4mmer: The daemon will get all transactions from the websocket, he needs some way to check if the tx he received is from the mempool or not (to start a mempool sync). This `first_block` metadata being null is enough to assert that the tx is from the mempool.

hathor/transaction/resources/mempool.py
r4mmer: I was mimicking the `block_at_height` params since the consumer will be the same, but i agree.

hathor/transaction/resources/mempool.py
r4mmer: Done.

hathor/transaction/resources/mempool.py
r4mmer: I'm putting a limit of 100 transactions for now, but i don't think the `truncated: True` is necessary, since the caller won't be able to fetch the rest of the mempool (in this version).

Pagination is a bit tricky, i had a conversation with @jansegre and @msbrogli about this and i don't think the intention is to add this feature to the index (for performance reasons).

I don't think we should change the "type" of the response based on an argument. If we need to send the full tx we should do it here, or add another api with this behavior. What do you think?

hathor/transaction/resources/mempool.py
r4mmer: Done

hathor/websocket/factory.py
pedroferreira1: So maybe we should remove the XXX from the comment? I think this is a normal comment and you could add the part of the wallet service daemon using this for the mempool

hathor/transaction/resources/mempool.py
pedroferreira1: I understand that they won't be able to fetch the rest of the mempool but maybe it's useful for them to know if they are not getting the full mempool.

About the pagination I agree to leave without it. And to return only one type.

hathor/websocket/factory.py
jansegre: I'm fine either with `XXX` or without, since this isn't surprising maybe without is better but it's subjective. I agree that just mentioning that the wallet service needs this is enough.

hathor/transaction/resources/mempool.py
jansegre: I'm fine with this as is. Something has to go really wrong for the mempool to grow too large, and iterating it is really fast (we don't do any processing, it should be `O(N)` for `N` txs on the mempool).

jansegre (commented):
Since I opened the PR on this repo Github won't let me approve it. But this has my symbolic approval.

hathor/websocket/factory.py
r4mmer: Removed the `XXX`

hathor/transaction/resources/mempool.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # Get a list of all txs on the mempool
+        tx_ids = map(
+            # get only tx_ids
+            lambda tx: tx.hash_hex,
+            # order by timestamp
+            sorted(list(self.manager.tx_storage.iter_mempool()), key=lambda tx: tx.timestamp),
+        )
+        # Only return up to 100 txs per call
+        data = {'success': True, 'transactions': list(islice(tx_ids, 100))}
```
pedroferreira1, line 58: This 100 should be a parameter in our config file. And it would be good to have this in a test. Have 101 txs in the mempool and return only 100.

pedroferreira1 (changes requested):

hathor/transaction/resources/mempool.py
r4mmer: Done.

pedroferreira1 (approved):

r4mmer (approved):
Symbolic approve from @jansegre

PR #242, opened 2021-06-22
chore: bump version

jansegre:


hathor/cli/openapi_files/openapi_base.json
```diff
@@ -7,7 +7,7 @@
         ],
 	"info": {
```
msbrogli, line 8: It seems useful to add the script that checks whether all versions are the same. @luislhl What do you think?

msbrogli (approved):

hathor/cli/openapi_files/openapi_base.json
jansegre: Done.

PR #246, opened 2021-06-25
fix(nginx): import correct resource, also improve Makefile

jansegre:
Also I think we should build the nginx docker image with the rest of the pipeline. We mainly use a private ECR repo, but we could use a public GHCR repo. What do you think? @luislhl @msbrogli @pedroferreira1


msbrogli (approved):

pedroferreira1 (approved):

pedroferreira1:
> Also I think we should build the nginx docker image with the rest of the pipeline. We mainly use a private ECR repo, but we could use a public GHCR repo. What do you think? @luislhl @msbrogli @pedroferreira1

I agree that we should include in the pipeline. We had forgotten to update nginx image more than once.

About the public repo I also think it's fine.

PR #254, opened 2021-07-01
feat(indexes): add RocksDBAddressIndex stored in disk

msbrogli:
## Acceptance criteria

1. Create `AddressesIndex` interface.
2. Implement `AddressesRocksDBIndex`.
3. Implement `RocksDBIndexesManager`.
4. Test `RocksDBIndexesManager`.
5. Skip add/del from `AddressesRocksDBIndex` during initialization.
6. Use `RocksDBIndexesManager` when using `TransactionRocksDBStorage`.

### DIFF
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -17,13 +17,17 @@ from typing import TYPE_CHECKING, Optional
 
 from structlog import get_logger
 
-from hathor.indexes.addresses_index import AddressesIndex
+from hathor.indexes.address_index import AddressIndex
+from hathor.indexes.memory_address_index import MemoryAddressIndex
+from hathor.indexes.rocksdb_address_index import RocksDBAddressIndex
 from hathor.indexes.timestamp_index import TimestampIndex
 from hathor.indexes.tips_index import TipsIndex
 from hathor.indexes.tokens_index import TokensIndex
 from hathor.transaction import BaseTransaction
 
 if TYPE_CHECKING:  # pragma: no cover
+    import rocksdb
+
     from hathor.pubsub import PubSubManager
 
 logger = get_logger()
@@ -38,12 +42,12 @@ class IndexesManager(ABC):
     sorted_blocks: 'TimestampIndex'
     sorted_txs: 'TimestampIndex'
 
-    addresses: Optional['AddressesIndex']
+    addresses: Optional['AddressIndex']
     tokens: Optional['TokensIndex']
 
     @abstractmethod
-    def enable_addresses_index(self, pubsub: 'PubSubManager') -> None:
-        """Enable addresses index. It does nothing if it has already been enabled."""
+    def enable_address_index(self, pubsub: 'PubSubManager') -> None:
+        """Enable address index. It does nothing if it has already been enabled."""
         raise NotImplementedError
 
     @abstractmethod
@@ -84,13 +88,13 @@ class MemoryIndexesManager(IndexesManager):
         self.sorted_blocks = TimestampIndex()
         self.sorted_txs = TimestampIndex()
 
-        self.addresses: Optional[AddressesIndex] = None
-        self.tokens: Optional[TokensIndex] = None
+        self.addresses = None
+        self.tokens = None
 
-    def enable_addresses_index(self, pubsub: 'PubSubManager') -> None:
-        """Enable addresses index."""
+    def enable_address_index(self, pubsub: 'PubSubManager') -> None:
+        """Enable address index."""
         if self.addresses is None:
-            self.addresses = AddressesIndex(pubsub)
+            self.addresses = MemoryAddressIndex(pubsub)
 
     def enable_tokens_index(self) -> None:
         """Enable tokens index."""
@@ -147,3 +151,14 @@ class MemoryIndexesManager(IndexesManager):
 
         if self.tokens:
             self.tokens.del_tx(tx)
+
+
+class RocksDBIndexesManager(MemoryIndexesManager):
+    def __init__(self, db: 'rocksdb.DB') -> None:
+        self._db = db
+        super().__init__()
+
+    def enable_address_index(self, pubsub: 'PubSubManager') -> None:
+        """Enable address index."""
+        if self.addresses is None:
+            self.addresses = RocksDBAddressIndex(self._db, pubsub=pubsub)
--- a/hathor/transaction/storage/rocksdb_storage.py
+++ b/hathor/transaction/storage/rocksdb_storage.py
@@ -13,8 +13,11 @@
 # limitations under the License.
 
 import os
-from typing import TYPE_CHECKING, Iterator, Optional
+from typing import TYPE_CHECKING, Any, Iterator, List, Optional
 
+from structlog import get_logger
+
+from hathor.indexes import IndexesManager, RocksDBIndexesManager
 from hathor.transaction.storage.exceptions import TransactionDoesNotExist
 from hathor.transaction.storage.transaction_storage import BaseTransactionStorage
 from hathor.util import json_dumpb, json_loadb
@@ -22,6 +25,7 @@ from hathor.util import json_dumpb, json_loadb
 if TYPE_CHECKING:
     from hathor.transaction import BaseTransaction, TransactionMetadata
 
+logger = get_logger()
 
 _DB_NAME = 'data_v2.db'
 _CF_NAME_TX = b'tx'
@@ -38,12 +42,13 @@ class TransactionRocksDBStorage(BaseTransactionStorage):
     def __init__(self, path: str = './', with_index: bool = True, cache_capacity: Optional[int] = None):
         import rocksdb
 
+        self.log = logger.new()
         self._path = path
 
         tx_dir = os.path.join(path, _DB_NAME)
         lru_cache = cache_capacity and rocksdb.LRUCache(cache_capacity)
         table_factory = rocksdb.BlockBasedTableFactory(block_cache=lru_cache)
-        opts = dict(
+        options = rocksdb.Options(
             table_factory=table_factory,
             write_buffer_size=83886080,  # 80MB (default is 4MB)
             compression=rocksdb.CompressionType.no_compression,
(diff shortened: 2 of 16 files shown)


jansegre:
It looks good. Maybe it only needs to use the `--data` dir as base for the database. Also if we could still have the option to use in-memory index, even if only for testing in some environments, would be great.

tests/unittest.py
```diff
@@ -123,7 +123,11 @@ def create_peer(self, network, peer_id=None, wallet=None, tx_storage=None, unloc
             if unlock_wallet:
                 wallet.unlock(b'MYPASS')
         if tx_storage is None:
-            tx_storage = TransactionMemoryStorage()
```
msbrogli, line 126: Our tests use `TransactionMemoryStorage()` by default. So we are not testing our in-disk indexes for most of the tests. Should we change it somehow?

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,178 @@
...
+    It works nicely because rocksdb uses a tree sorted by key under the hoods.
+
+    The timestamp must be serialized in big-endian, so ts1 > ts2 implies that bytes(ts1) > bytes(ts2),
+    hence the transactions are sorted by timestamp.
+    """
+    def __init__(self, db: rocksdb.DB, *, cf_name: Optional[bytes] = None,
+                 pubsub: Optional['PubSubManager'] = None) -> None:
+        self._db = db
+        self._cf_name = cf_name or _CF_NAME_ADDRESS_INDEX
+        self._cf = self._db.get_column_family(self._cf_name)
```
msbrogli, line 52: Do I have to create this column family before using it?

msbrogli, line 52: Do I have to create this column family before using it?

hathor/indexes/addresses_memory_index.py
```diff
@@ -0,0 +1,123 @@
...
+from hathor.transaction.scripts import parse_address_script
+
+if TYPE_CHECKING:  # pragma: no cover
+    from hathor.pubsub import EventArguments, PubSubManager
+    from hathor.transaction import TxOutput
+
+logger = get_logger()
+
+
+class AddressesMemoryIndex(AddressesIndex):
```
jansegre, line 32: I'm not sure about the infix these names, also maybe just "address index" instead of "addresses index", and instead of the infix, a prefix, like the managers: "rocksdb address index", "memory address index".

hathor/indexes/manager.py
```diff
@@ -147,3 +150,14 @@ def del_tx(self, tx: BaseTransaction, *, remove_all: bool = False, relax_assert:
 
         if self.tokens:
             self.tokens.del_tx(tx)
+
+
+class RocksDBIndexesManager(MemoryIndexesManager):
```
jansegre, line 155: I think this is a consequence of why I think the abstraction on `IndexesManager` isn't good. But anyway I'm not opposed to merging this as is. I'll make a proposal to refactor this tomorrow if this is merged.

tests/unittest.py
jansegre: Maybe the default can be configured though settings, or at least have some env var that the pytest setup will load to choose one or the other. RocksDB storage isn't available on Windows, it would be useful to make it configurable on the CI.

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
```
jansegre, line 52: Also, I think this might be an OK place to reset any existing data. If we don't reset it won't work with an existing database and the current set of changes. To me resetting would be the only blocker to merging this PR.

jansegre (approved):
I'm pre-approving this. But it will certainly not work if the index doesn't start fresh, which to me would be the only blocker for merging this.

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def get_sorted_from_address(self, address: str) -> List[bytes]:
+        """ Get a sorted list of transaction hashes of an address
+        """
+        return list(self._get_from_address_iter(address))
+
+    def is_address_empty(self, address: str) -> bool:
+        it = self._db.iterkeys(self._cf)
+        it.seek(self._to_key(address))
+        key = it.get()
```
msbrogli, line 172: What happens if it's empty? Raise exception?

hathor/indexes/manager.py
pedroferreira1: The main thing that you don't like is that we should be extending `IndexesManager` here and not `MemoryIndexesManager`?

tests/unittest.py
pedroferreira1: Is that the reason the CI is not ok for windows?

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+_CF_NAME_ADDRESS_INDEX = b'address-index'
+
+
+class AddressesRocksDBIndex(AddressesIndex):
+    """ Index of inputs/outputs by address.
+
+    This index uses rocksdb and the following key format:
+
+        key = [address][tx.timestamp][tx.hash]
+              |--34b--||--4 bytes---||--32b--|
```
pedroferreira1, line 41: Why do we have 34 bytes for address? I thought it was 25 bytes

hathor/indexes/addresses_rocksdb_index.py
pedroferreira1: You mean to reset the memory index data?

hathor/indexes/addresses_rocksdb_index.py
pedroferreira1: Ahhh we are not saving the bytes, we are saving b58, right?

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+        assert len(address) == 34
+        key = address.encode('ascii')
+        if tx:
+            assert tx.hash is not None
+            assert len(tx.hash) == 32
+            key += struct.pack('>I', tx.timestamp) + tx.hash
+            assert len(key) == 34 + 4 + 32
+        return key
+
+    def _from_key(self, key: bytes) -> Tuple[str, int, bytes]:
```
pedroferreira1, line 68: I thought about turning this return a namedtuple but given that we use only once this structure (and we get it twice), I think it's fine to keep it like this. But if we start using the same tuple in more places, having a namedtuple would be better in my opinion

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+    def subscribe_pubsub_events(self) -> None:
+        """ Subscribe wallet index to receive voided/winner tx pubsub events
+        """
+        assert self.pubsub is not None
+        # Subscribe to voided/winner events
+        events = [HathorEvents.STORAGE_TX_VOIDED, HathorEvents.STORAGE_TX_WINNER]
+        for event in events:
+            self.pubsub.subscribe(event, self.handle_tx_event)
+
+    def _get_addresses(self, tx: BaseTransaction) -> Set[str]:
```
pedroferreira1, line 88: This method is exactly the same as in the `AddressesMemoryIndex`. Should we refactor this to have a helper method for it?

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+            tx2 = tx.storage.get_transaction(txin.tx_id)
+            txout = tx2.outputs[txin.index]
+            add_address_from_output(txout)
+
+        for txout in tx.outputs:
+            add_address_from_output(txout)
+
+        return addresses
+
+    def publish_tx(self, tx: BaseTransaction, *, addresses: Optional[Iterable[str]] = None) -> None:
```
pedroferreira1, line 110: The same as `_get_addresses` method. This is exactly the same as in the other index class.

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def get_from_address(self, address: str) -> List[bytes]:
+        """ Get list of transaction hashes of an address
+        """
+        return list(self._get_from_address_iter(address))
+
+    def get_sorted_from_address(self, address: str) -> List[bytes]:
+        """ Get a sorted list of transaction hashes of an address
+        """
+        return list(self._get_from_address_iter(address))
```
pedroferreira1, line 167: Should we just use `get_from_address` and explain that `_get_from_address_iter` already returns it sorted?

hathor/indexes/addresses_rocksdb_index.py
```diff
@@ -0,0 +1,174 @@
...
+        for key in it:
+            addr, _, tx_hash = self._from_key(key)
+            if addr != address:
+                break
+            yield tx_hash
+
+    def get_from_address(self, address: str) -> List[bytes]:
+        """ Get list of transaction hashes of an address
+        """
+        return list(self._get_from_address_iter(address))
```
pedroferreira1, line 162: Does it return a list or an iterable?

hathor/indexes/addresses_rocksdb_index.py
pedroferreira1: I didn't get the question. What if what's empty? The db?

This method is used by ws to check if we are subscribed to empty addresses.

hathor/indexes/manager.py
jansegre: Not really. I think `IndexesManager` isn't really being usefully abstracted here, nor it will be after all on-disk indexes are implemented. This is how the `TransactionStorage` is right now, and we ended up growing a lot of complexity because of that. IMO we shouldn't do the same here.

A better approach IMO is to leave `IndexesManager` as a concrete class that has abstract properties on the signature but the concrete properties are instantiated by a builder (looks like a factory but is meant to construct a single instances, not many/repeated instances). This would solve the awkward instantiation inside the `enable_*` methods (that currently is the only abstraction we're actually using).

tests/unittest.py
jansegre: Yes.

hathor/indexes/addresses_rocksdb_index.py
msbrogli: But I haven't gotten an error just using it as it is.

tests/unittest.py
jansegre: This is fixed.

hathor/indexes/addresses_rocksdb_index.py
jansegre: I've implemented data-resetting (by dropping and creating a new column-family) for the rocksdb-address-index

hathor/indexes/addresses_rocksdb_index.py
jansegre: I've improved the opening of column families, it should work very well now.

hathor/indexes/addresses_rocksdb_index.py
jansegre: Yes, it's using the "encoded" representation. I think we could use the decoded pubkey part only if there is any noticeble performance advantage.

hathor/indexes/addresses_rocksdb_index.py
jansegre: `_get_from_address_iter` returns an iterator and `get_from_address` collects the items from it to return a list

hathor/indexes/addresses_rocksdb_index.py
jansegre: I think it's fine for both to have the same implementation instead of forwarding to `get_from_address`, if that's what you mean.

We seem to have two separate methods because the in-memory implementation needs manual sorting, and the sorted one ends up being a little more expensive and only used on the address_history resource (the rocksdb implementation naturally stores the the indexes sorted, so both methods are identical). I think we can maintain that for now.

hathor/indexes/addresses_rocksdb_index.py
jansegre: Makes sense. I'll move this to `BaseTransaction.get_related_addresses`, what do you think?

hathor/indexes/addresses_rocksdb_index.py
jansegre: I'm moving this to the base class.

hathor/indexes/addresses_rocksdb_index.py
jansegre: I think it's easy enough to make it a namedtuple and will improve readability and make serialization/deserialization stuff closer. I'll do that.

jansegre: While making this I realized it's a bit awkward because of the asymmetry of serialize/deserialize (because we use partial keys for iterating). I'm now leaning more towards leaving it as is.

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -83,6 +84,9 @@ def _load_from_bytes(self, tx_data: bytes, meta_data: bytes) -> 'BaseTransaction
         tx.storage = self
         return tx
 
+    def _build_indexes_manager(self) -> IndexesManager:
```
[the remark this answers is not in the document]
jansegre: Yes. I understand the limitation but I think it makes sense to do this for now. It might be simple for now to support mixing both, but it won't be trivial in the long run (we will need to be able to optionally rebuild indexes). I think we could revise this feature later.

hathor/indexes/addresses_rocksdb_index.py
pedroferreira1: Sounds good

hathor/indexes/manager.py
pedroferreira1: I think I agree with you. I guess we should leave this refactor for another PR. What do you think?

hathor/indexes/addresses_rocksdb_index.py
pedroferreira1: If we continue with this, any restart would also need to reload all index again, right? We are not using the already saved index in storage to speed up sync. Or did I miss something?

pedroferreira1 (approved):

hathor/indexes/addresses_rocksdb_index.py
jansegre: Yes, that's exactly it. This is done on `self._reset_cf()`, more precisely on `self._db.drop_column_family(_cf)`. Marcelo and I researched the performance implications and the conclusion is that the cost is basically O(1) and the data will be freed seamlessly.

hathor/indexes/addresses_rocksdb_index.py
jansegre: I just tested this and `it.get()` will return whatever was given to `it.seek(...)` when `db` is empty. Under that premise this method will still work as expected (`is_address_empty(...)` will return `True`).

PR #255, opened 2021-07-01
fix(tests): pytest warnings when collecting tests

jansegre:


msbrogli (approved):

hathor/daa.py
```diff
@@ -41,6 +41,8 @@
 
 
 class TestMode(IntFlag):
+    __test__ = False
```
msbrogli, line 44: I didn't understand what it is doing and why it fixes the warning. Do you have any documentation to link here? Maybe add as a comment in the code?

msbrogli: Never mind. I got it.

pedroferreira1 (approved):

PR #256, opened 2021-07-01
chore(ci): run tests on Windows, do not continue on error

jansegre:


hathor/cli/top.py
```diff
@@ -26,6 +24,12 @@
 from math import floor
 from typing import Any, Callable, DefaultDict, Dict, List, Optional, Tuple
 
+try:
+    import curses
+    import curses.ascii
+except ModuleNotFoundError:
+    pass
```
msbrogli, line 31: Shouldn't we print a friendly error message and exit?

tests/p2p/test_protocol.py
```diff
@@ -155,6 +157,7 @@ def test_valid_hello(self):
         self.assertFalse(self.conn.tr1.disconnecting)
         self.assertFalse(self.conn.tr2.disconnecting)
 
+    @pytest.mark.skipif(sys.platform == 'win32', reason='resolver._parseCall not defined on Windows')
```
msbrogli, line 160: It is used only to remove warnings. Maybe we should do the clean up in a different way or add the if inside the test. As far as I understood, the only consequence is that there will be a warning message after the test is finished.

tests/resources/test_profiler.py
```diff
@@ -14,6 +16,7 @@ def setUp(self):
         super().setUp()
         self.web = StubSite(ProfilerResource(self.manager))
 
+    @pytest.mark.skipif(sys.platform == 'win32', reason='shutil.rmtree fails on Windows')
```
msbrogli, line 19: Maybe this link is useful to get it fixed? https://docs.python.org/3.6/library/shutil.html#rmtree-example

tests/simulation/test_simulator.py
```diff
@@ -1,10 +1,14 @@
...
+import sys
 from typing import Optional
 
+import pytest
+
 from hathor.simulator import FakeConnection, MinerSimulator, RandomTransactionGenerator, Simulator
 from tests import unittest
 
 
+@pytest.mark.skipif(sys.platform == 'win32', reason='set_seed fails on Windows')
```
msbrogli, line 11: What's the error message? Maybe I can get it fixed.

tests/tx/test_prometheus.py
```diff
@@ -14,6 +17,7 @@ def setUp(self):
         self.network = 'testnet'
         self.manager = self.create_peer(self.network, unlock_wallet=True)
 
+    @pytest.mark.skipif(sys.platform == 'win32', reason='set_new_metrics fails on Windows')
```
msbrogli, line 20: Do you know the error message? I don't see anything obvious that might raise an exception.

msbrogli (changes requested):
Great PR!

hathor/cli/top.py
jansegre: If we exit here this can't even be imported on Windows. The obvious alternative would be to import inside the methods that use it, they there are a lot. I think it's a good compromise to let it fail with `curses undefined` when it is being used, than to fail immediately or give a warning when most of the time this isn't what you will be running.

This mostly happens because every subcommand module on `hathor/cli` is imported before any subcommand is selected.

tests/p2p/test_protocol.py
jansegre: This doesn't affect any warnings, it only prevents the test from running on Windows. The rationale is that all tests that are known to fail will be skipped so `make tests-lib` runs to completion on Windows, and we can then remove the `continue-on-error` on the CI, which will make the task actually fail when a test fails which wasn't known to fail.

tests/resources/test_profiler.py
jansegre: I actually haven't debugged it at all. I don't know if it's worth the time, I think we can leave this to a separate task.

tests/simulation/test_simulator.py
jansegre: I actually haven't debugged it at all. I don't know if it's worth the time, I think we can leave this to a separate task.

It has something to do with `numpy.random.set_seed`.

tests/tx/test_prometheus.py
jansegre: I actually haven't debugged it at all. I don't know if it's worth the time, I think we can leave this to a separate task.

tests/resources/test_profiler.py
jansegre: The reason that I'm not too worried about getting this to work is that I've never gotten rocksdb to work on Windows (at the very least it will need an update on the python lib, which isn't being well maintained so we might have to fork it), and this might make Windows not viable for the size of the database we have today (unless one uses a lot of RAM and waits for a full sync). So I'm taking it slow and only putting the absolute minimal amount of effort in this.

hathor/cli/top.py
msbrogli: You're right. And the alternative is bad because it is used everywhere in the code.

Ok. I agree this is the best method for now.

tests/resources/test_profiler.py
msbrogli: Ok. In this case, I think we should open an issue to remove these skips and fix theses tests in the future.

tests/simulation/test_simulator.py
msbrogli: Same as above. Thanks!

tests/tx/test_prometheus.py
msbrogli: Same as above.

msbrogli (approved):

tests/resources/test_profiler.py
jansegre: Created #260 to track this.

PR #259, opened 2021-07-02
chore(ci): use docker cache from master instead of dev

jansegre:
Docker build will prefer a cache from dev when there isn't one for the current branch. This can be a problem because the `dev` branch isn't always built as often as the `master` branch (which is built daily from a scheduled task). And an outdated cache can have some problems. This might be the cause of the recent docker build fails on the `dev` branch, even though the scheduled `master` builds are fine. I'm not sure if this will fix those builds though, because the `dev` builds might still try to load the cache from the `dev` branch.


msbrogli (approved):

jansegre:
I decided to remove the other 2 cache entries. Effectively this will now only restore the cache from the same exact commit (benefiting re-runs when it fails for network reasons) and use the cache from the master branch in any other case, which might increase build times of the dev branch when we change dependencies, but should be more reliable and will almost certainly fix the current problem.

PR #263, opened 2021-07-08
build: allow disabling rate-limits in 'generate_nginx_config' command

luislhl:
This PR changes the nginx.conf generator script to allow disabling the rate-limits we include there.
It is supposed to be used only with internal use cases that are not exposed to the internet.

### What was done
- Added a new parameter `--disable-rate-limits` to the `generate_nginx_config` command in hathor cli
- When this parameter is enabled, we do not set rate-limits, and set higher connection limits in websocket endpoints.
- Included targets in Makefile to build the new docker image without rate-limits.

### How to test
Go into `extras/nginx_docker` and run `make docker-use-cases` to check the generated `nginx_use_cases.conf` file.

I've setup the old `make docker` target to now build both images, with and without rate-limits, using different tags.
Run it if you need to test the whole process, but be aware that it will push the images if you have enough permissions and are logged in.


hathor/cli/nginx_config.py
```diff
@@ -192,10 +194,16 @@ def generate_nginx_config(openapi: Dict[str, Any], *, out_file: TextIO, rate_k:
...
+    if disable_rate_limits:
+        websocket_max_conn_global = 10000
+        websocket_max_conn_per_ip = 10000
+        mining_websocket_max_conn_global = 1000
+        mining_websocket_max_conn_per_ip = 1000
+    else:
+        websocket_max_conn_global = 4000
+        websocket_max_conn_per_ip = 10
+        mining_websocket_max_conn_global = 100
+        mining_websocket_max_conn_per_ip = 4
```
jansegre, line 206: I know this isn't great, but I think it's fine for now. Maybe we should create an issue to use jinja2 for the nginx config file, and also possibly move this out of this repo.

jansegre (commented):
On the `Makefile` and ECR repo, we use the term "use case" to separate this from the default image. I think this is a bit odd, outsiders might think that they need this image for their use case, when in fact _most_ (maybe) of the time you'll want to use the default.

What about just using "no ratelimit" (no-ratelimit-latest, no_ratelimit, ...) instead?

hathor/cli/nginx_config.py
msbrogli: I have the same feeling that we should move it out of this repo. But I also feel that it might be a good recommendation for anyone running a public node to use these limits.

msbrogli (approved):

hathor/cli/nginx_config.py
luislhl: Created an issue for using templating when generating the file: https://github.com/HathorNetwork/hathor-core/issues/265

jansegre (approved):

pedroferreira1 (approved):

PR #269, opened 2021-07-18
fix(websocket): bytes is not JSON serializable on 'first_block' field

jansegre:
I think this is a regression from the mempool API. I made this fix quickly, I haven't added any tests, I'm open to suggestions on how to test this (and possibly any JSON serialization error on events).

### DIFF
--- a/hathor/websocket/factory.py
+++ b/hathor/websocket/factory.py
@@ -197,7 +197,9 @@ class HathorAdminWebsocketFactory(WebSocketServerFactory):
             data = tx.to_json_extended()
             data['is_block'] = tx.is_block
             # needed to check if the transaction is from the mempool
-            data['first_block'] = tx.get_metadata().first_block
+            first_block = tx.get_metadata().first_block
+            first_block_hex: Optional[str] = None if first_block is None else first_block.hex()
+            data['first_block'] = first_block_hex
             return data
         elif event == HathorEvents.WALLET_BALANCE_UPDATED:
             data['balance'] = data['balance'][settings.HATHOR_TOKEN_UID]._asdict()
@@ -208,7 +210,11 @@ class HathorAdminWebsocketFactory(WebSocketServerFactory):
     def execute_send(self, data: Dict[str, Any], connections: Set[HathorAdminWebsocketProtocol]) -> None:
         """ Send data in ws message for the connections
         """
-        payload = json.dumps(data).encode('utf-8')
+        try:
+            payload = json.dumps(data).encode('utf-8')
+        except TypeError:
+            self.log.error('failed to serialize data', data=repr(data), exc_info=True)
+            return
         for c in connections:
             try:
                 c.sendMessage(payload, False)
--- a/tests/websocket/test_websocket.py
+++ b/tests/websocket/test_websocket.py
@@ -54,6 +54,9 @@ class TestWebsocket(_BaseResourceTest._ResourceTest):
         self.protocol.state = HathorAdminWebsocketProtocol.STATE_OPEN
 
         tx = self.genesis[1]
+        meta = tx.get_metadata()
+        meta.first_block = self.genesis[0].hash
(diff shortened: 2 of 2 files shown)


hathor/websocket/factory.py
```diff
@@ -197,7 +197,10 @@ def serialize_message_data(self, event, args):
             data = tx.to_json_extended()
             data['is_block'] = tx.is_block
             # needed to check if the transaction is from the mempool
-            data['first_block'] = tx.get_metadata().first_block
+            first_block = tx.get_metadata().first_block
```
msbrogli, line 200: We are changing the type of `first_block` along the code. Maybe rename this to `first_block_bytes` or the other to `first_block_hex`?

msbrogli (changes requested):

pedroferreira1 (approved):

hathor/websocket/factory.py
jansegre: Do you mean rename the JSON property `'first_block'` to `'first_block_hex'`? I don't think that's good because on all API surfaces we already use hex for all fields that contain a hash without suffixing it with `_hex`, like `'hash'`, `'tx_id'`, `'parents'`, ..., and on the other hand, we don't use `_bytes` on the "internal" code (not API surface). So I think this is best as is.

This change in "type" also doesn't affect any API because this `data` was never able to be serialized while that fields was using `bytes`. This is actually a bug that has to be fixed in order for this API to start working again.

What do you think? @msbrogli @pedroferreira1

hathor/websocket/factory.py
```diff
@@ -197,7 +197,10 @@ def serialize_message_data(self, event, args):
             data = tx.to_json_extended()
             data['is_block'] = tx.is_block
             # needed to check if the transaction is from the mempool
-            data['first_block'] = tx.get_metadata().first_block
```
msbrogli, line 200: Should we add a unittest in this PR?

msbrogli (changes requested):

hathor/websocket/factory.py
```diff
@@ -197,7 +197,10 @@ def serialize_message_data(self, event, args):
             data = tx.to_json_extended()
```
msbrogli, line 197: Why hasn't the mempool API unittest gotten it in the first place? Any issue in the simulator?

jansegre: I don't think the simulator should have caught this problem as it is fairly independent from the main flow. But regardless, I don't remember seeing this exception on the simulator logs (which I've been running a lot lately). Maybe it's setup in a way that doesn't run that part of the code.

hathor/websocket/factory.py
jansegre: That would certainly be great. But I invested no time into this. Should I do this?

hathor/websocket/factory.py
pedroferreira1: I think it's only the variable name.

`first_block = tx.get_metadata().first_block` and then you `first_block = first_block.hex()`

jansegre: Makes sense, I'll rename that variable.

jansegre: What do you think?

pedroferreira1: Perfect for me

hathor/websocket/factory.py
jansegre: @pedroferreira1 went ahead and patched a test to cover this regression, I also checked it and it does fail without the fix.

hathor/websocket/factory.py
jansegre: Also this isn't a feature of the mempool API itself. It's a feature that was included together with the mempool API because the wallet-service mempool-sync needed it.

hathor/websocket/factory.py
msbrogli: It was just the variable name. Thanks!

PR #270, opened 2021-07-19
refactor: use hathor.util.Random instances instead of global random

jansegre:
This makes it easier to reproduce simulations, although it may not be enough because of the use of some `Set`s and `Dict`s where the order of iteration will change across simulations. Ideally there would be a lint to prevent the use of `random` or `numpy.random` directly, but I haven't found a simple way to do that.


jansegre:
The CI is failing because of #271 (and a random failure, which I can re-run the tests after #271 is merged).

hathor/simulator/tx_generator.py
```diff
@@ -70,7 +73,8 @@ def schedule_next_transaction(self):
             self.manager.propagate_tx(self.tx, fails_silently=False)
             self.tx = None
 
-        dt = random.expovariate(self.rate)
+        dt = self.rng.exponential(1 / self.rate)
```
msbrogli, line 76: Maybe create a `self.mean = 1 / self.rate` and use it here.

hathor/simulator/tx_generator.py
```diff
@@ -84,9 +88,10 @@ def new_tx_step1(self):
         if not self.send_to:
             address = self.manager.wallet.get_unused_address(mark_as_used=False)
         else:
-            address = random.choice(self.send_to)
+            address = self.rng.choice(self.send_to)
 
-        value = random.randint(1, balance.available)
+        value = self.rng.integers(1, balance.available, endpoint=True)
```
msbrogli, line 93: Is it an `int`?

hathor/util.py
```diff
@@ -274,3 +276,19 @@ def not_none(optional: Optional[_T], message: str = 'Unexpected `None`') -> _T:
...
         raise AssertionError(message)
     return optional
+
+
+def random_choices(rng: Rng, options: Sequence[_T], size: int) -> List[_T]:
+    """Choose `size` randomly selected elements from `options` sequence."""
+    # XXX: explicitly convert to numpy array to prevent numpy from converting a NamedTuple (i.e.) to a numpy array too
+    _options = np.empty(len(options), dtype=object)
+    _options[:] = options
+    return list(rng.choice(_options, size, replace=False))
```
msbrogli, line 286: I think including numpy in the project is error prone. We will have to remember all these details when we decide to call `choice(...)` in the future.

msbrogli (changes requested):

hathor/simulator/simulator.py
```diff
@@ -12,137 +12,216 @@
...
+        """ This is used by when stopping instances of Simulator to determine when to run _remove_patches"""
+        assert cls._patches_rc >= 0
+        cls._patches_rc -= 1
+        if cls._patches_rc == 0:
+            # patches not needed anymore
+            cls._remove_patches()
+
+    def __init__(self, seed: Optional[int] = None):
+        self.log = logger.new()
+        self.seed = secrets.randbits(64)
```
msbrogli, line 102: `if seed is None`?

hathor/simulator/simulator.py
```diff
@@ -12,137 +12,227 @@
...
+            network = self._network
 
         wallet = HDWallet(gap_limit=2)
         wallet._manually_initialize()
 
-        # TODO FIXME Peer-id changes even when using the same seed.
-        peer_id = PeerId()
+        assert peer_id is not None  # XXX: temporary, for checking that tests are using the peer_id
+        if peer_id is None:
+            peer_id = PeerId()
```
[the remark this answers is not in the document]
jansegre: I used a pool of peer-ids and let the rng choose one, so they will remain the same across simulations. Generating a PeerId directly from the rng would be ideal, but I haven't found a way to do it. And maybe we shouldn't, as to minimize the chance of the running the portion of code that injects an "unsafe" random generator when generating a real peer-id.

hathor/manager.py
```diff
@@ -102,6 +103,10 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
 
         self.log = logger.new()
 
+        if rng is None:
+            rng = Rng(PCG64())
```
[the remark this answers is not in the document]
jansegre: Done, it is now using `Random()` which is the refactored default.

hathor/merged_mining/coordinator.py
```diff
@@ -1116,8 +1116,11 @@ class MergedMiningCoordinator:
     def __init__(self,  bitcoin_rpc: IBitcoinRPC, hathor_client: IHathorClient,
                  payback_address_bitcoin: Optional[str], payback_address_hathor: Optional[str],
                  address_from_login: bool = True, min_difficulty: Optional[int] = None,
-                 sequential_xnonce1: bool = False):
+                 sequential_xnonce1: bool = False, rng: Optional[Rng] = None):
         self.log = logger.new()
+        if rng is None:
+            rng = Rng(PCG64())
```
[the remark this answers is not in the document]
jansegre: Done, it is now using `Random()` which is the refactored default.

hathor/mining/block_template.py
```diff
@@ -70,12 +72,13 @@ def generate_mining_block(self, merge_mined: bool = False, address: Optional[byt
...
         return block
 
-    def get_random_parents(self) -> Tuple[bytes, bytes, bytes]:
+    def get_random_parents(self, rng: Rng) -> Tuple[bytes, bytes, bytes]:
         """ Get parents from self.parents plus a random choice from self.parents_any to make it 3 in total.
 
         Return type is tuple just to make it clear that the length is always 3.
         """
-        more_parents = [x for _, x in sorted(random.sample(list(enumerate(self.parents_any)), 3 - len(self.parents)))]
+        more_parents = random_choices(rng, self.parents_any, 3 - len(self.parents))
```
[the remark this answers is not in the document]
jansegre: Done. I've added `assert 1 <= len(self.parents) <= 3`.

hathor/simulator/miner.py
```diff
@@ -41,6 +42,8 @@ def __init__(self, manager: 'HathorManager', *, hashpower: float):
         self.clock = manager.reactor
         self.block = None
         self.delayedcall = None
+        self.log = logger.new()
+        self.rng = default_rng()
```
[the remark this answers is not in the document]
jansegre: I added it as a requirement on the constructor, as to make it harder to forget passing it (using `create_miner` the simulator will inject the rng arg, but one could do `MinerSimulator(manager, manager.rng)` instead) because it has to be explicit. What do you think?

hathor/simulator/simulator.py
```diff
@@ -12,137 +12,227 @@
...
+        self._peers: OrderedDict[str, 'HathorManager'] = OrderedDict()
+        self._connections: List['FakeConnection'] = []
+        # XXX: erasing the types of the random states is preferred as they aren't stable and may change on any version
+        self._py_random_state: Optional[Any] = None
+        self._numpy_random_state: Optional[Any] = None
+        self._started = False
+
+    @property
+    def seed(self) -> int:
+        return self._seed.entropy
```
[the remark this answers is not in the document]
jansegre: Refactored out. But originally I wanted to not expose the `SeedSequence` object.

hathor/simulator/tx_generator.py
jansegre: Refactored out since `rng` now has `expovariate`.

hathor/simulator/tx_generator.py
jansegre: Refactored back to `randint`. But you're right it would be a `numpy.int64`.

hathor/util.py
jansegre: Done. Removed numpy completely.

hathor/simulator/simulator.py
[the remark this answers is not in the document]
msbrogli: I guess your solution is really good. Thanks!

hathor/simulator/simulator.py
jansegre: Done. Added more or less that change.

hathor/util.py
```diff
@@ -274,3 +275,29 @@ def not_none(optional: Optional[_T], message: str = 'Unexpected `None`') -> _T:
...
+        [1] https://en.wikipedia.org/wiki/Inverse_transform_sampling
+        """
+        return math.ceil(math.log(self.random()) / math.log(1 - p))
+
+    def ordered_choices(self, seq: Sequence[T], k: int) -> List[T]:
+        """Like self.choices but preserve orginal order.
+
+        For example, ordered_choices([1, 2, 3]) will never return [3, 2] only [2, 3] instead."""
+        chosen = self.choices(seq, k=k)
+        chosen.sort(key=lambda x: seq.index(x))
```
msbrogli, line 296: It might be two problems: (i) repeated elements, and (ii) time complexity O(n^2 log(n)) if the sort algorithm does not cache the keys.

hathor/simulator/miner.py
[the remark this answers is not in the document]
msbrogli: Seems good. Thanks!

msbrogli (approved):

hathor/util.py
msbrogli: Fixed. Thanks!

pedroferreira1 (approved):

PR #276, opened 2021-07-21
chore(docker): enable support for sentry on CI-built images

jansegre:


.github/workflows/docker.yml
```diff
@@ -136,7 +136,7 @@ jobs:
     - name: Generate requirements
       run: |
         poetry remove -D grpcio-tools --dry-run  # workaround for Poetry bug
-        poetry export -o requirements.txt -E rocksdb
+        poetry export -o requirements.txt -E rocksdb -E sentry
```
msbrogli, line 139: Shouldn't we enable in Docker Hub's image as well?

msbrogli (approved):

.github/workflows/docker.yml
jansegre: This is the image the is pushed to both GHCR and DockerHub. We don't use the `Dockerfile` for any official image, but we do build it on DockerHub to check that it is building. Since sentry is optional, I'm not sure if it's useful or not to enable it, since it's up to whoever is using it.

jansegre: I wish setuptools had a way to have default extras so we could make it opt-out instead of opt-in. Since it doesn't have this kind of awkward situations arise.

jansegre: Anyway I think it makes sense, I've added it.

luislhl (approved):

PR #277, opened 2021-07-23
feat(debug): add APIs to allow forcing exceptions/prints/etc

jansegre:
Unresolved questions:

- Should these endpoints only by added when enabled from the command line? (like `--enable-debug-api`)
- Is the prefix `_debug/` OK? I wanted it to look as something clearly made for internal use, that's why the `_`
- Should these endpoints be registered to the openapi generator? I just added them without putting much thought into it, they're all missing the parameters supported and/or examples, like we have on other endpoints
- Is there any possibly useful functionality missing?

### DIFF
--- a/hathor/debug_resources.py
+++ b/hathor/debug_resources.py
@@ -1,0 +1,227 @@
+# Copyright 2021 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+from typing import Dict, List, TypeVar, cast
+
+from structlog import get_logger
+from twisted.internet import defer, reactor
+from twisted.web import resource
+
+from hathor.cli.openapi_files.register import register_resource
+from hathor.exception import HathorError
+from hathor.manager import HathorManager
+
+logger = get_logger()
+
+
+# XXX: this class must not inherit from Exception or its sub-classes so it purposely skips any `except Exception`
+class DebugException(BaseException):
+    pass
+
+
+T = TypeVar('T')
+
+
+def get_arg_default(args: Dict[bytes, List[bytes]], key: str, default: T) -> T:
+    assert isinstance(default, (type(None), str, int))
+    bkey = key.encode()
+    values = args.get(bkey)
+    if not values:
+        return cast(T, default)
+    value: bytes = values[0]
+    if isinstance(default, int):
+        return cast(T, int(value))
+    else:
+        return cast(T, value.decode())
+
+
+@register_resource
+class DebugRaiseResource(resource.Resource):
+    isLeaf = True
+    openapi = {
+        '/_debug/raise': {
+            'x-visibility': 'private',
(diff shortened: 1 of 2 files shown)


luislhl:
> * Should these endpoints only by added when enabled from the command line? (like `--enable-debug-api`)

I think so, otherwise, if a full-node has its API exposed, someone will be able to crash it. Even with the possibility of blocking requests in another level, the default should be the safest option to avoid mistakes.

jansegre:
> > * Should these endpoints only by added when enabled from the command line? (like `--enable-debug-api`)
> 
> I think so, otherwise, if a full-node has its API exposed, someone will be able to crash it. Even with the possibility of blocking requests to the `_debug` path in another level, the default should be the safest option to avoid mistakes.

Makes sense. I agree. Added `--enable-debug-api` now.

luislhl (approved):

pedroferreira1 (approved):
I think we should add more protections for those endpoints. We've discussed to allow `--test-mode` only with another parameter `--unsafe-mode`. Maybe we should do the same here.

I am especially worried about the last two endpoints, the one that exit the full node and corrupts the database and the other that sets storage to None. We have to make sure none of our public nodes have those endpoints available (even if it's private, we need to make sure we won't call this API by mistake in our wallets/explorer nodes).

What do you think?

About the API docs, I like having it, I think this should be as complete as possible, but maybe the public docs could have only the public endpoints, I don't know.

hathor/debug_resources.py
```diff
@@ -0,0 +1,223 @@
...
+        super().__init__()
+        self.manager = manager
+        # map is create here to be able to access the bound methods
+        self.mess_map = {
+            'storage': self.remove_storage,
+        }
+
+    def remove_storage(self):
+        # XXX: this is just used to cause a problem on another part of the fullnode
+        self.manager.tx_storage = None
```
msbrogli, line 192: Should we an API to revert it?

hathor/debug_resources.py
```diff
@@ -0,0 +1,223 @@
...
+                'operationId': 'debug_crash',
+                'summary': 'Used for forcing an unclean exit, this will almost certainly corrupt the database',
+            }
+        }
+    }
+
+    def run(self, code: int) -> None:
+        # XXX: sys.exit will raise a SystemExit exception that get's trapped by twisted
+        #      os._exit will bypass that by exiting directly, note that no cleanup methods will be called
+        os._exit(code)
```
msbrogli, line 218: Isn't it the same as using `kill -9`?

msbrogli (approved):

hathor/debug_resources.py
jansegre: The main difference is that `kill -9` isn't able to make the process seem it exited successfully. With `os._exit` we can exit with `0` and it will seem like a "successful" exit. I have no idea if this is actually useful, but might be (a successful exit may signal that the database was closed successfully).

hathor/debug_resources.py
jansegre: Maybe but I don't think we should expect things to go back to normal after reverting this, so I don't know.

pedroferreira1:
> If we do decide to add a --unsafe-mode for --test-mode, it makes sense to include this parameter and maybe others, but I think we should discuss the --unsafe-mode separately.

Agree

> Do you think it's worth it to add a mechanism to selectively register endpoints so we can optionally generate a full API doc? Or do you think it's enough for the class to have the OpenAPI description but simply not register it?

I think it's enough for now just not to register it.

hathor/debug_resources.py
msbrogli: You're right. The consequences are unpredictable and the full node must be restarted.

hathor/debug_resources.py
msbrogli: Ok. Thanks!

PR #278, opened 2021-07-27
feat(cli): Add --unsafe-mode argument to run_node

msbrogli:
## `--unsafe-mode` with no unsafe argument

image

## `--test-mode-tx-weight` with `--unsafe-mode`

image

## Error using `--test-mode-tx-weight` without `--unsafe-mode`

image

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -18,7 +18,7 @@ import os
 import platform
 import sys
 from argparse import ArgumentParser, Namespace
-from typing import Any, Dict, List
+from typing import Any, Callable, Dict, List, Tuple
 
 from autobahn.twisted.resource import WebSocketResource
 from structlog import get_logger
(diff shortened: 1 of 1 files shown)


hathor/cli/run_node.py
```diff
@@ -258,6 +305,10 @@ def create_wallet():
         self.start_manager(args)
         self.register_resources(args)
 
+    def unsafe_mode_required(self, argument: str):
```
jansegre, line 308: Missing `-> None`

hathor/cli/run_node.py
```diff
@@ -131,6 +136,46 @@ def prepare(self, args: Namespace) -> None:
...
+                '',
+                'You should never use --unsafe-mode in production environments.',
+                '',
+                'The --unsafe-mode parameter allows you to enable unsafe features in your full-node.',
+                '',
+            ]
+
+            fail = False
+            if not unsafe_args_found:
+                message.append('You have not enabled any unsafe feature. Remove the --unsafe-mode and try again.')
```
jansegre, line 161: The message may be a bit confusing, what about `'--unsafe-mode is not needed because you have not enabled any unsafe feature. Remove --unsafe-mode and try again.'`

jansegre (approved):

hathor/cli/run_node.py
msbrogli: Nice suggestion. Done!

jansegre (approved):

pedroferreira1 (approved):

PR #279, opened 2021-07-28
fix(simulation): test fails sometimes

jansegre:


msbrogli:
We can cancel this PR. These tests were added here: https://github.com/HathorNetwork/hathor-core/pull/282

Issue #280, opened 2021-07-28
Unhandled exception in a malformed Stratum packet


msbrogli:
It seems we just have to capture the `UnicodeDecodeError` here: https://github.com/HathorNetwork/hathor-core/blob/v0.40.2/hathor/stratum/stratum.py#L197-L201

```
  | 2021-07-28T08:22:57.818-07:00 | Traceback (most recent call last):
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/python/log.py", line 103, in callWithLogger
  | 2021-07-28T08:22:57.818-07:00 | return callWithContext({"system": lp}, func, *args, **kw)
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/python/log.py", line 86, in callWithContext
  | 2021-07-28T08:22:57.818-07:00 | return context.call({ILogContext: newCtx}, func, *args, **kw)
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/python/context.py", line 122, in callWithContext
  | 2021-07-28T08:22:57.818-07:00 | return self.currentContext().callWithContext(ctx, func, *args, **kw)
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/python/context.py", line 85, in callWithContext
  | 2021-07-28T08:22:57.818-07:00 | return func(*args,**kw)
  | 2021-07-28T08:22:57.818-07:00 | --- <exception caught here> ---
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
  | 2021-07-28T08:22:57.818-07:00 | why = selectable.doRead()
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/internet/tcp.py", line 243, in doRead
  | 2021-07-28T08:22:57.818-07:00 | return self._dataReceived(data)
  | 2021-07-28T08:22:57.818-07:00 | File "/usr/local/lib/python3.7/site-packages/twisted/internet/tcp.py", line 249, in _dataReceived
  | 2021-07-28T08:22:57.818-07:00 | rval = self.protocol.dataReceived(data)
[...]

PR #281, opened 2021-07-29
fix(util): Convert UnicodeDecodeError into JSONDecodeError in json_loadb

msbrogli:
Fixes https://github.com/HathorNetwork/hathor-core/issues/280.

### DIFF
--- a/hathor/transaction/resources/create_tx.py
+++ b/hathor/transaction/resources/create_tx.py
@@ -69,6 +69,7 @@ class CreateTxResource(resource.Resource):
         set_cors(request, 'POST')
 
         body_content = json_loadb(request.content.read())
+
         raw_inputs = body_content.get('inputs', [])
         raw_outputs = body_content.get('outputs', [])
 
--- a/hathor/transaction/resources/mining.py
+++ b/hathor/transaction/resources/mining.py
@@ -113,6 +113,7 @@ class SubmitBlockResource(resource.Resource):
         set_cors(request, 'GET')
 
         data = json_loadb(request.content.read())
+
         tx = tx_or_block_from_bytes(bytes.fromhex(data['hexdata']), storage=self.manager.tx_storage)
 
         if not tx.is_block:
--- a/hathor/transaction/resources/push_tx.py
+++ b/hathor/transaction/resources/push_tx.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import struct
+from json import JSONDecodeError
 from typing import TYPE_CHECKING, Any, Dict, Optional, cast
 
 from twisted.web import resource
@@ -150,7 +151,16 @@ class PushTxResource(resource.Resource):
         if not body_content:
             return error_ret
 
-        data = json_loadb(body_content)
+        try:
+            data = json_loadb(body_content)
+        except JSONDecodeError as exc:
+            return json_dumpb({
+                'success': False,
+                'message': str(exc),
+            })
+
+        if not isinstance(data, dict):
+            return error_ret
 
         # Need to do that because json_loadb returns an object, which is not compatible with Dict[str, Any]
         data = cast(Dict[str, Any], data)
@@ -192,7 +202,10 @@ PushTxResource.openapi = {
                 'content': {
                     'application/json': {
                         'schema': {
-                            'type': 'string',
+                            'type': 'object',
+                            'properties': {
+                                'hex_tx': 'string',
+                            }
                         }
                     }
                 }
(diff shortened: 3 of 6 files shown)


hathor/transaction/resources/push_tx.py
```diff
@@ -150,7 +151,17 @@ def render_POST(self, request: Request) -> bytes:
...
+            data = json_loadb(body_content)
+        except JSONDecodeError as exc:
+            return json_dumpb({
+                'success': False,
+                'message': str(exc),
+                'can_force': False,
+            })
+
+        if not isinstance(data, dict):
+            return error_ret
```
luislhl, line 164: Should we add this error to the API docs?

hathor/transaction/resources/mining.py
```diff
@@ -112,7 +113,11 @@ def render_POST(self, request):
         request.setHeader(b'content-type', b'application/json; charset=utf-8')
         set_cors(request, 'GET')
 
-        data = json_loadb(request.content.read())
+        try:
+            data = json_loadb(request.content.read())
+        except JSONDecodeError as exc:
+            raise APIError('Cannot decode JSON') from exc
```
luislhl, line 119: Should we add this error and its returned status_code to the API docs?

hathor/transaction/resources/create_tx.py
```diff
@@ -68,7 +69,14 @@ def render_POST(self, request):
...
 
-        body_content = json_loadb(request.content.read())
+        try:
+            body_content = json_loadb(request.content.read())
+        except JSONDecodeError as exc:
+            return json_dumpb({
+                'success': False,
+                'message': str(exc),
+            })
+
```
jansegre, line 79: Wouldn't it be better to have `api_catch_exceptions` catch all `JSONDecodeError` and give a uniform message. JSON parsing isn't really application specific or endpoint specific, so I think it's odd that each endpoint has a different error message.

msbrogli: Ok. I just reverted these changes.

hathor/transaction/resources/mining.py
msbrogli: I reverted these changes because the decorator `@api_catch_exceptions` will handle these exceptions. Even though I think we should rethink this decorator, we should do it in a separate PR.

hathor/transaction/resources/push_tx.py
msbrogli: We usually don't add basic verification errors in the API docs. I mean, users are not following the documentation and it can generate a multitude of errors. What do you think about this approach?

hathor/util.py
```diff
@@ -215,7 +215,12 @@ def json_loadb(raw: bytes) -> Dict:
 
     # XXX: from Python3.6 onwards, json.loads can take bytes
     #      See: https://docs.python.org/3/library/json.html#json.loads
-    return json.loads(raw)
+    try:
+        return json.loads(raw)
+    except UnicodeDecodeError as exc:
+        # We cannot do `doc=raw` because it expects a str and there
+        # is no way to decode it.
+        raise json.JSONDecodeError(msg=str(exc), doc='', pos=exc.start) from exc
```
jansegre, line 223: We could do `doc=raw.hex()`, if that would help identify an error it's better than nothing at all.

hathor/transaction/resources/create_tx.py
jansegre: LGTM

jansegre (approved):

hathor/transaction/resources/push_tx.py
luislhl: It seems ok for me, but I didn't really get what you mean by "users are not following the documentation and it can generate a multitude of errors.".

Another thing I noticed after taking a second look on the Docs is that it states that the Body for `push_tx` should be a "string", but the code is expecting an object with a `hex_tx` key, right?

luislhl (approved):

hathor/transaction/resources/push_tx.py
msbrogli: You're right. I think the documentation is wrong. It should say something like "a JSON with the following format: `"{ "hex_tx": "[HEX]"}"`.

hathor/util.py
msbrogli: Nice idea. Thanks!

hathor/transaction/resources/push_tx.py
msbrogli: Fixed! Can you review again, please? @luislhl

luislhl (approved):

hathor/transaction/resources/push_tx.py
msbrogli: Thanks!

PR #282, opened 2021-07-29
fix(tests): assertConsensusEqual() must convert conflict_with to a set for comparison

msbrogli:
Fixes https://github.com/HathorNetwork/hathor-core/pull/279.

### DIFF
--- a/tests/simulation/test_simulator_itself.py
+++ b/tests/simulation/test_simulator_itself.py
@@ -1,0 +1,258 @@
+from hathor.simulator import FakeConnection, Simulator
+from tests import unittest
+
+
+class BaseSimulatorSelfTestCase(unittest.TestCase):
+    """These tests will basically setup two independent simulator instances and assert that they generate the same
+    blocks and transactions to the hash throughout the execution.
+    """
+
+    __test__ = False
+
+    def setUp(self):
+        super().setUp()
+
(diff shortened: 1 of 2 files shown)


tests/simulation/test_simulator_itself.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+class BaseSimulatorSelfTestCase(unittest.TestCase):
+    __test__ = False
+
+    # XXX: these are known not to trigger known issues, this is fine for now because the point of these tests isn't to
+    #      uncover issues from random cases (that is for test_simulator.py), but to test if the simulator itself is
+    #      able to deterministically reproduce the exact same situations
+    sample_seeds: List[int] = [
+    ]
```
jansegre, line 18: I wonder if we still benfit from this list. I think we might, otherwise I'd just remove this.

jansegre (approved):

pedroferreira1 (approved):

tests/simulation/test_simulator_itself.py
msbrogli: I agree we can just remove it.

msbrogli: Removed. Thanks!

PR #283, opened 2021-07-30
fix: Fix rare conditions captured by random simulations

msbrogli:


hathor/transaction/storage/transaction_storage.py
```diff
@@ -1170,19 +1170,23 @@ def add_to_indexes(self, tx: BaseTransaction) -> None:
...
         if not self.with_index:
             raise NotImplementedError
         assert self.block_index is not None
         assert self.tx_index is not None
         if self.tokens_index:
             self.tokens_index.del_tx(tx)
         if tx.is_block:
-            self._cache_block_count -= 1
-            self.block_index.del_tx(tx, relax_assert=relax_assert)
+            # This is the code to remove the block from the indexes.
```
pedroferreira1, line 1181: I think it might get confusing here when we should set `del_blocks=True`. If I understood correctly, only when you want to completely remove the block from the storage is when you should set it, right? Or is there any other situation? Maybe we could have this as a comment just to prevent mistakes

I am saying that because on the consensus file we are not using `del_blocks=True` and on `remove_transaction` we are. Instead of re-reading all places to understand the reasoning  behind, we could just write here.

pedroferreira1 (approved):

hathor/transaction/storage/transaction_storage.py
jansegre: I don't know if this information adds anything: as far as I know there is only one case in practice where a fullnode will remove a transaction by itself (that is without you opening a shell and doing it manually), it is when there is a re-org that changes the reward-lock validity of a transaction. In all other cases the uses of `remove_transaction` are artificial which are basically used in tests.

PR #287, opened 2021-08-03
tests: Mark a few tests as flaky

msbrogli:


tests/simulation/test_simulator.py
```diff
@@ -112,6 +114,8 @@ def test_many_miners_since_beginning(self):
         for node in nodes[1:]:
             self.assertTipsEqual(nodes[0], node)
 
+    # retry the test once if it fails, see: https://github.com/box/flaky
```
jansegre, line 117: Comment could be "retry the test _more times_ if it fails..." because with runs=3 it's more than once.

jansegre (approved):
All in all, CI runs will get longer, which is annoying, but that's better than having to check failures and manually re-runs too often when we know there's nothing new.

tests/simulation/test_simulator.py
msbrogli: I'll fix it. Thanks!

PR #275, opened 2021-07-21
feat(sync-v2): sync-v2 implemented, sync-v1 still default

jansegre:
These changes were separated from  #236, it should now have mostly the sync-v2 base code itself. [RFC](https://github.com/HathorNetwork/rfcs/blob/master/text/0025-p2p-sync-v2.md)

## Acceptance Criteria

1. Fix `DepsIndex` and its both implementations (memory and rocksdb), allowing voided vertices and using the correct scope when retrieving vertices from the storage.
2. Change `SyncVersion.V2` to `'v2'` (previously `'v2-fake'`).
3. Add `SyncV2Factory` and use it on `ConnectionsManager._sync_factories`.
4. Add `NodeBlockSync` that sync blocks and its transactions.
5. Add `BlockchainStreaming` and `TransactionsStreaming` to stream vertices as requested by peers. They are used by `NodeBlockSync`.
6. Add `TransactionStorage.iter_mempool_tips_from_best_index()`.
7. Refactor some tests to run with both sync-v1 and sync-v2.

## Current issues

- [ ] Sync checkpoint seems to hang when syncing from zero (happens sometimes and not always at the same height);
- [ ] Sync checkpoint seems to not resume properly sometimes after restarting (happens sometimes and not always at the same height);
- [ ] Stats counters aren't being update (they might have to be redesigned since they were originally made for sync-v1 and some counters can't have the same semantics);
- [ ] When using rocksdb-indexes, the memory-deps index is not re-initialized correctly because the iterator will ignore partially validated transactions;
- [x] Deal with syncing soft-voided txs, which didn't exist when sync-v2 was designed, and since sync-v2 doesn't sync voided txs/blocks they need special treatment;
- [x] Some tx-reward-lock validation could fail during checkpoint-sync;
- [x] Sync blocks seems to hang requesting blocks from the same height: with checkpoints disabled the node was only able to sync about 30% of the mainnet;

PR #292, opened 2021-08-12
refactor: last preparations for sync-v2

jansegre:
Refactors and changes that don't alter any behavior extracted from #275


hathor/transaction/storage/transaction_storage.py
```diff
@@ -409,24 +418,35 @@ def update_block_height_cache_new_chain(self, height: int, block: Block) -> None
 
         block_height = height
         side_chain_block = block
-        add_to_cache: List[Tuple[int, bytes]] = []
+        add_to_cache: List[Tuple[int, bytes, int]] = []
```
msbrogli, line 421: Should we replace the tuple by a namedtuple? If I got it right, it should be `(height, blk_id, timestamp)`.

tests/tx/test_tx.py
```diff
@@ -94,6 +95,67 @@ def test_validation(self):
...
+        del block2._metadata
+        block2.storage = manager3.tx_storage
+        block2.set_height_hint(block2_height)
+        # failing silently should not raise an exception, but should return False
+        self.assertFalse(manager3.on_new_tx(block2, sync_checkpoints=True, partial=True, fails_silently=True))
+        # otherwise it should raise the correct exception
+        with self.assertRaises(InvalidNewTransaction):
+            manager3.on_new_tx(block2, sync_checkpoints=True, partial=True, fails_silently=False)
+
+    def test_set_height_hint(self):
```
msbrogli, line 142: I feel it should be moved to the sync-v2 unittest.

hathor/simulator/fake_connection.py
```diff
@@ -80,6 +80,35 @@ def disable_idle_timeout(self):
         self._proto1.disable_idle_timeout()
         self._proto2.disable_idle_timeout()
 
+    def is_synced_or_error(self) -> bool:
+        """Short-hand check that can be used to make "step loops" without having to guess the number of iterations."""
+        from hathor.p2p.states.ready import ReadyState
+        conn1_aborting = self._proto1.aborting
+        conn2_aborting = self._proto2.aborting
+        if conn1_aborting or conn2_aborting:
```
msbrogli, line 88: Is aborting considered an error?

tests/p2p/test_sync_mempool.py
```diff
@@ -0,0 +1,126 @@
...
+        conn = FakeConnection(self.manager1, manager2)
+        for _ in range(1000):
+            if conn.is_empty():
+                break
+            conn.run_one_step(debug=True)
+            self.clock.advance(1)
+
+        # node_sync1 = conn.proto1.state.get_sync_plugin()
+        # node_sync2 = conn.proto2.state.get_sync_plugin()
+        # self.assertEqual(node_sync1.peer_height, common_height)
```
msbrogli, line 99: Why are these asserts commented out?

tests/p2p/test_sync_mempool.py
```diff
@@ -0,0 +1,126 @@
...
+        # self.assertEqual(node_sync2.peer_height, common_height)
+        self.assertConsensusValid(self.manager1)
+        self.assertConsensusValid(manager2)
+        self.assertConsensusEqual(self.manager1, manager2)
+
+        # 3 genesis
+        # 25 blocks
+        # Unlock reward blocks
+        # 8 txs
+        # self.assertEqual(self.manager1.tx_storage.get_count_tx_blocks(), total_count)
```
msbrogli, line 110: Why are these asserts commented out?

msbrogli (changes requested):

hathor/consensus.py
```diff
@@ -242,23 +243,12 @@ def update_voided_info(self, block: Block) -> None:
...
-                    assert best_block.hash is not None
-                    best_meta = best_block.get_metadata()
-                    self.log.debug('index previous best block', height=best_meta.height, block=best_block.hash_hex)
-                    storage.add_new_to_block_height_index(best_meta.height, best_block.hash)
-
-        # Uncomment the following lines to check that the cache update is working properly.
-        # You shouldn't run this test in production because it dampens performance.
-        #     v = storage.get_best_block_tips(skip_cache=True)
-        #     assert v == storage._best_block_tips
+                    storage.update_best_block_tips_cache([not_none(blk.hash) for blk in heads])
```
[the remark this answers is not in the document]
jansegre: No, it isn't necessary. I'm not entirely sure if we have to update the height cache, but since those blocks are voided, when one of them is eventually unvoided the correct block will be added to the height cache.

hathor/manager.py
```diff
@@ -708,7 +709,8 @@ def propagate_tx(self, tx: BaseTransaction, fails_silently: bool = True) -> bool
     @cpu.profiler('on_new_tx')
     def on_new_tx(self, tx: BaseTransaction, *, conn: Optional[HathorProtocol] = None,
                   quiet: bool = False, fails_silently: bool = True, propagate_to_peers: bool = True,
-                  skip_block_weight_verification: bool = False, partial: bool = False) -> bool:
+                  skip_block_weight_verification: bool = False, sync_checkpoints: bool = False,
```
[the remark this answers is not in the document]
jansegre: `sync_checkpoints` should be set to `True` (defaults to `False`) only when adding a block/tx while in syncing checkpoints, which enables the checkpoint validation parts, I guess `skip_checkpoint_verification` would have the opposite semantic and would default to `True`, but I think it's less intuitive to set it to `False` and a bit imprecise (you aren't actually "skipping" checkpoint verification, just not expecting it.

hathor/p2p/manager.py
```diff
@@ -223,13 +232,24 @@ def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
                 self.connected_peers[protocol.peer.id] = existing_protocol
         self.pubsub.publish(HathorEvents.NETWORK_PEER_DISCONNECTED, protocol=protocol)
 
+    # XXX: maybe rename these to `iter_*` since they aren't returning a container
```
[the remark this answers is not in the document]
jansegre: OK.

hathor/p2p/manager.py
```diff
@@ -223,13 +232,24 @@ def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
...
-        return set(self.connected_peers.values())
+    def get_ready_connections(self) -> Iterable[HathorProtocol]:
+        """Iterate over ready connections."""
+        for conn in self.connected_peers.values():
+            yield conn
+
+    def get_not_ready_connections(self) -> Iterable[HathorProtocol]:
+        """Iterate over not-ready connections."""
+        for conn in self.connecting_peers.values():
+            yield conn
```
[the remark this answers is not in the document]
jansegre: I thought I fixed this already. I'll fix this on this PR, it isn't too trivial though.

hathor/p2p/node_sync.py
```diff
@@ -741,8 +745,12 @@ def on_tx_success(self, tx: 'BaseTransaction') -> 'BaseTransaction':
         # the parameter of the second callback is the return of the first
         # so I need to return the same tx to guarantee that all peers will receive it
         if tx:
-            # Add tx to the DAG.
-            success = self.manager.on_new_tx(tx)
+            assert tx.hash is not None
+            if self.manager.tx_storage.transaction_exists(tx.hash):
```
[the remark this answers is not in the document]
jansegre: This wasn't removed from `on_new_tx`, `on_new_tx` considers it a "soft error" if the tx already exists (it raises an exception unless `fails_silently=True`. But this case is perfectly normal and common, so I think it's best to just not call `on_new_tx` when it's perfectly fine for a "duplicate" tx to be received. Otherwise it will keep logging a bunch of warnings that are completely useless.

hathor/p2p/resources/add_peers.py
```diff
@@ -56,8 +56,8 @@ def already_connected(connection_string: str) -> bool:
             # determines if given connection string is already among connected or connecting peers
             endpoint_url = connection_string.replace('//', '')
 
-            # remove peers that we're already trying to connect
-            if endpoint_url in self.manager.connections.connecting_peers:
+            # ignore peers that we're already trying to connect
+            if endpoint_url in set(self.manager.connections.get_not_ready_connections()):
```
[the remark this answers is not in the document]
jansegre: Yes. I thought I fixed this but apparently not. There is also no test that is affected by this, which I'll try to add.

hathor/transaction/base_transaction.py
```diff
@@ -460,13 +461,27 @@ def can_validate_full(self) -> bool:
...
-            if not meta.validation.is_valid():
+            if not meta.validation.is_fully_connected():
                 all_valid = False
             if meta.validation.is_invalid():
                 # or any of them is invalid (which would make this one invalid too)
                 return True
         return all_exist and all_valid
 
+    def validate_checkpoint(self, checkpoints: List[Checkpoint]) -> bool:
+        """ Run basic validations (all that are possible without dependencies) and update the validation state.
```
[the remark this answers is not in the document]
jansegre: Fixed.

hathor/transaction/base_transaction.py
```diff
@@ -763,6 +795,24 @@ def start_mining(self, start: int = 0, end: int = MAX_NONCE, sleep_seconds: floa
...
+        validation state, that is, a height of `None` on a full validated transaction is final) and it doesn't match
+        the given hint. This is in place in order to prevent accidentally setting the height to a wrong value after the
+        value has been validated, the height must be checked when basic-validating this tx and effort should be made to
+        always set the correct height when using this method.
+        """
+        assert self.storage is not None
+        assert self.hash is not None
+        metadata = self.storage.get_metadata(self.hash)
+        if metadata is None:
+            metadata = TransactionMetadata(hash=self.hash, accumulated_weight=self.weight, height=height)
```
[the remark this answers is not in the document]
jansegre: > The only point I think we can do it safely is when syncing blocks between checkpoints from right to left. In this case, I feel this method should be placed in sync-v2 to avoid someone calling it from a different location.

That is exactly the only place where it is called from. What do you mean by placing it in sync-v2? Moving this change to another PR won't make any difference, or is it about moving the method? Since it touches "internals" (_metadata) I thought it was best to keep it on the `Transaction` class.

hathor/transaction/storage/transaction_storage.py
jansegre: Makes sense. I'll do that.

jansegre: Done.

tests/tx/test_tx.py
jansegre: What do you mean by "sync-v2 unitest"? I kept it on this PR to keep the coverage.

tests/p2p/test_sync_mempool.py
jansegre: I've no idea. I ported this from your sync-v2-mempool PR but haven't reviewed them. We are no long using the plugin pattern for the sync managers, so maybe these can be removed.

tests/p2p/test_sync_mempool.py
jansegre: I'm not sure.

hathor/simulator/fake_connection.py
jansegre: I think `is_synced_or_error` isn't a good name. What do you think of renaming it to `can_step` (with flipped booleans)? It returns `True` when either connection aborts because it doesn't make sense to step aborted connections.

tests/p2p/test_sync_mempool.py
jansegre: The `assertConsensusEqual` above already checks the `get_count_tx_blocks` and  `total_count` is not defined. I've removed these lines.

tests/p2p/test_sync_mempool.py
jansegre: I'm removing this lines.

hathor/simulator/fake_connection.py
jansegre: Renamed.

hathor/p2p/manager.py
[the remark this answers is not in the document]
jansegre: Done. Also put `_` in the properties name to indicate that they shouldn't be accessed directly from outside the class.

hathor/p2p/resources/add_peers.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/p2p/manager.py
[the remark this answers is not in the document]
jansegre: I undid the `_` prefix, too many tests use them and this broke, I don't think it's good to mix it with this PR.

hathor/p2p/resources/add_peers.py
```diff
@@ -56,8 +56,8 @@ def already_connected(connection_string: str) -> bool:
             # determines if given connection string is already among connected or connecting peers
             endpoint_url = connection_string.replace('//', '')
 
-            # remove peers that we're already trying to connect
-            if endpoint_url in self.manager.connections.connecting_peers:
+            # ignore peers that we're already trying to connect
+            if endpoint_url in set(self.manager.connections.iter_not_ready_endpoints()):
```
msbrogli, line 60: I guess we do not need this `set(...)` here.

hathor/simulator/fake_connection.py
msbrogli: Seems good. Thanks!

hathor/simulator/simulator.py
```diff
@@ -219,6 +219,20 @@ def _run(self, interval: float, step: float, status_interval: float) -> Generato
                 latest_time = self._clock.seconds()
             self._clock.advance(step)
 
+    def run_until_synced_or_error(self,
```
msbrogli, line 222: Maybe rename this method as well?

hathor/p2p/resources/add_peers.py
jansegre: Removed.

tests/p2p/test_connections.py
```diff
@@ -33,8 +33,8 @@ def test_manager_connections(self):
         endpoint = 'tcp://127.0.0.1:8005'
         manager.connections.connect_to(endpoint, use_ssl=True)
 
-        self.assertNotIn(endpoint, manager.connections.connecting_peers)
-        self.assertNotIn(endpoint, manager.connections.handshaking_peers)
-        self.assertNotIn(endpoint, manager.connections.connected_peers)
+        self.assertNotIn(endpoint, set(manager.connections.iter_not_ready_endpoints()))
```
msbrogli, line 36: I don't think we need these `set(...)` calls here.

hathor/simulator/simulator.py
jansegre: Renamed to `run_until_complete`.

tests/p2p/test_connections.py
jansegre: You're right, removing.

hathor/p2p/manager.py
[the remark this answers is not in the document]
msbrogli: Seems good. Well done!

hathor/p2p/resources/add_peers.py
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Great! Thanks!

hathor/simulator/simulator.py
```diff
@@ -189,7 +189,7 @@ def add_connection(self, conn: 'FakeConnection') -> None:
         self._connections.append(conn)
 
     def _run(self, interval: float, step: float, status_interval: float) -> Generator[None, None, None]:
-        """ Implementation of run, yields at every step to allow verifications like in run_until_synced
+        """ Implementation of run, yields at every step to allow verifications like in run_until_synced_or_error
```
msbrogli, line 192: Outdated comment. Need to update `run_until_synced_or_error` to `run_until_complete`.

hathor/consensus.py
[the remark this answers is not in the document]
msbrogli: We finished the discussion and you're right we don't need this piece of code.

hathor/simulator/simulator.py
jansegre: Fixed.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: Maybe add an explanation for each parameter of this method in the docstring? I feel it is confusing to get the meaning from the code.

hathor/p2p/manager.py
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

hathor/transaction/base_transaction.py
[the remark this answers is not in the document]
msbrogli: Thanks!

tests/p2p/test_sync_mempool.py
msbrogli: Ok!

tests/p2p/test_sync_mempool.py
msbrogli: Ok. Thanks!

tests/p2p/test_connections.py
msbrogli: Thanks!

hathor/simulator/simulator.py
msbrogli: Thanks!

hathor/manager.py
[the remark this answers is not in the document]
jansegre: Done.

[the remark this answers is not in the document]
msbrogli: Great! Thanks!

hathor/p2p/node_sync.py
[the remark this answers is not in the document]
msbrogli: I got it. Nice clean up in the logs.

tests/tx/test_tx.py
msbrogli: You moved to the next PR. Great!

hathor/transaction/base_transaction.py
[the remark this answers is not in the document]
msbrogli: You moved to the next PR. Great!

msbrogli (approved):

PR #295, opened 2021-08-20
feat(sync-v2): more changes, non-essential

jansegre:


hathor/transaction/storage/block_height_index.py
```diff
@@ -69,3 +69,46 @@ def get_height_tip(self) -> Tuple[int, bytes]:
         """
         height = len(self._index) - 1
         return height, self._index[height].hash
+
+    def search_previous_best_block(self, timestamp: float) -> bytes:
+        """Get the hash of the block with highest timestamp that isn't higher then the given timestamp."""
+        # XXX: commented simple O(N) search, use when in doubt
```
msbrogli, line 75: Maybe implement this in a separate method so we can use in the tests.

hathor/transaction/storage/block_height_index.py
```diff
@@ -69,3 +69,46 @@ def get_height_tip(self) -> Tuple[int, bytes]:
...
+            assert low <= mid <= high
+            mid_entry = self._index[mid]
+            if mid_entry.timestamp > timestamp:
+                high = mid - 1
+                high_entry = self._index[high]
+                if high_entry.timestamp < timestamp:
+                    # flipped, return new lower
+                    return high_entry.hash
+            elif mid_entry.timestamp < timestamp:
+                low = mid + 1
```
msbrogli, line 105: Watch out on infinite loop.

PR #300, opened 2021-08-25
feat(tx): improved reward lock verification algorithm

jansegre:
## Acceptance Criteria (msbrogli's proposal)

1. A block `B` is valid only if `B.height >= tx.min_height` for all `tx` such that `tx.first_block == B`.
2. All invalid blocks are discarded (and cannot be added to the storage).
2. A transaction `tx` is valid if it is verified by a block in the bestchain.
3. If a transaction `tx` is in the mempool and the next block cannot verify it, then it must be invalid.

Note: We are using the word "invalid" with two different meanings. On the one hand, invalid blocks must be discarded. On the other hand, invalid txs can be added to the DAG. We need a better word for each of theses cases to avoid misunderstandings.


hathor/transaction/base_transaction.py
```diff
@@ -886,8 +897,12 @@ def update_accumulated_weight(self, *, stop_value: float = inf, save_file: bool
 
         return metadata
 
-    def update_initial_metadata(self) -> None:
-        """Update the tx's initial metadata. It does not update the whole metadata.
+    def update_reward_lock_metadata(self) -> None:
+        metadata = self.get_metadata()
+        metadata.min_height = self.calculate_min_height()
```
msbrogli, line 902: You have to save it, right?

hathor/transaction/block.py
```diff
@@ -312,6 +325,12 @@ def verify_weight(self) -> None:
             raise WeightError(f'Invalid new block {self.hash_hex}: weight ({self.weight}) is '
                               f'smaller than the minimum weight ({block_weight})')
 
+    def verify_height(self) -> None:
```
msbrogli, line 328: Shouldn't we get the `min_height` of the tx parents and compare with this block's height?

hathor/transaction/block.py
```diff
@@ -138,6 +140,17 @@ def calculate_height(self) -> int:
         parent_block = self.get_block_parent()
         return parent_block.get_metadata().height + 1
 
+    def calculate_min_height(self) -> int:
```
msbrogli, line 143: Shouldn't it be always `0`?

hathor/transaction/transaction.py
```diff
@@ -155,6 +155,21 @@ def calculate_height(self) -> int:
...
+        """Calculates a min height the first block confirming this tx needs to have for reward lock verification.o
+
+        Assumes tx has been fully verified (parents and inputs exist and have complete metadata).
+        """
+        assert self.storage is not None
+        if self.is_genesis:
+            return 0
+        return max(chain(
+            # 1) don't drop the min height of any parent tx
+            (self.storage.get_transaction(tx).get_metadata().min_height for tx in self.get_tx_parents()),
```
msbrogli, line 168: I think we have to get the `meta.min_height` of the inputs as well.

hathor/transaction/block.py
```diff
@@ -138,6 +140,17 @@ def calculate_height(self) -> int:
...
+        """Same as height for a block.
+        """
+        assert self.storage is not None
+        return max(chain(
+            # 1) height of parent block + 1
+            [self.calculate_height()],
```
msbrogli, line 149: I got your logic. I don't like the structure itself but it might work. Anyway, why did you include the current height?

hathor/transaction/transaction_metadata.py
```diff
@@ -109,6 +109,10 @@ class TransactionMetadata:
     first_block: Optional[bytes]
     height: int
     validation: ValidationState
+    # XXX: this is only used to defer the reward-lock verification from the transaction spending a reward to the first
+    # block that confirming this transaction, it is important to always have this set to be able to distinguish an old
+    # metadata (that does not have this calculated, from a tx with a new format that does have this calculated)
+    min_height: int
```
msbrogli, line 115: Shouldn't we use `None` to distinguish from an old metadata? How do we know whether it is really zero or non-calculated?

hathor/transaction/transaction_metadata.py
```diff
@@ -168,6 +172,9 @@ def __init__(self, spent_outputs: Optional[Dict[int, List[bytes]]] = None, hash:
         # Height
         self.height = height
 
+        # Min height
```
msbrogli, line 175: I feel this commend should explain what it is used for.

hathor/manager.py
```diff
@@ -524,6 +525,22 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
         max_timestamp = max(int(i.begin) for i in can_include_intervals)
         must_include: List[bytes] = []
         assert len(can_include_intervals) > 0, f'invalid timestamp "{timestamp}", no tips found"'
+        if height is not None:
```
msbrogli, line 528: I feel we do not have to it if we handle it properly somewhere else.

msbrogli (changes requested):

hathor/transaction/transaction_metadata.py
jansegre: It should never be zero on this implementation.

hathor/transaction/base_transaction.py
jansegre: Maybe, it depends on when we use it. On this implementation we update this metadata _before_ validating, we don't want to save a potentially invalid transaction/block.

hathor/manager.py
```diff
@@ -524,6 +525,22 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
         max_timestamp = max(int(i.begin) for i in can_include_intervals)
         must_include: List[bytes] = []
         assert len(can_include_intervals) > 0, f'invalid timestamp "{timestamp}", no tips found"'
```
msbrogli, line 527: Maybe we should assert that all candidates are fully validated.

pedroferreira1 (approved):

msbrogli (approved):

PR #302, opened 2021-09-01
feat(sync-v2): improved sync version negotiation

jansegre:
Instead of having a capability that indicates sync-v2 support, we now have a capability that indicates `sync_versions` field support. The `sync_versions` field is added to the `HELLO` state data and is a list of the sync versions supported by the peer. Both peers will have the chance to explicitly indicate which versions of the sync algorithm they support, the chosen version will be the highest version supported by both peers, if there are none the connection is closed.

This change gives more flexibility by allowing a peer to explicitly state if it supports the sync-v1, which will help when phasing it out. But most importantly it allows for experimenting with different sync-v2 versions that don't interfere with each other. We could start provisionally using a `v2-alpha` string, and if we when we make an incompatible change we might use a `v2-alpha2` or `v2-beta` string, and peers already running with the previous string won't try to connect to each other.

It's currently the best time to implement this because no sync-v2 code has been deployed yet, and no peers are forcibly using sync-v2 by mocking sync-v1 (which is what is currently implemented with the unsafe arguments).

Since there is a change in capabilities, and when the capability is not provided the assumed supported list is only sync-v1, this change is also fully backwards compatible.


hathor/p2p/protocol.py
```diff
@@ -155,11 +150,9 @@ def __init__(self, network: str, my_peer: PeerId, connections: Optional['Connect
         self.aborting = False
 
         self.use_ssl: bool = use_ssl
-        self.enable_sync_v1: bool = enable_sync_v1
-        self.enable_sync_v2: bool = enable_sync_v2
 
         # Protocol version is initially unset
```
pedroferreira1, line 154: This is old comment?

hathor/p2p/sync_version.py
```diff
@@ -0,0 +1,50 @@
...
+    #      example, peers using `v2-fake` (which just uses sync-v1) will not connect to peers using `v2-alpha`, and so
+    #      on.
+    V1 = 'v1'
+    V2 = 'v2-fake'  # uses sync-v1 to mock sync-v2
+
+    def __str__(self):
+        return f'sync-{self.value}'
+
+    def get_priority(self) -> int:
+        """Numerical indication used to sort preferred versions, higherr values are preferred over lower values.
```
pedroferreira1, line 32: Typo: higherr

hathor/p2p/protocol.py
```diff
@@ -155,11 +150,9 @@ def __init__(self, network: str, my_peer: PeerId, connections: Optional['Connect
...
-        self.protocol_version = ProtocolVersion.UNSET
+        self.sync_version = None
```
pedroferreira1, line 155: Where do we set this attribute to the one selected?

pedroferreira1 (approved):

hathor/p2p/protocol.py
jansegre: Sort of, but it still applies.

hathor/p2p/sync_version.py
jansegre: Fixed.

hathor/p2p/protocol.py
jansegre: On `hathor/p2p/states/hello.py` line `129`.

hathor/manager.py
```diff
@@ -200,10 +199,8 @@ def __init__(self, reactor: IReactorCore, peer_id: Optional[PeerId] = None, netw
         # List of capabilities of the peer
         if capabilities is not None:
             self.capabilities = capabilities
-        elif enable_sync_v2:
-            self.capabilities = [settings.CAPABILITY_WHITELIST, settings.CAPABILITY_SYNC_V2]
         else:
-            self.capabilities = [settings.CAPABILITY_WHITELIST]
+            self.capabilities = [settings.CAPABILITY_WHITELIST, settings.CAPABILITY_SYNC_VERSION]
```
msbrogli, line 203: Should we move to a constant `DEFAULT_CAPABILITIES`? In this case, if a list of capabilities is given, the caller can choose to extend the default ones or not.

hathor/p2p/manager.py
```diff
@@ -133,6 +139,9 @@ def stop(self) -> None:
         if self.lc_reconnect.running:
             self.lc_reconnect.stop()
 
+    def get_sync_versions(self) -> Set[SyncVersion]:
```
msbrogli, line 142: Missing docstring.

hathor/p2p/protocol.py
```diff
@@ -86,15 +86,10 @@ class WarningFlags(str, Enum):
     aborting: bool
     diff_timestamp: Optional[int]
     idle_timeout: int
-    protocol_version: ProtocolVersion
+    sync_version: Optional[SyncVersion]
```
msbrogli, line 89: Ours or theirs?

hathor/p2p/states/hello.py
```diff
@@ -58,6 +59,21 @@ def _get_hello_data(self) -> Dict[str, Any]:
...
+        return settings.CAPABILITY_SYNC_VERSION in self.protocol.node.capabilities
+
+    def _get_sync_versions(self) -> Set[SyncVersion]:
+        if self._has_sync_version_capability():
+            connections_manager = self.protocol.connections
+            assert connections_manager is not None
+            return connections_manager.get_sync_versions()
+        else:
+            # XXX: this is to make it easy to simulate old behavior if we disable the sync-version capability
+            return {SyncVersion.V1}
```
msbrogli, line 76: Isn't it better to do this in the `connections_manager.get_sync_versions()`?

hathor/p2p/states/hello.py
```diff
@@ -168,3 +169,16 @@ def handle_hello(self, payload: str) -> None:
...
+    """Versions that are not recognized will not be included."""
+    if settings.CAPABILITY_SYNC_VERSION in hello_data['capabilities']:
+        if 'sync_versions' not in hello_data:
+            raise HathorError('protocol error, expected sync_versions field')
+        known_values = set(x.value for x in SyncVersion)
+        recognized_values = set(hello_data['sync_versions']) & known_values
+        return set(SyncVersion(x) for x in recognized_values)
+    else:
+        # XXX: implied value when sync-version capability isn't present
+        return {SyncVersion.V1}
```
msbrogli, line 184: It seems we should return `{SyncVersion.V1} & known_values`. I mean, my code doesn't work because they have different types, but we are filtering in one case and not filtering here. It seems like a bug.

hathor/p2p/states/ready.py
```diff
@@ -73,16 +70,13 @@ def __init__(self, protocol: 'HathorProtocol') -> None:
...
-            self.log.debug('loading fake sync-v2 (actually sync-v1)')
-            sync_factory = connections.sync_v2_factory
-        else:
-            raise HathorError(f'wrong protocol version: {self.protocol.protocol_version}')
-        self.sync_manager: SyncManager = sync_factory.create_sync_manager(self.protocol, reactor=self.reactor)
+
+        sync_version = self.protocol.sync_version
+        self.log.debug(f'loading {sync_version}')
+        if sync_version not in connections.sync_factories:
+            raise HathorError(f'wrong protocol version: {self.protocol.sync_version}')
```
msbrogli, line 77: Why raise an exception? Should we do `self.send_error_and_close_connection(...)` instead?

Or we can create a `connections.get_sync_factory(...)` method that will raise the exception, and here we will capture the exception and close the connection gracefully.

hathor/p2p/sync_version.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def get_priority(self) -> int:
+        """Numerical indication used to sort preferred versions, higher values are preferred over lower values.
+
+        In practice this is used to sort versions.
+        """
+        # XXX: these values are only used internally and in memory, there is no need to keep them consistency, for
+        #      example, if we need more granularity, we can just add a 0 to all values and use the values in between,
+        #      although this shouldn't really be necessary
+        if self is SyncVersion.V1:
```
msbrogli, line 39: Should we create a dict in the class?

msbrogli (changes requested):

hathor/manager.py
jansegre: I think we should.

hathor/p2p/protocol.py
jansegre: This is the version chosen to be used on this protocol. So, both?

hathor/p2p/states/hello.py
jansegre: I agree, as well as `_has_sync_version_capability`.

hathor/p2p/states/hello.py
jansegre: I don't understand. That'd be redundant because `SyncVersion.V1` is always "known", by definition. So even if we made `known_values = set(SyncVersion)` (which would skip the SyncVersion -> str -> SyncVersion), it wouldn't make any difference because `SyncVersion.V1 in SyncVersion` is _always_ `True`.

hathor/p2p/states/ready.py
jansegre: I like `connections.get_sync_factory(...)` better. But this should be a harder error than just a `self.send_error_and_close_connection(...)`. I'd say this should be an assert (at the very least `assert self.protocol.sync_version is not None`, because we negotiated `self.protocol.sync_version` earlier and it shouldn't be possible to negotiate a version we don't have.

hathor/p2p/sync_version.py
jansegre: Makes sense. I think it might not be completely straightforward (because we can't access the class before creating it and accessing the members (`V1`, `V2`) will give their pre-metaclass values (`str`). But I think a dict right after the class would be great. Ideally in a few years when we can use [Python 3.10's match statements](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching) this would be much cleaner IMO.

hathor/manager.py
jansegre: Done.

hathor/p2p/manager.py
jansegre: Done.

hathor/p2p/states/hello.py
jansegre: Done.

hathor/p2p/protocol.py
msbrogli: I agree it still applies.

hathor/manager.py
msbrogli: Great! Thanks!

hathor/p2p/manager.py
msbrogli: Great! Thanks!

hathor/p2p/protocol.py
msbrogli: Should we add a note about it? Maybe `# version chosen to be used on this connection`?

hathor/p2p/states/hello.py
msbrogli: Great! Thanks!

hathor/p2p/states/hello.py
msbrogli: I got it. Should it be in the `SyncVersion` enum?

hathor/p2p/sync_version.py
msbrogli: I agree. This implementation seems good enough. Forget it. Thanks!

hathor/p2p/states/ready.py
jansegre: Done.

msbrogli: Oh! Ok. It seems using an assert is the best way to go.

hathor/p2p/states/hello.py
jansegre: Maybe, it makes sense to be a classmethod there.

msbrogli: Now that I'm looking the whole diff, I think it is good here. It needs the `hello_data` and is specific to this step of the connection.

msbrogli (approved):

hathor/p2p/protocol.py
msbrogli: Thanks!

PR #311, opened 2021-09-16
feat: allow push nft creation tx

pedroferreira1:


tests/utils.py
```diff
@@ -376,6 +380,9 @@ def create_tokens(manager: 'HathorManager', address_b58: Optional[str] = None, m
     script = P2PKH.create_output_script(address)
 
     deposit_amount = get_deposit_amount(mint_amount)
+    if nft_data:
+        # NFT creation needs 0.01 HTR of fee
+        deposit_amount += 1
```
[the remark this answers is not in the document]
pedroferreira1: With the NFT creation we need HTR for 2 different reasons, (i) the deposit HTR for creating the new token amount and (ii) the 0.01 HTR fee to be used as the DataScript output.

This `deposit_amount += 1` is for the 0.01 HTR fee and maybe the name is misleading but I am just using this to select the utxos and calculate the change output considering the correct amount of HTR that will be "lost" in this transaction

pyproject.toml
```diff
@@ -78,6 +78,7 @@ typing-extensions = "^3.7.4"  # indirect dependency, should not be needed on Pyt
 setproctitle = "^1.2.2"
 sentry-sdk = {version = "^1.1.0", optional = true}
 structlog-sentry = {version = "^1.4.0", optional = true}
+hathorlib = {version = "^0.1.0", extras = ["client"]}
```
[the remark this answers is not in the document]
pedroferreira1: Makes sense. Changed here https://github.com/HathorNetwork/hathor-core/pull/311/commits/c9f9b5c53809f582cf2ff4abe745c6381f72d2a9

tests/utils.py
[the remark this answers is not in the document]
jansegre: Oh! Let me see if I understand:

- `deposit_amount` is burnt/locked by the mint
- `0.01 HTR` is used to create an utxo (change output?) that has an unspendable script (which is why it will be "lost") that has the nft data

Does that make sense?

jansegre (approved):

hathor/transaction/resources/push_tx.py
```diff
@@ -90,6 +91,12 @@ def handle_push_tx(self, params: Dict[str, Any]) -> Dict[str, Any]:
                 'can_force': False
             }
 
+        # We are using here the method from lib because the property
+        # to identify a nft creation transaction was created on the lib
+        # to be used in the full node and tx mining service
+        tx_from_lib = lib_tx_or_block_from_bytes(tx_bytes)
```
msbrogli, line 97: Shouldn't we replace the method from hathor-core with this one?

hathor/transaction/resources/push_tx.py
```diff
@@ -98,7 +105,7 @@ def handle_push_tx(self, params: Dict[str, Any]) -> Dict[str, Any]:
                     'message': 'Transaction has an output script that is too big.',
                     'can_force': False,
                 }
-            if not self.allow_non_standard_script:
+            if not self.allow_non_standard_script and not is_nft_creation:
```
msbrogli, line 108: Should we add NFT Creation as part of the standard scripts? I mean, change the `txout.is_standard_script()`? In fact, maybe we should move it to `tx.is_standard()` which will call `txout.is_standard_script()` by default.

Another approach would be to add a `is_standard_tx(...)` in hathorlib and use it here.

msbrogli (changes requested):

tests/utils.py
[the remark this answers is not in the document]
pedroferreira1: Yes, exactly.

hathor/transaction/resources/push_tx.py
pedroferreira1: I think replacing methods from hathor-core with the lib should be a different task. It must be done carefully because I've checked that the lib and hathor core files already have differences.

That's why I decided to use the lib method only in the new part of the code

hathor/transaction/resources/push_tx.py
pedroferreira1: I guess the best solution for this would be to add the method in hathorlib, in order to use it on the full node and tx mining service. Should I create a new PR in the lib?

hathor/transaction/resources/push_tx.py
msbrogli: If they have differences, how do we know it is safe to use this new method here? I mean, applying different rules in different parts of the code might be an attack vector.

hathor/transaction/resources/push_tx.py
msbrogli: I feel this code is still outdated, right?

hathor/transaction/resources/push_tx.py
pedroferreira1: I agree with you but we are still using the same code to validate the transactions, which is the critical part of the code. This new code here is to decide if a tx is standard or not, in order to use our public nodes APIs to push this tx.

I agree with you about the differences and we have an issue to refactor the hathor-core/hathorlib code to be only one. I can do it now if you think it's relevant but I guess it would take some days to have everything done. I think it's okay to continue with this. What do you think?

hathor/transaction/resources/push_tx.py
pedroferreira1: Yes, can you see again, please?

hathor/transaction/resources/push_tx.py
msbrogli: You're right. This method is basically being used to let our public APIs check whether a tx is standard or not. I agree we can move forward. Would you open an issue in hathor-core to get this refactor done in future, please?

msbrogli (approved):

hathor/transaction/resources/push_tx.py
jansegre: Also mypy _should_ be able to complain if we're using a `hathorlib.Transaction` instead of a `hathor.transaction.Transaction` or vice-versa. But I know we can't completely rely on mypy to catch that.

jansegre (approved):

PR #317, opened 2021-09-27
doc: update packages for compiling on ubuntu 20.04

mbnunes:
this packages is necessary to build rocksdb on Ubuntu

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ First, you need to have Python 3.6 installed. If you don't, we recommend you to 
   ```
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt update
-  sudo apt install python3 python3-dev python3-pip build-essential python3-structlog liblz4-dev libbz2-dev libsnappy-dev
+  sudo apt install python3 python3-dev python3-pip build-essential liblz4-dev libbz2-dev libsnappy-dev
   pip install -U poetry
   ```
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ First, you need to have Python 3.6 installed. If you don't, we recommend you to 
   ```
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt update
-  sudo apt install python3.6 python3.6-dev python3.6-pip build-essential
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ First, you need to have Python 3.6 installed. If you don't, we recommend you to
   ```
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt update
-  sudo apt install python3.6 python3.6-dev python3.6-pip build-essential
+  sudo apt install python3 python3-dev python3-pip build-essential python3-structlog liblz4-dev libbz2-dev libsnappy-dev
```
jansegre, line 45: `python3-structlog` should not be needed, it will be installed from poetry

jansegre (changes requested):

jansegre (approved):

README.md
```diff
@@ -42,7 +42,7 @@ First, you need to have Python 3.6 installed. If you don't, we recommend you to
   ```
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt update
-  sudo apt install python3.6 python3.6-dev python3.6-pip build-essential
+  sudo apt install python3 python3-dev python3-pip build-essential liblz4-dev libbz2-dev libsnappy-dev
```
luislhl, line 45: If we use `python3` instead of `python3.6`, we don't have guarantees about what version will be installed. It will depend on the Ubuntu version.

For 20.04, apparently it's Python 3.8.

I believe this shouldn't be a problem, since Python itself should keep backwards compatibility. The possibility of some compatibility errors when installing the dependencies exists.

---
I personally think we should use `pyenv` in the docs, like we did in macOs below.

luislhl (approved):

README.md
jansegre: This is something that needs updating because `python3.6` is not available on Ubuntu 20.04 unless you add a third-party repository. I think `python3` is better, for now at least. since we support Python 3.6 through 3.9.

jansegre: But I agree that `pyenv` would be great.

jansegre:
Manually merged on ea9f5af because I couldn't push to the upstream branch for rebasing/squashing.

PR #339, opened 2021-12-08
chore: drop support for Python 3.6

jansegre:
Acceptance criteria: don't run any tests/build on Python 3.6, remove any mentions to Python 3.6, and instead have all texts (and internal settings) use "Python 3.7 (or later)" (or equivalent).


luislhl (approved):

pedroferreira1 (approved):
It's important that we remember to mention this support drop in the release text

PR #342, opened 2021-12-09
feat(db): testnet-golf and prevent mixing networks

jansegre:
Acceptance criteria: after this version is in use (i.e., after nodes have upgraded), a database used on the mainnet should be rejected immediately when trying to use it for the testnet, and vice-versa; additionally when upgrading mainnet should be implied, but testnet-golf must be explicit (i.e., do not try to "import" a testnet-foxtrot database).

### DIFF
--- a/hathor/transaction/storage/transaction_storage.py
+++ b/hathor/transaction/storage/transaction_storage.py
@@ -108,6 +108,9 @@ class TransactionStorage(ABC):
         # Initialize cache for genesis transactions.
         self._genesis_cache: Dict[bytes, BaseTransaction] = {}
 
+        # Key storage attribute to save if the network stored is the expected network
+        self._network_attribute: str = 'network'
+
         # Key storage attribute to save if the full node is running a full verification
         self._running_full_verification_attribute: str = 'running_full_verification'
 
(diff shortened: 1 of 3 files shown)


luislhl (approved):
That looks great

hathor/transaction/storage/transaction_storage.py
```diff
@@ -898,6 +937,16 @@ def get_value(self, key: str) -> Optional[str]:
         """
         raise NotImplementedError
 
+    def get_network(self) -> Optional[str]:
+        """ Save full verification on storage
```
msbrogli, line 941: Get network from storage?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -898,6 +937,16 @@ def get_value(self, key: str) -> Optional[str]:
...
+        """
+        return self.get_value(self._network_attribute)
+
+    def set_network(self, network: str) -> None:
+        """ Save full verification on storage
```
msbrogli, line 946: Save network to storage?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -458,7 +461,43 @@ def is_empty(self) -> bool:
 
     def pre_init(self) -> None:
         """Storages can implement this to run code before transaction loading starts"""
-        pass
+        self._check_and_set_network()
+
+    def _check_and_set_network(self) -> None:
```
msbrogli, line 466: Missing docstring.

msbrogli (changes requested):

hathor/transaction/storage/transaction_storage.py
```diff
@@ -458,7 +461,43 @@ def is_empty(self) -> bool:
...
+        from hathor.transaction.storage.exceptions import WrongNetworkError
+        network = settings.NETWORK_NAME
+        stored_network = self.get_network()
+        if stored_network is None:
```
msbrogli, line 470: The logic seems great. I don't like these nested if statements because they are hard to read. Not sure if we can refactor for an easier reading.

msbrogli (changes requested):

hathor/transaction/storage/transaction_storage.py
jansegre: Oh! I copied the method and forgot to write the docstring.

hathor/transaction/storage/transaction_storage.py
jansegre: Here too.

hathor/transaction/storage/transaction_storage.py
jansegre: What if I split out the inside of `if stored_network is None` into a `def _try_to_infer_network(self)` method?

hathor/transaction/storage/transaction_storage.py

hathor/transaction/storage/transaction_storage.py

hathor/transaction/storage/transaction_storage.py
jansegre: Refactored.

msbrogli (approved):

hathor/transaction/storage/transaction_storage.py
msbrogli: Great! Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Thanks!

PR #348, opened 2021-12-14
refactor(indexes): Move all indexes to MemoryIndexesManager

msbrogli:
## Acceptance Criteria

1. Move current indexes to `MemoryIndexesManager`.
2. Remove direct access to internal parts of the indexes (because it will have a different implementation soon).
3. Turn `IndexesManager` into an interface that will be implemented by `DiskIndexesManager` in the near future.

### DIFF
--- a/hathor/indexes.py
+++ b/hathor/indexes.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from abc import ABC, abstractmethod
 from collections import defaultdict
 from math import inf
 from typing import TYPE_CHECKING, DefaultDict, Dict, Iterable, List, NamedTuple, Optional, Set, Tuple, cast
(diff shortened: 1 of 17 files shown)


hathor/indexes.py
```diff
@@ -44,57 +44,72 @@ class IndexesManager:
...
+        self.all_tips = TipsIndex()
+        self.block_tips = TipsIndex()
+        self.tx_tips = TipsIndex()
+
+        self.sorted_all = TimestampIndex()
+        self.sorted_blocks = TimestampIndex()
+        self.sorted_txs = TimestampIndex()
+
+        self.addresses: Optional[AddressesIndex] = None
+        self.tokens: Optional[TokensIndex] = None
```
jansegre, line 57: I think all of these should also be here that are currently in `TransactionStorage`:

```
# Initialize cache for genesis transactions.
self._genesis_cache: Dict[bytes, BaseTransaction] = {}

# Cache of block hash by height
self._block_height_index = BlockHeightIndex()

# Direct and reverse dependency mapping (i.e. needs and needed by)
self._dir_dep_index: Dict[bytes, _DirDepValue] = {}
self._rev_dep_index: Dict[bytes, Set[bytes]] = {}
self._txs_with_deps_ready: Set[bytes] = set()

# Needed txs (key: tx missing, value: requested by)
self._needed_txs_index: Dict[bytes, Tuple[int, bytes]] = {}

# Hold txs that have not been confirmed
self._mempool_tips_index: Set[bytes] = set()

# Hold blocks that can be used as the next parent block
# XXX: if there is more than one they must all have the same score, must always have at least one hash
self._parent_blocks_index: Set[bytes] = {BLOCK_GENESIS.hash}
```

msbrogli: Nice catch! I'll move these indexes in a separate PR.

hathor/indexes.py
```diff
@@ -37,64 +38,124 @@ class TransactionIndexElement(NamedTuple):
     hash: bytes
 
 
-class IndexesManager:
-    """ IndexesManager manages all the indexes that we will have in the system
+class IndexesManager(ABC):
```
jansegre, line 41: I think that this isn't the right class to be abstract. Since this isn't introducing any other implementation besides  `MemoryIndexesManager`, I'd prefer if this abstraction wasn't introduced right now, because I want to discuss a better approach of which classes to make abstract and how to construct them. Aside from that, everything else looks great to me.

jansegre (approved):

pedroferreira1 (approved):

hathor/wallet/resources/thin_wallet/address_history.py
```diff
@@ -49,7 +49,7 @@ def render_POST(self, request: Request) -> bytes:
         request.setHeader(b'content-type', b'application/json; charset=utf-8')
         set_cors(request, 'POST')
 
-        if not self.manager.tx_storage.wallet_index:
+        if not self.manager.tx_storage.indexes.wallet_index:
```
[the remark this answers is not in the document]
msbrogli: You're right. It's weird that the tests passed.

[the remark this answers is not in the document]
msbrogli: Added a test to capture this bug, and then fixed the bug.

pedroferreira1 (approved):

hathor/wallet/resources/thin_wallet/address_history.py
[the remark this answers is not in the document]
pedroferreira1: Great, thanks.

hathor/indexes.py
msbrogli: As we discussed, I'll move forward with this abstraction, and we will update it soon.

hathor/indexes.py
msbrogli: I'll mark as resolved and it will be executed soon.

PR #355, opened 2022-01-01
feat(indexes): add RocksDBHeightIndex stored in disk

jansegre:


hathor/indexes/rocksdb_height_index.py
```diff
@@ -0,0 +1,145 @@
...
+if TYPE_CHECKING:  # pragma: no cover
+    import rocksdb
+
+settings = HathorSettings()
+logger = get_logger()
+
+_CF_NAME_HEIGHT_INDEX = b'height-index'
+
+
+class _IndexEntry(NamedTuple):
```
pedroferreira1, line 34: Can we reuse the one from `hathor.indexes.height_index` file here and in the memory height index? If yes, then we should reuse `_IndexEntry` and `BLOCK_GENESIS_ENTRY`

pedroferreira1 (approved):
For me the PR is great, except for the comment I left and maybe add some more docstrings to some internal methods in the new rocksdb height index class

hathor/indexes/rocksdb_height_index.py
jansegre: Good idea. Done.

pedroferreira1 (approved):

hathor/indexes/rocksdb_height_index.py
```diff
@@ -0,0 +1,150 @@
...
+        (height,) = struct.unpack('>I', key)
+        return height
+
+    def _to_value(self, entry: IndexEntry) -> bytes:
+        """ Serialize entry to the value used internally"""
+        import struct
+        assert len(entry.hash) == 32
+        value = bytearray(entry.hash)
+        value.extend(struct.pack('>I', entry.timestamp))
+        assert len(value) == 32 + 4
```
obiyankenobi, line 83: Will this give us `[tx.timestamp][tx.hash]` or `[tx.hash][tx.timestamp]`? The docstring in the beginning says `[tx.timestamp][tx.hash]`.

hathor/indexes/height_index.py
```diff
@@ -0,0 +1,94 @@
...
+            new_block_height = side_chain_block.get_metadata().height
+            assert new_block_height + 1 == block_height
+            block_height = new_block_height
+
+        # Reverse the data because I was adding in the array from the highest block
+        reversed_add_to_index = reversed(add_to_index)
+
+        for item in reversed_add_to_index:
+            # Add it to the index
+            self.add_reorg(item.height, item.hash, item.timestamp)
```
obiyankenobi, line 94: The logic here is definitely simpler, but I wonder if it will have any impact on performance. The deletion of the old heights will occur as a batch, but adding new indexes will be done one by one. Anyway, this is probably a corner case and we shouldn't worry about it. Maybe even there won't be much impact, as the reorgs are likely only a few blocks. Long reorgs should be very rare.

obiyankenobi (approved):

hathor/indexes/height_index.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class _AddToIndexItem(NamedTuple):
+    height: int
+    hash: bytes
+    timestamp: int
+
+
+class HeightIndex(ABC):
+    """Store the block hash for each given height
```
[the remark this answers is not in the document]
obiyankenobi: From the rest of the file, I understand it only stores the height and associated blocks for the best chain.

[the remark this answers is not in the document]
jansegre: Yes. Only the best chain.

hathor/indexes/rocksdb_height_index.py
jansegre: The docstring is wrong. I'm fixing it. It's actually `[tx.hash][tx.timestamp]`, although the order doesn't really matter here.

hathor/indexes/height_index.py
jansegre: This logic is not new at all, it's exactly the same as before, just on a different place. While I agree we could have done some optimization before to avoid adding one-by-one, we didn't do this before. It could be possible on the memory index, but I don't see how to do that on the rocksdb index. Anyway, since we aren't changing anything, it seemed fine for me. But we could take the opportunity to rethink if it makes sense.

hathor/indexes/height_index.py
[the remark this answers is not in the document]
obiyankenobi: Ok, I'd just add it to the comment as well.

hathor/indexes/height_index.py
obiyankenobi: I was actually thinking for the rocks db index. It seems they have batch operations. Anyway, I don't think it's worth doing it. I just brought it up so you'd also think if it was worth it.

jansegre: Ah, got it. I think this could be just a matter of refactor so this operation would be done in a batch instead. It shouldn't be too hard. I think there is a good advantage, not in the sense of performance, it probably won't have any noticeable impact, but it would be slightly better in case of a crash.

jansegre: I thought it would be simpler but since this is on the base class it would need more abstractions. I'll leave it as is right now.

PR #358, opened 2022-01-07
feat(indexes): add RocksDBDepsIndex stored in disk

jansegre:


hathor/indexes/rocksdb_deps_index.py
```diff
@@ -0,0 +1,413 @@
...
+        cousins = list(self._iter_rev_deps_of(tx))
+        for cousin in cousins:
+            key = self._to_key_direct(cousin, tx)
+            value = self._db.get((self._cf, key))
+            if value != new_value:
+                batch.put((self._cf, key), new_value)
+        # XXX: this only makes sense if the tx validation is complete
+        if validation.is_fully_connected():
+            for cousin in cousins:
+                if all(v.is_fully_connected() for _, v in self._iter_deps_of(cousin)):
```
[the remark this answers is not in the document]
jansegre: Because `validation` will most often not be "fully connected", and the if condition `all(v.is_fully_connected() for _, v in self._iter_deps_of(cousin))` is not very cheap. So I think it only makes sense to make those checks when `validation` is complete (aka "fully connected"). I think my `XXX` comment might have been too brief. Do you have a suggestion of a better comment?

hathor/indexes/rocksdb_deps_index.py
```diff
@@ -0,0 +1,413 @@
...
+            assert tx_dep_hash is not None
+            self.log.debug('seek found', tx_dep_hash=tx_dep_hash.hex())
+            validation = self._from_value_direct(value)
+            yield tx_dep_hash, validation
+        self.log.debug('seek end')
+
+    def _del_from_deps(self, tx_hash: bytes, batch: 'rocksdb.WriteBatch') -> None:
+        for tx_dep_hash, _ in self._iter_deps_of(tx_hash):
+            batch.delete((self._cf, self._to_key_direct(tx_hash, tx_dep_hash)))
+            batch.delete((self._cf, self._to_key_direct(tx_dep_hash, tx_hash)))
```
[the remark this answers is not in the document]
jansegre: This is a typo. I'll fix it.

hathor/indexes/rocksdb_deps_index.py
```diff
@@ -0,0 +1,413 @@
...
+from hathor.util import not_none
+
+if TYPE_CHECKING:
+    import rocksdb
+
+    from hathor.transaction.storage import TransactionStorage
+
+logger = get_logger()
+
+_CF_NAME_TOKENS_INDEX = b'deps-index'
```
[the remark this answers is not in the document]
jansegre: Yeah. This is a typo.

hathor/indexes/rocksdb_utils.py
```diff
@@ -85,13 +85,42 @@ def _fresh_cf(self, cf_name: bytes) -> 'rocksdb.ColumnFamilyHandle':
 
 # XXX: should be `Collection[bytes]`, which only works on Python 3.9+
 class RocksDBSimpleSet(Collection, RocksDBIndexUtils):
-    def __init__(self, db: 'rocksdb.DB', log: 'structlog.stdlib.BoundLogger', *, cf_name: bytes) -> None:
+    def __init__(self,
+                 db: 'rocksdb.DB',
+                 log: 'structlog.stdlib.BoundLogger',
+                 *,
+                 prefix: Optional[bytes] = None,
```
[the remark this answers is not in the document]
jansegre: I'll revert this change. I was going to implement it with this structure, but ended up not using it.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        self._txs_with_deps_ready.discard(tx)
+
+    def next_ready_for_validation(self, tx_storage: 'TransactionStorage', *, dry_run: bool = False) -> Iterator[bytes]:
+        if dry_run:
+            cur_ready = self._txs_with_deps_ready.copy()
+        else:
+            cur_ready, self._txs_with_deps_ready = self._txs_with_deps_ready, set()
+        while cur_ready:
+            yield from sorted(cur_ready, key=lambda tx_hash: tx_storage.get_transaction(tx_hash).timestamp)
```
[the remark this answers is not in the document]
jansegre: No, it's an optimization that I forgot to port.

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,96 @@
...
+class DepsIndex(ABC):
+    @abstractmethod
+    def update_validation(self, tx: bytes, validation: ValidationState) -> None:
+        raise NotImplementedError
+
+    # partial=True, formerly tx_storage.add_to_deps_index + tx_storage.add_needed_deps
+    # partial=False, formerly tx_storage.del_from_deps_index
+    @abstractmethod
+    def update(self, tx: BaseTransaction, partial: bool = True) -> None:
+        """Call to remove tx from all reverse dependencies, for example when validation is complete."""
```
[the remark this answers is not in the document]
jansegre: I'll improve the docstrings. The refactor merged some functions and I think I haven't updated the docstrings to reflect that.

hathor/manager.py
```diff
@@ -419,27 +419,22 @@ def _initialize_components(self) -> None:
             try:
                 if self._full_verification:
                     # TODO: deal with invalid tx
-                    if self.tx_storage.is_tx_needed(tx.hash):
```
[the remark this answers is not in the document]
jansegre: Because this changes were made on top of improved-reward-lock + sync-v2 where `tx._height_cache` is not needed anymore. When I ported the changes back to the current `dev` branch, I also removed `tx._height_cache` references.

hathor/indexes/rocksdb_deps_index.py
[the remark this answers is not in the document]
jansegre: Fixed.

hathor/indexes/rocksdb_deps_index.py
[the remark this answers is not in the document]
jansegre: Fixed.

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,96 @@
...
+    """Query database for the validation state of a transaction, returns INITIAL when tx does not exist."""
+    tx_meta = tx_storage.get_metadata(tx)
+    if tx_meta is None:
+        return ValidationState.INITIAL
+    return tx_meta.validation
+
+
+class DepsIndex(ABC):
+    @abstractmethod
+    def update_validation(self, tx: bytes, validation: ValidationState) -> None:
```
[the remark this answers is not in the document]
jansegre: Added.

hathor/indexes/deps_index.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/indexes/rocksdb_deps_index.py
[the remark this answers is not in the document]
pedroferreira1: Now I get it. Maybe you can add just like you've written here

> Because validation will most often not be "fully connected", and the if condition all(v.is_fully_connected() for _, v in self._iter_deps_of(cousin)) is not very cheap. So I think it only makes sense to make those checks when validation is complete (aka "fully connected").

pedroferreira1 (approved):

hathor/indexes/rocksdb_deps_index.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,94 @@
...
+        """Call to update the index when tx's validation state changes."""
+        raise NotImplementedError
+
+    # partial=True, formerly tx_storage.add_to_deps_index + tx_storage.add_needed_deps
```
msbrogli, line 42: Maybe add to the docstring?

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,94 @@
...
+
+def get_validation_state(tx_storage: 'TransactionStorage', tx: bytes) -> ValidationState:
+    """Query database for the validation state of a transaction, returns INITIAL when tx does not exist."""
+    tx_meta = tx_storage.get_metadata(tx)
+    if tx_meta is None:
+        return ValidationState.INITIAL
+    return tx_meta.validation
+
+
+class DepsIndex(ABC):
```
msbrogli, line 36: Add docstring explaining what kind of queries are accelerated by this index and where it's used.

hathor/transaction/base_transaction.py
```diff
@@ -1037,6 +1037,18 @@ def clone(self) -> 'BaseTransaction':
     def get_token_uid(self, index: int) -> bytes:
         raise NotImplementedError
 
+    @abstractmethod
+    def get_requested_from_height(self) -> int:
```
msbrogli, line 1041: It seems completely related to the operation of the index. Why don't you put this inside the index instead of here?

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import TYPE_CHECKING, Iterator, List
+
+from hathor.transaction import BaseTransaction
+from hathor.transaction.transaction_metadata import ValidationState
+
+if TYPE_CHECKING:  # pragma: no cover
+    from hathor.transaction.storage import TransactionStorage
+
+
+INF_HEIGHT: int = 2**32 - 1
```
msbrogli, line 25: Add a comment explaining that we can't increase this number directly since it would generate an error when writing it in the RocksDB index.

msbrogli (changes requested):

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def get_validation_state(tx_storage: 'TransactionStorage', tx: bytes) -> ValidationState:
```
msbrogli, line 28: Should it be part of the `BaseTransaction`?

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+            raise ValueError('partial=False will only accept fully connected transactions')
+        else:
+            self._add_deps(tx)
+            self._add_needed(tx)
+
+    def _add_deps(self, tx: BaseTransaction) -> None:
+        assert tx.hash is not None
+        assert tx.storage is not None
+        # deps are immutable for a given hash
+        _deps = _DirDepValue((dep, get_validation_state(tx.storage, dep)) for dep in tx.get_all_dependencies())
```
msbrogli, line 81: It seems we are caching the state of the transactions. Isn't it better to just hit the database and avoid dirty values?

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        """True if all deps' validation are fully connected."""
+        return all(val.is_fully_connected() for val in self.values())
+
+
+class MemoryDepsIndex(DepsIndex):
+    def __init__(self):
+        self.log = logger.new()
+
+        # Direct and reverse dependency mapping (i.e. needs and needed by)
+        self._dir_dep_index: Dict[bytes, _DirDepValue] = {}
```
msbrogli, line 43: Why don't we cache in a separate dict? This data structure seems more complicated than it has to be.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+                rev_deps.remove(tx)
+            if not rev_deps:
+                del self._rev_dep_index[rev_dep]
+
+    def remove_ready_for_validation(self, tx: bytes) -> None:
+        """ Removes from ready for validation set.
+        """
+        self._txs_with_deps_ready.discard(tx)
+
+    def next_ready_for_validation(self, tx_storage: 'TransactionStorage', *, dry_run: bool = False) -> Iterator[bytes]:
```
msbrogli, line 111: This code seems really confusing. It seems the caller might change `_txs_with_deps_ready`, so you have to copy it before yielding. In order to iterate over the new additions to the `_txs_with_deps_ready`, you need this while loop that updated `cur_ready` at the end. Is that right?

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        else:
+            cur_ready, self._txs_with_deps_ready = self._txs_with_deps_ready, set()
+        while cur_ready:
+            yield from sorted(cur_ready, key=lambda tx_hash: tx_storage.get_transaction(tx_hash).timestamp)
+            if dry_run:
+                cur_ready = self._txs_with_deps_ready - cur_ready
+            else:
+                cur_ready, self._txs_with_deps_ready = self._txs_with_deps_ready, set()
+
+    def iter(self) -> Iterator[bytes]:
```
msbrogli, line 123: Why not `iter_rev_dep()`?

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        return tx in self._needed_txs_index
+
+    def remove_from_needed_index(self, tx: bytes) -> None:
+        self._needed_txs_index.pop(tx, None)
+
+    def get_next_needed_tx(self) -> bytes:
+        # This strategy maximizes the chance to download multiple txs on the same stream
+        # find the tx with highest "height"
+        # XXX: we could cache this onto `needed_txs` so we don't have to fetch txs every time
+        height, start_hash, tx = max((h, s, t) for t, (h, s) in self._needed_txs_index.items())
```
msbrogli, line 151: O(n)? Maybe we should use a separate data structure to keep priorities in order.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+    def _add_needed(self, tx: BaseTransaction) -> None:
+        assert tx.storage is not None
+        tx_storage = tx.storage
+
+        height = tx.get_requested_from_height()
+        self.log.debug('add needed deps', tx=tx.hash_hex, height=height, type=type(tx).__name__)
+        # get_tx_parents is used instead of get_tx_dependencies because the remote will traverse the parent
+        # tree, not # the dependency tree, eventually we should receive all tx dependencies and be able to validate
+        # this transaction
+        for tx_hash in tx.get_tx_parents():
```
msbrogli, line 165: What about `tx.inputs`?

msbrogli (changes requested):

hathor/manager.py
```diff
@@ -492,19 +487,17 @@ def _initialize_components(self) -> None:
                 sys.exit()
 
         # restart all validations possible
-        deps_size = self.tx_storage.count_deps_index()
-        if deps_size > 0:
-            self.log.debug('run pending validations', deps_size=deps_size)
+        if self.tx_storage.indexes.deps.has_needed_tx():
+            self.log.debug('run pending validations')
```
msbrogli, line 491: run or running?

hathor/manager.py
```diff
@@ -778,9 +772,6 @@ def on_new_tx(self, tx: BaseTransaction, *, conn: Optional[HathorProtocol] = Non
 
         # if partial=False (the default) we don't even try to partially validate transactions
         if not partial or (metadata.validation.is_fully_connected() or tx.can_validate_full()):
-            if isinstance(tx, Transaction) and self.tx_storage.is_tx_needed(tx.hash):
-                tx._height_cache = self.tx_storage.needed_index_height(tx.hash)
```
msbrogli, line 782: Why don't we need it anymore?

hathor/manager.py
```diff
@@ -419,27 +419,22 @@ def _initialize_components(self) -> None:
             try:
                 if self._full_verification:
                     # TODO: deal with invalid tx
-                    if self.tx_storage.is_tx_needed(tx.hash):
-                        assert isinstance(tx, Transaction)
-                        tx._height_cache = self.tx_storage.needed_index_height(tx.hash)
```
msbrogli, line 424: Why don't we need it anymore?

hathor/indexes/deps_index.py
jansegre: Do you think so? I actually forgot those two lines there, I had them for all the methods to aid in the refactor (especially because of the sync-v2 rebase situation), but removed after I was done.

hathor/indexes/deps_index.py
jansegre: That would be great. Will do.

hathor/transaction/base_transaction.py
jansegre: It's more ergonomic to have this on the transaction. But I can move this to an isolated standalone function somewhere on the indexes module (maybe `hathor.indexes.deps_index`).

hathor/indexes/deps_index.py
jansegre: OK!

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        # Direct and reverse dependency mapping (i.e. needs and needed by)
+        self._dir_dep_index: Dict[bytes, _DirDepValue] = {}
+        self._rev_dep_index: Dict[bytes, Set[bytes]] = {}
+        self._txs_with_deps_ready: Set[bytes] = set()
+
+        # Needed txs (key: tx missing, value: requested by)
+        self._needed_txs_index: Dict[bytes, Tuple[int, bytes]] = {}
+
+    def update_validation(self, tx: bytes, validation: ValidationState) -> None:
+        if tx in self._rev_dep_index:
```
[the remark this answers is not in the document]
jansegre: Makes sense, this is probably because of how it was refactored I didn't see that a separate `_update_validation` method became useless, will merge them into one.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+    def is_ready(self) -> bool:
+        """True if all deps' validation are fully connected."""
+        return all(val.is_fully_connected() for val in self.values())
+
+
+class MemoryDepsIndex(DepsIndex):
+    def __init__(self):
+        self.log = logger.new()
+
+        # Direct and reverse dependency mapping (i.e. needs and needed by)
```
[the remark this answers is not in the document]
jansegre: There is a very brief general explanation (before the specific part) on `hathor.indexes.rocksdb_deps_indexes`. I'm thinking that can be the base for a general explanation on `hathor.indexes.deps_indexes`, this could either refer to the general explanation or have a copy of that brief explanation. What do you think?

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        # Needed txs (key: tx missing, value: requested by)
+        self._needed_txs_index: Dict[bytes, Tuple[int, bytes]] = {}
+
+    def update_validation(self, tx: bytes, validation: ValidationState) -> None:
+        if tx in self._rev_dep_index:
+            self._update_validation(tx, validation)
+
+    def _update_validation(self, tx: bytes, validation: ValidationState) -> None:
+        """Propagate the new validation state of a given dep."""
+        for cousin in self._rev_dep_index[tx].copy():
```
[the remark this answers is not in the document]
jansegre: We need a copy because `self._del_from_deps` will mutate `self._rev_deps_index` and we never want to iterate while we mutate.

The list of dependencies is never too big (the max is 258 I think) and transactions with too many dependencies should be sparse.

I know it's entirely possible to build a list of changes and apply them in a separate loop, but I think this would be too much in this case.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+    def update(self, tx: BaseTransaction, partial: bool = True) -> None:
+        assert tx.hash is not None
+        assert tx.storage is not None
+        validation = get_validation_state(tx.storage, tx.hash)
+        if validation.is_fully_connected():
+            self._del_from_deps_index(tx.hash)
+        elif not partial:
+            raise ValueError('partial=False will only accept fully connected transactions')
+        else:
+            self._add_deps(tx)
```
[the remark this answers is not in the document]
jansegre: They should be idempotent, although there isn't any test or explicit construction to make sure it is. But it's basically just calling `add` on `set`s and creating those `set`s when they don't exist, so it should be simple enough.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,170 @@
...
+        for cousin in self._rev_dep_index[tx].copy():
+            deps = self._dir_dep_index[cousin]
+            # XXX: this check serves to avoid calling is_ready() when nothing changed
+            if deps[tx] != validation:
+                deps[tx] = validation
+                if deps.is_ready():
+                    self._del_from_deps_index(cousin)
+                    self._txs_with_deps_ready.add(cousin)
+
+    def update(self, tx: BaseTransaction, partial: bool = True) -> None:
```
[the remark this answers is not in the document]
jansegre: It should be too.

hathor/indexes/deps_index.py
jansegre: This is somewhat similar to `BaseTransaction.get_requested_from_height`, although this _could_ be useful for other places (while `get_requested_from_height` is much less likely to be useful in other places) I'm still unsure, so I'm OK either way.

hathor/indexes/memory_deps_index.py
jansegre: I think that is the case. Removing the "direct dependencies+validation_state" portion of this index could be a very good clean up. But it might be best to validate this on top of sync-v2 first.

hathor/indexes/memory_deps_index.py
jansegre: Yeah, maybe, although that might be subjective.

 ```python
map_x: Dict[Key, ValueX]
map_y: Dict[Key, ValueY]
```

To me, seems more complex to maintain than:

```python
map_xy: Dict[Key, Tuple[ValueX, ValueY]]
```

When `map_x` and `map_y` must always have the same keys.

Which is exactly what this dict is doing, except it's using a named tuple for the sake of less verbose typing.

hathor/indexes/memory_deps_index.py
jansegre: This method is designed to allow being called only once and still yield new items that might be added because of external side effects (since the iterator will be consumed specifically while the transactions it yields are validated and as a side effect more transactions become ready for validation).

I agree that it is confusing, because there are some non-obvious caveats, if dry_run=False and you don't consume the whole iterator, the list of "transactions ready" will be cleared and we would not validate some transactions. Although the only place that uses this method consumes the iterator until the end.

These caveats can certainly be remove, although maybe with a small performance loss (currently it will sort the transactions everytime the list is exhausted so it should be something like `O(k.n.log(n))` and `k << n`, so it's more like `O(n.log(n))`, and the trivial fix for the caveat would just take the `min` (or `max`, not sure) of the list at each iteration, so it would be something like `O(m.n)` but since `m ~ n`, it's more like `O(n2)`, but this is really from the top of my head and lacking a lot of explanation, I may be missing something).

The `copy` is only for `dry_run=True`, which I think might not be in use anymore (I think I only used for testing). So it doesn't affect its real use. The `while` is just for allowing the iterator to be used until it is exhausted (otherwise the user of this function would basically need a part of the structure that is inside of this function.

Anyway, I think I remember it better. This function was refactored to simplify the caller, I think the `def method() -> Iterator[bytes]` does this well for the caller, but maybe we can drop the `dry_run` and make some refactors and remove the caveats for this implementation, even if at a slightly less efficient way, since we'll eventually want the rocksdb implementation to be the default (which may be easier to optimize).

hathor/indexes/memory_deps_index.py
jansegre: That seems just as confusing. I think the docstring's explanation isn't great, I'll improve it. But basically this is almost like the "main", most obvious iterator, although it is only used in one place (during initialization), but the point of this iterator is to yield any transaction that is not fully validated in the index, so that we can check if we can step the validation.

This is likely redundant, because if everything is consistent within the index, downloading the "needed txs" and stepping through the "next to validate", would be enough to step through all the transactions.

hathor/indexes/memory_deps_index.py
jansegre: I think that makes sense, a sorted container seems easy enough to add. The rocksdb would also need to be adapted, but it shouldn't be too hard.

hathor/indexes/memory_deps_index.py
jansegre: There is a brief explanation above. But this was from when I had a different understanding about how we should treat inputs. I'm now convinced we should have `tx.inputs` here too. There are a few details about this that are very tied to sync-v2 that could be discussed there though.

hathor/manager.py
jansegre: To me it makes more sense to say `do X` when we decide to do X and before actually doing X, and say `doing X` when we're in the middle of doing X (and possible log `doing X` multiple times). But we don't have a consistent logging style, so I don't mind either way.

hathor/manager.py
jansegre: The implementation is not using `_height_cache` anymore, so this isn't needed.

hathor/manager.py
jansegre: Also because `tx._height_cache` not being in use anymore.

hathor/indexes/deps_index.py
msbrogli: Great. Thanks!

hathor/indexes/deps_index.py
msbrogli: Thanks!

hathor/indexes/manager.py
```diff
@@ -165,6 +171,8 @@ def __init__(self, db: 'rocksdb.DB') -> None:
         self.tokens = None
         self.height = RocksDBHeightIndex(self._db)
         self.mempool_tips = RocksDBMempoolTipsIndex(self._db)
+        # self.deps = RocksDBDepsIndex(self._db)
```
msbrogli, line 174: Should we use this one?

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Can you add to the docstring that it must be idempotent?

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Got it. I feel we should explicitly say it's idempotent in the docstring. And we should add some tests to make sure it's working properly.

hathor/transaction/base_transaction.py
jansegre: I had missed this, I just updated to move `get_requested_from_height` to a standalone function in the `hathor.indexes.deps_index`. What do you think?

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
jansegre: I've refactored both `memory_deps_index` and `rocksdb_deps_index` to not use `_update_validation`.

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
jansegre: I've added an extended explanation including some examples.

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
jansegre: I added the docstring. But I still have to add some tests. While I'm not sure of an useful way to test for idempotency I think some basic interface testing of the deps indexes would be good and I have a much clearer idea of how to do that (in fact all indexes would benefit from a basic interface testing, but I'll leave that for the future).

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/indexes/deps_index.py
jansegre: I ended up removing this function. It made sense before when the index had uses from places where it had the hash of a transaction and needed the validation state. Now that it is refactored in a way in which that does not happen, it ends up being easier to just use `tx.get_metadata().validation` instead of `get_validation_state(tx_hash, tx_storage)` (because we had tx_hash and tx_storage separately).

hathor/indexes/memory_deps_index.py
jansegre: I've removed those, it is significantly simpler now.

hathor/indexes/memory_deps_index.py
jansegre: Since these info is now removed, the proposed change no longer applies.

hathor/indexes/memory_deps_index.py
jansegre: I did not implement that. But I still think it would be great.

hathor/indexes/memory_deps_index.py
jansegre: I've included `tx.inputs` too (indirectly) by using `tx.get_all_dependencies()` instead of `tx.get_tx_parents()`.

hathor/indexes/manager.py
jansegre: Yes. I've reversed this now. I had it because I was only testing MemoryDepsIndex while I finished the interface refactor.

hathor/indexes/deps_index.py
msbrogli: That's fine. Thanks!

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,166 @@
...
+    else:
+        block = tx.storage.get_transaction(first_block)
+        assert isinstance(block, Block)
+        return block.get_metadata().height
+
+
+class DepsIndex(ABC):
+    """ Index of dependencies between transactions
+
+    This index exists to accelerate queries related to the partial validation mechanism needed by sync-v1. More
```
msbrogli, line 55: Sync-v2, right?

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/indexes/deps_index.py
jansegre: Yes, thanks!

hathor/transaction/base_transaction.py
msbrogli: Seems good. Thanks!

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,169 @@
...
+
+class DepsIndex(ABC):
+    """ Index of dependencies between transactions
+
+    This index exists to accelerate queries related to the partial validation mechanism needed by sync-v2. More
+    specifically these queries:
+
+    - Which transactions need to be downloaded next? That is, all the transactions which are a reverse dependency of
+      all the transactions that aren't fully validated;
+    - Which tranasctions can we validate next? That is, all the transactions which are not fully validated but can be
```
msbrogli, line 60: Minor typo "tranasctions".

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Great explanation, thanks!

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Got it. Thanks! I agree the max is 258 (256 inputs + 2 parents, right?).

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
jansegre: These idempotency adds were added.

hathor/indexes/deps_index.py
jansegre: Fixed.

hathor/indexes/deps_index.py
msbrogli: Ok. Thanks!

hathor/indexes/memory_deps_index.py
msbrogli: Great! Thanks!

hathor/indexes/memory_deps_index.py
msbrogli: Makes sense. Thanks!

hathor/manager.py
msbrogli: Ok. Thanks!

hathor/manager.py
msbrogli: Got it. Just to double check: Is it safe to remove it now? Or should we remove it together with the reward lock PR?

hathor/indexes/manager.py
```diff
@@ -133,9 +138,12 @@ def del_tx(self, tx: BaseTransaction, *, remove_all: bool = False, relax_assert:
         if self.tokens:
             self.tokens.del_tx(tx)
 
+        self.deps.del_tx(tx)
```
msbrogli, line 141: Is this idempotent as well?

hathor/indexes/deps_index.py
```diff
@@ -0,0 +1,169 @@
...
+    assert tx.storage is not None
+    if tx.is_block:
+        return tx.get_metadata().height
+    first_block = tx.get_metadata().first_block
+    if first_block is None:
+        # XXX: consensus did not run yet to update first_block, what should we do?
+        #      I'm defaulting the height to `inf` (practically), this should make it heightest priority when
+        #      choosing which transactions to fetch next
+        return INF_HEIGHT
+    else:
```
msbrogli, line 46: You don't need this `else`, right?

tests/tx/test_indexes.py
```diff
@@ -136,6 +136,35 @@ def test_genesis_not_in_mempool(self):
...
             self.assertNotIn(tx, mempool_txs)
 
+    def test_deps_index(self):
+        from hathor.indexes.memory_deps_index import MemoryDepsIndex
+
+        add_new_blocks(self.manager, 5, advance_clock=15)
+        add_blocks_unlock_reward(self.manager)
+
+        # XXX: this test makes use of the internals of the memory deps-index implementation
+        deps_index: MemoryDepsIndex = self.manager.tx_storage.indexes.deps
```
msbrogli, line 146: Shouldn't we test the RocksDB implementation as well?

hathor/indexes/memory_deps_index.py
msbrogli: Leave a comment to get it improved in a future PR (possibly the sync-v2 PR).

hathor/indexes/rocksdb_deps_index.py
```diff
@@ -0,0 +1,329 @@
...
+    def _add_needed(self, tx: BaseTransaction, batch: 'rocksdb.WriteBatch') -> None:
+        assert tx.hash is not None
+        assert tx.storage is not None
+        tx_storage = tx.storage
+
+        height = get_requested_from_height(tx)
+        self.log.debug('add needed deps', tx=tx.hash_hex, height=height, type=type(tx).__name__)
+        # get_tx_parents is used instead of get_tx_dependencies because the remote will traverse the parent tree, not
+        # the dependency tree, eventually we should receive all tx dependencies and be able to validate this tx
+        for tx_dep_hash in tx.get_tx_parents():
```
msbrogli, line 218: I guess you changed it on the memory implementation. You are using the `get_tx_dependencies()`.

hathor/indexes/rocksdb_deps_index.py
```diff
@@ -0,0 +1,329 @@
...
+    def next_ready_for_validation(self, tx_storage: 'TransactionStorage', *, dry_run: bool = False) -> Iterator[bytes]:
+        import rocksdb
+        batch = rocksdb.WriteBatch()
+        ready = self._drain_all_sorted_ready(tx_storage, batch)
+        if not dry_run:
+            self._db.write(batch)
+        while ready:
+            yield from ready
+            batch = rocksdb.WriteBatch()
+            ready = self._drain_all_sorted_ready(tx_storage, batch)
```
msbrogli, line 237: What happens if the full node is stopped after the generator has started, all items have been drained, but the caller hasn't finished consuming the generator? Would it recover nicely after the full node is restarted?

hathor/indexes/rocksdb_deps_index.py
```diff
@@ -0,0 +1,329 @@
...
+            key_any = self._from_key_any(key)
+            if key_any.tag is not _Tag.REVERSE:
+                break
+            tx_dep_hash = key_any.tx_dep_hash
+            assert tx_dep_hash is not None
+            self.log.debug('seek found', tx_dep=tx_dep_hash.hex())
+            yield tx_dep_hash
+            # XXX: do this seek to skip directly to the next tx_dep_hash, otherwise we would have to iterate until the
+            #      found key has a different tx_dep_hash
+            seek_key = incr_key(self._to_key_reverse(tx_dep_hash))
```
msbrogli, line 280: What does `incr_key(...)` do?

msbrogli (changes requested):

hathor/indexes/manager.py
jansegre: While I'm not sure (I'll check it in a bit), I don't believe wider `IndexesManager.del_tx` is meant to be idempotent. So I don't think there should be any problem.

jansegre: The implementations seem to be in fact idempotent, but I also don't think it's intentional, they just happen to be idempotent.

But also taking a better look again into the indexes as a whole, it looks like `IndexesManager.del_tx` _is_ meant to be _sort of_ idempotent, its just to clear at all. I think we could definitely improve the documentation of the indexes and how we expect everything to work.

Right now event the word "idempotent" is only present in the documentation of some of the methods involved in this PR. I think this might not give a clear view to someone new to the code, it might give the impression that only this index in particular require that property, while actually most, if not all, of `add/del` methods from the indexes need it as well.

From what I understand it basically boils down to a transaction being able to be "removed twice" or "added twice" to the indexes. It can _technically_ be "removed twice" when first becoming voided and second being completely removed , although there is no codepath that leads to a transaction being "completely removed". And it can also _techincally_ be "added twice" when first being newly received and second when being unvoided, although that implies it was voided when it was received (which I don't think is possible, I think we add as if it wasn't voided, then run the consensus, which can then void it), however it could also be: 1) received, 2) voided, 3) unvoided, and this would map to: 1) add 2) remove *not all* 3) add, if we didn't have the distinction between "remove *all*" and "remove *not all*", then there wouldn't be a requirement for an "idempotant add" on some indexes, but because we do, at least the `all_tips` and `sorted_all` indexes need an "idempotent add" (since their "remove" will not be called on "remove *not all*").

Anyway, I don't think this is something special to the dependencies indexes, and I'd think it's best to have a front to improve the documentation and testing of the whole indexing mechanism.

hathor/indexes/deps_index.py
jansegre: Right. I'll remove it, but I don't think it matters too much. I'd prefer if we used `black` and have it pick a style to use uniformly. I think I'm more used to do `if: ... return; else: ... return` than `if: ... return; ... return`, but it's just a matter of style anyway.

hathor/indexes/rocksdb_deps_index.py
jansegre: You're right, I'll fix it.

hathor/indexes/rocksdb_deps_index.py
jansegre: `incr_key(x: bytes)` will basically treat `x` as an integer and increment it by 1, it mainly does byte-by-byte increments dealing with overflows, the main usefulness is when navigating keys that are like this:

Consider a key that is `[2 bytes hash][2 bytes timestamp]`, with entries:

- `00feffff`
- `00ff0001`
- `00ff0002`
- `00ffffff`
- `01000000`

If you want all transactions with hash `00ff` in _ascending_ order, you have to use a "forward" iterator, seek to `00ff` and start from there (the first yield will be `00ff0001`), if you want a all of the same transactions in _descending_ order, you have to use a "reverse" iterator, "seek for the prev" of `0100` and start from there (the first yield will be `00ffffff`). This is done a lot in `rocksdb_timestamp_index` and it can be easier to understand taking a look at that implementation.

hathor/indexes/memory_deps_index.py
```diff
@@ -0,0 +1,139 @@
...
+        return start_hash
+
+    def _add_needed(self, tx: BaseTransaction) -> None:
+        """This method is idempotent, because self.update needs it to be indempotent."""
+        assert tx.storage is not None
+        tx_storage = tx.storage
+
+        height = get_requested_from_height(tx)
+        self.log.debug('add needed deps', tx=tx.hash_hex, height=height, type=type(tx).__name__)
+        for tx_hash in tx.get_all_dependencies():
```
pedroferreira1, line 134: In the previous implementation, we were iterating only through the parents here. What changed?

hathor/indexes/memory_deps_index.py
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

hathor/indexes/memory_deps_index.py
msbrogli: Great! Thanks!

hathor/manager.py
msbrogli: Same as in the other conversation thread, right? Thanks!

hathor/indexes/manager.py
msbrogli: Makes a lot of sense. And I agree that we haven't clearly discussed and documented which methods are idempotent. I feel we should create a new issue to do it properly in the future.

hathor/indexes/deps_index.py
msbrogli: I agree. I prefer the second way because I feel the if statement is a guard clause.

pedroferreira1 (approved):

hathor/indexes/rocksdb_deps_index.py
msbrogli: Got it. Thanks for the explanation. In which byte does it start the incrementing? The first or the last byte? For example, `incr_key(b'abcde')` would be `bbcde` or `abcdf`?

I feel we should add it to the comments. Not the forward and reverse iterator part, but why we are using `incr_key(...)` in the first place.

jansegre: `incr_key(b'abcde') -> b'abcdf'`, which is how rocksdb sorts keys, the docstring of `incr_key` is fairly complete IMO

the XXX comment above is the explanation, but maybe it wasn't clear enough

hathor/indexes/rocksdb_deps_index.py
jansegre: Yeah, that might be a problem, I'm not sure if it would recover nicely, and even if moving the `_db.write(batch)` to _after_ the `yield from`, so maybe batching them isn't good for recovery, I also found other recovery problems when testing something else with rocksdb indexes. It might not be so trivial to work around these. Maybe switching back to a memory-only index for these one could be the short-term solution.

hathor/indexes/manager.py
jansegre: Done: https://github.com/HathorNetwork/hathor-core/issues/409

hathor/indexes/memory_deps_index.py
jansegre: Done.

hathor/indexes/deps_index.py
jansegre: Done.

tests/tx/test_indexes.py
jansegre: Done.

hathor/indexes/rocksdb_deps_index.py
jansegre: Done.

hathor/indexes/memory_deps_index.py
jansegre: Although using the parents is sufficient, using all the dependencies leads to a slightly better order and should also reach a known tx sooner (because this only makes sense for sync-v2).

jansegre: I just remembered that because in sync-v2 we don't sync voided blocks, I think there can be cases where a tx is only reachable if you also follow the inputs (that is, using dependencies instead of only parents).

hathor/indexes/rocksdb_deps_index.py
msbrogli: Great! Did you check if any test was failing because of this problem?

tests/tx/test_indexes.py
msbrogli: Thanks!

hathor/indexes/deps_index.py
msbrogli: Thanks!

hathor/indexes/manager.py
msbrogli: Thanks!

hathor/indexes/memory_deps_index.py
msbrogli: Thanks!

hathor/indexes/rocksdb_deps_index.py
msbrogli: Which other problems have you found? In which indexes?

I'm ok to go ahead and merge this PR without this fix. But I feel it is a mandatory fix before we start using the RocksDBDepsIndex class in production. Maybe we should raise an exception on `__init__()` just to make sure we don't use it by mistake.

hathor/indexes/rocksdb_deps_index.py
msbrogli: Got it. Maybe you can state that you will do a reverse traversal (from end to beginning).

jansegre: But it is not a reverse traversal though, `self._to_key_reverse(tx_dep_hash)` is to format `tx_dep_hash` as a key to get its _reverse dependency_. The `incr_key` is used to skip everything under `tx_dep_hash`, which is to say, if it was some list like my previous comment, if `tx_dep_hash=b'00ff'`, the skip will make it so the iterator will get `01000000` on the next iteration (of which we extract `0100` from). That way we don't need to visit `00ff0002` and `00ffffff` and skip them manually, we can let rocksdb's iterator engine skip them.

hathor/indexes/rocksdb_deps_index.py
jansegre: I don't know if it's a problem on other indexes in particular, but when testing with rocksdb indexes and exiting correctly (SIGINT/Ctrl+C) _while it was syncing_, I got a transaction not found at the end of the loading phase because the transaction was added to the indexes/metadata used for the loading (happens when using either the children metadata or the timestamp index, which are the two "fast iterators" that we have). So, there's some path (and it might be very common) that when interrupted, it will not have saved a new transaction, but will have added it to the indexes.

Prohibiting `__init__()` of the class isn't great because we won't be able to test it. Maybe prefixing the name with something like `DoNotUse` is enough, I think it just has to be evident enough to not be missed in a review.

hathor/indexes/rocksdb_deps_index.py
jansegre: No tests fail because of this. I still don't have a clear picture of the specific case that would cause a problem, I just believe there might be one.

msbrogli: Got it. Let's keep an open issue about it so we can get back to it later. I don't feel it's critical but I think we should have it properly tested (even if it means that we can change back to `get_tx_parents()`).

hathor/indexes/rocksdb_deps_index.py
msbrogli: Got it. It's important to keep consistency when gracefully stopping the full node. Should we keep a separate issue to investigate it?

hathor/indexes/memory_deps_index.py
msbrogli: I understood how it works. We have other discussion threads covering potential issues. So I'll close this thread.

hathor/indexes/rocksdb_deps_index.py
msbrogli: I think I have a scenario where `get_tx_parents()` would fail.

The solid lines are for parents, and the dotted lines are for inputs.

image

```
txA.inputs = [txC]
txB.parents = [txA, txC]
```

In this case, txA would never have all dependencies if we don't include txC in the needed list.

hathor/indexes/rocksdb_deps_index.py
jansegre: I think it makes sense. Created #411

hathor/indexes/rocksdb_deps_index.py
msbrogli: I finally understood how it works. Thanks for clarifying. We need to double check that `it.seek(...)` simply walks through the iterator until it finds a match. Otherwise, this method would be `O(m log n)`.

hathor/indexes/rocksdb_deps_index.py
msbrogli: I feel that we should add another method to mark the tx validation as completed (instead of draining the records from disk, moving them to memory and taking the risk of losing it all if the daemon is terminated). What do you think?

hathor/indexes/rocksdb_deps_index.py
jansegre: That's not exactly the case that will trigger the problem, but it's very very close. In your case `txC` _will_ be pulled when we get the parents from `txB`, which in the context of sync-v2 we go through `txB` _before_ going through `txA`. But if `txC` is a tip, then we have the issue. In the context of sync-v2, if `txC` is a tip, and there is no block that confirms it, then we would not try to fetch it by only using the parents.

msbrogli: Hum got it. Nice catch! What if the block confirms A instead of B? In this case, B would be in the mempool.

jansegre: Hmm, yes, if there is no block confirming B, then the block-sync state of sync-v2 (which has to finish first before we start syncing the mempool) will not pull B, so that would in fact lead to the problem.

hathor/indexes/rocksdb_deps_index.py
jansegre: That would work too. I'm to sure that by itself would fix the problem, but it could simplify the implementation of `next_ready_for_validation`. What I imagine is that `remove_ready_for_validation` would become something like `mark_as_complete`, which itself will trickle the updating of the state of the dependencies to check which have become ready to be validated. But we still need a mechanism to not update the list that is currently being iterated through, which I think is the main point that causes issues. The current solution copies the list to prevent modifying it while iterating over it. Whether we calculate when a transaction can be validated inside _or_ outside of this function, it doesn't change the fact that we still have a common "logical list" that we're adding to and removing from in the same loop.

hathor/indexes/rocksdb_deps_index.py
msbrogli: We might just leave a note and investigate it further it we have performance issues here. I'm good to go with the current implementation.

hathor/indexes/rocksdb_deps_index.py
msbrogli: I think we can add a note about it with the link to the issue and go ahead. This index is not in use right now, and we will go through it on the sync-v2 PR.

hathor/indexes/rocksdb_deps_index.py
jansegre: Just pushed it, it got a little bigger than I imagined it would be. Either we really could use some ergonomic improvements to make these, or I'm really out of touch with how to use the ergonomics we have.

hathor/indexes/rocksdb_deps_index.py
msbrogli: Great! Thanks!

msbrogli (approved):

hathor/indexes/rocksdb_deps_index.py
andreabadesso: @jansegre @msbrogli 

We have a real-life example of this situation on the mainnet:

[This transaction](https://explorer.hathor.network/transaction/000000008cd231281dc44c7c7ba8acc4da95848211bb6ba6cd67e45b538fa87f) is confirmed by a block with height [1727402](https://explorer.hathor.network/transaction/00000000000000001936b86fe501adcf29df93c5b0324cff0d432c5179705b4c) and uses funds from [a transaction](https://explorer.hathor.network/transaction/0000000090db1a72b14d7af871f72345689de27739bc374e44f724ba3d121146) that is confirmed by a block with height [1727403](https://explorer.hathor.network/transaction/00000000000000003902ef2592e5d30712c4b249f1eb5623360fc9cd8dfa54b3)

PR #360, opened 2022-01-10
refactor(indexes): simplify concrete managers

jansegre:


pedroferreira1 (approved):

hathor/indexes/manager.py
```diff
@@ -36,6 +32,12 @@
 
 
 class IndexesManager(ABC):
+    """ IndexesManager manages all the indexes that we will have in the system
+
+    The ideia is for the manager to handle all method calls to indexes,
```
obiyankenobi, line 37: `ideia` misspelled.

obiyankenobi (approved):

hathor/indexes/manager.py
jansegre: Fixed.

PR #363, opened 2022-01-11
refactor(indexes): remove parent-blocks-index

jansegre:
This index was supposed to be used on sync-v2, but it will not be needed now and will just get in the way of the new index structure.


pedroferreira1 (approved):
This is not being used even on sync-v2?

obiyankenobi (approved):

PR #364, opened 2022-01-13
chore: add some settings configuration log when starting the node

pedroferreira1:
### Motivation

There are some configurations in the full node that must be customized depending on the use case and to know if the custom configuration worked the user needs to wait until the node full load, which may take a lot of time.

### Acceptance criteria

We should log some configuration values when starting the manager

### Unresolved questions

Should we add more config values? I've added the ones that are commonly customized by the use cases.

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -215,6 +215,11 @@ class HathorManager:
         self.is_started = True
 
         self.log.info('start manager', network=self.network)
+        self.log.info(
+            'Settings configuration',
+            ws_max_subs_addrs_conn=settings.WS_MAX_SUBS_ADDRS_CONN,
+            ws_max_subs_addrs_empty=settings.WS_MAX_SUBS_ADDRS_EMPTY
+        )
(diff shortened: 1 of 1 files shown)


hathor/manager.py
```diff
@@ -215,6 +215,11 @@ def start(self) -> None:
         self.is_started = True
 
         self.log.info('start manager', network=self.network)
+        self.log.info(
+            'Settings configuration',
+            ws_max_subs_addrs_conn=settings.WS_MAX_SUBS_ADDRS_CONN,
+            ws_max_subs_addrs_empty=settings.WS_MAX_SUBS_ADDRS_EMPTY
+        )
```
jansegre, line 222: I think this is weird. Can't we make this a debug log? Or what if we log the settings module path instead, that way anyone using custom settings will know if their settings is actually being used.

pedroferreira1:
Closed because the following PR was implemented to solve the same problem in a better way: https://github.com/HathorNetwork/hathor-core/pull/367

PR #370, opened 2022-01-24
chore: update depedencies and refactor accordingly

jansegre:
Acceptance Criteria:

- Use latest dependencies of at least the most important packages (twisted, cryptography, ...)
- Do not change any behavior
- Make refactors to pass linters and tests regressions
- Optionally add more Python versions to the matrix, since they work now

Docker build tested here: https://github.com/jansegre/hathor-core/runs/4975542239

Notes:

- `json.loads/json.dumps: str->object/object->str` is a little limited (mostly because of `object` but it is also annoying to call encode/decode for working with bytes), these were replaced by `json_loads/json_dumps: str->Dict/Dict->str` or `json_loadb/json_dumpb: bytes->Dict/Dict->bytes`, otherwise use `json.load/json.dump` when dealing directly with files
- `request.args` is `Optional[Dict]`, which is missing `Dict` typing and is inconvenient when it is `None`, a `get_args` function was added to help with that
- `twisted.web.resource.Resource` does not have `openapi` attribute, so `hathor.api_util.Resource` was created to declare it
- the situation with `reactor/clock` is a bit more complicated, `hathor.util.reactor` and `hathor.util.Reactor` were created to help with that (as well as some other changes)
- interafaces from `zope.interfaces` will fail `isinstance` checks, `hathor.util.verified_cast` was made to help with that
- `from multiprocessing.sharedctypes` typing was wrong
- various other small changes, mostly to deal with the fact that twisted is now typed, surprisingly very little is because of the mypy update


Makefile
```diff
@@ -56,7 +55,7 @@ tests-full:
 
 .PHONY: mypy
 mypy:
-	mypy $(mypy_flags) ./hathor ./tests
+	mypy -p hathor -p tests
```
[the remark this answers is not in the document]
pedroferreira1: Are them the ones listed in `pyproject.toml`? In mypy tool?

tests/p2p/test_connections.py
```diff
@@ -26,15 +24,14 @@ def test_connections(self):
...
         manager = HathorManager(self.clock, tx_storage=tx_storage, wallet=wallet)
 
         endpoint = 'tcp://127.0.0.1:8005'
         manager.connections.connect_to(endpoint, use_ssl=True)
 
-        self.assertNotIn(endpoint, manager.connections.iter_not_ready_endpoints())
-        self.assertNotIn(endpoint, manager.connections.iter_ready_connections())
-        self.assertNotIn(endpoint, manager.connections.iter_all_connections())
-
-        shutil.rmtree(tmpdir)
```
pedroferreira1, line 40: No need to remove anymore?

tests/resources/wallet/test_thin_wallet.py
```diff
@@ -65,16 +65,24 @@ def test_post(self):
         tx = Transaction(inputs=[i], outputs=[o])
 
         data_to_sign = tx.get_sighash_all()
+        self.log.debug('XXXXX', x=1)
```
pedroferreira1, line 68: Should we remove the debug logs?

pedroferreira1 (approved):

Makefile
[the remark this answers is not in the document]
jansegre: I ported the flags previously on `$(mypy_flags)` to `pyproject.toml`, this makes it so you can just run `mypy foo/bar.py` to run mypy on a single file with the same checks that would be run when checking the whole project, but much faster, which makes it possible to have a much agile flow for fixing typing errors. The use of `-p hathor` instead of `./hathor` makes no difference though, I could have kept `./hathor` but I was testing `-p hathor` and left it there (same for tests).

.github/workflows/main.yml
```diff
@@ -21,17 +21,16 @@ jobs:
...
-          'os': ['ubuntu-latest', 'macos-latest'],
-          'include': [
-            {'os': 'windows-latest', 'python': '3.7'},
-            # these still have some errors:
-            # {'os': 'windows-latest', 'python': '3.8'},
-            # {'os': 'windows-latest', 'python': '3.9'},
-          ],
+          'python': ['3.7', '3.8', '3.9', '3.10'],
+          'os': ['ubuntu-latest', 'macos-latest', 'windows-latest'],
+          # 'include': [
```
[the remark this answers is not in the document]
jansegre: I intentionally left this commented to remember what to look at next time, could have left a comment. Those ones still fail, not sure why, a tried to debug but it got non-trivial very quickly.

.github/workflows/main.yml
```diff
@@ -83,11 +75,7 @@ jobs:
       run: brew install -q graphviz rocksdb
     - name: Install Poetry
       run: pip -q --no-input install poetry
-    - name: Install Poetry dependencies (with rocksdb)
```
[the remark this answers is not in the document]
jansegre: Not really, but I learnt that I can use a marker to skip it from installing on Windows. The main benefit is that now it won't be required to `-E rocksdb` to `poetry install` and such.

hathor/wallet/resources/thin_wallet/address_history.py
```diff
@@ -51,19 +51,22 @@ def render_POST(self, request: Request) -> bytes:
...
+            return json_dumpb({'success': False})
 
-        raw_body_bytes = request.content.read() or b''
-        raw_body_str = raw_body_bytes.decode('utf-8')
+        assert request.content is not None
+        raw_body = request.content.read() or b''
         try:
-            post_data = json.loads(raw_body_str)
-        except json.JSONDecodeError:
+            print(repr(raw_body))
```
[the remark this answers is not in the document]
jansegre: Oh, good catch! Will remove.

hathor/wallet/resources/thin_wallet/send_tokens.py
```diff
@@ -173,32 +193,34 @@ def _responseFailed(self, err, tx, request):
...
+        context.tx = self.manager.stratum_factory.mined_txs[funds_hash]
         # Delete it to avoid memory leak
         del(self.manager.stratum_factory.mined_txs[funds_hash])
 
-        deferred = threads.deferToThreadPool(reactor, self.manager.pow_thread_pool, self._stratum_thread_verify, tx)
-        deferred.addCallback(self._cb_tx_resolve, request)
-        deferred.addErrback(self._err_tx_resolve, request, 'stratum_resolve')
+        deferred = threads.deferToThreadPool(reactor, self.manager.pow_thread_pool,
+                                             self._stratum_thread_verify, context)
+        deferred.addCallback(self._cb_tx_resolve)
```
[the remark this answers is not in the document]
jansegre: No, `_stratum_thread_verify` returns a `_Context`, so it will become the first argument to `_cb_tx_resolve`. I really wish we could use `async/await` and avoid these convoluted functions and argument passing though.

hathor/wallet/resources/thin_wallet/send_tokens.py
```diff
@@ -173,32 +193,34 @@ def _responseFailed(self, err, tx, request):
...
+        deferred.addErrback(self._err_tx_resolve, 'stratum_resolve')
```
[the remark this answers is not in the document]
jansegre: I'm actually not sure about this one. Because it's an "errback" it will probably not receive the argument returned from the deferred. I'll take a better look.

tests/p2p/test_connections.py
jansegre: We _do_ need to remove it, however doing it on the test itself is bad because if the test fails it will not reach that line. I was going to port it to setUp/tearDown which is where these things should be, but I noticed that we already have a mechanism to do that, anything that is added to `self.tmpdirs` will be removed on `tearDown`, so I did just that.

tests/resources/wallet/test_thin_wallet.py
jansegre: Oh! Good catch, I forgot it from a test I was doing. Will remove.

hathor/wallet/resources/thin_wallet/address_history.py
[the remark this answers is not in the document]
jansegre: Removed.

tests/resources/wallet/test_thin_wallet.py
jansegre: Removed.

hathor/p2p/netfilter/factory.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, connections: 'ConnectionsManager', wrappedFactory: 'IProtocol
         super().__init__(wrappedFactory)
         self.connections = connections
 
-    def buildProtocol(self, addr: IAddress) -> Protocol:
+    def buildProtocol(self, addr: IAddress) -> Optional[Protocol]:
```
msbrogli, line 34: Optional?

hathor/p2p/netfilter/matches.py
```diff
@@ -83,7 +83,7 @@ def match(self, context: 'NetfilterContext') -> bool:
             return False
 
         try:
-            host = ip_address(context.addr.host)
+            host = ip_address(getattr(context.addr, 'host', ''))
```
msbrogli, line 86: Maybe `host unknown`?

hathor/p2p/rate_limiter.py
```diff
@@ -68,17 +68,19 @@ def add_hit(self, key: str, weight: int = 1) -> bool:
...
         """
+        assert self.reactor is not None
+
         if key not in self.keys:
             return True
         (max_hits, window_seconds) = self.keys[key]
 
         if key not in self.hits:
-            self.hits[key] = RateLimiterLimit(weight, self.reactor.seconds())
+            self.hits[key] = RateLimiterLimit(weight, int(self.reactor.seconds()))
```
msbrogli, line 78: Why round down the time? I feel we can work with the float number here.

hathor/p2p/resources/mining.py
```diff
@@ -46,15 +46,18 @@ def render_POST(self, request):
 
             :rtype: bytes
         """
+        raw_data = request.content.read()
+        if raw_data is None:
+            return b'0'
```
msbrogli, line 51: Why would it happen? Add a comment here explaining, please.

hathor/p2p/states/base.py
```diff
@@ -26,7 +27,7 @@
 
 class BaseState:
     protocol: 'HathorProtocol'
-    cmd_map: Dict[ProtocolMessages, Callable[[str], None]]
+    cmd_map: Dict[ProtocolMessages, Union[Callable[[str], None], Callable[[str], Deferred[None]]]]
```
msbrogli, line 30: Arght! Really?

pyproject.toml
```diff
@@ -36,48 +36,50 @@ exclude = ["tests", "tests.*"]
...
-graphviz = "~0.16"
-ipython = "<7.17"  # 7.17 drops Python 3.6 compatibility
-jedi = "<0.18"  # indirect dependency restriction (0.18 breaks ipython)
-mnemonic = "~0.19"
-prometheus_client = "~0.9.0"
-pyopenssl = "~20.0.1"
+configargparse = "~1.5.3"
+cryptography = "~36.0.1"
+graphviz = "~0.19.1"
+ipython = "<8"  # ipython 8.0 drops compatibility with Python 3.7
```
msbrogli, line 60: Should we drop compatibility with Python 3.7 as well? I'm currently using 3.7.10 in my dev environment.

tests/conftest.py
```diff
@@ -1,7 +1,13 @@
 import asyncio
 import os
+import sys
 
 from twisted.internet import asyncioreactor
 
 os.environ['HATHOR_CONFIG_FILE'] = os.environ.get('HATHOR_TEST_CONFIG_FILE', 'hathor.conf.unittests')
+
+if sys.platform == 'win32':
+    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
msbrogli, line 10: Why do we need this? Add a comment explaining, please.

tests/conftest.py
```diff
@@ -1,7 +1,13 @@
...
+    os.environ['HATHOR_TEST_MEMORY_STORAGE'] = 'true'
```
msbrogli, line 11: Why only for Windows?

hathor/wallet/resources/thin_wallet/send_tokens.py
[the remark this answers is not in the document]
msbrogli: Shouldn't we pass `context` to the errback?

msbrogli (changes requested):

hathor/wallet/resources/thin_wallet/send_tokens.py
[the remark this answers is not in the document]
jansegre: I think we should. But I still have to look into the deferred system to better understand how it works.

hathor/p2p/manager.py
```diff
@@ -307,7 +306,7 @@ def reconnect_to_all(self) -> None:
         for peer in list(self.peer_storage.values()):
             self.connect_to_if_not_connected(peer, now)
 
-    def update_whitelist(self) -> Deferred:
+    def update_whitelist(self) -> Deferred[None]:
```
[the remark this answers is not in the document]
msbrogli: I got it. ;)

hathor/p2p/states/base.py
msbrogli: It's ok to go like this. Maybe we should refactor so all methods return `Deferred[None]`.

Dockerfile.pypy
```diff
@@ -1,31 +1,31 @@
...
+RUN poetry build -f wheel
+RUN poetry run pip install dist/hathor-*.whl
 
 # finally: use production .venv from before
 # lean and mean: this image should be about ~50MB, would be about ~470MB if using the whole stage-1
-FROM pypy:$PYTHON-$DEBIAN
+FROM pypy:$PYTHON-slim-$DEBIAN
 RUN apt-get -qy update
-RUN apt-get -qy install libssl1.1 graphviz librocksdb5.17
+RUN apt-get -qy install libssl1.1 graphviz librocksdb6.11
```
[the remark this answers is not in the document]
jansegre: Yes. It is compatible. We've been using rocksdb-6.x for a while now on the non-pypy image, it's just not explicit what version we use.

pyproject.toml
jansegre: At the very least I'd wait a few releases. It is still the default version of the docker image when no version is given, which means is what we use by default on our deploys. Also, but much less relevant, the latest version of PyPy is only compatible with Python-3.7 and Python-3.8, but unfortunately PyPy-Python-3.8 cannot build python-rocksdb, it has something to do with Cython, which may be updated in the future to support that version combination.

tests/conftest.py
jansegre: I'll do. It boils down to the default asyncio loop policy changing on Python-3.8 and twisted not supporting the new default loop policy, so we have to manually set the old default (which is still well supported up to the latest version of Python).

tests/conftest.py
jansegre: Because python-rocksdb does not install on Windows. I considered making this based on whether rocksdb is available or not, but it would have this downside: if rocksdb is accidentally not installed correctly, the tests would fully run, but not cover the parts that we wanted.

hathor/debug_resources.py
```diff
@@ -67,14 +52,15 @@ class DebugRaiseResource(resource.Resource):
     }
     default_msg = 'exception raised for debugging purposes'
 
-    def run(self, exc_cls: BaseException, msg: str) -> None:
-        raise exc_cls(msg)  # type: ignore
+    def run(self, exc_cls: Type[BaseException], msg: str) -> None:
+        raise exc_cls(msg)
 
     def render_GET(self, request):
```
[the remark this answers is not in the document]
jansegre: > Missing typing?

These are not the only ones missing types. I avoided touching these to keep the PR simpler. I'll add them to this file, but I'd prefer to keep the rest as is.

> Do we need to use `get_args(request)` after the right type is placed here?

What do you mean? We already use `get_args(request)` here? Are we looking at different parts of the code?

hathor/debug_resources.py
```diff
@@ -151,22 +137,23 @@ class DebugLogResource(resource.Resource):
     default_log_msg = 'debugging logging system'
 
     def render_GET(self, request):
```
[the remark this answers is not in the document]
jansegre: Same as my reply above.

hathor/manager.py
```diff
@@ -911,11 +908,13 @@ def tx_fully_validated(self, tx: BaseTransaction) -> None:
             self.wallet.on_new_tx(tx)
 
     def listen(self, description: str, use_ssl: Optional[bool] = None) -> None:
-        endpoint = self.connections.listen(description, use_ssl)
+        self.connections.listen(description, use_ssl)
+        # XXX: although endpoint._port has the port it is private/internal, description also has the port
```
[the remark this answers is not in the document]
jansegre: No, mostly because the way endpoint is typed, it may not even have a port depending on the concrete class (for example, it could be listening on a unix socket, which does not have a port).

What do you think about using `getattr(endpoint, `_port`, None)`?

hathor/p2p/factory.py
```diff
@@ -32,7 +33,7 @@ class HathorServerFactory(protocol.ServerFactory):
     """
 
     manager: Optional[ConnectionsManager]
-    protocol = MyServerProtocol
+    protocol: Optional[Type[MyServerProtocol]] = MyServerProtocol
```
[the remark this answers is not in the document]
jansegre: Because the superclass `protocol.ServerFactory` also has a `protocol` attribute, and it is `Optional`, and even though we can use `Type[MyServerProtocol]` which is more specific than the superclass's non-optional type, we cannot remove the `Optional` from the type.

hathor/p2p/netfilter/factory.py
jansegre: I think this came up because of a superclass, but I'll revise it.

hathor/p2p/states/base.py
jansegre: Some of the functions on the map will return a `Deferred`, so this is needed.

hathor/p2p/netfilter/factory.py
jansegre: I just revised it, it's not because of a superclass, we literally return `None` in the body of the function, so the de-facto type is `Optional`. Is it a mistake to return `None`?

hathor/p2p/netfilter/matches.py
jansegre: I just noticed that we check for a `hasattr` a few lines above, so the value here doesn't matter at all, I'll just remove this parameter, since its optional.

hathor/p2p/rate_limiter.py
jansegre: It was annotated with `int` before. But I checked and it seems that I added the `int` annotation which is wrong. I'll change the annotation instead.

hathor/p2p/resources/mining.py
jansegre: I'm not sure how I ended up with that. I meant to check whether request.content is not None. I'll fix this.

hathor/api_util.py
```diff
@@ -99,3 +102,28 @@ def validate_tx_hash(hash_hex: str, tx_storage: TransactionStorage) -> Tuple[boo
...
+
+def get_args(request: Request) -> Dict[bytes, List[bytes]]:
+    """Type-friendly way to access request.args, also always returns a dict instead of None"""
+    args = cast(Optional[Dict[bytes, List[bytes]]], request.args)
+    if args is None:
+        return {}
+    return args
+
+
+def get_arg_default(args: Dict[bytes, List[bytes]], key: str, default: T) -> T:
```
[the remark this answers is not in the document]
jansegre: Done.

hathor/debug_resources.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/debug_resources.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/p2p/netfilter/matches.py
jansegre: Done.

hathor/p2p/rate_limiter.py
jansegre: Done.

hathor/p2p/resources/mining.py
jansegre: Done.

tests/conftest.py
jansegre: Done.

hathor/wallet/resources/thin_wallet/send_tokens.py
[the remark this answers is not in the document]
jansegre: Done.

Dockerfile.pypy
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/api_util.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/debug_resources.py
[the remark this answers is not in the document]
msbrogli: I mean, `get_args(...)` returns the right type for `request.args`. Maybe typing `request` would already give us the right type without the need to do it in a separate method. Anyway, it's ok to do like you did.

hathor/debug_resources.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/p2p/factory.py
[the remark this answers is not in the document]
msbrogli: Ok! Thanks!

hathor/p2p/netfilter/factory.py
msbrogli: Oh! You're right. We do it to refuse a new connection. Thanks!

hathor/p2p/netfilter/matches.py
msbrogli: Great! Thanks!

hathor/p2p/rate_limiter.py
msbrogli: Great! Thanks!

hathor/p2p/resources/mining.py
msbrogli: Ok. Thanks!

hathor/p2p/states/base.py
msbrogli: I see. We can refactor it later. Thanks!

pyproject.toml
msbrogli: Good points. Let's keep it until everything works fine with 3.8+. Thanks!

tests/conftest.py
msbrogli: Got it. Thanks for the explanation.

tests/conftest.py
msbrogli: Interesting. I didn't know about the change in the loop policy on Python 3.8. Thanks!

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: I see. The `listen(...)` method accepts many different descriptions. I feel it's safer to use `getattr(endpoint, '_port', None)`.

hathor/wallet/resources/thin_wallet/send_tokens.py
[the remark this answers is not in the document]
msbrogli: Thanks!

msbrogli (approved):

hathor/p2p/resources/mining.py
jansegre: I just "remembered" how I ended up with that. When I changed to test if `request.content` is None, these tests broke: `pytest tests/resources/p2p/test_mining.py -k 'v1 and invalid_data' -s`, I'm not entirely sure why, it might have to do with this: https://docs.python.org/3/library/io.html#io.RawIOBase.read But I haven't checked if `content` has that base.

jansegre: I've kept the `if request.content is None` check because it popped up when I added annotations to the method. Which is why I thought I meant that in the first place.

hathor/manager.py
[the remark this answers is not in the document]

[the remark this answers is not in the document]

hathor/p2p/resources/mining.py
msbrogli: It's funny that it worked nicely before the upgrade. I feel that your changes are harmless and fix issues related to the upgrade itself.

jansegre: Actually I think that it worked before because it raised an `AttributeError` when trying to access `.decode` from the `None` returned by `request.content.read()`, but because I moved it out of the `try` block, this no longer fell to the `execpt` block and needed its own treatment.

msbrogli: Got it! Better this way. Thanks for diving deep to find out what changed.

PR #375, opened 2022-01-28
docs(indexes): Add in-disk indexes with RocksDB design

jansegre:


pedroferreira1 (approved):

docs/rocksdb-indexes.md
```diff
@@ -0,0 +1,53 @@
+# Summary
+
```
luislhl, line 2: Do you think it makes sense to add a link to the original issue?

It has some discussions that could yield more info

luislhl (approved):

PR #392, opened 2022-04-12
feat(init): speed up initialization

jansegre:
Using the latest snapshot (`data-20220303175324.tar.gz`) as a base, initialization is about 60% faster when using all existing in-disk indexes (some indexes are still in-memory only), and about 120% faster than using all in-disk indexes but rebuilding them on every initialization (the current behavior before this PR):

- cold run (all indexes rebuilt, in-disk will be reused later):
  ```
  2022-05-11 20:55:29 [info     ] [hathor.indexes.manager] loaded                         blocks=2267901 height=2259960 total_dt=~3509s tx_count=2895834 tx_rate=825.1686592546121 txs=627933
  ```` 
- hot run (only in-memory indexes are rebuilt):
  ```
  2022-05-11 21:29:53 [info     ] [hathor.indexes.manager] loaded                         blocks=2267901 height=2259960 total_dt=~1596s tx_count=2895834 tx_rate=1814.152552360921 txs=627933
  ```
- hot run (all in-memory indexes are rebuilt, no in-disk index is used):
  ```
  2022-05-11 22:19:42 [info     ] [hathor.indexes.manager] loaded                         blocks=2267901 height=2259960 total_dt=~2615s tx_count=2895834 tx_rate=1107.2753905973273 txs=627933
  ```

But still in-disk indexes are not enabled by default, they require cli argument `--x-rocksdb-indexes`.

Last bits:

- [x] `hathor/indexes/address_index.py:36` tx_meta.voided_by should not be skipped, add a test for that
- [x] `hathor/indexes/deps_index.py:110` check whether the it really makes sense to filter-out voided txs
- [x] `hathor/indexes/tips_index.py:63` check whether looking at validation.is_final makes sense for all indexes
- [x] `hathor/indexes/manager.py:44` check if soft-voided txs are handled correctly
- [x] `hathor/indexes/manager.py:147` move iterator selector to the storage and only consider DFS sort when a flag is used

### DIFF
--- a/hathor/indexes/address_index.py
+++ b/hathor/indexes/address_index.py
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from typing import TYPE_CHECKING, Iterable, List, Optional
 
 from structlog import get_logger
 
+from hathor.indexes.base_index import BaseIndex
 from hathor.transaction import BaseTransaction
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -25,10 +26,13 @@ if TYPE_CHECKING:  # pragma: no cover
 logger = get_logger()
 
 
-class AddressIndex(ABC):
+class AddressIndex(BaseIndex):
     """ Index of inputs/outputs by address
     """
     pubsub: Optional['PubSubManager']
+
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        self.add_tx(tx)
 
     def publish_tx(self, tx: BaseTransaction, *, addresses: Optional[Iterable[str]] = None) -> None:
         """ Publish WALLET_ADDRESS_HISTORY for all addresses of a transaction.
--- a/hathor/indexes/base_index.py
+++ b/hathor/indexes/base_index.py
@@ -1,0 +1,47 @@
+# Copyright 2021 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from hathor.transaction.base_transaction import BaseTransaction
+
+
+class BaseIndex(ABC):
+    """ All indexes must inherit from this index.
+
+    This class exists so we can interact with indexes without knowing anything specific to its implemented. It was
+    created to generalize how we initialize indexes and keep track of which ones are up-to-date.
+    """
+
+    @abstractmethod
+    def get_db_name(self) -> Optional[str]:
+        """ The returned string is used to generate the relevant attributes for storing an indexe's state in the db.
+
+        If None is returned, the database will not store the index initialization state and they will always be
+        initialized. This is the expected mode that memory-only indexes will use.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        """ When the index needs to be initialized, this function will be called for every tx in topological order.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def force_clear(self) -> None:
+        """ Clear any existing data in the index.
+        """
+        raise NotImplementedError
--- a/hathor/indexes/deps_index.py
+++ b/hathor/indexes/deps_index.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from typing import TYPE_CHECKING, Iterator, List
 
+from hathor.indexes.base_index import BaseIndex
 from hathor.transaction import BaseTransaction, Block
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -48,7 +49,7 @@ def get_requested_from_height(tx: BaseTransaction) -> int:
     return block.get_metadata().height
 
 
-class DepsIndex(ABC):
+class DepsIndex(BaseIndex):
     """ Index of dependencies between transactions
 
     This index exists to accelerate queries related to the partial validation mechanism needed by sync-v2. More
@@ -103,6 +104,12 @@ class DepsIndex(ABC):
     - The "needed" and "ready" concepts should be easier to understand, but are harder to ascii-draw, thus I skipped
       them.
     """
+
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if tx_meta.voided_by:
+            return
+        self.add_tx(tx, partial=False)
 
     @abstractmethod
     def add_tx(self, tx: BaseTransaction, partial: bool = True) -> None:
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -13,11 +13,13 @@
 # limitations under the License.
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Optional
+from enum import Enum, auto
+from typing import TYPE_CHECKING, Iterator, List, Optional, Tuple
 
 from structlog import get_logger
 
 from hathor.indexes.address_index import AddressIndex
+from hathor.indexes.base_index import BaseIndex
 from hathor.indexes.deps_index import DepsIndex
 from hathor.indexes.height_index import HeightIndex
 from hathor.indexes.mempool_tips_index import MempoolTipsIndex
@@ -25,6 +27,7 @@ from hathor.indexes.timestamp_index import TimestampIndex
 from hathor.indexes.tips_index import TipsIndex
 from hathor.indexes.tokens_index import TokensIndex
 from hathor.transaction import BaseTransaction
+from hathor.util import progress
 
 if TYPE_CHECKING:  # pragma: no cover
     import rocksdb
@@ -35,12 +38,20 @@ if TYPE_CHECKING:  # pragma: no cover
 logger = get_logger()
 
 
+class _IndexFilter(Enum):
+    ALL = auto()  # block or tx, voided or not
+    ALL_BLOCKS = auto()  # only blocks that are not voided
+    VALID_TXS = auto()  # only transactions that are not voided
+
+
 class IndexesManager(ABC):
     """ IndexesManager manages all the indexes that we will have in the system
 
     The idea is for the manager to handle all method calls to indexes,
     so it will know which index is better to use in each moment
     """
+
+    log = get_logger()
 
     all_tips: TipsIndex
     block_tips: TipsIndex
@@ -56,6 +67,39 @@ class IndexesManager(ABC):
     addresses: Optional[AddressIndex]
     tokens: Optional[TokensIndex]
 
+    def __init_checks__(self):
+        """ Implementations must call this at the **end** of their __init__ for running ValueError checks."""
+        # check if every index has a unique db_name
+        indexes_db_names = set()
+        for index in self.iter_all_indexes():
+            index_db_name = index.get_db_name()
+            if index_db_name is None:
+                continue
+            if index_db_name in indexes_db_names:
+                raise ValueError(f'duplicate index name "{index_db_name}", already in use by another index')
+            indexes_db_names.add(index_db_name)
+
+    def iter_all_indexes(self) -> Iterator[BaseIndex]:
+        """ Iterate over all of the indexes abstracted by this manager, hiding their specific implementation details"""
+        for _, index in self._iter_all_indexes_with_filter():
+            yield index
+
+    def _iter_all_indexes_with_filter(self) -> Iterator[Tuple[_IndexFilter, BaseIndex]]:
+        """ Same as `iter_all_indexes()`, but includes a filter for what transactions an index is interested in."""
+        yield _IndexFilter.ALL, self.all_tips
+        yield _IndexFilter.ALL_BLOCKS, self.block_tips
+        yield _IndexFilter.VALID_TXS, self.tx_tips
+        yield _IndexFilter.ALL, self.sorted_all
+        yield _IndexFilter.ALL_BLOCKS, self.sorted_blocks
+        yield _IndexFilter.VALID_TXS, self.sorted_txs
+        yield _IndexFilter.ALL, self.deps
+        yield _IndexFilter.ALL, self.height
+        yield _IndexFilter.ALL, self.mempool_tips
+        if self.addresses is not None:
+            yield _IndexFilter.ALL, self.addresses
+        if self.tokens is not None:
+            yield _IndexFilter.ALL, self.tokens
+
     @abstractmethod
     def enable_address_index(self, pubsub: 'PubSubManager') -> None:
         """Enable address index. It does nothing if it has already been enabled."""
@@ -66,23 +110,74 @@ class IndexesManager(ABC):
         """Enable tokens index. It does nothing if it has already been enabled."""
         raise NotImplementedError
 
-    def _manually_initialize_tips_indexes(self, tx_storage: 'TransactionStorage') -> None:
-        """ Initialize the tips indexes, populating them from a tx_storage that is otherwise complete.
-
-        XXX: this method requires timestamp indexes to be complete and up-to-date with the rest of the database
-        XXX: this method is not yet being used
+    def force_clear_all(self) -> None:
+        """ Force clear all indexes.
         """
-        for tx in tx_storage._topological_sort_timestamp_index():
-            tx_meta = tx.get_metadata()
-            if not tx_meta.validation.is_final():
+        for index in self.iter_all_indexes():
+            index.force_clear()
+
+    def _manually_initialize(self, tx_storage: 'TransactionStorage') -> None:
+        """ Initialize the indexes, checking the indexes that need initialization, and the optimal iterator to use.
+        """
+        from hathor.transaction.genesis import BLOCK_GENESIS
+        from hathor.transaction.storage.transaction_storage import NULL_INDEX_LAST_STARTED_AT
+
+        db_last_started_at = tx_storage.get_last_started_at()
+
+        indexes_to_init: List[Tuple[_IndexFilter, BaseIndex]] = []
+        for index_filter, index in self._iter_all_indexes_with_filter():
+            index_db_name = index.get_db_name()
+            if index_db_name is None:
+                indexes_to_init.append((index_filter, index))
                 continue
+            index_last_started_at = tx_storage.get_index_last_started_at(index_db_name)
+            if db_last_started_at != index_last_started_at:
+                indexes_to_init.append((index_filter, index))
 
-            self.all_tips.add_tx(tx)
+        if indexes_to_init:
+            self.log.debug('there are indexes that need initialization',
+                           indexes_to_init=[i for _, i in indexes_to_init])
+        else:
+            self.log.debug('there are no indexes that need initialization')
 
+        # make sure that all the indexes that we're rebuilding are cleared
+        for _, index in indexes_to_init:
+            index_db_name = index.get_db_name()
+            if index_db_name:
+                tx_storage.set_index_last_started_at(index_db_name, NULL_INDEX_LAST_STARTED_AT)
+            index.force_clear()
+
+        block_count = 0
+        tx_count = 0
+        latest_timestamp = BLOCK_GENESIS.timestamp
+        first_timestamp = BLOCK_GENESIS.timestamp
+
+        for tx in progress(tx_storage.topological_iterator(), log=self.log):
+            # XXX: these would probably make more sense to be their own simple "indexes" instead of how it is here
+            latest_timestamp = max(tx.timestamp, latest_timestamp)
+            first_timestamp = min(tx.timestamp, first_timestamp)
             if tx.is_block:
-                self.block_tips.add_tx(tx)
-            elif not tx_meta.voided_by:
-                self.tx_tips.add_tx(tx)
+                block_count += 1
+            else:
+                tx_count += 1
+
+            tx_meta = tx.get_metadata()
+
+            # feed each transaction to the indexes that they are interested in
+            for index_filter, index in indexes_to_init:
+                if index_filter is _IndexFilter.ALL:
+                    index.init_loop_step(tx)
+                elif index_filter is _IndexFilter.ALL_BLOCKS:
+                    if tx.is_block:
+                        index.init_loop_step(tx)
+                elif index_filter is _IndexFilter.VALID_TXS:
+                    # XXX: all indexes that use this filter treat soft-voided as voided, nothing special needed
+                    if tx.is_transaction and not tx_meta.voided_by:
+                        index.init_loop_step(tx)
+                else:
+                    assert False, 'impossible filter'
+
+        tx_storage._update_caches(block_count, tx_count, latest_timestamp, first_timestamp)
 
     def add_tx(self, tx: BaseTransaction) -> bool:
         """ Add a transaction to the indexes
@@ -162,6 +257,9 @@ class MemoryIndexesManager(IndexesManager):
         self.mempool_tips = MemoryMempoolTipsIndex()
         self.deps = MemoryDepsIndex()
 
+        # XXX: this has to be at the end of __init__, after everything has been initialized
+        self.__init_checks__()
+
     def enable_address_index(self, pubsub: 'PubSubManager') -> None:
         from hathor.indexes.memory_address_index import MemoryAddressIndex
         if self.addresses is None:
@@ -176,8 +274,8 @@ class MemoryIndexesManager(IndexesManager):
 class RocksDBIndexesManager(IndexesManager):
     def __init__(self, db: 'rocksdb.DB') -> None:
         from hathor.indexes.memory_deps_index import MemoryDepsIndex
+        from hathor.indexes.memory_mempool_tips_index import MemoryMempoolTipsIndex
         from hathor.indexes.rocksdb_height_index import RocksDBHeightIndex
-        from hathor.indexes.rocksdb_mempool_tips_index import RocksDBMempoolTipsIndex
         from hathor.indexes.rocksdb_timestamp_index import RocksDBTimestampIndex
 
         self._db = db
@@ -186,15 +284,18 @@ class RocksDBIndexesManager(IndexesManager):
         self.block_tips = TipsIndex()
         self.tx_tips = TipsIndex()
 
-        self.sorted_all = RocksDBTimestampIndex(self._db, cf_name=b'timestamp-sorted-all')
-        self.sorted_blocks = RocksDBTimestampIndex(self._db, cf_name=b'timestamp-sorted-blocks')
-        self.sorted_txs = RocksDBTimestampIndex(self._db, cf_name=b'timestamp-sorted-txs')
+        self.sorted_all = RocksDBTimestampIndex(self._db, 'all')
+        self.sorted_blocks = RocksDBTimestampIndex(self._db, 'blocks')
+        self.sorted_txs = RocksDBTimestampIndex(self._db, 'txs')
 
         self.addresses = None
         self.tokens = None
         self.height = RocksDBHeightIndex(self._db)
-        self.mempool_tips = RocksDBMempoolTipsIndex(self._db)
+        self.mempool_tips = MemoryMempoolTipsIndex()  # use of RocksDBMempoolTipsIndex is very slow and was suspended
         self.deps = MemoryDepsIndex()  # use of RocksDBDepsIndex is currently suspended until it is fixed
+
+        # XXX: this has to be at the end of __init__, after everything has been initialized
+        self.__init_checks__()
 
     def enable_address_index(self, pubsub: 'PubSubManager') -> None:
         from hathor.indexes.rocksdb_address_index import RocksDBAddressIndex
--- a/hathor/indexes/memory_address_index.py
+++ b/hathor/indexes/memory_address_index.py
@@ -30,11 +30,20 @@ logger = get_logger()
 class MemoryAddressIndex(AddressIndex):
     """ Index of inputs/outputs by address
     """
+
+    index: DefaultDict[str, Set[bytes]]
+
     def __init__(self, pubsub: Optional['PubSubManager'] = None) -> None:
-        self.index: DefaultDict[str, Set[bytes]] = defaultdict(set)
         self.pubsub = pubsub
+        self.force_clear()
         if self.pubsub:
             self.subscribe_pubsub_events()
+
+    def get_db_name(self) -> Optional[str]:
+        return None
+
+    def force_clear(self) -> None:
+        self.index = defaultdict(set)
 
     def subscribe_pubsub_events(self) -> None:
         """ Subscribe wallet index to receive voided/winner tx pubsub events
--- a/hathor/indexes/rocksdb_timestamp_index.py
+++ b/hathor/indexes/rocksdb_timestamp_index.py
@@ -38,13 +38,16 @@ class RocksDBTimestampIndex(TimestampIndex, RocksDBIndexUtils):
     It works nicely because rocksdb uses a tree sorted by key under the hood.
     """
 
-    def __init__(self, db: 'rocksdb.DB', *, cf_name: bytes) -> None:
-        RocksDBIndexUtils.__init__(self, db)
+    def __init__(self, db: 'rocksdb.DB', name: str) -> None:
         self.log = logger.new()
+        RocksDBIndexUtils.__init__(self, db, f'timestamp-sorted-{name}'.encode())
+        self._name = name
 
-        # column family stuff
-        self._cf_name = cf_name
-        self._cf = self._fresh_cf(self._cf_name)
+    def get_db_name(self) -> Optional[str]:
+        return f'timestamp_{self._name}'
+
+    def force_clear(self) -> None:
+        self.clear()
 
     def _to_key(self, timestamp: int, tx_hash: Optional[bytes] = None) -> bytes:
         """Make a key for a timestamp and optionally tx_hash, the key represents the membership itself."""
--- a/hathor/indexes/tips_index.py
+++ b/hathor/indexes/tips_index.py
@@ -13,17 +13,18 @@
 # limitations under the License.
 
 from math import inf
-from typing import Dict, Set
+from typing import Dict, Optional, Set
 
 from intervaltree import Interval, IntervalTree
 from structlog import get_logger
 
+from hathor.indexes.base_index import BaseIndex
 from hathor.transaction import BaseTransaction
 
 logger = get_logger()
 
 
-class TipsIndex:
+class TipsIndex(BaseIndex):
     """ Use an interval tree to quick get the tips at a given timestamp.
 
     The interval of a transaction is in the form [begin, end), where `begin` is
@@ -48,8 +49,20 @@ class TipsIndex:
 
     def __init__(self) -> None:
         self.log = logger.new()
+        self.force_clear()
+
+    def get_db_name(self) -> Optional[str]:
+        return None
+
+    def force_clear(self) -> None:
         self.tree = IntervalTree()
-        self.tx_last_interval = {}  # Dict[bytes(hash), Interval]
+        self.tx_last_interval = {}
+
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if not tx_meta.validation.is_final():
+            return
+        self.add_tx(tx)
 
     def add_tx(self, tx: BaseTransaction) -> bool:
         """ Add a new transaction to the index
--- a/hathor/indexes/tokens_index.py
+++ b/hathor/indexes/tokens_index.py
@@ -15,6 +15,7 @@
 from abc import ABC, abstractmethod
 from typing import Iterator, List, NamedTuple, Optional, Tuple
 
+from hathor.indexes.base_index import BaseIndex
 from hathor.transaction import BaseTransaction
 
 
@@ -57,9 +58,15 @@ class TokenIndexInfo(ABC):
         raise NotImplementedError
 
 
-class TokensIndex(ABC):
+class TokensIndex(BaseIndex):
     """ Index of tokens by token uid
     """
+
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if tx_meta.voided_by:
+            return
+        self.add_tx(tx)
 
     @abstractmethod
     def add_tx(self, tx: BaseTransaction) -> None:
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -27,7 +27,7 @@ from hathor import daa
 from hathor.checkpoint import Checkpoint
 from hathor.conf import HathorSettings
 from hathor.consensus import ConsensusAlgorithm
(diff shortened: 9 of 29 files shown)


hathor/indexes/base_index.py
```diff
@@ -0,0 +1,47 @@
...
+        """ The returned string is used to generate the relevant attributes for storing an indexe's state in the db.
+
+        If None is returned, the database will not store the index initialization state and they will always be
+        initialized. This is the expected mode that memory-only indexes will use.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        """ When the index needs to be initialized, this function will be called for each tx in the database.
```
msbrogli, line 39: Should we say it will be called in topological order?

hathor/manager.py
```diff
@@ -385,88 +384,91 @@ def _initialize_components(self) -> None:
...
-                        # If the hash is different from checkpoint hash, we stop the node
-                        self.log.error('Error initializing the node. Checkpoint validation error.')
-                        sys.exit()
-            else:
-                tx_count += 1
-
-            if time.time() - t2 > 1:
-                dt = LogDuration(time.time() - t2)
-                self.log.warn('tx took too long to load', tx=tx.hash_hex, dt=dt)
+        # FIXME: full-verification does not work alongside the new index initialization
```
msbrogli, line 387: Has it been fixed? There's a lot of commented code after this comment.

hathor/manager.py
```diff
@@ -475,7 +477,7 @@ def _initialize_components(self) -> None:
 
         # Check if all checkpoints in database are ok
         my_best_height = self.tx_storage.get_height_best_block()
-        if checkpoint_heights:
+        if checkpoint_heights and False:
```
msbrogli, line 480: Skipping verification?

hathor/manager.py
```diff
@@ -497,15 +499,14 @@ def _initialize_components(self) -> None:
             self.step_validations(depended_final_txs)
             self.log.debug('pending validations finished')
 
-        best_height = self.tx_storage.get_height_best_block()
-        if best_height != h:
-            self.log.warn('best height doesn\'t match', best_height=best_height, max_height=h)
+        # best_height = self.tx_storage.get_height_best_block()
+        # if best_height != h:
+        #     self.log.warn('best height doesn\'t match', best_height=best_height, max_height=h)
```
msbrogli, line 504: Why have you removed this check?

hathor/manager.py
jansegre: This is debug code that wasn't removed.

hathor/manager.py
jansegre: Because there's no `h` variable anymore now, since the loop is inside the indexes and theoretically might not always run (in practice it will because there still are memory indexes that need to be initialized).

hathor/cli/quick_test.py
```diff
@@ -21,15 +21,10 @@ class QuickTest(RunNode):
 
     def register_resources(self, args: Namespace) -> None:
```
msbrogli, line 22: Why did you change this method?

hathor/indexes/address_index.py
```diff
@@ -25,11 +26,17 @@
...
 
-class AddressIndex(ABC):
+class AddressIndex(BaseIndex):
     """ Index of inputs/outputs by address
     """
     pubsub: Optional['PubSubManager']
 
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if tx_meta.voided_by:
```
msbrogli, line 36: Why do you skip voided transactions? What about soft voided ones?

hathor/indexes/deps_index.py
```diff
@@ -104,6 +105,12 @@ class DepsIndex(ABC):
       them.
     """
 
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if tx_meta.voided_by:
```
msbrogli, line 110: Why do you skip voided transactions? What about soft voided ones?

hathor/indexes/memory_address_index.py
```diff
@@ -30,11 +30,20 @@
...
     """
+
+    index: DefaultDict[str, Set[bytes]]
+
     def __init__(self, pubsub: Optional['PubSubManager'] = None) -> None:
-        self.index: DefaultDict[str, Set[bytes]] = defaultdict(set)
         self.pubsub = pubsub
         if self.pubsub:
             self.subscribe_pubsub_events()
+        self.force_clear()
```
msbrogli, line 40: I guess we should call this method at the beginning of `__init__`. I know it's unliked that this index will be created after the full node has already been initialized. So pubsub would be empty anyway. But the call to `subscribe_pubsub_event()` might trigger events right away, which could cause an exception.

hathor/transaction/base_transaction.py
```diff
@@ -832,7 +832,8 @@ def get_metadata(self, *, force_reload: bool = False, use_storage: bool = True)
             #        which requires the use of a storage, this is a workaround that should be fixed, places where this
             #        happens include generating new mining blocks and some tests
             height = self.calculate_height() if self.storage else 0
-            metadata = TransactionMetadata(hash=self.hash, accumulated_weight=self.weight, height=height)
+            score = self.weight if self.is_genesis else 0
+            metadata = TransactionMetadata(hash=self.hash, accumulated_weight=self.weight, height=height, score=score)
```
msbrogli, line 836: Why do we need this change?

hathor/transaction/base_transaction.py
```diff
@@ -844,7 +845,8 @@ def reset_metadata(self) -> None:
         recalculating all metadata.
         """
         assert self.storage is not None
-        self._metadata = TransactionMetadata(hash=self.hash,
+        score = self.weight if self.is_genesis else 0
+        self._metadata = TransactionMetadata(hash=self.hash, score=score,
```
msbrogli, line 849: Why do we need this change?

hathor/indexes/rocksdb_timestamp_index.py
```diff
@@ -38,13 +38,16 @@ class RocksDBTimestampIndex(TimestampIndex, RocksDBIndexUtils):
     It works nicely because rocksdb uses a tree sorted by key under the hood.
     """
 
-    def __init__(self, db: 'rocksdb.DB', *, cf_name: bytes) -> None:
-        RocksDBIndexUtils.__init__(self, db)
+    def __init__(self, db: 'rocksdb.DB', name: str) -> None:
         self.log = logger.new()
+        RocksDBIndexUtils.__init__(self, db, f'timestamp-sorted-{name}'.encode())
```
msbrogli, line 43: Why is it different from `get_db_name()`?

hathor/indexes/tips_index.py
```diff
@@ -48,8 +49,20 @@ class TipsIndex:
...
+        return None
+
+    def force_clear(self) -> None:
         self.tree = IntervalTree()
-        self.tx_last_interval = {}  # Dict[bytes(hash), Interval]
+        self.tx_last_interval = {}
+
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if not tx_meta.validation.is_final():
```
msbrogli, line 63: Should we take this into consideration in other indexes?

hathor/indexes/tokens_index.py
```diff
@@ -57,10 +58,16 @@ def iter_melt_utxos(self) -> Iterator[TokenUtxoInfo]:
...
 
 
-class TokensIndex(ABC):
+class TokensIndex(BaseIndex):
     """ Index of tokens by token uid
     """
 
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        if tx_meta.voided_by:
```
msbrogli, line 67: Why are we skipping voided transactions?

tests/tx/test_indexes3.py
```diff
@@ -62,24 +60,20 @@ def test_tips_index_initialization(self):
         base_block_tips_tree = tx_storage.indexes.block_tips.tree.copy()
         base_tx_tips_tree = tx_storage.indexes.tx_tips.tree.copy()
 
-        # reset the indexes and force a manual initialization
-        tx_storage._reset_cache()
-        self.manager._initialize_components()
+        # reset the indexes, which will force a re-initialization of all indexes
+        tx_storage._manually_initialize()
 
         reinit_all_tips_tree = tx_storage.indexes.all_tips.tree.copy()
```
msbrogli, line 66: Instead of doing each one manually, should we iterate over all indexes?

msbrogli (changes requested):

hathor/indexes/manager.py
```diff
@@ -35,13 +38,21 @@
 logger = get_logger()
 
 
+class _IndexFilter(Enum):
```
msbrogli, line 41: Why don't you move it to a class attribute for each index?

hathor/indexes/manager.py
```diff
@@ -35,13 +38,21 @@
...
+    ALL = auto()  # block or tx, voided or not
+    VALID_BLOCKS = auto()  # only blocks that are not voided
```
msbrogli, line 43: What about soft-voided blocks?

hathor/indexes/manager.py
```diff
@@ -35,13 +38,21 @@
...
+    VALID_TXS = auto()  # only transactions that are not voided
```
msbrogli, line 44: What about soft-voided transactions?

hathor/indexes/manager.py
```diff
@@ -186,16 +292,19 @@ def __init__(self, db: 'rocksdb.DB') -> None:
...
-        self.sorted_txs = RocksDBTimestampIndex(self._db, cf_name=b'timestamp-sorted-txs')
+        self.sorted_all = RocksDBTimestampIndex(self._db, 'all')
+        self.sorted_blocks = RocksDBTimestampIndex(self._db, 'blocks')
+        self.sorted_txs = RocksDBTimestampIndex(self._db, 'txs')
 
         self.addresses = None
         self.tokens = None
         self.height = RocksDBHeightIndex(self._db)
-        self.mempool_tips = RocksDBMempoolTipsIndex(self._db)
+        self.mempool_tips = MemoryMempoolTipsIndex()  # use of RocksDBMempoolTipsIndex is very slow and was suspended
```
msbrogli, line 302: How much slow?

hathor/indexes/manager.py
```diff
@@ -66,23 +110,82 @@ def enable_tokens_index(self) -> None:
...
+                           indexes_to_init=[i for _, i in indexes_to_init])
+        else:
+            self.log.debug('there are no indexes that need initialization')
+
+        iter_tx: Iterator[BaseTransaction]
+        if can_use_timestamp_index:
+            iter_tx = tx_storage._topological_sort_timestamp_index()
+        # XXX: this mostly exist because some tests will break when not using a caching-storage because of metadata
+        #      objects being duplicated which make the topological metadata iterator not work
+        elif tx_storage.is_caching_storage():
```
msbrogli, line 147: Why is this the criteria to use `_topological_sort_metadata`?

hathor/indexes/manager.py
```diff
@@ -66,23 +110,82 @@ def enable_tokens_index(self) -> None:
...
+            iter_tx = tx_storage._topological_sort_metadata()
+        else:
+            iter_tx = tx_storage._topological_sort_dfs()
+
+        # make sure that all the indexes that we're rebuilding are cleared
+        for _, index in indexes_to_init:
+            index_db_name = index.get_db_name()
+            if index_db_name:
+                tx_storage.set_index_last_started_at(index_db_name, 0)
```
msbrogli, line 156: Why `0`? What does it mean?

hathor/indexes/manager.py
```diff
@@ -66,23 +110,82 @@ def enable_tokens_index(self) -> None:
...
-            elif not tx_meta.voided_by:
-                self.tx_tips.add_tx(tx)
+                block_count += 1
+            else:
+                tx_count += 1
+
+            tx_meta = tx.get_metadata()
+
+            # feed each transaction to the indexes that they are interested in
+            for index_filter, index in indexes_to_init:
```
msbrogli, line 176: I feel we should call `init_loop_step(...)` for all transactions, and let the index decide whether it should skip it or not.

hathor/indexes/manager.py
```diff
@@ -176,8 +282,8 @@ def enable_tokens_index(self) -> None:
 class RocksDBIndexesManager(IndexesManager):
     def __init__(self, db: 'rocksdb.DB') -> None:
         from hathor.indexes.memory_deps_index import MemoryDepsIndex
+        from hathor.indexes.memory_mempool_tips_index import MemoryMempoolTipsIndex
```
msbrogli, line 285: Why are we using Memory for mempool instead of RocksDB?

hathor/indexes/manager.py
```diff
@@ -176,8 +282,8 @@ def enable_tokens_index(self) -> None:
 class RocksDBIndexesManager(IndexesManager):
     def __init__(self, db: 'rocksdb.DB') -> None:
         from hathor.indexes.memory_deps_index import MemoryDepsIndex
```
msbrogli, line 284: Add a note that the RocksDB implementation is not ready to be used yet.

hathor/manager.py
```diff
@@ -385,88 +384,91 @@ def _initialize_components(self) -> None:
             for tx in self.tx_storage.get_all_transactions():
                 tx.reset_metadata()
 
-        self.log.debug('load blocks and transactions')
```
msbrogli, line 388: Where was this whole block moved to? If it's just removed, why don't we need it anymore?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -50,6 +50,18 @@ class TransactionStorage(ABC):
...
+    _network_attribute: str = 'network'
+
+    # Key storage attribute to save if the full node is running a full verification
+    _running_full_verification_attribute: str = 'running_full_verification'
+
+    # Key storage attribute to save if the manager is running
+    _manager_running_attribute: str = 'manager_running'
+
+    # Ket storage attribute to save the last time the node started
+    _last_start_attribute: str = 'last_start'
```
msbrogli, line 63: Maybe rename to `db_last_start`?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -643,6 +655,40 @@ def is_running_manager(self) -> bool:
...
+
+        Using this mehtod ensures that the same timestamp is being used and the correct indexes are being selected.
+        """
+        assert self.indexes is not None
+        self.set_last_started_at(timestamp)
+        for index in self.indexes.iter_all_indexes():
+            index_db_name = index.get_db_name()
+            if index_db_name is None:
+                continue
+            self.add_value(index_db_name, str(timestamp))
```
msbrogli, line 690: Maybe add a prefix to avoid key collision?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -677,6 +729,8 @@ def _build_indexes_manager(self) -> IndexesManager:
     def _reset_cache(self) -> None:
         """Reset all caches. This function should not be called unless you know what you are doing."""
         assert self.with_index, 'Cannot reset cache because it has not been enabled.'
+
+        # XXX: these fields would be better of being migrated to the IndexesManager, and probably as proper indexes
```
msbrogli, line 733: I agree! Let's create an issue to get it done.

msbrogli (changes requested):

hathor/cli/quick_test.py
jansegre: I was testing the load times when not using all the indexes and when it finished loading adding a transaction doesn't work. But also the sync itself would also break, depending on what I removed. This changes shouldn't have been committed though, I'll remove them (although it could be somewhat useful if I added an option to change the behavior, I'd rather do it separately, and also optionally not enable the p2p, one of the tests I made caused an alert because of how the node interacted with other nodes on the p2p).

hathor/indexes/address_index.py
jansegre: As far as I understand, this index is not interested in soft voided transactions, even if they are soft voided.

hathor/indexes/deps_index.py
jansegre: I'll have to review this one, I think I should let `add_tx` deal with the tx state.

hathor/indexes/manager.py
jansegre: soft-voided blocks don't exist, soft-voided only apply to transactions

hathor/indexes/manager.py
jansegre: I'm not really sure about that actually. I'll have to review what timestamp-index and tips-index need for this to work correctly.

hathor/indexes/manager.py
jansegre: Mainly because of tips-index and timestamp-index. The interface can't be a classmethod/classattribute because we use different instances of tips-index/timestamp-index with different intended filtering. The interface can be a method/attribute, where tips-index/timestamp-index could have a parameter to store their preference, but I avoided adding this change to those indexes, currently they don't know if they are being fed all txs, only blocks or only txs. I think the way we use tips-index/timestamp-index, even if not straightforward, is consistent, if we _partially_ add the knowledge about what types of transactions they are being fed to inside the implementation, it will significantly confusing IMO.

I'd rather either _fully_ move that knowledge to the index (as in, the manager will not give any special treatment to these indexes), or not move them at all, and let the manager continue being the only place where that knowledge is. I opted for the latter because it's less changes overall, but I do think that an ideal indexes-manager would not know the specifics about the indexes, but that can be refactored later.

hathor/indexes/manager.py
jansegre: It's not like this **is the criteria**, but as I explained in the comment above, some tests will break when using `_topological_sort_metadata` here, it's tests that do not use a caching-storage, and it's related to the sharing of metadata objects. This is _probably_ a bug on how we handle `TransactionMetadata` objects, but I'd rather not deal with this bug right now, since _apparently_ it only affects tests. But I do think this deserves at least an issue to investigate later.

hathor/indexes/manager.py
jansegre: `0` is the smallest timestamp, it's also not a valid timestamp, it's used to indicate that the index is invalid, it also will never match the database timestamp because we reset that to `1` instead. I'll move these two values (`0` and `1`) to a module constant to give it more meaning and add an explanation.

hathor/indexes/manager.py
jansegre: This goes back to the other comment above: https://github.com/HathorNetwork/hathor-core/pull/392/files#r878182397

Although I think that would be better in the end, for now I think this is better. The implementation is also self-contained to `hathor/indexes/manager.py` and no details leak. The way it is right now makes it so the only place that has a decision on what type of transaction goes into an index is private and within `hathor/indexes/manager.py`. I prefer to avoid simultaneously changing both: how indexes are implemented and how indexes are managed. This PR has to change how indexes are managed, and I think it's best to do that while _not_ changing how indexes are implemented.

But I think it would be great if a future PR refactored this and move that to inside the index.

hathor/indexes/manager.py
jansegre: For performance reasons, I added a comment bellow.

hathor/indexes/manager.py
jansegre: There already is a comment bellow.

hathor/indexes/memory_address_index.py
jansegre: OK, that makes sense.

hathor/indexes/rocksdb_timestamp_index.py
jansegre: This name pattern already exist, it doesn't need to be the same. We only use `get_db_name` for generating the name of the attribute used to track an index state, and this pattern is used to name the column-family that the implementation of the index uses. I don't think the two should be tied, an index could in theory use multiple column-families (which is already a refactor that we discussed that could improve the rocksdb-deps-index implementation).

Maybe `get_db_name` could be called `get_attr_name` or `get_index_state_name` or something long those lines.

hathor/indexes/tips_index.py
jansegre: I didn't think too much about this, but possibly we should. I'll review this along with all the other validation related checks.

hathor/indexes/tokens_index.py
jansegre: As far as I understand, this index isn't interested in voided transactions, even soft-voided ones.

hathor/manager.py
jansegre: This block was mostly just simplified. But this diff is outdated.

hathor/transaction/base_transaction.py
jansegre: Basically to have a correct `accumulated_weight` metadata for freshly generated metadata of genesis transactions. This appeared on some tests because of the changes to the manager initialization. And honestly, I'm not completely sure how it was working before.

hathor/transaction/base_transaction.py
jansegre: Same as here the previous comment: https://github.com/HathorNetwork/hathor-core/pull/392/files#r878208656

hathor/transaction/storage/transaction_storage.py
jansegre: But it's also the last time the node itself started. I don't think there's any difference between the "node" and the "node's database", so this can also be used for logging when was the last time the node started.

hathor/transaction/storage/transaction_storage.py
jansegre: I think it makes sense, I'll move the "index" word from all the `get_db_name` to a common prefix, but not to here, this isn't a "centralized" place that controls the attribute name generation. Also, this is probably doing something that is already implemented somewhere else, so I'll refactor that too.

tests/tx/test_indexes3.py
```diff
@@ -62,24 +60,20 @@ def test_tips_index_initialization(self):
...
         reinit_block_tips_tree = tx_storage.indexes.block_tips.tree.copy()
         reinit_tx_tips_tree = tx_storage.indexes.tx_tips.tree.copy()
 
         self.assertEqual(reinit_all_tips_tree, base_all_tips_tree)
         self.assertEqual(reinit_block_tips_tree, base_block_tips_tree)
+        self.assertEqual(len(reinit_tx_tips_tree), len(base_tx_tips_tree))
```
jansegre, line 72: I'll remove this, this was just for helping in debugging.

tests/tx/test_indexes3.py
jansegre: I don't think that works very well. This is implementation specific, and iterating over indexes means erasing the specific class and only looking at the interface.

I don't think it'd be good to make all indexes implement some form of copying or serialization that can be used for comparisons.

Maybe there could be some improvement we could make in this front, but I think it would need better thought, I'd rather avoid this changing this right now.

hathor/indexes/manager.py
jansegre: I'll have to run again to confirm, from what I remember, on my machine, using the rocksdb indexes on everything _except_ the mempool (including using the rocksdb-deps index, which are not used by default) it processed at an average rate of ~2000 tx/s, by only adding the rocksdb-mempool-tips-index (instead of the memory implementation), this rate fell to around ~150 tx/s (more than a 90% slowdown), which makes it take 10x longer to load.

Because of how the mempool-tips-index works, I think it should be able to have a specialized implementation that does not work by having a method called on "every" transaction, but instead will either make a walk or use other indexes _after_ they have been initialized.

hathor/transaction/storage/transaction_storage.py
jansegre: Done: #415

hathor/transaction/storage/transaction_storage.py

hathor/indexes/manager.py

hathor/indexes/memory_address_index.py
jansegre: Done.

hathor/cli/quick_test.py
jansegre: Done.

hathor/indexes/base_index.py
jansegre: Done.

tests/tx/test_indexes3.py
jansegre: Done.

hathor/indexes/base_index.py
msbrogli: Thanks!

hathor/cli/quick_test.py
msbrogli: Thanks!

hathor/indexes/manager.py
msbrogli: I just don't think we should add a criteria to pass the tests. Should we add a parameter to IndexesManager to pick which topological sort it will use?

We should also create an issue to investigate duplicate metadata. Please, add the steps to generate the issue there.

hathor/indexes/manager.py
msbrogli: Have you reviewed those?

hathor/indexes/manager.py
msbrogli: I know we don't have soft-voided blocks. I'm wondering whether our current implementation would support them. But there's nothing to do here anyway. Thanks!

hathor/indexes/manager.py
msbrogli: Got it. I feel we should move these decisions to the indexes and let the manager just coordinate them. I guess we can do it later but still on the speed up initialization project. Would you create an issue to get this done, please?

hathor/indexes/manager.py
msbrogli: Got it. I agree we can do it in a separate PR. Thanks!

hathor/indexes/manager.py
msbrogli: Thanks!

hathor/indexes/manager.py
msbrogli: Thanks!

hathor/indexes/manager.py
msbrogli: Got it. I feel we should create an issue to investigate this slowness. Please add these numbers to the issue as a starting point for the investigation.

tests/tx/test_indexes3.py
msbrogli: Makes sense. Thanks!

msbrogli (changes requested):

hathor/manager.py
msbrogli: I guess it was fixed, right? Just double checking...

hathor/manager.py
msbrogli: I guess you have already removed it. Just double checking.

hathor/manager.py
msbrogli: I feel you fixed it in the last refactor of the initialization, right?

hathor/indexes/address_index.py
msbrogli: As far as I understood, this index has all transactions, including the voided ones.

hathor/indexes/deps_index.py
msbrogli: I'm not sure whether this index should have voided transactions or not. I'd say we should add soft-voided txs so we keep downloading all dependencies correctly.

hathor/indexes/memory_address_index.py
msbrogli: Thanks!

hathor/transaction/base_transaction.py
msbrogli: Make sense. Thanks!

hathor/manager.py
```diff
@@ -507,6 +510,126 @@ def _initialize_components(self) -> None:
...
+                if settings.SOFT_VOIDED_ID not in voided_set:
+                    self.log.error(
+                        'Error initializing node. Your database is not compatible with the current version of the'
+                        ' full node. You must use the latest available snapshot or sync from the beginning.'
+                    )
+                    sys.exit(-1)
+
+                assert {soft_voided_id, settings.SOFT_VOIDED_ID}.issubset(voided_set)
+
+        # TODO: re-implement _full_verification, currently it doesn't work anymore
```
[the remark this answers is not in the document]
jansegre: I'll update the comment, full verification works, but it has a lot of redundant code that could be greatly improved in the future.

hathor/transaction/base_transaction.py
msbrogli: Sure. Thanks!

hathor/indexes/rocksdb_timestamp_index.py
msbrogli: I guess it's ok. We can refactor it later if needed. I was just wondering why they were different.

hathor/indexes/tips_index.py
msbrogli: Have you reviewed this?

hathor/indexes/tokens_index.py
msbrogli: I guess they should. Maybe we should check with Pedro.

hathor/indexes/manager.py
msbrogli: Thanks!

hathor/manager.py
msbrogli: That's fine. I've already reviewed the new changes. Good job! Thanks!

hathor/transaction/storage/transaction_storage.py

hathor/indexes/manager.py
jansegre: I think I got what you mean, now it might make sense to move that to the storage, since it's related to the storage I think it's better to have a flag/toggle there, which would be temporarily used on tests that fail when using `_topological_sort_metadata`. I'd say the storage "owns" the indexes so it might make more sense to have an iterator selector there anyway.

hathor/indexes/manager.py
jansegre: OK.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: I saw that. You kept the whole block that was previously used to initialize the full node. Let's create an issue to get this refactored later, please.

hathor/indexes/manager.py
jansegre: Not deeply enough, I'll add a test for testing how the indexes rebuild when voided txs exist, I think it might help me understand what actually happens.

hathor/indexes/manager.py
jansegre: OK

hathor/indexes/address_index.py
jansegre: Sorry, you're right! I'll add a test to catch this bug and fix it.

hathor/manager.py
jansegre: Yes.

hathor/manager.py
jansegre: Yes.

hathor/manager.py
jansegre: Yes, full-verification now works again.

hathor/indexes/manager.py
jansegre: Done: #418

hathor/indexes/manager.py
jansegre: Done: #419

hathor/manager.py
jansegre: This was updated, the old method is now used fully unmodified when full-verification is enabled.

hathor/manager.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/indexes/manager.py
msbrogli: Thanks!

hathor/indexes/manager.py
msbrogli: Thanks!

hathor/indexes/tips_index.py
jansegre: Yes, and I found out there's a problem: we're not guaranteed to visit not-fully-connected transactions, we will only visit them if they are partially linked through metadata when using the metadata topological iterator, and not at all when using the other iterators.

I think I prefer to address this in sync-v2.

hathor/indexes/manager.py
jansegre: Done.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -674,13 +773,21 @@ def _save_transaction(self, tx: BaseTransaction, *, only_metadata: bool = False)
     def _build_indexes_manager(self) -> IndexesManager:
         return MemoryIndexesManager()
 
-    def _reset_cache(self) -> None:
+    def reset_indexes(self) -> None:
+        assert self.indexes is not None
+        self.indexes.force_clear_all()
+        self._reset_cache(build_indexes=False)
+
+    def _reset_cache(self, *, build_indexes: bool = True) -> None:
```
pedroferreira1, line 781: Should we reset here `_latest_timestamp` and `_first_timestamp` as well that you also set in the `_update_caches` method?

hathor/indexes/manager.py
```diff
@@ -66,23 +110,74 @@ def enable_tokens_index(self) -> None:
...
+                tx_count += 1
+
+            tx_meta = tx.get_metadata()
+
+            # feed each transaction to the indexes that they are interested in
+            for index_filter, index in indexes_to_init:
+                if index_filter is _IndexFilter.ALL:
+                    index.init_loop_step(tx)
+                elif index_filter is _IndexFilter.VALID_BLOCKS:
+                    if tx.is_block:
```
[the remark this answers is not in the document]
jansegre: I was doing that initially, but then realized that the indexes that use that filter actually want the voided blocks, so maybe I should rename the filter.

hathor/indexes/memory_address_index.py
```diff
@@ -30,12 +30,21 @@
...
+    index: DefaultDict[str, Set[bytes]]
+
     def __init__(self, pubsub: Optional['PubSubManager'] = None) -> None:
-        self.index: DefaultDict[str, Set[bytes]] = defaultdict(set)
         self.pubsub = pubsub
+        self.force_clear()
         if self.pubsub:
             self.subscribe_pubsub_events()
 
+    def get_db_name(self) -> Optional[str]:
```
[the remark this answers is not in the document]
jansegre: I see what you mean. I avoided doing anything that would make it easy to forget explicitly adding that method to non-memory indexes. Maybe if all memory indexes inherited from a `BaseMemoryIndex` instead of `BaseIndex`, that could be explicit enough.

hathor/indexes/address_index.py
msbrogli: Have you added it?

hathor/indexes/tips_index.py
msbrogli: But this index is not used only by sync-v2, right? I feel that it is ok not to have transactions that are not connected to the genesis.

hathor/indexes/tokens_index.py
msbrogli: I guess we should discuss it further out of this PR. The goal here is only to keep the same behavior.

hathor/indexes/manager.py
msbrogli: Have you done it?

hathor/manager.py
```diff
@@ -507,6 +510,128 @@ def _initialize_components(self) -> None:
...
+
+        # XXX: last step before actually starting is updating the last started at timestamps
+        self.tx_storage.update_last_started_at(started_at)
+        self.state = self.NodeState.READY
+        self.log.info('ready')
+
+    def _verify_checkpoints(self) -> None:
+        """ Method to verify if all checkpoints that exist in the database have the correct hash and are winners.
+
+        This method needs the essential indexes to be already initialized.
```
msbrogli, line 582: Which ones are the essential ones?

hathor/indexes/tips_index.py
jansegre: You mean it's _not_ ok to have transactions that are not connected the to the genesis, right? This index definitely shouldn't have transactions that are not connected to the genesis. While it is used by the current sync-v2 implementation, it's use is small and will be easy to replace with either data from other indexes or a much simpler index.

But I don't think that is the point. Maybe we could temporarily guard all indexes from receiving transactions that aren't fully connected, or at least make a guard inside the iterator to make sure we're not visiting cases that aren't fully supported, having transactions that aren't fully connected isn't properly supported right now, and I think it will only make sense to fix this support on sync-v2, it's already hard enough to test this without actually using any of the sync-v2 code that actually make use of the stepped validation system.

msbrogli (approved):

hathor/indexes/tokens_index.py
jansegre: I agree. The de-facto behavior is to leave out voided txs, so that's mainly why this should be done here too. But this is certainly something that we can re-visit and discuss outside this PR.

hathor/manager.py
jansegre: All the non-optional, what should I call them? I tried to avoid enumerating all of them. Or rather, it might only _actually_ need a subset of them, but we don't have the notion of splitting them, I mostly meant that this has to be called after the indexes were initialized, but we don't care about the optional indexes.

hathor/indexes/manager.py
jansegre: Yes.

hathor/indexes/manager.py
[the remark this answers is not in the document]
jansegre: Sounds good.

hathor/indexes/address_index.py
msbrogli: Thanks!

hathor/transaction/storage/transaction_storage.py
jansegre: I don't get what you mean,`_reset_cache` does in fact reset `_latest_timestamp` and `_first_timestamp`.

hathor/indexes/manager.py
msbrogli: Just reviewed the code. Thanks!

hathor/indexes/tips_index.py
msbrogli: Agreed. Let's do it in the sync-v2 PR. Thanks!

hathor/manager.py
msbrogli: Got it. That's ok. We could be more explicit here but it's ok to leave to the reader the task to dive deep into the index dependencies.

pedroferreira1 (approved):

hathor/indexes/manager.py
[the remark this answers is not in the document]
jansegre: There is actually a problem with this line. The neither filter name is completely accurate, because neither excluding all voided blocks nor including all voided blocks will _necessarily_ rebuild the index to the same state it was before.

The problematic case is when there are multiple valid heads: they are voided _and_ tips. Which makes it so they are not removed from the block tips index when they are marked as voided, but they are marked as voided anyway.

I'm not completely sure that the workaround should be. I think this is just one more case of a situation that could have been avoided if we never allowed ties. However that change itself must be done more carefully, but I really think we should consider it and prioritize it soon, it's probably possible to do it in a completely backwards way.

We also probably have that bug already, the current initialization does not deal with this correctly (currently we add everything to the indexes and immediately remove what's voided, the remove function will skip the indexes that should not be affected when `remove_all=False`, but that's it).

For now I prefer to have some sort of workaround to make sure the index is restored to the desired state, although I'm not really sure what's best. I tested using this condition: `tx.is_block and (not tx_meta.voided_by or not tx_meta.children)`, it works for all test cases, but it's still wrong because old voided tips that aren't currently heads will end up as block tips.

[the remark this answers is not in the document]
jansegre: From what we talked the de-facto implementation _is_ to add all blocks, voided or not, to the index. But there's also some small inconsistency in how add and remove blocks to the blocks_tips, but this can be dealt separately.

PR #393, opened 2022-04-13
feat(storage): add fast topological iterator

jansegre:
## Measure example

I ran this on ipython after initializing the indexes, on a database with about 2.9M objects.

```
In [9]: def countit(it):
   ...:     return sum(1 for _ in it)
   ...:

In [10]: %time countit(tx_storage._topological_fast())
CPU times: user 11min 57s, sys: 0 ns, total: 11min 57s
Wall time: 11min 57s
Out[10]: 2895834

In [11]: %time countit(tx_storage._topological_sort())
CPU times: user 26min 40s, sys: 2.45 s, total: 26min 43s
Wall time: 26min 43s
Out[11]: 2895834

In [12]: %time countit(tx_storage._topological_fast())
CPU times: user 12min 4s, sys: 19.5 ms, total: 12min 4s
Wall time: 12min 4s
Out[12]: 2895834

In [13]: %time countit(tx_storage._topological_sort())
CPU times: user 23min 35s, sys: 3.75 s, total: 23min 39s
Wall time: 23min 39s
Out[13]: 2895834
```

There's some variation, but it's roughly about twice as fast, also not shown here is memory usage, which should be constant versus some significant amount that grows with the number of objects.

## Acceptance criteria

- do not change any current behavior
- have a new method for iterating that will not be used right now
- must iterate through all transactions in topological order
- add test for the current iterator
- avoid fixing possible newly discovered bugs in this PR

### DIFF
--- a/tests/simulation/test_simulator.py
+++ b/tests/simulation/test_simulator.py
@@ -6,6 +6,58 @@ from tests.simulation.base import SimulatorTestCase
 
 
 class BaseRandomSimulatorTestCase(SimulatorTestCase):
+    def test_topological_iterators(self):
+        # BEGIN: build random blockchain
+        manager1 = self.create_peer()
+
+        # FIXME: this second peer is only needed because of some problem on the simulator
+        manager2 = self.create_peer()
+        conn12 = FakeConnection(manager1, manager2, latency=0.150)
+        self.simulator.add_connection(conn12)
+        self.simulator.run(10)
+
+        miner1 = self.simulator.create_miner(manager1, hashpower=100e6)
+        miner1.start()
+        self.simulator.run(10)
+
+        miner2 = self.simulator.create_miner(manager1, hashpower=100e6)
+        miner2.start()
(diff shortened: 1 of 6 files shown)


luislhl (approved):

tests/simulation/test_simulator.py
```diff
@@ -17,6 +69,9 @@ def test_one_node(self):
         gen_tx1.start()
         self.simulator.run(60 * 60)
 
+        # FIXME: the setup above produces 0 new blocks and transactions
+        # self.assertGreater(manager1.tx_storage.get_count_tx_blocks(), 3)
```
msbrogli, line 73: Should we create an issue to handle it later?

msbrogli (approved):

tests/simulation/test_simulator.py
jansegre: Definitely, just did it: #397

PR #395, opened 2022-04-18
feat(indexes): add manual tips-index initialization

jansegre:
## Measure example

A standalone test suggests that there initialization time should drop by about half:

This was run on a modified version of `HathorManager.start` to not reset the existing rocksdb indexes and to use the `_topological_fast` iterator:

```
In [1]: %time manager.start()
2022-04-14 19:19:33 [info     ] [hathor.manager] start manager                  network=mainnet
2022-04-14 19:19:33 [info     ] [hathor.p2p.manager] update whitelist
2022-04-14 19:19:33 [info     ] [hathor.manager] initialize
2022-04-14 19:49:42 [info     ] [hathor.manager] ready                          blocks=0 height=0 total_dt=<1ms tx_count=0 tx_rate=? txs=0
2022-04-14 19:49:42 [info     ] [hathor.p2p.manager] listen on                      endpoint=tcp:9101
CPU times: user 30min 3s, sys: 4.51 s, total: 30min 7s
Wall time: 30min 8s
```

While on the same hardware initialization using all rocksdb indexes takes about double the time:

```
In [1]: manager.start()
2022-04-14 18:09:54 [info     ] [hathor.manager] start manager                  network=mainnet
2022-04-14 18:09:54 [info     ] [hathor.p2p.manager] update whitelist
2022-04-14 18:09:54 [info     ] [hathor.manager] initialize
2022-04-14 18:12:28 [info     ] [hathor.manager] load transactions...           dt=~30s height=58185 latest_ts=2020-01-24 01:40:05 total=58539 tx_new=58539 tx_rate=1951.270225979218
...
2022-04-14 19:11:45 [info     ] [hathor.manager] load transactions...           dt=~30s height=2259960 latest_ts=2022-03-03 17:53:24 total=2895818 tx_new=123 tx_rate=4.06934933520013
2022-04-14 19:11:55 [info     ] [hathor.manager] ready                          blocks=2267901 height=2259960 total_dt=~3716s tx_count=2895834 tx_rate=779.1259755028108 txs=627933
2022-04-14 19:11:55 [info     ] [hathor.p2p.manager] listen on                      endpoint=tcp:9101
```

However this is just a peek into what should be expected in the near future, this PR does not include everything required to enable this speed up, that will be left to the next PR on this series.

## Acceptance criteria

- do not change any current behavior
- have a new method for initializing the tips indexes that will not be used right now
- the new initializer must produce the same exact interval trees
- test both the current and new initializers
- avoid fixing possible newly discovered bugs in this PR

### DIFF
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -29,6 +29,7 @@ if TYPE_CHECKING:  # pragma: no cover
     import rocksdb
 
     from hathor.pubsub import PubSubManager
+    from hathor.transaction.storage import TransactionStorage
 
 logger = get_logger()
 
@@ -62,6 +63,24 @@ class IndexesManager(ABC):
     def enable_tokens_index(self) -> None:
         """Enable tokens index. It does nothing if it has already been enabled."""
         raise NotImplementedError
+
+    def _manually_initialize_tips_indexes(self, tx_storage: 'TransactionStorage') -> None:
+        """ Initialize the tips indexes, populating them from a tx_storage that is otherwise complete.
+
+        XXX: this method requires timestamp indexes to be complete and up-to-date with the rest of the database
+        XXX: this method is not yet being used
+        """
+        for tx in tx_storage._topological_fast():
+            tx_meta = tx.get_metadata()
+            if not tx_meta.validation.is_final():
+                continue
+
+            self.all_tips.add_tx(tx)
+
+            if tx.is_block:
+                self.block_tips.add_tx(tx)
+            elif not tx_meta.voided_by:
+                self.tx_tips.add_tx(tx)
 
     def add_tx(self, tx: BaseTransaction) -> bool:
         """ Add a transaction to the indexes
--- a/slow_tests/test_simulator.py
+++ b/slow_tests/test_simulator.py
@@ -6,8 +6,8 @@ from hathor.transaction import BaseTransaction
 from hathor.transaction.genesis import genesis_transactions
 from hathor.wallet import HDWallet
 from tests import unittest
-from tests.utils import FakeConnection, MinerSimulator, RandomTransactionGenerator, Simulator
 from tests.clock import HeapClock
+from tests.utils import FakeConnection, MinerSimulator, RandomTransactionGenerator, Simulator
 
 
 class HathorSyncMethodsTestCase(unittest.TestCase):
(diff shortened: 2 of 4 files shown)


hathor/indexes/manager.py
```diff
@@ -63,6 +64,22 @@ def enable_tokens_index(self) -> None:
         """Enable tokens index. It does nothing if it has already been enabled."""
         raise NotImplementedError
 
+    def _manually_initialize(self, tx_storage: 'TransactionStorage') -> None:
```
luislhl, line 67: I saw that we have a `BaseTransactionStorage._manually_initialize` method.

Does it make sense to make it call this new method?

luislhl (approved):

hathor/indexes/manager.py
jansegre: Yes, but not right now. That will make sense when we actually want to use this initialization. This PR aims to create and tests the new initialization function, but not enable it, because that will require more changes.

hathor/indexes/manager.py
```diff
@@ -63,6 +64,22 @@ def enable_tokens_index(self) -> None:
...
+        """ Initialize the tips indexes, populating them from a tx_storage that is otherwise complete.
```
msbrogli, line 68: Maybe add a note that this method is still not being used?

hathor/indexes/manager.py
```diff
@@ -63,6 +64,22 @@ def enable_tokens_index(self) -> None:
...
+        """
+        # XXX: use tx_storage._topological_fast when it is available
```
msbrogli, line 70: Should we create a `tx_storage.topological_sort()` that points to the desired method? The caller can either choose to use a specific method or the default one.

hathor/indexes/manager.py
msbrogli, line 67: Maybe rename to `_manually_initialize_tips_index(...)`?

msbrogli (approved):

hathor/indexes/manager.py
jansegre: I prefer not to. Both iterators still have very specific caveats, I think we still have to be careful with "recommending" an iterator. I think we could leave them internal for now.

hathor/indexes/manager.py
jansegre: Makes sense, I'll rename it to `_manually_initialize_tips_indexes`.

hathor/indexes/manager.py
jansegre: Done.

hathor/indexes/manager.py

PR #404, opened 2022-05-05
chore: update dependencies

jansegre:
Test run of docker CI updates: https://github.com/jansegre/hathor-core/runs/6309343035?check_suite_focus=true

Together with the updating of the dependencies is making the fixes necessary for the updated linters and also updating the PyPy images to 3.8 and 3.9 (with PyPy 3.7 removed). We can now start dropping support for Python 3.7 if we want to.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,17 +20,19 @@ jobs:
       matrix:
         python-impl:
         - python
+        - pypy
         python-version:
         - '3.7'
         - '3.8'
         - '3.9'
         - '3.10'
-        include:
+        exclude:
+        # XXX: pypy-3.7 is not compatible with Dockerfile.pypy anymore
         - python-impl: pypy
           python-version: '3.7'
-        # XXX: this will still fail (probably because of the code Cython generates on python-rocksdb)
-        # - python-impl: pypy
-        #   python-version: '3.8'
+        # XXX: pypy-3.10 does not exist yet
+        - python-impl: pypy
+          python-version: '3.10'
     steps:
     - name: Checkout
       uses: actions/checkout@v2
--- a/hathor/cli/top.py
+++ b/hathor/cli/top.py
@@ -24,774 +24,760 @@ from collections import defaultdict
 from math import floor
 from typing import Any, Callable, DefaultDict, Dict, List, Optional, Tuple
 
-try:
+# XXX: as annoying as it is, a simple `if: raise` is not enough, but putting the whole module inside works
+if sys.platform != 'win32':
     import curses
     import curses.ascii
-except ModuleNotFoundError:
-    pass
 
-from aiohttp import ClientSession
+    from aiohttp import ClientSession
 
-Key = Tuple[str, ...]
-ProcGroup = DefaultDict[Key, List['ProcItem']]
+    Key = Tuple[str, ...]
+    ProcGroup = DefaultDict[Key, List['ProcItem']]
 
-# Global color variable.
-Color: Optional['DefaultColor'] = None
+    # Global color variable.
+    Color: Optional['DefaultColor'] = None
 
+    class ProfileData:
+        """Profile data."""
+        hostname: str
+        version: str
+        network: str
+        enabled: bool
+        last_update: float
+        error: str
+        proc_list: List['ProcItem']
 
-class ProfileData:
-    """Profile data."""
-    hostname: str
-    version: str
-    network: str
-    enabled: bool
-    last_update: float
-    error: str
-    proc_list: List['ProcItem']
+        @classmethod
+        def create_from_api(cls, data: Dict[str, Any]) -> 'ProfileData':
+            self = cls()
+            self.hostname = data['hostname']
+            self.version = data['version']
+            self.network = data['network']
+            self.enabled = data['enabled']
+            self.last_update = data['last_update']
+            self.error = data['error']
+            self.proc_list = []
+            for proc_dict in data['proc_list']:
+                proc = ProcItem.create_from_api(proc_dict)
+                self.proc_list.append(proc)
+            return self
 
-    @classmethod
-    def create_from_api(cls, data: Dict[str, Any]) -> 'ProfileData':
-        self = cls()
-        self.hostname = data['hostname']
-        self.version = data['version']
-        self.network = data['network']
-        self.enabled = data['enabled']
-        self.last_update = data['last_update']
-        self.error = data['error']
-        self.proc_list = []
-        for proc_dict in data['proc_list']:
-            proc = ProcItem.create_from_api(proc_dict)
-            self.proc_list.append(proc)
-        return self
+    class ProcItem:
+        """Process item."""
+        key: Key
+        percent_cpu: float
+        total_time: float
 
+        @classmethod
+        def create_from_api(cls, data: Tuple[Key, Dict[str, Any]]) -> 'ProcItem':
+            self = cls()
+            self.key = tuple(data[0])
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -38,32 +39,31 @@ hathor-cli = 'hathor.cli.main:main'
 [tool.poetry.dev-dependencies]
 flake8 = "~4.0.1"
 isort = {version = "~5.10.1", extras = ["colors"]}
-yapf = "^0.32"
```
luislhl, line 41: We still have yapf's settings in `setup.cfg`.

Should we remove them?

luislhl: Also, would it make sense to start using Black instead? Maybe we could create an issue for this.

hathor/merged_mining/coordinator.py
```diff
@@ -1307,7 +1307,7 @@ async def _update_hathor_block(self, mining: IMiningChannel) -> None:
             # self.log.debug('hathor.get_block_template response', block=block, height=height)
             self.hathor_coord_job = HathorCoordJob(block, height)
             self.log.debug('new Hathor block template', height=height, weight=block.weight)
-            self.update_merged_block()
+            await self.update_merged_block()
```
luislhl, line 1310: Why was this needed?

pyproject.toml
jansegre: I forgot about `setup.cfg`, I'll remove yapf's options from there. I think it totally makes sense to start using Black. But I think that CI checks are a must. At some point we stopped using `yapf` and because there was no CI checks we did not realize how much we had drifted. It would also be great to facilitate adding pre-commit hooks. I think we can add that to the issue.

hathor/merged_mining/coordinator.py
jansegre: Oh, this is actually a bugfix that the updated `mypy` caught.

jansegre: The bug itself is pretty minor and does not affect any part of the fullnode itself. And also there shouldn't be anyone using our merged mining coordinator, so the impact is very minimal.

luislhl (approved):

.github/workflows/docker.yml
```diff
@@ -20,17 +20,19 @@ jobs:
...
         - '3.10'
-        include:
+        exclude:
+        # XXX: pypy-3.7 is not compatible with Dockerfile.pypy anymore
         - python-impl: pypy
           python-version: '3.7'
-        # XXX: this will still fail (probably because of the code Cython generates on python-rocksdb)
-        # - python-impl: pypy
-        #   python-version: '3.8'
+        # XXX: pypy-3.10 does not exist yet
```
msbrogli, line 33: Should we remove it then?

hathor/merged_mining/coordinator.py
msbrogli, line 1310: Is it a bug fix?

hathor/cli/top.py
```diff
@@ -24,774 +24,760 @@
 from math import floor
 from typing import Any, Callable, DefaultDict, Dict, List, Optional, Tuple
 
-try:
+# XXX: as annoying as it is, a simple `if: raise` is not enough, but putting the whole module inside works
+if sys.platform != 'win32':
```
msbrogli, line 28: I prefer to raise an exception and let the importer method handle it instead of doing it. So, I think we should remove this if statement. What do you think?

msbrogli (changes requested):

.github/workflows/docker.yml
jansegre: Yes, that's what is being done now. The diff got a bit confusing, but I basically added an "exclude" section, and it looks like "include" became "exclude".

hathor/merged_mining/coordinator.py
jansegre: Yes, it's a bug fix, although it's probably very minor.

hathor/cli/top.py
jansegre: I prefer that too! But, although mypy correctly understands `sys.platform`, it doesn't understand that a global exception should make it ignore the rest of the file. So the only simple solution I found was to put everything inside an `if`. I also tried having both the `if` and a `raise`, like `if sys.platform == 'win32': raise Exception; else: ...`, but that also makes mypy complaint. Maybe there a better solution, I just don't want to spend too much time on this.

pyproject.toml
jansegre: Done.

hathor/merged_mining/coordinator.py
msbrogli: No tests to update?

jansegre: The updated `mypy` complained about it. No tests are affected by it. I'm not really sure how to test this though, so I'd rather leave it as is.

.github/workflows/docker.yml
msbrogli: Oh! I got it now. Thanks for the clarification.

hathor/merged_mining/coordinator.py
msbrogli: I feel there's a small behavior change. Before adding the `await`, the method would keep executing regardless of the full node response. Now, it will wait for an update. What if the update fails? Will the method break?

hathor/cli/top.py
msbrogli: Should we move the code to `_top.py` and here just do the following?

```python
if sys.platform != 'win32':
    from _top import *
```

hathor/merged_mining/coordinator.py
jansegre: That makes sense, I'll change it to keep the old behavior but not have a loose future, instead I'll add it to the event loop directly. The main complaint that mypy makes is that the future was not awaited, it's technically not a mistake to not await a future, but the proper way to do it is through the event loop, because the handling of that is very implementation specific.

hathor/cli/top.py
jansegre: That also doesn't work. Because mypy will type-check `_top.py` and make the same complaints it makes here. Also that is exactly how the `top` module is used in `hathor/cli/main.py`:

```python
# ...
        if sys.platform != 'win32':
            from . import top
            self.add_cmd('hathor', 'top', top, 'CPU profiler viewer')
# ...
```

hathor/merged_mining/coordinator.py
jansegre: I've reviewed the code. It was supposed to have an `await`, I prefer to have the small behavior change with the added `await`. If there's an exception it will be propagated to the function that called `update_merged_block`, which wouldn't before, but this isn't an issue by itself, the outer function `update_bitcoin_block` is called periodically and can fail and will recover. The only other place we call `update_merged_block` is within `update_hathor_block` (indirectly, and directly inside `_update_hathor_block`), which correctly had an `await`, and also has the behavior we should have had on `update_bitcoin_block`.

msbrogli: That's great! Thanks!

hathor/cli/top.py
msbrogli: To be honest, I hate not failing on Windows. But I agree that we can refactor it later. Just to confirm: Except for the indentation, no other change was made on this file, right?

msbrogli (approved):

hathor/cli/top.py
jansegre: Yes, the only change is the indentation.

PR #410, opened 2022-05-12
feat(storage): add an alternative fast topological iterator

jansegre:
This alternative topological iterator that is significantly faster over the traditional one, nearly as fast as the other one added recently and has a good advantage in that is doesn't need the timestamp indexes to be up-to-date. Which in turn makes it so it can be used for speeding up the rebuilding of indexes, and also other cases where we don't want to rely on the timestamp index, however it cannot be used to rebuild the metadata.

### DIFF
--- a/hathor/transaction/storage/transaction_storage.py
+++ b/hathor/transaction/storage/transaction_storage.py
@@ -683,18 +683,7 @@ class TransactionStorage(ABC):
         pass
 
     @abstractmethod
-    def _topological_fast(self) -> Iterator[BaseTransaction]:
-        """Return an iterable of the transactions in topological ordering, i.e., from genesis to the most recent
-        transactions. The order is important because the transactions are always valid --- their parents and inputs
-        exist. This method makes use of the timestamp index, so it is crucial that that index is correct and complete.
-
-        XXX: blocks are still prioritized over transactions, but only within the same timestamp, which means that it
-        will yield a different sequence than _topological_sort, but the sequence is still topological.
-        """
-        raise NotImplementedError
-
-    @abstractmethod
-    def _topological_sort(self) -> Iterator[BaseTransaction]:
+    def _topological_sort_dfs(self) -> Iterator[BaseTransaction]:
         """Return an iterable of the transactions in topological ordering, i.e., from genesis to the most recent
         transactions. The order is important because the transactions are always valid --- their parents and inputs
         exist. This method is designed to be used for rebuilding metadata or indexes, that is, it does not make use of
@@ -702,6 +691,27 @@ class TransactionStorage(ABC):
 
         XXX: blocks are prioritized so as soon as a block can be yielded it will, which means that it is possible for a
         block to be yielded much sooner than an older transaction that isn't being confirmed by that block.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def _topological_sort_timestamp_index(self) -> Iterator[BaseTransaction]:
+        """Return an iterable of the transactions in topological ordering, i.e., from genesis to the most recent
+        transactions. The order is important because the transactions are always valid --- their parents and inputs
+        exist. This method makes use of the timestamp index, so it is crucial that that index is correct and complete.
+
+        XXX: blocks are still prioritized over transactions, but only within the same timestamp, which means that it
+        will yield a different sequence than _topological_sort_dfs, but the sequence is still topological.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def _topological_sort_metadata(self) -> Iterator[BaseTransaction]:
+        """Return an iterable of the transactions in topological ordering, using only info from metadata.
+
+        This is about as good as _topological_sort_timestamp_index but only needs the transaction's metadata to be
+        consistent and up-to-date. It could replace _topological_sort_timestamp_index if we can show it is faster or at
+        least not slower by most practical cases.
         """
         raise NotImplementedError
 
@@ -962,10 +972,10 @@ class BaseTransactionStorage(TransactionStorage):
 
         # We need to construct a topological sort, then iterate from
         # genesis to tips.
-        for tx in self._topological_sort():
+        for tx in self._topological_sort_dfs():
             self.add_to_indexes(tx)
 
-    def _topological_fast(self) -> Iterator[BaseTransaction]:
+    def _topological_sort_timestamp_index(self) -> Iterator[BaseTransaction]:
         assert self.indexes is not None
 
         cur_timestamp: Optional[int] = None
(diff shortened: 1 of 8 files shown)


hathor/transaction/storage/transaction_storage.py
```diff
@@ -693,6 +693,16 @@ def _topological_fast(self) -> Iterator[BaseTransaction]:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def _topological_fast2(self) -> Iterator[BaseTransaction]:
+        """Return an iterable of the transactions in topological ordering, using only info form metadata.
```
msbrogli, line 698: Typo: ... from metadata.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -988,6 +998,58 @@ def _topological_fast(self) -> Iterator[BaseTransaction]:
...
 
+    def _topological_fast2(self) -> Iterator[BaseTransaction]:
+        import heapq
+        from dataclasses import dataclass, field
+        cur_timestamp = settings.GENESIS_TIMESTAMP
+
+        @dataclass(order=True)
+        class Item:
+            timestamp: int
+            is_transaction: bool
```
msbrogli, line 1009: Add a comment explaining that `False < True`.

msbrogli, line 1009: We did it on the other topological sort method because it used a DFS. I feel we don't need this here.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -988,6 +998,58 @@ def _topological_fast(self) -> Iterator[BaseTransaction]:
...
+                self.is_transaction = tx.is_transaction
+                self.tx = tx
+
+        to_visit: List[Item] = list(map(Item, self.get_all_genesis()))
+        heapq.heapify(to_visit)
+        visited: Set[bytes] = set()
+        to_check: List[Item] = []
+        to_yield: List[Item] = []
+        while (to_visit or to_check or to_yield):
+            while (to_visit and to_visit[0].timestamp <= cur_timestamp) or (not to_check):
```
msbrogli, line 1023: I'm almost sure we don't need this `cur_timestamp` to control the order. I guess it would simplify this implementation. Let me think more about it.

tests/tx/test_indexes3.py
```diff
@@ -85,26 +91,31 @@ def test_tips_index_initialization(self):
...
+        # XXX: sanity check that the children metadata is properly set (this is needed for one of the iterators)
+        for tx in tx_storage.get_all_transactions():
+            assert tx.hash is not None
+            for parent_tx in map(tx_storage.get_transaction, tx.parents):
+                self.assertIn(tx.hash, parent_tx.get_metadata().children)
 
         # test iterators, name is used to aid in assert messages
         iterators = [
             ('traditional', tx_storage._topological_sort()),
             ('fast', tx_storage._topological_fast()),
```
msbrogli, line 109: Should we rename them to `_topological_sort_dfs`, `_topological_sort_timestamp_index`, and `_topological_sort_metadata`?

msbrogli (changes requested):

hathor/transaction/storage/transaction_storage.py
```diff
@@ -988,6 +998,58 @@ def _topological_fast(self) -> Iterator[BaseTransaction]:
...
+                pop_item = heapq.heappop(to_visit)
+                heapq.heappush(to_check, pop_item)
+                new_children = set(pop_item.tx.get_metadata().children) - visited
+                visited.update(new_children)
+                for child in new_children:
+                    try:
+                        child_tx = self.get_transaction(child)
+                    except TransactionDoesNotExist:
```
msbrogli, line 1031: If we have the metadata, shouldn't we have the transaction?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -988,6 +998,58 @@ def _topological_fast(self) -> Iterator[BaseTransaction]:
         yield from cur_blocks
         yield from cur_txs
 
+    def _topological_fast2(self) -> Iterator[BaseTransaction]:
```
msbrogli, line 1001: I feel the following implementation should work:

```python
to_visit: List[Item] = list(map(Item, self.get_all_genesis()))
seen: Set[bytes] = set()
heapq.heapify(to_visit)
while to_visit:
    item = heapq.heappop(to_visit)
    yield item.tx
    # We can safely discard because no other tx will try to visit this one.
    seen.discard(item.tx.hash)
    for child_tx_hash in item.tx.get_metadata().children:
        if child_tx_hash in seen:
            continue
        child_tx = self.get_transaction(child_tx_hash)
        heapq.heappush(to_visit, Item(child_tx))
        seen.add(child_tx_hash)
```

Note that both my implementation and yours assume that for all tx there exists a path between genesis and that tx. In other words, if a transaction is disconnected, it will not be yielded.

jansegre: Your assumption is part of the protocol, so I'd say it's an implied assumption as much as the structure of the transaction (parents, inputs, metadata, etc).

I think your implementation is correct, I had a somewhat similar previous implementation (but using `sortedcontainers.SortedKeyList` instead of `list+heapq`) and I had some issue with yielding transactions not in timestamp order, which I didn't investigate too well and went with an implementation based off the indexes one (there it makes more sense to have a cur_timestamp). I think your proposal is much better than what's currently implemented, I'll use that and test if everything is OK.

tests/tx/test_indexes3.py
jansegre: That would be great. I tried to avoid calling the other two implementations a "topological **sort**" because no **sorting** is taking place, at least not on the timestamp indexes that has everything _already_ sorted, well, the second implementation also doesn't _really_ sort anything, it just walks taking into consideration the `timestamp` and because of the "parent's have strictly smaller timestamp" rule, the walk is naturally in topological order. But I think this is a small detail to fuss about with the naming, since they're still (and probably will for a while) considered internal, I think it's fine. I'll go with your suggestions since they give a much better idea of what the methods are doing.

hathor/transaction/storage/transaction_storage.py
jansegre: We don't, it was trying to fix a problem that wouldn't need fixing if we just include the type of tx before.

hathor/transaction/storage/transaction_storage.py
jansegre: Do you mean that we wouldn't need to prioritize blocks over transactions when they have the same timestamp? It doesn't cause any harm (it should have nearly undetectable performance downside), so I just did it. But I still think there could be a case where a transaction with the same timestamp as the block would need the block to add enough height to the best chain so the reward lock verification passes, so I'd rather just do it anyway.

hathor/transaction/storage/transaction_storage.py
jansegre: Fixed.

hathor/transaction/storage/transaction_storage.py
jansegre: Done.

hathor/transaction/storage/transaction_storage.py

It seems to be working great and it's much simpler! I'll test if there's any performance gained (which probably there won't be because complexity is the same and rocksdb reads are the bottleneck).

hathor/transaction/storage/transaction_storage.py
msbrogli: Thanks!

msbrogli: I agree. Thanks!

hathor/transaction/storage/transaction_storage.py

hathor/transaction/storage/transaction_storage.py
msbrogli: Ok!

tests/tx/test_indexes3.py
msbrogli: Great! Thanks! I agree that they are not really sorting but what matters is that the traversal order is a valid topological order.

hathor/transaction/storage/transaction_storage.py
msbrogli: Great! Thanks for testing it.

hathor/transaction/storage/transaction_storage.py
msbrogli: Ok. This case you described might be a common cause of corruption. Maybe we can fix it during initialization in a separate PR. We could use a similar structure to journaling. Before adding the tx to the db and the indexes, we could add it to a journaling db. After the tx is successfully added, we remove it from the the journal. When the full node starts, we check the journal and remove all partially added tx. What do you think?

msbrogli (approved):

hathor/transaction/storage/transaction_storage.py
jansegre: I think that would be great, and fairly independent from this. I'll link to this suggestion from the tracking issue that we have.

jansegre: Here it is: https://github.com/HathorNetwork/hathor-core/issues/411#issuecomment-1128825322

hathor/transaction/storage/transaction_storage.py
```diff
@@ -988,7 +998,40 @@ def _topological_fast(self) -> Iterator[BaseTransaction]:
...
+                self.tx = tx
+
+        to_visit: List[Item] = list(map(Item, self.get_all_genesis()))
+        seen: Set[bytes] = set()
+        heapq.heapify(to_visit)
+        while to_visit:
+            item = heapq.heappop(to_visit)
+            assert item.tx.hash is not None
+            yield item.tx
+            # We can safely discard because no other tx will try to visit this one.
```
luislhl, line 1025: This was not so obvious to me, I had to stop and think for a while about this.

Maybe if this comment explanation was expanded it could help some people (especially the ones not so used to the code/architecture)

jansegre:
Some tests failed because of #414, will merge anyway since their fix will come next.

PR #413, opened 2022-05-13
feat(cli): add experimental db export/import commands

jansegre:
This PR depends on #410, it's basically two commands (prefixed with `x-` to indicate they are experimental): `x-import` and `x-export`, that will export raw blocks+transactions with no metadata with a trivial serialization. The format is intended to be simple enough so it can be stable, it's useful for recreating a database without having to sync to the network, and also useful for running experiments related to the loading of transactions.

### DIFF
--- a/hathor/cli/db_export.py
+++ b/hathor/cli/db_export.py
@@ -1,0 +1,146 @@
+# Copyright 2021 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+import struct
+from argparse import ArgumentParser, FileType, Namespace
+from typing import TYPE_CHECKING, Iterator, Optional
+
+from hathor.cli.run_node import RunNode
+
+if TYPE_CHECKING:
+    from hathor.transaction import BaseTransaction
+
+MAGIC_HEADER = b'HathDB'
+
+
+class DbExport(RunNode):
+    def start_manager(self, args: Namespace) -> None:
+        pass
+
+    def register_signal_handlers(self, args: Namespace) -> None:
+        pass
+
+    def register_resources(self, args: Namespace) -> None:
+        pass
+
+    def create_parser(self) -> ArgumentParser:
+        from hathor.conf import HathorSettings
+        settings = HathorSettings()
+
+        def max_height(arg: str) -> Optional[int]:
+            if arg.lower() == 'checkpoint':
+                if not settings.CHECKPOINTS:
+                    raise ValueError('There are no checkpoints to use')
+                return settings.CHECKPOINTS[-1].height
+            elif arg:
+                return int(arg)
+            else:
+                return None
+
+        parser = super().create_parser()
+        parser.add_argument('--export-file', type=FileType('wb', 0), required=True,
+                            help='Save the export to this file')
+        parser.add_argument('--export-iterator', choices=['metadata', 'timestamp_index', 'dfs'], default='metadata',
+                            help='Which method of iterating to use, don\'t change unless you know what it does')
+        parser.add_argument('--export-max-height', type=max_height,
+                            help='Make no assumption about the mempool when using this option. It may be partially'
+                            'exported or not, depending on the timestamps and the traversal algorithm.')
+        parser.add_argument('--export-skip-voided', action='store_true', help='Do not export voided txs/blocks')
+        return parser
+
+    def prepare(self, args: Namespace) -> None:
+
+        super().prepare(args)
+
+        # allocating io.BufferedWriter here so we "own" it
+        self.out_file = io.BufferedWriter(args.export_file)
+        if not self.out_file.seekable():
+            raise ValueError('file cannot be used because it is not seekable')
+
+        self._iter_tx: Iterator['BaseTransaction']
+        if args.export_iterator == 'metadata':
+            self._iter_tx = self.tx_storage._topological_sort_metadata()
+        elif args.export_iterator == 'timestamp_index':
+            self._iter_tx = self.tx_storage._topological_sort_timestamp_index()
+        elif args.export_iterator == 'dfs':
+            self._iter_tx = self.tx_storage._topological_sort_dfs()
+        else:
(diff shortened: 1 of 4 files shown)


hathor/cli/db_export.py
```diff
@@ -0,0 +1,99 @@
...
+        elif args.export_iterator == 'slow':
+            self.iter_tx = self.tx_storage._topological_sort()
+        else:
+            raise ValueError(f'unknown iterator "{args.export_iterator}"')
+
+        self.export_height = args.export_max_height
+
+    def run(self) -> None:
+        from hathor.util import progress
+        self.out_file.write(MAGIC_HEADER)
```
msbrogli, line 80: Should we add some extra information such as the total number of transactions and blocks?

hathor/cli/db_export.py
```diff
@@ -0,0 +1,99 @@
...
+        self.log.info('export')
+        for tx in progress(self.iter_tx, log=self.log):
+            assert tx.hash is not None
+            tx_meta = tx.get_metadata()
+            # write tx
+            if tx.is_genesis:
+                continue
+            tx_bytes = bytes(tx)
+            self.out_file.write(struct.pack('!I', len(tx_bytes)) + tx_bytes)
```
msbrogli, line 89: Should we add a separator between transactions so we could partially recover from damaged files?

msbrogli, line 89: I feel we should call write twice instead of concatenating `len(tx_bytes)` with `tx_bytes` before doing it. Such a concatenation will require memory allocation and two memcpy.

hathor/cli/db_export.py
```diff
@@ -0,0 +1,99 @@
...
+            else:
+                return None
+
+        parser = super().create_parser()
+        parser.add_argument('--export-file', type=FileType('wb', 0), required=True,
+                            help='Save the export to this file')
+        parser.add_argument('--export-iterator', choices=['fast2', 'fast', 'slow'], default='fast2',
+                            help='Which method of iterating to use, don\'t change unless you know what it does')
+        parser.add_argument('--export-max-height', type=max_height,
+                            help='Stop exporting after a certain height, use "checkpoint" stop at the last one')
```
msbrogli, line 55: block height? What happens with the mempool?

hathor/cli/db_export.py
```diff
@@ -0,0 +1,99 @@
...
+        for tx in progress(self.iter_tx, log=self.log):
+            assert tx.hash is not None
+            tx_meta = tx.get_metadata()
+            # write tx
+            if tx.is_genesis:
+                continue
+            tx_bytes = bytes(tx)
+            self.out_file.write(struct.pack('!I', len(tx_bytes)) + tx_bytes)
+            # stop as soon as we reach our target height (if any) and after writing it
+            if self.export_height is not None and tx.is_block and tx_meta.height >= self.export_height:
```
msbrogli, line 91: It would be weirdly rare, but I feel that a voided blocks might stop the loop before all confirmed blocks are there.

hathor/cli/db_export.py
```diff
@@ -0,0 +1,99 @@
...
+            if arg.lower() == 'checkpoint':
+                if not settings.CHECKPOINTS:
+                    raise ValueError('There are no checkpoints to use')
+                return settings.CHECKPOINTS[-1].height
+            elif arg:
+                return int(arg)
+            else:
+                return None
+
+        parser = super().create_parser()
```
msbrogli, line 49: Should we allow skipping voided blocks and transactions?

msbrogli (commented):
I started reviewing it but I noticed that it has changes from other PRs. I'll wait for a rebase to finish the review.

hathor/cli/db_export.py
jansegre: That makes sense, I'll do that.

hathor/cli/db_export.py
jansegre: I'd rather keep it simple. I don't know what a successful recovery would look like, if we lose transactions the dataset is screwed anyway. If we want to add any complexity on this front, I'd much prefer to have a concrete motivation first. After all, this isn't meant to be a final format, it's an experimental format just to help with experimenting.

jansegre: You're right, although it's a non-issue currently because iterating is several orders of magnitudes slower, so this doesn't make much of a difference.

hathor/cli/db_export.py
jansegre: I see how that can be useful, I'll add that, although I won't skip soft-voided txs.

msbrogli: Makes sense.

hathor/cli/db_export.py
msbrogli: Got it. Maybe we should clarify it on the help message.

hathor/cli/db_export.py
msbrogli: Great!

hathor/cli/db_export.py
msbrogli: Nice implementation. Good job!

hathor/cli/db_export.py
msbrogli: You have a good point. As we are writing bytes, including a separator might not be trivial.

msbrogli: Thanks for doing it.

hathor/cli/db_export.py
jansegre: I tried to clarify, but I also realized it's not so cut and dry. I don't really know what to clarify, actually. The original mempool may or may not be partially exported when using a max-height, and the resulting export file, may or may not contain transactions that will be in the mempool. It mostly depends on the timestamps and on the iterator used (I suspect that when using the dfs iterator and a max-height, the mempool will always be empty, but I'm also not 100% sure). Let me know if you have any suggestions of what to add to the help explanation, because I think there are too many details to explain without turning the help into a full guide.

Most importantly, this command is experimental, so I'm not too worried if the help is unclear or incomplete.

msbrogli: Hum, your comments about the mempool make sense. I feel we can use something like: "Make no assumption about the mempool. It may be partially exported or not, depending on the timestamps and the traversal algorithm.".

hathor/cli/db_export.py
jansegre: I've added it, I made an export of our latest snapshot while excluding voided txs and I'm waiting for the import to finish to check if the soft-voided txs were correctly handled.

pedroferreira1 (approved):

hathor/cli/db_export.py
msbrogli: I found the condition. Good job! Thanks!

msbrogli (approved):

jansegre:
Last test runs were OK, this run just has the commits signed.

hathor/cli/db_export.py
jansegre: Yes, it worked well.

PR #423, opened 2022-06-06
feat(indexes): add UTXO index

jansegre:
Mini-design:

## High-level

Provide an API that can be used to query UTXOs given an address/tokenid, and provide some type of filtering by amount that can be used to allow for users to easily implement their own UTXO selection algorithm. It is important to note that some UTXOs may not be "spendable", since the API should be primarily used for cases where "spendable" outputs are the only that make sense, the API will not include them. We could in the future add an option to include locked (not spendable) UTXOs as well, but this is out of the scope of this initial implementation.

### Making queries

User inputs:

- `address`
- `token_uid`
- `target_amount`
- optional: `target_timestamp` (defaults to the timestamp of the best block)
- optional: `target_height` (defaults to the height of the best block)
- optional: `max_count` (defaults to 256)

API outputs:

- list of UTXOs with amounts varying from the next amount higher than `target_amount` and then amounts from equal to `target_amount` until it sums up to `target_amount`, each entry will have:
  - `txid` 
  - `index`
  - `amount`
  - optional: `timelock`
  - optional: `heightlock`
- the returned list will have at most `max_count` elements
- `has_more`, whether there are more spendable UTXOs that weren't included on the list due to either `max_count` or the `target_amount`

### Index implementation

- in-memory:
  ```
  Dict[Tuple[token_uid, bytes], Tuple[NoLockUtxos, TimelockUtxos, HeightlockUtxos]]
  NoLockUtxos = SortedKeyList[Tuple[bytes, int]]  # tuples are (txid, index)
  TimelockUtxos = SortedKeyList[Tuple[int, bytes, int]]  # tuples are (timelock, txid, index)
  HeightlockUtxos = SortedKeyList[Tuple[int, bytes, int]]  # tuples are (heightlock, txid, index)
  ```
  (note: this annotation is not exactly what will be in code, but it's just to give an idea) where the dict key is a tuple of token_uid and byte-serialized address
- in-disk: Only use RocksDB keys, with the following format:
  ```
  [token_uid][address][lock_tag][amount][tx_id][index] lock_tag=NO_LOCK
  |--32b----||--25b--||--1b----||--4b--||-32b-||-1b--|

  [token_uid][address][lock_tag][timelock][amount][tx_id][index] lock_tag=TIME_LOCK
  |--32b----||--25b--||--1b----||--4b----||--4b--||-32b-||-1b--|

  [token_uid][address][lock_tag][heightlock][amount][tx_id][index] lock_tag=HEIGHT_LOCK
  |--32b----||--25b--||--1b----||--4b------||--4b--||-32b-||-1b--|
  ```
  since entries are naturally sorted in RocksDB this already has the property that we want (important to note that the amount must be always serialized with big-endian, because rocksdb does byte-oriented sorting, this is already done in other indexes though, so it's not special to this one in particular)

Both implementations will have 3 "heads" to iterate on when searching for UTXOs, we can trivially make an iterator that merges these 3 into one by always yielding from the head with the next sorted value.

### API definition

Deferred to #425

## Unresolved questions

- Should we ignore token-creation/authority/data-only utxos? I'm leaning towards _yes we should_, this could easily be supported later (amount/value would make no sense, so the API will need specific parameters for that).
- While implementing the index I realized that an output can be BOTH timelocked and heightlocked, specifically only when a mined block uses a timelock script for its output. This is case never happened and will most likely never happen (I can't see any reason miners wouldn't want to unlock an output as fast as possible), so to keep everything consistent, when there is BOTH a timelock and a heightlock, I'll ignore the timelock and only use the heightlock, my guess is that this is might not even be properly handled on the wallets, ignoring the timelock seems like a fair and consistent compromise. It's also perfectly fine to improve this in the future.
- Another problem that I found was that just sorting UTXOs with either timelock/heightlock by (timelock, amount) or (heightlock, amount), makes it so they aren't sorted by amount for different values of timelock/heightlock, meaning that it will break the latter assumption when sort-merging. However this may not be very big deal _for now_, the results will certainly be out of order in a few cases, but for the end-user, and in particular, for the use-cases that we know are going to use this when it is released. But we have to make a decision, I'd say it's between:
  1. Leaving the implementation as is (with cases where the ordering by amount is not respected);
  2. Excluding timelocked/heightlocked UTXOs for now;
  3. Putting them together with the nolock UTXOs and post-filtering (which is a complete solution, but will not be efficient on some situations);
  4. Designing a complete and efficient solution; Which, while I think it _may_ be possible, I don't have a good idea of what that would be.

### DIFF
--- a/hathor/consensus.py
+++ b/hathor/consensus.py
@@ -29,6 +29,28 @@ cpu = get_cpu_profiler()
 _base_transaction_log = logger.new()
 
 
+class ConsensusAlgorithmContext:
+    """ An instance of this class holds all the relevant information related to a single run of a consensus update.
+    """
+
+    consensus: 'ConsensusAlgorithm'
+    block_algorithm: 'BlockConsensusAlgorithm'
+    transaction_algorithm: 'TransactionConsensusAlgorithm'
+    txs_affected: Set[BaseTransaction]
+
+    def __init__(self, consensus: 'ConsensusAlgorithm') -> None:
+        self.consensus = consensus
+        self.block_algorithm = consensus.block_algorithm_factory(self)
+        self.transaction_algorithm = consensus.transaction_algorithm_factory(self)
+        self.txs_affected = set()
+
+    def save(self, tx: BaseTransaction) -> None:
+        """Only metadata is ever saved in a consensus update."""
+        assert tx.storage is not None
+        self.txs_affected.add(tx)
+        tx.storage.save_transaction(tx, only_metadata=True)
+
+
 class ConsensusAlgorithm:
     """Execute the consensus algorithm marking blocks and transactions as either executed or voided.
 
@@ -56,19 +78,32 @@ class ConsensusAlgorithm:
     """
 
     def __init__(self, soft_voided_tx_ids: Set[bytes]) -> None:
-        self.soft_voided_tx_ids = soft_voided_tx_ids
-        self.block_algorithm = BlockConsensusAlgorithm(self, soft_voided_tx_ids)
-        self.transaction_algorithm = TransactionConsensusAlgorithm(self, soft_voided_tx_ids)
+        self.soft_voided_tx_ids = frozenset(soft_voided_tx_ids)
+        self.block_algorithm_factory = BlockConsensusAlgorithmFactory()
+        self.transaction_algorithm_factory = TransactionConsensusAlgorithmFactory()
+
+    def create_context(self) -> ConsensusAlgorithmContext:
+        """Handy method to create a context that can be used to access block and transaction algorithms."""
+        return ConsensusAlgorithmContext(self)
 
     @cpu.profiler(key=lambda self, base: 'consensus!{}'.format(base.hash.hex()))
     def update(self, base: BaseTransaction) -> None:
+        """Run a consensus update with its own context, indexes will be updated accordingly."""
         from hathor.transaction import Block, Transaction
+
+        # this context instance will live only while this update is running
+        context = self.create_context()
         if isinstance(base, Transaction):
-            self.transaction_algorithm.update_consensus(base)
+            context.transaction_algorithm.update_consensus(base)
         elif isinstance(base, Block):
-            self.block_algorithm.update_consensus(base)
+            context.block_algorithm.update_consensus(base)
         else:
             raise NotImplementedError
+        # finally signal an index update for all affecetd transactions
+        for tx in context.txs_affected:
+            assert tx.storage is not None
+            assert tx.storage.indexes is not None
+            tx.storage.indexes.update(tx)
 
     def filter_out_soft_voided_entries(self, tx: BaseTransaction, voided_by: Set[bytes]) -> Set[bytes]:
         if not (self.soft_voided_tx_ids & voided_by):
@@ -93,9 +128,8 @@ class ConsensusAlgorithm:
 class BlockConsensusAlgorithm:
     """Implement the consensus algorithm for blocks."""
 
-    def __init__(self, consensus: ConsensusAlgorithm, soft_voided_tx_ids: Set[bytes]) -> None:
-        self.consensus = consensus
-        self.soft_voided_tx_ids = soft_voided_tx_ids
+    def __init__(self, context: ConsensusAlgorithmContext) -> None:
+        self.context = context
 
     @classproperty
     def log(cls):
@@ -180,14 +214,14 @@ class BlockConsensusAlgorithm:
             tx = storage.get_transaction(h)
             tx_meta = tx.get_metadata()
             tx_meta.accumulated_weight = sum_weights(tx_meta.accumulated_weight, block.weight)
-            storage.save_transaction(tx, only_metadata=True)
+            self.context.save(tx)
 
         # Check conflicts of the transactions voiding us.
         for h in voided_by:
             tx = storage.get_transaction(h)
             if not tx.is_block:
                 assert isinstance(tx, Transaction)
-                self.consensus.transaction_algorithm.check_conflicts(tx)
+                self.context.transaction_algorithm.check_conflicts(tx)
 
         parent = block.get_block_parent()
         parent_meta = parent.get_metadata()
@@ -293,7 +327,7 @@ class BlockConsensusAlgorithm:
                     # the blocks themselves.
                     voided_by2 = voided_by2.copy()
                     voided_by2.discard(parent.hash)
-                voided_by.update(self.consensus.filter_out_soft_voided_entries(parent, voided_by2))
+                voided_by.update(self.context.consensus.filter_out_soft_voided_entries(parent, voided_by2))
         return voided_by
 
     def update_voided_by_from_parents(self, block: Block) -> bool:
@@ -307,7 +341,7 @@ class BlockConsensusAlgorithm:
                 meta.voided_by.update(voided_by)
             else:
                 meta.voided_by = voided_by.copy()
-            block.storage.save_transaction(block, only_metadata=True)
+            self.context.save(block)
             block.storage.del_from_indexes(block, relax_assert=True)
             return True
         return False
@@ -437,13 +471,13 @@ class BlockConsensusAlgorithm:
         self.log.debug('add_voided_by', block=block.hash_hex, voided_hash=voided_hash.hex())
 
         meta.voided_by.add(voided_hash)
-        storage.save_transaction(block, only_metadata=True)
+        self.context.save(block)
 
         spent_by: Iterable[bytes] = chain(*meta.spent_outputs.values())
         for tx_hash in spent_by:
             tx = storage.get_transaction(tx_hash)
             assert isinstance(tx, Transaction)
-            self.consensus.transaction_algorithm.add_voided_by(tx, voided_hash)
+            self.context.transaction_algorithm.add_voided_by(tx, voided_hash)
         return True
 
     def remove_voided_by(self, block: Block, voided_hash: Optional[bytes] = None) -> bool:
@@ -469,13 +503,13 @@ class BlockConsensusAlgorithm:
         meta.voided_by.remove(voided_hash)
         if not meta.voided_by:
             meta.voided_by = None
-        block.storage.save_transaction(block, only_metadata=True)
+        self.context.save(block)
 
         spent_by: Iterable[bytes] = chain(*meta.spent_outputs.values())
         for tx_hash in spent_by:
             tx = storage.get_transaction(tx_hash)
             assert isinstance(tx, Transaction)
-            self.consensus.transaction_algorithm.remove_voided_by(tx, voided_hash)
+            self.context.transaction_algorithm.remove_voided_by(tx, voided_hash)
         return True
 
     def remove_first_block_markers(self, block: Block) -> None:
@@ -497,7 +531,7 @@ class BlockConsensusAlgorithm:
                 continue
 
             meta.first_block = None
-            storage.save_transaction(tx, only_metadata=True)
+            self.context.save(tx)
 
     def _score_block_dfs(self, block: BaseTransaction, used: Set[bytes],
                          mark_as_best_chain: bool, newest_timestamp: int) -> float:
@@ -543,7 +577,7 @@ class BlockConsensusAlgorithm:
                     if mark_as_best_chain:
                         assert meta.first_block is None
                         meta.first_block = block.hash
-                        storage.save_transaction(tx, only_metadata=True)
+                        self.context.save(tx)
 
                     score = sum_weights(score, tx.weight)
 
@@ -551,7 +585,7 @@ class BlockConsensusAlgorithm:
         meta = block.get_metadata()
         if not meta.score:
             meta.score = score
-            storage.save_transaction(block, only_metadata=True)
+            self.context.save(block)
         else:
             # The score of a block is immutable since the sub-DAG behind it is immutable as well.
             # Thus, if we have already calculated it, we just check the consistency of the calculation.
@@ -573,7 +607,7 @@ class BlockConsensusAlgorithm:
             if mark_as_best_chain:
                 meta = block.get_metadata()
                 meta.score = block.weight
-                block.storage.save_transaction(block, only_metadata=True)
+                self.context.save(block)
             return block.weight
 
         parent = self._find_first_parent_in_best_chain(block)
@@ -583,12 +617,16 @@ class BlockConsensusAlgorithm:
         return self._score_block_dfs(block, used, mark_as_best_chain, newest_timestamp)
 
 
+class BlockConsensusAlgorithmFactory:
+    def __call__(self, context: ConsensusAlgorithmContext) -> BlockConsensusAlgorithm:
+        return BlockConsensusAlgorithm(context)
+
+
 class TransactionConsensusAlgorithm:
     """Implement the consensus algorithm for transactions."""
 
-    def __init__(self, consensus: ConsensusAlgorithm, soft_voided_tx_ids: Set[bytes]) -> None:
-        self.consensus = consensus
-        self.soft_voided_tx_ids = soft_voided_tx_ids
+    def __init__(self, context: ConsensusAlgorithmContext) -> None:
+        self.context = context
 
     @classproperty
     def log(cls):
@@ -632,7 +670,7 @@ class TransactionConsensusAlgorithm:
                 meta.conflict_with.extend(set(spent_by) - set(meta.conflict_with))
             else:
                 meta.conflict_with = spent_by.copy()
-        tx.storage.save_transaction(tx, only_metadata=True)
+        self.context.save(tx)
 
         for h in spent_by:
             # Update meta.conflict_with of our conflict transactions.
@@ -644,11 +682,11 @@ class TransactionConsensusAlgorithm:
                     tx_meta.conflict_with.append(tx.hash)
             else:
                 tx_meta.conflict_with = [tx.hash]
-            tx.storage.save_transaction(conflict_tx, only_metadata=True)
+            self.context.save(conflict_tx)
 
         # Add ourselves to meta.spent_by of our input.
         spent_by.append(tx.hash)
-        tx.storage.save_transaction(spent_tx, only_metadata=True)
+        self.context.save(spent_tx)
 
     def set_conflict_twins(self, tx: Transaction) -> None:
         """ Get all transactions that conflict with self
@@ -716,9 +754,9 @@ class TransactionConsensusAlgorithm:
                 meta.twins.append(candidate.hash)
                 tx_meta = candidate.get_metadata()
                 tx_meta.twins.append(tx.hash)
-                tx.storage.save_transaction(candidate, only_metadata=True)
+                self.context.save(candidate)
 
-        tx.storage.save_transaction(tx, only_metadata=True)
+        self.context.save(tx)
 
     def update_voided_info(self, tx: Transaction) -> None:
         """ This method should be called only once when the transactions is added to the DAG.
@@ -732,9 +770,9 @@ class TransactionConsensusAlgorithm:
         for parent in tx.get_parents():
             parent_meta = parent.get_metadata()
             if parent_meta.voided_by:
-                voided_by.update(self.consensus.filter_out_soft_voided_entries(parent, parent_meta.voided_by))
+                voided_by.update(self.context.consensus.filter_out_soft_voided_entries(parent, parent_meta.voided_by))
         assert settings.SOFT_VOIDED_ID not in voided_by
-        assert not (self.soft_voided_tx_ids & voided_by)
+        assert not (self.context.consensus.soft_voided_tx_ids & voided_by)
 
         # Union of voided_by of inputs
         for txin in tx.inputs:
@@ -753,14 +791,13 @@ class TransactionConsensusAlgorithm:
             tx2 = tx.storage.get_transaction(h)
             tx2_meta = tx2.get_metadata()
             tx2_meta.accumulated_weight = sum_weights(tx2_meta.accumulated_weight, tx.weight)
-            assert tx2.storage is not None
-            tx2.storage.save_transaction(tx2, only_metadata=True)
+            self.context.save(tx2)
 
         # Then, we add ourselves.
         meta = tx.get_metadata()
         assert not meta.voided_by or meta.voided_by == {tx.hash}
         assert meta.accumulated_weight == tx.weight
-        if tx.hash in self.soft_voided_tx_ids:
+        if tx.hash in self.context.consensus.soft_voided_tx_ids:
             voided_by.add(settings.SOFT_VOIDED_ID)
             voided_by.add(tx.hash)
         if meta.conflict_with:
@@ -770,7 +807,7 @@ class TransactionConsensusAlgorithm:
         # the conflicting tx might affect this tx's voided_by metadata.
         if voided_by:
             meta.voided_by = voided_by.copy()
-            tx.storage.save_transaction(tx, only_metadata=True)
+            self.context.save(tx)
             tx.storage.del_from_indexes(tx)
 
         # Check conflicts of the transactions voiding us.
@@ -880,7 +917,7 @@ class TransactionConsensusAlgorithm:
         meta = tx.get_metadata()
         assert bool(meta.conflict_with)  # FIXME: this looks like a runtime guarantee, MUST NOT be an assert
         assert meta.voided_by == {tx.hash}
-        assert tx.hash not in self.soft_voided_tx_ids
+        assert tx.hash not in self.context.consensus.soft_voided_tx_ids
         self.remove_voided_by(tx, tx.hash)
         self.assert_valid_consensus(tx)
 
@@ -917,7 +954,7 @@ class TransactionConsensusAlgorithm:
             if not meta2.voided_by:
                 meta2.voided_by = None
                 tx.storage.add_to_indexes(tx2)
-            tx2.storage.save_transaction(tx2, only_metadata=True)
+            self.context.save(tx2)
             self.assert_valid_consensus(tx2)
 
         from hathor.transaction import Transaction
@@ -954,7 +991,7 @@ class TransactionConsensusAlgorithm:
         self.log.debug('add_voided_by', tx=tx.hash_hex, voided_hash=voided_hash.hex())
 
         is_dag_verifications = True
-        if meta.voided_by and bool(self.soft_voided_tx_ids & meta.voided_by):
+        if meta.voided_by and bool(self.context.consensus.soft_voided_tx_ids & meta.voided_by):
             # If tx is soft voided, we can only walk through the DAG of funds.
             is_dag_verifications = False
 
@@ -968,7 +1005,7 @@ class TransactionConsensusAlgorithm:
 
             if tx2.is_block:
                 assert isinstance(tx2, Block)
-                self.consensus.block_algorithm.mark_as_voided(tx2)
+                self.context.block_algorithm.mark_as_voided(tx2)
 
             assert not meta2.voided_by or voided_hash not in meta2.voided_by
             if tx2.hash != tx.hash and meta2.conflict_with and not meta2.voided_by:
@@ -982,10 +1019,15 @@ class TransactionConsensusAlgorithm:
                 self.mark_as_voided(tx2)
                 # All voided transactions with conflicts must have their accumulated weight calculated.
                 tx2.update_accumulated_weight(save_file=False)
-            tx2.storage.save_transaction(tx2, only_metadata=True)
+            self.context.save(tx2)
             tx2.storage.del_from_indexes(tx2, relax_assert=True)
             self.assert_valid_consensus(tx2)
 
         for tx2 in check_list:
             self.check_conflicts(tx2)
         return True
+
+
+class TransactionConsensusAlgorithmFactory:
+    def __call__(self, context: ConsensusAlgorithmContext) -> TransactionConsensusAlgorithm:
+        return TransactionConsensusAlgorithm(context)
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -26,6 +26,7 @@ from hathor.indexes.mempool_tips_index import MempoolTipsIndex
 from hathor.indexes.timestamp_index import TimestampIndex
 from hathor.indexes.tips_index import TipsIndex
 from hathor.indexes.tokens_index import TokensIndex
+from hathor.indexes.utxo_index import UtxoIndex
 from hathor.transaction import BaseTransaction
 from hathor.util import progress
 
@@ -66,6 +67,7 @@ class IndexesManager(ABC):
     mempool_tips: MempoolTipsIndex
     addresses: Optional[AddressIndex]
     tokens: Optional[TokensIndex]
+    utxo: Optional[UtxoIndex]
 
     def __init_checks__(self):
         """ Implementations must call this at the **end** of their __init__ for running ValueError checks."""
@@ -99,6 +101,8 @@ class IndexesManager(ABC):
             yield _IndexFilter.ALL, self.addresses
         if self.tokens is not None:
             yield _IndexFilter.ALL, self.tokens
+        if self.utxo is not None:
+            yield _IndexFilter.ALL, self.utxo
 
     @abstractmethod
     def enable_address_index(self, pubsub: 'PubSubManager') -> None:
@@ -108,6 +112,11 @@ class IndexesManager(ABC):
     @abstractmethod
     def enable_tokens_index(self) -> None:
         """Enable tokens index. It does nothing if it has already been enabled."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def enable_utxo_index(self) -> None:
+        """Enable UTXO index. It does nothing if it has already been enabled."""
         raise NotImplementedError
 
     def force_clear_all(self) -> None:
@@ -179,6 +188,15 @@ class IndexesManager(ABC):
 
         tx_storage._update_caches(block_count, tx_count, latest_timestamp, first_timestamp)
 
+    def update(self, tx: BaseTransaction) -> None:
+        """ This is the new update method that indexes should use instead of add_tx/del_tx
+        """
+        # XXX: this _should_ be here, but it breaks some tests, for now this is done explicitly in hathor.manager
+        # self.mempool_tips.update(tx)
+        self.deps.update(tx)
+        if self.utxo:
+            self.utxo.update(tx)
+
     def add_tx(self, tx: BaseTransaction) -> bool:
         """ Add a transaction to the indexes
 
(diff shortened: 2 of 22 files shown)


hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+    """
+
+    def __init__(self):
+        self.log = logger.new()
+
+    # interface methods provided by the base class
+
+    def init_loop_step(self, tx: BaseTransaction) -> None:
+        tx_meta = tx.get_metadata()
+        # we don't want ANY type of voided transaction
```
msbrogli, line 70: What if a transaction moves from validated to voided? Or the other way around?

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+            spent_tx = tx.get_spent_tx(tx_input)
+            spent_tx_meta = spent_tx.get_metadata()
+            spent_tx_output = spent_tx.outputs[tx_input.index]
+            address_script = parse_address_script(spent_tx_output.script)
+            if not address_script:
+                # XXX: completely ignore scripts that aren't "addressable"
+                continue
+            if spent_tx.is_block:
+                # XXX: timelock forced to None when there is a heightlock
+                timelock = None
```
msbrogli, line 107: That's an interesting case. Why don't we add the transaction to both lists?

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+                heightlock = spent_tx_meta.height + settings.REWARD_SPEND_MIN_BLOCKS
```
msbrogli, line 206: Same logic again.

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+                   target_height: Optional[int] = None) -> Iterator[UtxoIndexItem]:
+        iter_nolock = self._iter_utxos_nolock(token_uid=token_uid, address=address,
+                                              target_amount=target_amount)
+        iter_timelock = self._iter_utxos_timelock(token_uid=token_uid, address=address,
+                                                  target_amount=target_amount,
+                                                  target_timestamp=target_timestamp)
+        iter_heightlock = self._iter_utxos_heightlock(token_uid=token_uid, address=address,
+                                                      target_amount=target_amount,
+                                                      target_height=target_height)
+        iter_utxos = peekable(sorted_merger(
```
msbrogli, line 232: Why `peekable(...)`? I can't see where it's in use here.

hathor/indexes/manager.py
```diff
@@ -179,11 +188,18 @@ def _manually_initialize(self, tx_storage: 'TransactionStorage') -> None:
 
         tx_storage._update_caches(block_count, tx_count, latest_timestamp, first_timestamp)
 
-    def add_tx(self, tx: BaseTransaction) -> bool:
+    def add_tx(self, tx: BaseTransaction, *, post_consensus: bool = False) -> bool:
```
msbrogli, line 191: Why don't we remove this method and let the caller choose between `_add_tx_post_consensus` and `_add_tx_pre_consensus`?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -244,15 +245,13 @@ def _disable_weakref(self) -> None:
         self._tx_weakref_disabled = True
 
     @abstractmethod
-    def save_transaction(self: 'TransactionStorage', tx: BaseTransaction, *, only_metadata: bool = False,
-                         add_to_indexes: bool = False) -> None:
+    def save_transaction(self: 'TransactionStorage', tx: BaseTransaction, *, only_metadata: bool = False) -> None:
```
msbrogli, line 248: Have you checked all calls to this method?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -599,11 +595,20 @@ def _topological_sort_metadata(self) -> Iterator[BaseTransaction]:
         raise NotImplementedError
 
     @abstractmethod
-    def add_to_indexes(self, tx: BaseTransaction) -> None:
+    def add_to_indexes_complete(self, tx: BaseTransaction) -> None:
```
msbrogli, line 598: Missing docstring. What does complete mean?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -599,11 +595,20 @@ def _topological_sort_metadata(self) -> Iterator[BaseTransaction]:
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_to_indexes_pre_consensus(self, tx: BaseTransaction) -> None:
```
msbrogli, line 602: Missing docstring. How is this method different from `post_consensus(...)`? And why it matters for the indexes?

Shouldn't the consensus algorithm "notify" the indexes that a transaction has been marked as voided or unmarked as voided, so the indexes can update accordingly?

hathor/manager.py
```diff
@@ -149,6 +149,10 @@ def __init__(self, reactor: Reactor, peer_id: Optional[PeerId] = None, network:
             self.log.debug('enable wallet indexes')
             self.tx_storage.indexes.enable_address_index(self.pubsub)
             self.tx_storage.indexes.enable_tokens_index()
+        if utxo_index and self.tx_storage.with_index:
```
msbrogli, line 152: I feel we should add a method here `enable_utxo_index()`, instead of adding one more parameter to the constructor. I'm good to move forward like this since we need to refactor the whole initialization structure and improve the dependencies injection.

hathor/indexes/memory_utxo_index.py
```diff
@@ -0,0 +1,97 @@
...
+@dataclass(frozen=True)
+class _IndexItem:
+    nolock: 'SortedList[Tuple[int, bytes, int]]' = field(default_factory=SortedList)
+    timelock: 'SortedList[Tuple[int, int, bytes, int]]' = field(default_factory=SortedList)
+    heightlock: 'SortedList[Tuple[int, int, bytes, int]]' = field(default_factory=SortedList)
+
+
+class MemoryUtxoIndex(UtxoIndex):
+    _txs: Set[bytes]
+    _index: DefaultDict[Tuple[bytes, str], _IndexItem]
```
msbrogli, line 36: Maybe use a NamedTuple for the key?

hathor/indexes/memory_utxo_index.py
```diff
@@ -0,0 +1,97 @@
...
+from sortedcontainers import SortedList
+from structlog import get_logger
+
+from hathor.indexes.utxo_index import UtxoIndex, UtxoIndexItem
+
+logger = get_logger()
+
+
+@dataclass(frozen=True)
+class _IndexItem:
```
msbrogli, line 28: Use NamedTuples for each of these items.

hathor/util.py
```diff
@@ -503,3 +503,157 @@ def progress(iter_tx: Iterator['BaseTransaction'], *, log: Optional['structlog.s
...
+    Even if they are not sorted, they will still be yielded eventually, but there are no guarantees about the order
+    they will come out
+
+    >>> list(sorted_merger([1,2,3],[4,3,2]))
+    [1, 2, 3, 4, 3, 2]
+    """
+
+    def __init__(self, *iterators: Iterator[T], key: Optional[Callable[[T], Any]] = None,
+                 reverse: bool = False) -> None:
+        self._iterators = [peekable(it) for it in iterators]
```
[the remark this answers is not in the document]
jansegre: While I agree with that, I think it's much easier to use a `peekable` than to juggle temporary values, specially because I'd have to juggle them in a very similar way for more than one case. I considered adding a dependency that would provide this `peekable` (more-itertools provides _exactly _ that), but I think it's simple enough that it could live in `hathor.util`.

hathor/util.py
```diff
@@ -503,3 +503,157 @@ def progress(iter_tx: Iterator['BaseTransaction'], *, log: Optional['structlog.s
...
+    [1, 1, 1, 1, 1, 1, 1, 1, 1]
+
+    Even if they are not sorted, they will still be yielded eventually, but there are no guarantees about the order
+    they will come out
+
+    >>> list(sorted_merger([1,2,3],[4,3,2]))
+    [1, 2, 3, 4, 3, 2]
+    """
+
+    def __init__(self, *iterators: Iterator[T], key: Optional[Callable[[T], Any]] = None,
```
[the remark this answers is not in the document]
jansegre: Did you see the explanation on the class docstring? There are examples too on how to use it with reverse order.

hathor/util.py
```diff
@@ -503,3 +503,157 @@ def progress(iter_tx: Iterator['BaseTransaction'], *, log: Optional['structlog.s
...
+
+
+def _identity(x: T) -> T:
+    return x
+
+
+class sorted_merger(Iterator[T]):
+    """ Adaptor class to merge multiple sorted iterators into a single iterator that is also sorted.
+
+    Note: for this adaptor to work as expected the input iterators have to already be sorted, but if they aren't, the
```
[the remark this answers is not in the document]
jansegre: Yes, however not doing that was an explicit decision. Because of the limitation that UTXOs with either timelock or heightlock won't have the amount in sort order, I chose to make this not a HARD assumption, but a SOFT assumption. The end result won't be significantly affected, but we can no longer guarantee sorted order for the results when the results include timelock/heightlocks.

I think we should further decide what to do with the timelock/heightlock situation and revisit this when we have a decision.

hathor/indexes/utxo_index.py
jansegre: A transaction cannot change from voided -> valid or valid -> voided _during_ initialization, unless using full-verification, but full-verification does not use `init_loop_step`, so this is not affected.

hathor/indexes/utxo_index.py
jansegre: We definitely _can_, and maybe we a complete solution would be doing that, but that creates a problem to avoid duplicated UTXOs, it means we have to check the resulting list for every addition to the list (or at least to the additions that can have a duplication).

IMO we don't lose much in practice by having this restriction and we can make a fully complete solution in the future that properly migrates the existing index.

hathor/indexes/utxo_index.py
jansegre: You're right, it's not in use. I think I was using it in an early implementation but ended up doing it without it.

hathor/indexes/manager.py
jansegre: I wanted to support something like `both=True`, and possible have that as the default. But I also wanted to minimize changes elsewhere.

hathor/manager.py
jansegre: I don't really like the `enable_*` structure, I really think we shouldn't move in that direction. Mostly because we can't just `enable_*` and `disable_*` freely, and running `enable_*` _after_ initialization will cause errors that are hard to track, to properly support that we'd need more structures (like storing which indexes have already been initialized, and if an index is enabled after initialization it would have to either be rejected with an error or properly reinitialized).

Particularly, if we had an `HathorManager.enable_utxo_index` method, it would be really tempting to use it on tests that already have a manager instantiated, but this will cause problems, because it's very common for blocks and transactions to be created immediately after instantiating a manager, and the `enable_utxo_index` (without careful protections) will miss those transctions, and even if none was added, it would have missed the genesis.

We sort of have this "problem" on the TransactionStorage, however `enable_*` is _exclusively_ used in very few situations, but not without causing problems, the index resetting mechanism had to be changed to account for indexes that were enabled and the information on which indexes were enabled was "lost" or not accessible without breaking encapsulation patterns.

I think that a better structure is having constructor classes (similar to factory, but only for one instantiation, instead of multiple, which simplifies some things).

While we don't have a better structure, I prefer to use a pattern that is the least disruptive as possible, instead of adding yet another pattern half-implemented. So, for now, as unfortunate as it is, I really prefer adding one more parameter to `HathorManager.__init__`

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+                heightlock=heightlock,
+            ))
+        # pre-calculate heightlock, if any, because if there is one it's the same for all outputs
+        if tx.is_block:
+            heightlock = tx_meta.height + settings.REWARD_SPEND_MIN_BLOCKS
+        else:
+            heightlock = None
+        # add new outputs to the index
+        for index, tx_output in enumerate(tx.outputs):
+            address_script = parse_address_script(tx_output.script)
```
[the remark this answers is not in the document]
jansegre: I'll try, but although it is _very_ similar, there are subtle differences that make each of the 4 cases slightly different from the other. I think I can at most reduce it to 2 cases, something like: `_generate_utxo_item_index_from_inputs(tx)` and `_generate_utxo_item_index_from_outputs(tx)`.

hathor/transaction/storage/transaction_storage.py
pedroferreira1: If I understood correctly, for the index to work fine when a tx is marked as voided, you must guarantee that in your example with `tx_A` and `tx_B` in conflict, you will never have `add(tx_A) add(tx_B) void(tx_A)`, right? You must guarantee that you are deleting `tx_A` before adding `tx_B`.

Is this what the indexes add split is doing?

hathor/indexes/rocksdb_utxo_index.py
```diff
@@ -0,0 +1,412 @@
...
+_DB_NAME: str = 'utxos'
+_CF_NAME_UTXO_INDEX = b'utxo-index'
+
+# the following type is used to help a little bit to distinguish when we're using a byte sequence that should only be
+# internally used
+_InternalUid = NewType('_InternalUid', bytes)
+_INTERNAL_HATHOR_TOKEN_UID = _InternalUid(b'\x00' * 32)
+
+
+def _to_internal_token_uid(token_uid: bytes) -> _InternalUid:
```
pedroferreira1, line 42: We already have this in other index, right? Should we try to reuse it?

hathor/indexes/rocksdb_utxo_index.py
```diff
@@ -0,0 +1,412 @@
...
+        return _KeyTxExists(tag=tag, tx_hash=tx_hash)
+    elif tag is _Tag.NOLOCK:
+        assert len(key) == 1 + 32 + 25 + 8 + 32 + 1
+        return _KeyNoLock(
+            tag=tag,
+            token_uid=_from_internal_token_uid(_InternalUid(key[1:33])),
+            address=get_address_b58_from_bytes(key[33:58]),
+            amount=struct.unpack('>Q', key[58:66])[0],
+            tx_id=key[66:98],
+            index=key[98],
```
pedroferreira1, line 269: `index` should be an int in the `_KeyNoLock` class, isn't it?

hathor/indexes/utxo_index.py
jansegre: Done: #426

hathor/indexes/manager.py
```diff
@@ -209,6 +225,22 @@ def add_tx(self, tx: BaseTransaction) -> bool:
...
 
+    def _add_tx_post_consensus(self, tx: BaseTransaction) -> bool:
+        assert tx.hash is not None
+        tx_meta = tx.get_metadata()
+        assert tx_meta.validation.is_fully_connected()
+        if self.utxo:
+            self.utxo.add_tx(tx)
+
+        # XXX: this _should_ be here, but it breaks some tests, for now this is done explicitly in hathor.manager
+        # self.mempool_tips.update(tx)
```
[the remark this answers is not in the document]
jansegre: We are calling the index's method explicitly. Mostly because it kinds of falls in the same situation, I tried to just move it here, which would be the "more correct" place for it to be, but I had an issue with some tests, so I chose to not do this right now, but this is something we should do eventually.

hathor/indexes/manager.py
```diff
@@ -99,6 +101,8 @@ def _iter_all_indexes_with_filter(self) -> Iterator[Tuple[_IndexFilter, BaseInde
             yield _IndexFilter.ALL, self.addresses
         if self.tokens is not None:
             yield _IndexFilter.ALL, self.tokens
+        if self.utxo is not None:
+            yield _IndexFilter.ALL, self.utxo
```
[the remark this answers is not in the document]
jansegre: I avoided creating a new filter for that (because it would have to include non-voided blocks AND transactions). The index initialization function already filters-out voided tx/blocks so it works correctly when using `ALL`.

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+                # XXX: completely ignore scripts that aren't "addressable"
+                continue
+            if heightlock is None:
+                timelock = address_script.get_timelock()
+            else:
+                # XXX: timelock forced to None when there is a heightlock
+                timelock = None
+            # XXX: that is, at least one of them must but None
+            assert timelock is None or heightlock is None
+            self._add_utxo(UtxoIndexItem(
```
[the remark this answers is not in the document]
jansegre: Makes sense. Also, what about data-only outputs? Is there a more general indicator? Maybe checking if the amount is 0?

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+                token_uid=spent_tx.get_token_uid(spent_tx_output.get_token_index()),
+                tx_id=tx_input.tx_id,
+                index=tx_input.index,
+                address=address_script.address,
+                amount=spent_tx_output.value,
+                timelock=timelock,
+                heightlock=heightlock,
+            ))
+        self._remove_tx(tx.hash)
```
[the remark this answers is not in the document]
jansegre: I think it's possible and that's not really being dealt correctly.

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+        # now that the next higher is out of the way, continue from where we were and stop when we reach a sum enough
+        # to cover the target amount
+        for utxo_item in iter_utxos:
+            if amount_sum >= target_amount:
+                break
+            # there's also no point in yielding more outputs than can be used as inputs, if there are more, the caller
+            # will just have to use the larger than target_amount UTXO (if there is any), or consolidate UTXOs first
+            if count_utxos >= settings.MAX_NUM_INPUTS:
+                break
+            amount_sum += utxo_item.amount
```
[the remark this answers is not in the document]
jansegre: Makes sense

hathor/indexes/memory_utxo_index.py
```diff
@@ -0,0 +1,97 @@
...
+            subindex.timelock.remove((item.timelock, item.amount, item.tx_id, item.index))
+        elif item.heightlock is not None:
+            subindex.heightlock.remove((item.heightlock, item.amount, item.tx_id, item.index))
+        else:
+            subindex.nolock.remove((item.amount, item.tx_id, item.index))
+
+    def _iter_utxos_nolock(self, *, token_uid: bytes, address: str, target_amount: int) -> Iterator[UtxoIndexItem]:
+        subindex = self._index[(token_uid, address)].nolock
+        # this will point to the next value that is equal or higher than target_amount
+        idx_next_amount = subindex.bisect((target_amount,)) + 1
```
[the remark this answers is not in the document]
jansegre: I think it would return an index that points to outside the list, this has to be fixed.

hathor/indexes/memory_utxo_index.py
```diff
@@ -0,0 +1,97 @@
...
+        yield from subindex.islice(stop=idx_next_amount, reverse=True)
```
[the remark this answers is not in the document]
jansegre: reverse, in this context, basically means _descending order_

hathor/indexes/rocksdb_utxo_index.py
jansegre: Makes sense, since this could be recurring in rocksdb indexes implementations I'll move it to `hathor.index.rocksdb_utils`

hathor/indexes/rocksdb_utxo_index.py
jansegre: Yes, and it is. This works because in Python a `bytes` instance is basically a `list[int]`, so `key[98]` is an `int`.

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
pedroferreira1: For data outputs the amount is actually 1. We should ignore data script output in this index because they can't be spent and you are already doing it in the address check.

If you prefer being more explicit about it, you can do `DataScript.parse_script(output_script)`. We have this class in the python hathorlib. But you are already skipping them, so I feel it's fine

hathor/indexes/memory_utxo_index.py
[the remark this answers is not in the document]
pedroferreira1: The fix was to add this `for` instead of yielding the list?

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+        count_utxos = 0
+        # we may have up to 3 items with amount higher than target_amount, skip until we get one lower, then yield it
+        # and start counting the amount_sum
+        for utxo_item in iter_utxos:
+            if utxo_item.amount >= target_amount:
+                next_higher = utxo_item
+            else:
+                amount_sum += utxo_item.amount
+                count_utxos += 1
+                if next_higher is not None:
```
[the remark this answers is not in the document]
pedroferreira1: In your first example, why don't you return 7? If you can fulfill the requested amount with a single utxo, this is good, isn't it?

hathor/indexes/memory_utxo_index.py
jansegre: Done.

hathor/consensus.py
```diff
@@ -983,9 +984,14 @@ def add_voided_by(self, tx: Transaction, voided_hash: bytes) -> bool:
...
             tx2.storage.save_transaction(tx2, only_metadata=True)
-            tx2.storage.del_from_indexes(tx2, relax_assert=True)
+            tx2.storage.del_from_indexes_pre_consensus(tx2, relax_assert=True)
+            reverse_index_update.append(tx2)
             self.assert_valid_consensus(tx2)
 
+        for tx2 in reversed(reverse_index_update):
+            assert tx2.storage is not None
+            tx2.storage.del_from_indexes_post_consensus(tx2, relax_assert=True)
+
```
jansegre, line 994: This is what I did to fix the updating of the index when there is a re-org.

jansegre: This was removed per our discussion.

hathor/indexes/memory_utxo_index.py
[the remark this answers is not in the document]
jansegre: A little bit more than that, but mainly yes.

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
jansegre: This has been fixed and is now being dealt with correctly.

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
jansegre: It's now explicitly ignoring authority outputs, but I left data outputs to be implicitly being ignored (since the script parser won't be able to get an address out of it).

hathor/indexes/manager.py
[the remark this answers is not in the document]
jansegre: We were, and still are, calling it directly in `hathor/manager.py`, in several places.

hathor/transaction/storage/transaction_storage.py
jansegre: These methods have been removed.

hathor/indexes/utxo_index.py
jansegre: Removed.

hathor/indexes/utxo_index.py
jansegre: This changed slightly, and now it doesn't even need to filter-out the voided ones.

hathor/indexes/memory_utxo_index.py
msbrogli: Thanks!

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
jansegre: I think this is part of the behavior we want. It could be more useful to use a range of smaller outputs instead of spending a larger or exact size one. If we only return 7, it's harder to choose.

hathor/transaction/storage/transaction_storage.py
msbrogli: Thanks!

hathor/util.py
```diff
@@ -503,3 +503,157 @@ def progress(iter_tx: Iterator['BaseTransaction'], *, log: Optional['structlog.s
...
+
+    def __iter__(self) -> Iterator[T]:
+        return self
+
+    def __next__(self) -> T:
+        self._clear_empty()
+        if not self._iterators:
+            raise StopIteration
+        # XXX: this line bellow is correct, but it's just really hard to convince mypy of that, ignoring for now
+        best_it = min(iter(self._iterators), key=lambda it: self._key(it.peek()))  # type: ignore
```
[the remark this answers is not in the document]
jansegre: I defaulted to `min` because that's how the built-in `sorted` method behaves, by default sorting for ascending order. I added the `reverse` parameter to match that as well. I think I can improve a little by choosing between `min/max` based on the `reverse` parameter though. But it is currently working as expected, so these are just improvements.

hathor/indexes/rocksdb_utxo_index.py
```diff
@@ -0,0 +1,365 @@
...
+                             target_timestamp: Optional[int] = None) -> Iterator[UtxoIndexItem]:
+        seek = _SeekKeyTimeLock(token_uid_internal=to_internal_token_uid(token_uid), address=decode_address(address),
+                                amount=target_amount, timelock=(target_timestamp or 0xfffffffe) + 1)
+        for key in self._iter_keys(seek):
+            if not isinstance(key, _KeyTimeLock):
+                break
+            if key.token_uid_internal != seek.token_uid_internal or key.address != seek.address:
+                break
+            i = key.to_index_item()
+            # it might happen that the first one is out of the heightlock range
```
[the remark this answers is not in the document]
jansegre: Fixed.

tests/tx/test_indexes.py
```diff
@@ -283,6 +297,306 @@ def _test_confirmed_tx_that_spends_unconfirmed_tx(self, debug=False):
...
+        block2.parents[1:] = [txA2.hash, txB2.hash]
+        block2.timestamp = block1.timestamp
+        block2.weight = 1.2
+        block2.resolve()
+        block2.validate_full()
+        self.manager.propagate_tx(block2, fails_silently=False)
+        self.graphviz.labels[block2.hash] = 'block2'
+
+        # make sure a reorg did happen as expected
+        check_utxos((block2.hash, 0, 6400, 36), (txB2.hash, 0, value, None))
```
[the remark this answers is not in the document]
jansegre: The `check_utxo` function tests for the exact output, which on this test it already accounts that the outputs that have been voided are not in the index.

tests/tx/test_indexes.py
```diff
@@ -283,6 +297,306 @@ def _test_confirmed_tx_that_spends_unconfirmed_tx(self, debug=False):
...
+                token_uid=settings.HATHOR_TOKEN_UID,
+                tx_id=tx.hash,
+                index=1,
+                address=address,
+                amount=amount,
+                timelock=None,
+                heightlock=None,
+            ) for tx, amount in txs_and_values[-1:-(max_outputs + 1):-1]  # these are the last 255 utxos
+        ]
+        # print('expected = [')
```
msbrogli, line 590: Same here.

msbrogli (approved):

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
jansegre: I'm a bit confused then.

I thought @pedroferreira1 meant that, when querying for 7, instead of returning 7, 5, 3, we would only return 7. Which I don't think it's what the API is meant for.

From what @msbrogli is saying, I'm now thinking that both of you mean that in the query for 5, instead of 5, 3, 1 it should return 7, 5. Is that it?

That's a simple change to make. The criteria I added was that the "extra UTXO" is one where `amount >= target_amount` and I continue walking the sorted list and start to count until `sum(value) >= target_amount`. The change would be for the "extra UTXO" to be one where `amount > target_amount`, if any.

I don't think there's any downside to doing that other than potentially ending up returning less UTXOs.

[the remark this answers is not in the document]
pedroferreira1: It makes total sense now, I misunderstood the return of the method and the use of `next_higher`.

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,281 @@
...
+        return True
+    return False
+
+
+class UtxoIndex(BaseIndex):
+    """ Index of UTXOs by address+token
+
+    This index is currently optional and only used for an optional API.
+
+    It is expected to be called with only non-voided transactions and blocks, including the genesis. It ignores
```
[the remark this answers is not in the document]
pedroferreira1: Yeah, I agree with you. We are handling correctly in the code, we are counting the created token output and ignoring the authority. My comment was just to update the docstring because we don't ignore new created tokens.

pedroferreira1 (approved):

hathor/indexes/memory_utxo_index.py
msbrogli: Good job! Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Great! Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Ok!

hathor/indexes/manager.py
msbrogli: Ok!

hathor/indexes/utxo_index.py
msbrogli: Thanks!

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
msbrogli: I like the new code. Thanks for the refactor!

hathor/indexes/utxo_index.py
```diff
@@ -0,0 +1,326 @@
...
+            return
+        tx_meta = tx.get_metadata()
+        # outputs are no longer UTXOs, remove them from the index
+        # pre-calculate heightlock, if any, because if there is one it's the same for all outputs
+        if tx.is_block:
+            heightlock = tx_meta.height + settings.REWARD_SPEND_MIN_BLOCKS
+        else:
+            heightlock = None
+        for index, tx_output in enumerate(tx.outputs):
+            address_script = parse_address_script(tx_output.script)
```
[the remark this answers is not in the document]
msbrogli: Refactored! Thanks!

hathor/indexes/utxo_index.py
msbrogli: Refactored! Thanks!

hathor/indexes/utxo_index.py
msbrogli: Great! Thanks!

hathor/indexes/utxo_index.py
msbrogli: Great!

hathor/util.py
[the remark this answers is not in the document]
msbrogli: You're right. We shouldn't enforce it due to some anomalies caused by the way we are sorting UTXOs.

hathor/util.py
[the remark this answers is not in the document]
msbrogli: I see it now. Thanks!

msbrogli (approved):

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
jansegre: Oh, got it! Will fix it.

hathor/indexes/manager.py
```diff
@@ -179,6 +188,15 @@ def _manually_initialize(self, tx_storage: 'TransactionStorage') -> None:
 
         tx_storage._update_caches(block_count, tx_count, latest_timestamp, first_timestamp)
 
+    def update(self, tx: BaseTransaction) -> None:
+        """ This is the new update method that indexes should use instead of add_tx/del_tx
+        """
+        # XXX: this _should_ be here, but it breaks some tests, for now this is done explicitly in hathor.manager
```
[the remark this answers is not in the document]
jansegre: Yes, this isn't too trivial because there are theoretically (but I don't think we have it in practice) cases where we want to remove a transaction but there's no metadata or data that indexes could use to infer that the transaction should be removed regardless of its validation/voided state.

I definitely think there are more improvements to the indexes structure that include this, that we could make on a follow up PR.

hathor/manager.py
```diff
@@ -919,6 +925,8 @@ def on_new_tx(self, tx: BaseTransaction, *, conn: Optional[HathorProtocol] = Non
                 return False
             else:
                 assert tx.validate_full(skip_block_weight_verification=True)
+                self.tx_storage.indexes.update(tx)
+                self.tx_storage.indexes.mempool_tips.update(tx)
```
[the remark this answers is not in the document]
jansegre: Will do.

hathor/manager.py
```diff
@@ -993,11 +1001,11 @@ def step_validations(self, txs: Iterable[BaseTransaction]) -> None:
...
             else:
-                self.tx_storage.save_transaction(tx, only_metadata=True, add_to_indexes=True)
+                self.tx_storage.save_transaction(tx, only_metadata=True)
+                self.tx_storage.add_to_indexes(tx)
                 self.consensus_algorithm.update(tx)
-                # save and process its dependencies even if it became invalid
-                # because invalidation state also has to propagate to children
-                self.tx_storage.indexes.deps.remove_ready_for_validation(tx.hash)
+                self.tx_storage.indexes.update(tx)
+                self.tx_storage.indexes.mempool_tips.update(tx)
```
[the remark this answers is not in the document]
jansegre: Maybe it will work now, it didn't work when I split it into pre/pos. But I'd rather not experiment with this at this point. I'll add a comment that it should be moved to `update`.

hathor/indexes/utxo_index.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/manager.py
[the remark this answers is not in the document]

tests/tx/test_indexes.py
```diff
@@ -283,6 +297,306 @@ def _test_confirmed_tx_that_spends_unconfirmed_tx(self, debug=False):
...
+                    tx_id=tx.hash,
+                    index=1,
+                    address=address,
+                    amount=amount,
+                    timelock=None,
+                    heightlock=None,
+                ) for tx, amount in reversed(txs_window)
+            ]
+            actual = list(utxo_index.iter_utxos(address=address, target_amount=target_amount))
+            # print('expected = [')
```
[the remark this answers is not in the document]
jansegre: Done.

tests/tx/test_indexes.py
jansegre: Done.

Issue #405, opened 2022-05-05
[Design] Reliable integrations with the full node


lucas3003:
_This design has been moved and updated [here](https://github.com/HathorNetwork/rfcs/blob/master/projects/reliable-integration/0001-high-level-design.md)._

## Problem

Currently, applications that want to interact with the full-node must write its own sync algorithm and handle all use cases (Like reorganization). This algorithm can become very complex and consume several days of development.

There are two alternatives when applications want to interact with the full node:

- Initialize a bidirectional communication via websocket.
- Call the REST APIs.

However, the main concerns with these approaches are:

- When a reorganization happens, the applications must know how to query all the affected txs and update them on their databases.
- There is no guarantee on the order the events happened.
- When the websocket is disconnected, events might be lost and a resync must be executed.

## Solution

To tackle the problems presented above, we must implement a built-in event management system, where all events will be sent in the order they occurred. This system have the following requirements:

1. Detect events that are important to applications. (Look for `Event Types` below).
1. Persist each event and give it an unique incremental ID.
1. Give users an REST API and Websocket connection to query for events

To set-up this system, the user must provide, during the full-node initialization, the `--enable-event-queue` flag.

Due to the necessary flags and the events that can be emitted, we will provide a document on how to understand this new mechanism.

This project, however, does **NOT** intend to eliminate the necessity of a sync algorithm, but to make it much more simpler.

## Out of Scope

These features will not be part of the first phase of this project:

- Sync pause/resume
  - An API to manipulate the sync algorithm.
- Event filter
  - Give users the choice to receive only a subset of events, according to some criteria.
- The `--flush-events` flag. By default, all events are retained. In the future, the user could provide a `--flush-events` flag to enable the flushing of events after each event is sent to the client. 

## Event generation during the full-node cycle

Considering this full-node cycle:

![Full-Node-Cycle drawio](https://user-images.githubusercontent.com/3287486/179076836-ab6ae036-b0c8-449c-a582-2f29ca58ab2b.png)

Where:
- `Load` is the period right after the full-node is started, where the local database is read.
- `Sync` is the period where `Load` is finished and the full-node continuously receive/send txs to other peers until full node is stopped.

By default, the events generated during load will be emitted. If user does not want this behavior, it must provide `--skip-load-events` flag.

## Flow

![Event Flow drawio](https://user-images.githubusercontent.com/3287486/178827445-cc4975f7-3b75-4bc9-a79c-e1042a8a64be.png)

## API

[...]

PR #434, opened 2022-07-13
feat: add flags related to the event mechanism

lucas3003:
### Acceptance Criteria

User can now start the full node with `--enable-event-queue` and `--retain-events`. The full-node manager will handle these flags. Subsequent PR's will use this value to manage the events inside the full node.

More information can be found on [the design](https://github.com/HathorNetwork/hathor-core/issues/405).

### Testing

Manual tests on Manager were performed. These were the results:

| INPUT - Flags        |                    |OUTPUT - Manager value |    |           
|----------------------|-----------------|------------------------|--------------
| --enable-event-queue | --retain-events | enable_event_queue     | retain_events |
| TRUE | TRUE | TRUE | TRUE 
| TRUE | FALSE | TRUE | FALSE 
| FALSE | FALSE | FALSE | FALSE

These results just state that `--retain-events` will only be considered if `--enable-event-queue` flag is given. If only `--retain-events` is given, we will throw an error and the full node will not be initialized.

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -101,7 +101,6 @@ class RunNode:
         parser.add_argument('--x-rocksdb-indexes', action='store_true', help='Use RocksDB indexes (currently opt-in)')
         parser.add_argument('--enable-event-queue', action='store_true', help='Enable event queue mechanism')
         parser.add_argument('--retain-events', action='store_true', help='Retain all events in the local database')
-        parser.add_argument('--skip-load-events', action='store_true', help='Skip events during full-node loading')
         return parser
 
     def prepare(self, args: Namespace) -> None:
@@ -308,7 +307,6 @@ class RunNode:
                            The events detected by the full node will be stored and retrieved to clients')
 
             self.manager.retain_events = args.retain_events is True
-            self.manager.skip_load_events = args.skip_load_events is True
 
         for description in args.listen:
             self.manager.add_listen_address(description)
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -303,7 +303,7 @@ class RunNode:
 
         if args.enable_event_queue:
             self.manager.enable_event_queue = True
-            self.log.info('--enable-event-queue flag provided. ' \
+            self.log.info('--enable-event-queue flag provided. '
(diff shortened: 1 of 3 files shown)


hathor/cli/run_node.py
```diff
@@ -296,6 +298,13 @@ def create_wallet():
         if args.x_fast_init_beta:
             self.log.warn('--x-fast-init-beta is now the default, no need to specify it')
 
+        if args.enable_event_queue:
+            self.manager.enable_event_queue = True
+            self.log.info('--enable-event-queue flag provided." \
+                           The events detected by the full node will be stored and retrieved to clients')
+
+            self.manager.retain_events = args.retain_events is True
```
jansegre, line 306: There is one case, where `--retain-events` is passed, but `--enable-event-queue` is not, that gives an outcome where `retain_events == False` (thanks for the table, it helped spot this easily).

I think that behavior isn't good. I think the better behavior would be to refuse to start, since it's very likely that the use forgot about `--enable-event-queue` in that case.

jansegre (changes requested):

hathor/cli/run_node.py
jansegre: I see this was addressed bellow.

jansegre (approved):
Looks good to me.

hathor/cli/run_node.py
```diff
@@ -99,6 +99,8 @@ def create_parser(self) -> ArgumentParser:
                             help='Disable support for running sync-v1. DO NOT ENABLE, IT WILL BREAK.')
         parser.add_argument('--x-localhost-only', action='store_true', help='Only connect to peers on localhost')
         parser.add_argument('--x-rocksdb-indexes', action='store_true', help='Use RocksDB indexes (currently opt-in)')
+        parser.add_argument('--enable-event-queue', action='store_true', help='Enable event queue mechanism')
```
msbrogli, line 102: Should we prefix with `--x-` (e.g., --x-enable-event-queue) while it's under development? It might be misleading for sysadmins to see these new flags available.

We might also exit the process with an error message saying that this feature is not available yet.

lucas3003: I agree. I've sent a new commit with this change

msbrogli (approved):

PR #436, opened 2022-07-14
fix(cache): unexpected crash when using cache and no snapshot

jansegre:
When we use a "compound" storage (i.e. when enabling the cache layer), the inner storage instance will not have indexes, only the outer storage instance, and when the database is fresh, while running `_save_or_verify_genesis` the genesis will not be found on the database so `add_to_indexes` will be called, however this happens during construction phase, so because the inner storage is constructed before the outer storage, it doesn't have the context to know if it's OK to skip this or not.

This PR simply changes the behavior from raising an exception to returning. Raising an exception is useful to prevent situations that were constructed incorrectly, but in this case the situation is correct but it's not possible to know that from within the `add_to_indexes`.

In the future, a better solution would be using a builder pattern, so when it's time to instantiate, we have all the context needed to know whether it's OK to skip either `add_to_indexes` or `_save_or_verify_genesis`. But this will require a larger refactor.

### DIFF
--- a/hathor/transaction/storage/transaction_storage.py
+++ b/hathor/transaction/storage/transaction_storage.py
@@ -105,6 +105,9 @@ class TransactionStorage(ABC):
         # Internal toggle to choose when to select topological DFS iterator, used only on some tests
         self._always_use_topological_dfs = False
 
+        # Only used in self.add_to_indexes to bypass raising an exception
+        self._saving_genesis = False
+
     @abstractmethod
     def _update_caches(self, block_count: int, tx_count: int, latest_timestamp: int, first_timestamp: int) -> None:
         """Update ephemeral caches, should only be used internally."""
@@ -190,6 +193,7 @@ class TransactionStorage(ABC):
 
     def _save_or_verify_genesis(self) -> None:
         """Save all genesis in the storage."""
+        self._saving_genesis = True
         for tx in self._get_genesis_from_settings():
             try:
                 assert tx.hash is not None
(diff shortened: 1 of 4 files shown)


hathor/transaction/storage/transaction_storage.py
```diff
@@ -1019,7 +1019,7 @@ def _run_topological_sort_dfs(self, root: BaseTransaction, visited: Dict[bytes,
 
     def add_to_indexes(self, tx: BaseTransaction) -> None:
         if not self.with_index:
-            raise NotImplementedError
+            return  # XXX: avoid failing on some situations where this is called before we know it's OK to skip
```
luislhl, line 1022: Maybe it will be difficult for some reading the code to know what situations are these.

Should we link to this PR's description, or create an issue to implement the builder pattern you suggested, describing the problem there, then link to that issue?

jansegre: I think linking to the PR makes sense and should be enough, I'll do that.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -1019,7 +1024,12 @@ def _run_topological_sort_dfs(self, root: BaseTransaction, visited: Dict[bytes,
 
     def add_to_indexes(self, tx: BaseTransaction) -> None:
         if not self.with_index:
-            raise NotImplementedError
```
msbrogli, line 1022: I'm wondering whether we should add a unit test that would fail before this fix.

msbrogli (approved):

luislhl (approved):

hathor/transaction/storage/transaction_storage.py
jansegre: Hmmm. I'll do that I think it should be simple to make.

jansegre: I didn't actually create a test, but I made the situation appear in the tests, and added a protection to make sure the cache storage is set up correctly.

PR #442, opened 2022-07-22
fix(indexes): RocksDB empty address test implementation was broken

jansegre:


luislhl (approved):

msbrogli (approved):
I'm wondering how the other tests have passed. It seems like we are not only fixing an edge case (no keys at all) but we are also fixing the `is_empty = addr != address`.

PR #445, opened 2022-07-26
feat(loading): optimize GC so pauses only happen when we want

jansegre:

This optimization was made during memory usage analysis. During the initialization phase, when loading transactions, we have a timer what will log a warning when the processing of a transaction takes longer than a certain threshold. It will very often go off for random transactions that aren't particularly slow to be processed and it almost never goes off for the same transaction on subsequent runs. What was noticed is that the longer processing time is actually caused by GC pauses, and has nothing to do with the transaction itself.

This PR introduces a helper that will pause automatic garbage collection within a region and uses that helper to make collection happen only during a phase where it won't interfere with the timing of processing a transaction. In practice it seems to work really well and the slow processing warning doesn't even seem to go off anymore, which may justify revisiting the threshold in the future.

### DIFF
--- a/hathor/util.py
+++ b/hathor/util.py
@@ -13,11 +13,13 @@
 # limitations under the License.
 
 import datetime
+import gc
 import json
 import math
 import time
 import warnings
 from collections import OrderedDict
+from contextlib import AbstractContextManager
 from enum import Enum
 from functools import partial, wraps
 from random import Random as PyRandom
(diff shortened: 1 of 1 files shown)


msbrogli (approved):

hathor/util.py
```diff
@@ -482,6 +492,11 @@ def progress(iter_tx: Iterator['BaseTransaction'], *, log: Optional['structlog.s
                 log.info(f'loading... {math.floor(progress * 100):2.0f}%', progress=progress, **kwargs)
             else:
                 log.info('loading...', **kwargs)
+            # XXX: this collections will happen every _DT_LOG_PROGRESS (=30s) on average, which is good, and because
+            #      automatic collection should be disabled, it won't happen during processing of transactions, which
+            #      can make it seem like a transaction took more time to be processed when has nothing to do with the
+            #      transaction itself
+            gc.collect()
```
pedroferreira1, line 499: Does it make sense to have another `gc.collect()` in the end of the method after the while breaks? We might break the while right before the `_DT_LOG_PROGRESS`.

It probably won't affect that much, given that after this method ends, the automatic gc will start again and clean it soon but I'm not sure.

pedroferreira1 (approved):

hathor/util.py
jansegre: Yeah, it totally makes sense, at some point I thought of doing that too but then I forgot. I'll do that, thanks.

jansegre: Done.

PR #447, opened 2022-07-26
feat(cli): add --no-wait option to quick_test subcommand

jansegre:

This PR was made because I needed to run quick tests and did not want to wait for it to receive a transaction (which is what it does by default).

### DIFF
--- a/hathor/cli/quick_test.py
+++ b/hathor/cli/quick_test.py
@@ -12,12 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


msbrogli (approved):
I'm ok with this change. I would like more information about what's the background behind it?

jansegre:
> I'm ok with this change. I would like more information about what's the background behind it?

While running some memory tests I just wanted to load the database and skip any side effects from running p2p, even if only for a while.

lucas3003 (approved):

PR #449, opened 2022-07-27
fix(api): validate token_uid before using it internally

jansegre:
This was noticed after the latest deploy. The RocksDB index implementation is stricter than the Memory implementation, so when using a token_uid of invalid length an assertion error is raised when it wasn't before. This PR fixes this behavior.


pedroferreira1 (approved):

luislhl (approved):

msbrogli (changes requested):
I feel we should fix the memory index and make it raise an exception just like the rocksdb index.

msbrogli (approved):

PR #450, opened 2022-07-27
chore(python): drop support for Python 3.7

jansegre:

This PR includes all the necessary changes to drop support for Python 3.7, which hasn't been used in a while. While this hasn't been discussed too much, impact should be minimal as there are no known deployments using this version. It should also help move closer enabling us to use features not available on Python 3.7, although most of the features that we recently been interested in would require also dropping support for Python 3.8, but it's probably best to do drop support one by one so we could at least start by dropping Python 3.7.

- [ ] communicate with partners that Python 3.7 support will be dropped

Some of the features of Python 3.9 we're interested in:

 - [PEP 585](https://www.python.org/dev/peps/pep-0585), type hinting generics in standard collections;
 - maybe [PEP 593](https://www.python.org/dev/peps/pep-0593), flexible function and variable annotations;

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ For more information about our HTTP API, check out our [API Documentation](https
 
 ## From source-code
 
-First, you need to have Python >=3.7 installed. If you don't, we recommend you to install using `pyenv` (check this
+First, you need to have Python >=3.8 installed. If you don't, we recommend you to install using `pyenv` (check this
 [link](https://github.com/pyenv/pyenv#installation)).
 
 ### System dependencies
@@ -59,11 +59,11 @@ First, you need to have Python >=3.7 installed. If you don't, we recommend you t
   brew install pyenv
   ```
 
-  then Python 3.7 (you could check the latest 3.7.x version with `pyenv install --list`):
+  then Python 3.10 (you could check the latest 3.10.x version with `pyenv install --list`):
 
   ```
-  pyenv install 3.7.11
-  pyenv local 3.7.11
+  pyenv install 3.10.6
+  pyenv local 3.10.6
   pip install -U poetry
   ```
 
@@ -75,7 +75,7 @@ First, you need to have Python >=3.7 installed. If you don't, we recommend you t
 - on Windows 10 (using [winget](https://github.com/microsoft/winget-cli)):
 
   ```
-  winget install python-3.7
+  winget install python-3.10
   pip install -U poetry
   ```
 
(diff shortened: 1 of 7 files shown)


msbrogli:
I like dropping Python 3.7 (even though that's the version I've been using in my local dev environment for a while). Have we checked with exchanges, mining pools, miners, and other use cases which Python version they use? I feel we should communicate this change before dropping support.

jansegre:
> 

Yeah, I think that makes sense. I'll add a task to double check with partners if there's anyone using Python 3.7 and recommend they update (possibly to 3.9 or 3.10).

msbrogli (approved):

README.md
```diff
@@ -59,11 +59,11 @@ First, you need to have Python >=3.7 installed. If you don't, we recommend you t
   brew install pyenv
   ```
 
-  then Python 3.7 (you could check the latest 3.7.x version with `pyenv install --list`):
+  then Python 3.8 (you could check the latest 3.8.x version with `pyenv install --list`):
```
pedroferreira1, line 62: Should we say here 3.10, since below we have `pyenv install 3.10.5`?

hathor/cli/run_node.py
```diff
@@ -644,18 +644,18 @@ def check_unsafe_arguments(self, args: Namespace) -> None:
...
-        MIN_STABLE = (3, 8)
+        MIN_VER = (3, 8)
         RECOMMENDED_VER = (3, 9)
         cur = sys.version_info
-        min_pretty = '.'.join(map(str, MIN_STABLE))
+        min_pretty = '.'.join(map(str, MIN_VER))
         cur_pretty = '.'.join(map(str, cur))
         recommended_pretty = '.'.join(map(str, RECOMMENDED_VER))
-        if cur < MIN_STABLE:
+        if cur < MIN_VER:
```
pedroferreira1, line 653: Should we exit here? The python version is not supported

README.md
jansegre: That makes sense, I'll do that.

hathor/cli/run_node.py
jansegre: Yeah, that'd be better than just warning.

README.md
jansegre: Done.

hathor/cli/run_node.py
jansegre: Done.

pedroferreira1 (approved):

PR #451, opened 2022-07-27
feat: add storage mechanism for events

lucas3003:

### Acceptance Criteria

Implement a persistence layer for events, so it can be retrieved by the event management layer.

### DIFF
--- a/hathor/event/base_event.py
+++ b/hathor/event/base_event.py
@@ -1,0 +1,33 @@
+# Copyright 2022 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import Dict, Optional
+
+
+@dataclass
+class BaseEvent:
+    # Full node id, because different full nodes can have different sequences of events
+    peer_id: str
+    # Event unique id, determines event order
+    id: int
+    # Timestamp in which the event was emitted, this follows the unix_timestamp format
+    timestamp: int
+    # One of the event types
+    type: str
+    # Variable for event type
+    data: Dict
+    # Used to link events, for example, many TX_METADATA_CHANGED will have the same group_id when they belong to the
+    # same reorg process
+    group_id: Optional[int] = None
--- a/hathor/event/storage/event_storage.py
+++ b/hathor/event/storage/event_storage.py
@@ -1,0 +1,30 @@
+# Copyright 2022 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from hathor.event.base_event import BaseEvent
+
+
+class EventStorage(ABC):
+    @abstractmethod
+    def save_event(self, event: BaseEvent) -> None:
+        """ Saves an event in the storage"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_event(self, key: int) -> Optional[BaseEvent]:
+        """ Get a stored event by key"""
+        raise NotImplementedError
--- a/hathor/event/storage/rocksdb_storage.py
+++ b/hathor/event/storage/rocksdb_storage.py
@@ -1,0 +1,61 @@
+# Copyright 2022 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+from typing import Optional
+
+from hathor.event.base_event import BaseEvent
+from hathor.event.storage.event_storage import EventStorage
+from hathor.storage.rocksdb_storage import RocksDBStorage
+from hathor.transaction.util import int_to_bytes
+
+_CF_NAME_EVENT = b'event'
+
+
+class EventRocksDBStorage(EventStorage):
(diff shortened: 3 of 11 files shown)


tests/event/test_event_storage.py
```diff
@@ -0,0 +1,42 @@
...
+import pytest
+
+from hathor.event.storage.event_storage import EventStorage
+from hathor.event.storage.memory_storage import EventMemoryStorage
+from hathor.event.storage.rocksdb_storage import EventRocksDBStorage
+from hathor.storage.rocksdb_storage import RocksDBStorage
+from tests.utils import HAS_ROCKSDB, generate_mocked_event
+
+
+class EventStorageBaseTest():
```
jansegre, line 16: We usually have a different pattern for these, which would basically be:

```python
class EventStorageBaseTest(unittest.TestCase):
    __test__ = False

    # ...

class EventStorageMemoryTest(EventStorageBaseTest):
    __test__ = True
```

I'm not sure if there are advantages or disadvantages of doing it differently, but I'd prefer to stick to the current pattern for consistency.

hathor/event/storage/event_storage.py
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Optional
+
+from hathor.event.base_event import BaseEvent
+
+
+class EventStorage(ABC):
```
jansegre, line 21: Shouldn't we have methods for reading/writing metadata?

tests/event/test_event_storage.py
jansegre: Done.

hathor/event/storage/event_storage.py
jansegre: I'm not sure how this will be needed, so I prefer to skip this for now.

hathor/event/base_event.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@dataclass
+class BaseEvent:
+    def __init__(self, id: int, peer_id: str, timestamp: int, type: str, data: Dict):
+        self.id = id
+        self.peer_id = peer_id
+        self.timestamp = timestamp
+        self.type = type
+        self.data = data
```
luislhl, line 26: We had a `group_id` field described in the design. Was it discontinued or we just don't need it yet?

hathor/event/base_event.py
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import Dict
+
+
+@dataclass
+class BaseEvent:
+    def __init__(self, id: int, peer_id: str, timestamp: int, type: str, data: Dict):
```
luislhl, line 21: We do we need this `__init__` method instead of using what the `dataclass` would provide us?

hathor/event/storage/rocksdb_storage.py
```diff
@@ -0,0 +1,59 @@
...
+
+_CF_NAME_EVENT = b'event'
+
+
+class EventRocksDBStorage(EventStorage):
+
+    def __init__(self, rocksdb_storage: RocksDBStorage):
+        rocksdb_storage.get_or_create_column_family(_CF_NAME_EVENT)
+        self._db = rocksdb_storage.get_db()
+        self._cf_event = rocksdb_storage.get_or_create_column_family(_CF_NAME_EVENT)
```
luislhl, line 31: Why do you need to run this twice?

hathor/pubsub.py
```diff
@@ -90,6 +124,14 @@ class HathorEvents(Enum):
 
     CONSENSUS_TX_UPDATE = 'consensus:tx_update'
 
+    NETWORK_NEW_TRANSACTION_ACCEPTED = 'network:new_transaction_accepted'
```
luislhl, line 127: Is this just an alias for `NETWORK_NEW_TX_ACCEPTED`, or does the latter also include blocks?

tests/event/test_event_storage.py
```diff
@@ -0,0 +1,46 @@
...
+    __test__ = False
+
+    def test_save_event_and_retrieve(self):
+        event = generate_mocked_event()
+        self.event_storage.save_event(event)
+        event_retrieved = self.event_storage.get_event(event.id)
+
+        assert event_retrieved == event
+
+    def _generate_random_word(self, length):
```
luislhl, line 24: This is not used anymore, right?

hathor/event/base_event.py
jansegre: Oh, you're right, I didn't notice this.

hathor/event/storage/rocksdb_storage.py
jansegre: We don't, this is an oversight.

tests/event/test_event_storage.py
jansegre: No. I'll remove it.

hathor/pubsub.py
jansegre: Oh, this is definitely a merge artifact that I missed.

hathor/event/base_event.py
jansegre: That's missing here.

tests/event/test_event_storage.py
jansegre: Not really. I'll remove it.

hathor/event/base_event.py
jansegre: Fixed.

hathor/event/base_event.py
jansegre: Fixed.

hathor/event/storage/rocksdb_storage.py
jansegre: Fixed.

hathor/pubsub.py
jansegre: Fixed.

tests/event/test_event_storage.py
jansegre: Removed.

luislhl (approved):

hathor/pubsub.py
```diff
@@ -108,6 +144,18 @@ class HathorEvents(Enum):
 
     WALLET_ELEMENT_VOIDED = 'wallet:element_voided'
 
+    LOAD_STARTED = 'manager:load_started'
+
+    LOAD_FINISHED = 'manager:load_finished'
+
+    REORG_STARTED = 'regorg:started'
```
pedroferreira1, line 151: Typo reorg*

tests/event/test_event_storage.py
```diff
@@ -0,0 +1,40 @@
...
+from hathor.event.storage.rocksdb_storage import EventRocksDBStorage
+from hathor.storage.rocksdb_storage import RocksDBStorage
+from tests import unittest
+from tests.utils import HAS_ROCKSDB, generate_mocked_event
+
+
+class EventStorageBaseTest(unittest.TestCase):
+    __test__ = False
+
+    def test_save_event_and_retrieve(self):
```
pedroferreira1, line 15: Maybe add also a `self.event_storage.get_event(-1)` just to guarantee that getting the event with a random id won't raise any errors, just return `None`.

pedroferreira1 (approved):

hathor/event/base_event.py
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+from typing import Dict
+
+
+@dataclass
+class BaseEvent:
```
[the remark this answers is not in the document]
jansegre: I think that's great. I added descriptions for all fields.

hathor/pubsub.py
jansegre: Fixed.

PR #452, opened 2022-07-27
refactor: rocksdb storage

lucas3003:
### Acceptance Criteria

In order to enable other use cases other than `tx` to use the storage, we must generalize the RocksDB database creation, so it can be passed as a dependency and used by clients to create new column families, insert data, retrieve data, and so on.


hathor/storage/rocksdb_storage.py
```diff
@@ -0,0 +1,68 @@
...
+            cf_names = []
+
+        # we need to open all column families
+        column_families = {cf: rocksdb.ColumnFamilyOptions() for cf in cf_names}
+
+        # finally, open the database
+        self._db = rocksdb.DB(db_path, options, column_families=column_families)
+        self.log.debug('open db', cf_list=[cf.name.decode('ascii') for cf in self._db.column_families])
+
+    def get_db(self):
```
jansegre, line 60: I think this can be typed, at least with something like `def get_db(self) -> 'rocksdb.DB'` or something like that, I don't remember the exact type name.

hathor/storage/rocksdb_storage.py
```diff
@@ -0,0 +1,68 @@
...
+        return self._db
+
+    def get_or_create_column_family(self, cf_name):
```
jansegre, line 63: This should also be typed, something like `def get_or_create_column_family(self, cf_name: bytes) -> rocksdb.ColumnFamily`, not sure about the exact return type name though.

hathor/cli/run_node.py
```diff
@@ -211,8 +213,9 @@ def create_wallet():
                 self.log.warn('--rocksdb-storage is now implied, no need to specify it')
             cache_capacity = args.rocksdb_cache
             use_memory_indexes = not args.x_rocksdb_indexes
-            tx_storage = TransactionRocksDBStorage(path=args.data, with_index=(not args.cache),
-                                                   cache_capacity=cache_capacity,
+            rocksdb_storage = RocksDBStorage(path=args.data, cache_capacity=cache_capacity)
+            tx_storage = TransactionRocksDBStorage(rocksdb_storage=rocksdb_storage,
```
jansegre, line 217: This looks like such a trivial argument that I think it might even be better to just not pass it by keyword, `TransactionRocksDBStorage(rocksdb_storage, ...)` seems a little bit better for me. This also applies for a few other cases where the keyword is being used.

jansegre (changes requested):
I think this is a move in the right direction. I only have very minor comments.

Although I'm not sure if we need to make it in this PR, but the `RocksDBStorage` class could very well be used on RocksDB indexes classes, which I think are currently using the db object directly.

msbrogli (approved):
Another refactor might be to let the indexes use the RocksDBStorage directly instead of getting it from the TransactionRocksDBStorage. Have you created an issue to do it?

jansegre (approved):

lucas3003:
> Another refactor might be to let the indexes use the RocksDBStorage directly instead of getting it from the TransactionRocksDBStorage. Have you created an issue to do it?

I created the issue: https://github.com/HathorNetwork/hathor-core/issues/453

PR #454, opened 2022-08-08
feat(events): add management layer

lucas3003:

### DIFF
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -1,0 +1,98 @@
+# Copyright 2022 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+from structlog import get_logger
+
+from hathor.event.base_event import BaseEvent
+from hathor.event.storage import EventStorage
+from hathor.pubsub import EventArguments, HathorEvents, PubSubManager
+from hathor.util import Reactor
+
+logger = get_logger()
+
+_GROUP_START_EVENTS = {
+    HathorEvents.REORG_STARTED,
+}
+
+_GROUP_END_EVENTS = {
+    HathorEvents.REORG_FINISHED,
+}
+
+_SUBSCRIBE_EVENTS = [
+    HathorEvents.NETWORK_NEW_TX_ACCEPTED,
+    HathorEvents.NETWORK_NEW_TX_VOIDED,
+    HathorEvents.NETWORK_BEST_BLOCK_FOUND,
+    HathorEvents.NETWORK_ORPHAN_BLOCK_FOUND,
+    HathorEvents.LOAD_STARTED,
+    HathorEvents.LOAD_FINISHED,
+    HathorEvents.REORG_STARTED,
+    HathorEvents.REORG_FINISHED,
+    HathorEvents.TX_METADATA_CHANGED,
+    HathorEvents.BLOCK_METADATA_CHANGED,
+]
+
+
+class EventManager:
+    def __init__(self, event_storage: EventStorage, reactor: Reactor, peer_id: str):
+        self.log = logger.new()
+        self.clock = reactor
+        self.event_storage = event_storage
+        last_event = event_storage.get_last_event()
+        last_event_type = HathorEvents(last_event.type) if last_event is not None else None
+        # XXX: we must check that the last event either does not belong to an event group or that it just closed an
+        #      event group, because we cannot resume an open group of events that wasn't properly closed before exit
+        assert (
+            last_event is None or
+            last_event.group_id is None or
+            last_event_type in _GROUP_END_EVENTS
+        ), 'an unclosed event group was detected, which indicates the node crashed, cannot resume'
+        self._next_event_id = 0 if last_event is None else last_event.id + 1
+        last_group_id = event_storage.get_last_group_id()
+        self._next_group_id = 0 if last_group_id is None else last_group_id + 1
+        self._current_group_id: Optional[int] = None
+        self._peer_id = peer_id
+
+    def subscribe(self, pubsub: PubSubManager) -> None:
+        """ Subscribe to defined events for the pubsub received
+        """
+        for event in _SUBSCRIBE_EVENTS:
+            pubsub.subscribe(event, self._persist_event)
+
+    def _persist_event(self, event: HathorEvents, args: EventArguments) -> None:
+        event_data = args.__dict__['event']
+        group_id: Optional[int]
+        if event in _GROUP_START_EVENTS:
+            assert self._current_group_id is None, 'cannot start an event group before the last one is ended'
+            group_id = self._next_group_id
+        else:
+            group_id = self._current_group_id
+        if event in _GROUP_END_EVENTS:
+            assert self._current_group_id is not None, 'cannot end group twice'
+        event_to_store = BaseEvent(
+            id=self._next_event_id,
+            peer_id=self._peer_id,
+            timestamp=self.clock.seconds(),
+            type=event.value,
+            data=event_data,
+            group_id=group_id,
+        )
+        self.event_storage.save_event(event_to_store)
+        self._next_event_id += 1
+        if event in _GROUP_START_EVENTS:
+            self._current_group_id = self._next_group_id
+            self._next_group_id += 1
+        if event in _GROUP_END_EVENTS:
+            self._current_group_id = None
--- a/hathor/event/storage/event_storage.py
+++ b/hathor/event/storage/event_storage.py
@@ -28,3 +28,13 @@ class EventStorage(ABC):
     def get_event(self, key: int) -> Optional[BaseEvent]:
         """ Get a stored event by key"""
         raise NotImplementedError
+
+    @abstractmethod
+    def get_last_event(self) -> Optional[BaseEvent]:
+        """ Get the last event that was emitted, this is used to help resume when restarting."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_last_group_id(self) -> Optional[int]:
+        """ Get the last group-id that was emitted, this is used to help resume when restarting."""
+        raise NotImplementedError
--- a/hathor/event/storage/memory_storage.py
+++ b/hathor/event/storage/memory_storage.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, Optional
+from typing import List, Optional
 
 from hathor.event.base_event import BaseEvent
 from hathor.event.storage.event_storage import EventStorage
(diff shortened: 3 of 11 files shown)


hathor/event/event_manager.py
```diff
@@ -0,0 +1,55 @@
...
+from hathor.pubsub import EventArguments, HathorEvents, PubSubManager
+
+logger = get_logger()
+
+
+class EventManager:
+    def __init__(self, event_storage: EventStorage, peer_id: str):
+        self.event_storage = event_storage
+        last_event = event_storage.get_last_event()
+        self.next_event_index = 1 if last_event is None else last_event.id+1
```
msbrogli, line 16: It's weird that our linter didn't get it. I feel we need to add some space here: `last_event.id + 1`.

hathor/event/event_manager.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def subscribe(self, pubsub: PubSubManager) -> None:
+        """ Subscribe to defined events for the pubsub received
+        """
+        self.log.error(pubsub)
+        events = [
+            HathorEvents.NETWORK_NEW_TX_ACCEPTED,
+            HathorEvents.NETWORK_NEW_TX_VOIDED,
+            HathorEvents.NETWORK_BEST_BLOCK_FOUND,
+            HathorEvents.NETWORK_ORPHAN_BLOCK_FOUND,
```
msbrogli, line 28: What if we have a tie? Which event would be emitted?

hathor/event/event_manager.py
```diff
@@ -0,0 +1,55 @@
...
+
+        for event in events:
+            pubsub.subscribe(event, self._persist_event)
+
+    def _persist_event(self, event: HathorEvents, args: EventArguments) -> None:
+        event_data = args.__dict__['event']
+        event_to_store = BaseEvent(
+            id=self.next_event_index,
+            peer_id=self.peer_id,
+            timestamp=int(time.time()),  # Current epoch
```
msbrogli, line 45: Should we store microseconds to order them correctly?

msbrogli, line 45: We cannot use `time.time()` here. We must use `reactor.seconds()` so our simulator and tests would work accordingly.

hathor/event/event_manager.py
```diff
@@ -0,0 +1,55 @@
...
+            type=event.name,
+            data=event_data,
+            group_id=0,  # TODO
+        )
+        try:
+            self.event_storage.save_event(event_to_store)
+        except Exception:
+            self.log.error('Error while saving event', event=event_to_store)
```
msbrogli, line 53: Should we kill the node when it happens? I mean, it could cause integrations to lose events.

hathor/event/storage/event_storage.py
```diff
@@ -28,3 +28,7 @@ def save_event(self, event: BaseEvent) -> None:
     def get_event(self, key: int) -> Optional[BaseEvent]:
         """ Get a stored event by key"""
         raise NotImplementedError
+
+    @abstractmethod
+    def get_last_event(self) -> Optional[BaseEvent]:
```
msbrogli, line 33: Missing docstring.

hathor/event/storage/memory_storage.py
```diff
@@ -33,3 +33,10 @@ def get_event(self, key: int) -> Optional[BaseEvent]:
             return None
 
         return self.events[key]
+
+    def get_last_event(self) -> Optional[BaseEvent]:
+        if not self.events:
+            return None
+
+        last_id = list(self.events)[-1]
```
msbrogli, line 41: That's O(n). I feel we should store the last event in a private attribute (e.g., `_last_event`) to make this O(1).

msbrogli, line 41: What's the type of `self.events`? I mean, the keys are integers but what are their meaning? Index? Why not a list instead?

hathor/manager.py
```diff
@@ -165,6 +165,9 @@ def __init__(self, reactor: Reactor, peer_id: Optional[PeerId] = None, network:
             assert self.tx_storage.indexes is not None
             self.log.debug('enable utxo index')
             self.tx_storage.indexes.enable_utxo_index()
+        if event_storage is not None:
+            self.event_manager = EventManager(event_storage, self.my_peer.id or "")
```
msbrogli, line 169: I feel we should create this object in the builder and just inject it here. What do you think?

msbrogli (changes requested):

hathor/event/event_manager.py
jansegre: A tie means both blocks are voided, and neither is the best. While it's not too precise to say they are orphans, I think we could go with `NETWORK_ORPHAN_BLOCK_FOUND` and then let a re-org deal with it later.

In practice, I really think we should get rid of ties. There's no benefit in allowing ties to exist, it only adds weird edge cases.

hathor/event/event_manager.py
jansegre: Yes. I fixed this in a later commit, I'm pulling those changes, since I think they make sense.

hathor/event/event_manager.py
jansegre: Maybe it makes sense to use floats here, but this isn't an issue for ordering, events are only guaranteed to be sorted by id. If we add any sorting requirement by timestamp, we'd have to make sure an event never has a time lower than the previous event. I'd rather just make the timestamp informative.

jansegre: Yes, I'll fix this.

hathor/event/storage/event_storage.py
jansegre: I'll add it.

hathor/event/storage/memory_storage.py
jansegre: I'll fix this. I'm also pulling `_last_group_id` from a later PR, it uses the same strategy that you suggested.

tests/event/test_event_manager.py
```diff
@@ -0,0 +1,28 @@
...
+        self.event_storage = EventMemoryStorage()
+        self.network = 'testnet'
+        self.manager = self.create_peer(self.network, event_storage=self.event_storage)
+        self.event_manager = self.manager.event_manager
+
+    def test_if_event_is_persisted(self):
+        self.manager.pubsub.publish(HathorEvents.NETWORK_BEST_BLOCK_FOUND,
+                                    event={"test": "test1"})
+        self.run_to_completion()
+        assert self.event_storage.get_event(1) is not None
```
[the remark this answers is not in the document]
jansegre: I'll change to `assertIsNotNone`, but in practice pytest makes it behave the same way. Comparing to the event itself would work if we had the event object, but we only have the event's content that was just epublished.

hathor/event/event_manager.py
jansegre: I agree, I think we should let the error bubble up and just not catch it here.

hathor/event/event_manager.py
jansegre: Fixed.

hathor/event/event_manager.py
jansegre: Fixed.

jansegre: Fixed.

hathor/event/event_manager.py
jansegre: Done. I removed the try/except.

hathor/event/storage/event_storage.py
jansegre: Added.

hathor/event/storage/memory_storage.py
jansegre: Fixed.

jansegre: Done.

tests/event/test_event_manager.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/event/event_manager.py
msbrogli: Makes sense to use the id for ordering and the timestamp only for reference. Thanks for the info. Please consider adding this note somewhere, maybe in a comment.

msbrogli: Thanks!

hathor/event/event_manager.py
msbrogli: Great!

hathor/event/storage/event_storage.py
msbrogli: Thanks!

hathor/event/storage/memory_storage.py
msbrogli: Great! Thanks!

msbrogli: Great! I'll review it again. Thanks!

hathor/event/event_manager.py
```diff
@@ -0,0 +1,93 @@
...
+        self._next_event_id = 0 if last_event is None else last_event.id + 1
+        last_group_id = event_storage.get_last_group_id()
+        self._next_group_id = 0 if last_group_id is None else last_group_id + 1
+        self._current_group_id: Optional[int] = None
+        self._peer_id = peer_id
+
+    def subscribe(self, pubsub: PubSubManager) -> None:
+        """ Subscribe to defined events for the pubsub received
+        """
+        self.log.error(pubsub)
```
luislhl, line 58: What is this log?

hathor/event/event_manager.py
```diff
@@ -0,0 +1,93 @@
...
+
+        for event in events:
+            pubsub.subscribe(event, self._persist_event)
+
+    def _persist_event(self, event: HathorEvents, args: EventArguments) -> None:
+        event_data = args.__dict__['event']
+        if event in _GROUP_START_EVENTS:
+            assert self._current_group_id is None, 'cannot start an event group before the last one is ended'
+            self._current_group_id = self._next_group_id
+            self._next_group_id += 1
```
luislhl, line 80: Isn't it better to increase this only after the call to `self.event_storage.save_event`?

If an exception happens when running it, we don't want to increase the ids, right?

hathor/event/storage/memory_storage.py
```diff
@@ -12,24 +12,37 @@
...
     def __init__(self):
-        self.events: Dict[int, BaseEvent] = {}
+        self._events: List[BaseEvent] = []
+        self._last_event: Optional[BaseEvent] = None
+        self._last_group_id: Optional[int] = None
 
     def save_event(self, event: BaseEvent) -> None:
-        self.events[event.id] = event
+        if event.id != len(self._events):
+            raise ValueError('invalid event.id, ids must be sequential and leave no gaps')
```
luislhl, line 29: Should we have a validation like this for the RocksDB storage?
Maybe not based on the number of keys, but on the id of the last event

tests/event/test_event_manager.py
```diff
@@ -0,0 +1,28 @@
...
+        self.network = 'testnet'
+        self.manager = self.create_peer(self.network, event_storage=self.event_storage)
+        self.event_manager = self.manager.event_manager
+
+    def test_if_event_is_persisted(self):
+        self.manager.pubsub.publish(HathorEvents.NETWORK_BEST_BLOCK_FOUND,
+                                    event={"test": "test1"})
+        self.run_to_completion()
+        self.assertIsNotNone(self.event_storage.get_event(0))
+
```
luislhl, line 21: Should we have tests for some edge cases, like receiving 2 "group start events" in sequence, or 2 "group end events", or trying to save an event that has a non-sequential id?

hathor/event/event_manager.py
jansegre: Yeah, that makes sense.

hathor/event/storage/memory_storage.py
jansegre: That'd be good.

tests/event/test_event_manager.py
jansegre: Yeah. I'll add tests for those.

hathor/event/event_manager.py
msbrogli: Thanks!

hathor/event/event_manager.py
msbrogli: We have to discuss how to get rid of it. Let's say we keep the latest best block. In this case, the consensus would depend on the downloading order. Even though it would be rare, it seems like an issue.

hathor/event/event_manager.py
msbrogli: Thanks!

hathor/manager.py
msbrogli: You're right. It's still an open PR. We can do it later in a refactor. Thanks!

tests/event/test_event_manager.py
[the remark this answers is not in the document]
msbrogli: Thanks!

msbrogli (approved):

hathor/event/event_manager.py
jansegre: Yeah, definitely. This should definitely be discussed better, but there are certainly many ways to do it without making it depend on order of operations, like using the numeric hash of the block.

luislhl (approved):

msbrogli (approved):

hathor/event/event_manager.py
msbrogli: That's right! Let's do it somewhere else. Thanks!

PR #455, opened 2022-08-08
fix(indexes): RocksDB empty address test fail for some multisig addresses

jansegre:
These issue was caught recently. In short it happens because the behavior of the RocksDB key iterator's `get` method is to return the key that was given to seek when it falls through the end of the database, so the method that was prepared to only accept keys that can be found in the database would fail because the seek key is always a prefix and not a key that could be in the database.

In practice this was found happening on wallets that have a multisig address which when they start loading will open a websocket connection and send a subscribe command, which will end up calling the empty address check which before this fix would raise an AssertionError and cause the alert that generated this incident: https://github.com/HathorNetwork/on-call-incidents/issues/50

The test added fails before the fix. The fix adds an additional query (a database get) to the empty key test implementation, but this seems unavoidable, and also it is as fast as a seek so the complexity is not affected.

### DIFF
--- a/hathor/indexes/rocksdb_address_index.py
+++ b/hathor/indexes/rocksdb_address_index.py
@@ -148,11 +148,15 @@ class RocksDBAddressIndex(AddressIndex, RocksDBIndexUtils):
     def is_address_empty(self, address: str) -> bool:
         self.log.debug('seek to', address=address)
         it = self._db.iterkeys(self._cf)
-        it.seek(self._to_key(address))
-        res = it.get()
-        if not res:
+        seek_key = self._to_key(address)
+        it.seek(seek_key)
+        cf_key = it.get()
+        if not cf_key:
             return True
-        _cf, key = res
+        _cf, key = cf_key
+        # XXX: this means we reached the end it did not found any key
+        if key == seek_key:
+            return True
         addr, _, _ = self._from_key(key)
         is_empty = addr != address
         self.log.debug('seek empty', is_empty=is_empty)
--- a/tests/tx/test_indexes.py
+++ b/tests/tx/test_indexes.py
@@ -609,6 +609,41 @@ class BaseIndexesTest(unittest.TestCase):
         self.assertTrue(addresses_indexes.is_address_empty(address))
         self.assertEqual(addresses_indexes.get_sorted_from_address(address), [])
 
+    def test_addresses_index_last(self):
+        """
+        See these for more context on why this test was added:
+        - https://github.com/HathorNetwork/hathor-core/pull/455
+        - https://github.com/HathorNetwork/on-call-incidents/issues/50
+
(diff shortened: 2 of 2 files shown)


pedroferreira1 (approved):

hathor/indexes/rocksdb_address_index.py
```diff
@@ -148,11 +148,14 @@ def get_sorted_from_address(self, address: str) -> List[bytes]:
...
         it = self._db.iterkeys(self._cf)
-        it.seek(self._to_key(address))
-        res = it.get()
-        if not res:
+        key = self._to_key(address)
+        it.seek(key)
+        cf_key = it.get()
+        if not cf_key:
+            return True
+        if self._db.get(cf_key) is None:
```
msbrogli, line 156: Why don't we do `key == cf_key` instead of hitting the database? I'd also add a note explaining that the `it.get()` returns the seek key when it reaches the end of the database.

msbrogli (changes requested):

tests/tx/test_indexes.py
```diff
@@ -609,6 +609,23 @@ def test_addresses_index_empty(self):
...
+        assert settings.P2PKH_VERSION_BYTE[0] < settings.MULTISIG_VERSION_BYTE[0]
+
+        # generating a multisig address:
+        address = generate_multisig_address(generate_multisig_redeem_script(2, [
+            bytes.fromhex('0250bf5890c9c6e9b4ab7f70375d31b827d45d0b7b4e3ba1918bcbe71b412c11d7'),
+            bytes.fromhex('02d83dd1e9e0ac7976704eedab43fe0b79309166a47d70ec3ce8bbb08b8414db46'),
+        ]))
+        assert address is not None
+
+        self.assertTrue(addresses_indexes.is_address_empty(address))
```
msbrogli, line 626: Maybe also check for the address `b'\xff' * 34` as well since it's guaranteed to reach the end of the records in the index.

jansegre: Done, but I'm using `'\x7f' * 34` instead, basically because the method takes an `str` and `\xff` is out of the range of what can be encoded to bytes.

hathor/indexes/rocksdb_address_index.py
jansegre: That makes sense, I changed the method to use that check instead of a `db.get`.

msbrogli (approved):

hathor/indexes/rocksdb_address_index.py
msbrogli: Great! Thanks!

tests/tx/test_indexes.py
msbrogli: Got it! Great!

PR #458, opened 2022-08-11
feat: add client timeout to whitelist update GET request

pedroferreira1:
### Acceptance Criteria

- An exception when updating the p2p whitelist stops the looping call nowadays. We must handle this case, restarting the looping call in case it's stopped by an exception.
- The whitelist update GET request must have a client timeout in order to prevent hanging forever waiting for a server response.

Root Cause Analysis: https://github.com/HathorNetwork/hathor-core/issues/406#issuecomment-1213019357

Closes https://github.com/HathorNetwork/on-call-incidents/issues/35
Closes https://github.com/HathorNetwork/hathor-core/issues/406

### DIFF
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -137,9 +137,9 @@ class ConnectionsManager:
     def start(self) -> None:
         self.lc_reconnect.start(5, now=False)
         if settings.ENABLE_PEER_WHITELIST:
-            self.start_whitelist_reconnect()
+            self._start_whitelist_reconnect()
 
-    def start_whitelist_reconnect(self) -> None:
+    def _start_whitelist_reconnect(self) -> None:
         # The deferred returned by the LoopingCall start method
         # executes when the looping call stops running
         # https://docs.twistedmatrix.com/en/stable/api/twisted.internet.task.LoopingCall.html
@@ -152,7 +152,7 @@ class ConnectionsManager:
             We log the error and start the looping call again.
         """
         self.log.error('whitelist reconnect had an exception. Start looping call again.', args=args, kwargs=kwargs)
-        self.reactor.callLater(30, self.start_whitelist_reconnect)
+        self.reactor.callLater(30, self._start_whitelist_reconnect)
 
     def stop(self) -> None:
         if self.lc_reconnect.running:
@@ -350,7 +350,7 @@ class ConnectionsManager:
         return d
 
     def _update_whitelist_timeout(self, param: Union[Failure, Optional[bytes]],
-                                         timeout_call: 'IDelayedCall') -> Union[Failure, Optional[bytes]]:
+                                  timeout_call: 'IDelayedCall') -> Union[Failure, Optional[bytes]]:
         """ This method is always called for both cb and errback in the update whitelist get request deferred.
             Because of that, the first parameter type will depend, will be a failure in case of errback
             or optional bytes in case of cb (see _update_whitelist_cb).
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -36,9 +36,10 @@ from hathor.transaction import BaseTransaction
 from hathor.util import Random, Reactor
 
 if TYPE_CHECKING:
+    from twisted.internet.interfaces import IDelayedCall
+
     from hathor.manager import HathorManager
     from hathor.p2p.factory import HathorClientFactory, HathorServerFactory
-    from twisted.internet.interfaces import IDelayedCall
 
 logger = get_logger()
 settings = HathorSettings()
@@ -147,7 +148,7 @@ class ConnectionsManager:
             and ends up stopping the looping call.
             We log the error and start the looping call again.
         """
-        self.log.error('whitelist reconnect had an exception. Will start looping call again.', args=args, kwargs=kwargs)
+        self.log.error('whitelist reconnect had an exception. Start looping call again.', args=args, kwargs=kwargs)
         self.reactor.callLater(30, self.start_whitelist_reconnect)
 
     def stop(self) -> None:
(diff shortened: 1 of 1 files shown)


hathor/p2p/manager.py
```diff
@@ -132,7 +134,22 @@ def __init__(self, reactor: Reactor, my_peer: PeerId, server_factory: 'HathorSer
     def start(self) -> None:
         self.lc_reconnect.start(5, now=False)
         if settings.ENABLE_PEER_WHITELIST:
-            self.wl_reconnect.start(30)
+            self.start_whitelist_reconnect()
+
+    def start_whitelist_reconnect(self) -> None:
```
msbrogli, line 139: I'd rename it to `start_whitelist_update(...)`. It seems simpler to name it after what it does. The timeout is just part of the starting process.

hathor/p2p/manager.py
```diff
@@ -132,7 +134,22 @@ def __init__(self, reactor: Reactor, my_peer: PeerId, server_factory: 'HathorSer
...
+        # The deferred returned by the LoopingCall start method
+        # executes when the looping call stops running
+        # https://docs.twistedmatrix.com/en/stable/api/twisted.internet.task.LoopingCall.html
+        d = self.wl_reconnect.start(30)
+        d.addErrback(self._handle_whitelist_reconnect_err)
+
+    def _handle_whitelist_reconnect_err(self, *args: Any, **kwargs: Any) -> None:
```
msbrogli, line 146: I'd rename it to `_handle_whitelist_update_error(...)`. What do you think?

hathor/p2p/manager.py
```diff
@@ -132,7 +134,22 @@ def __init__(self, reactor: Reactor, my_peer: PeerId, server_factory: 'HathorSer
...
+        """ This method will be called when an exception happens inside the whitelist update
+            and ends up stopping the looping call.
+            We log the error and start the looping call again.
+        """
+        self.log.error('whitelist reconnect had an exception. Start looping call again.', args=args, kwargs=kwargs)
+        self.reactor.callLater(30, self.start_whitelist_reconnect)
```
msbrogli, line 152: Why should we wait 30 seconds? Why not at once?

hathor/p2p/manager.py
```diff
@@ -317,11 +334,29 @@ def update_whitelist(self) -> Deferred[None]:
             settings.WHITELIST_URL.encode(),
             Headers({'User-Agent': ['hathor-core']}),
             None)
+        # Twisted Agent does not have a direct way to configure the HTTP client timeout
+        # only a TCP connection timeout. The callLater below is a manual client timeout
```
msbrogli, line 338: I agree that a TCP connection timeout is not the best for us. But I feel we should enhance this comment to explain that we want a timeout that encompasses the connection and download time.

hathor/p2p/manager.py
```diff
@@ -317,11 +334,29 @@ def update_whitelist(self) -> Deferred[None]:
...
+        # that will cancel the deferred in case it's called
+        timeout_call = self.reactor.callLater(10, d.cancel)
```
msbrogli, line 340: Move it to a constant? It's in seconds, right? I feel we should let it go for 30-45 seconds.

hathor/p2p/manager.py
```diff
@@ -317,11 +334,29 @@ def update_whitelist(self) -> Deferred[None]:
...
+        """ This method is always called for both cb and errback in the update whitelist get request deferred.
+            Because of that, the first parameter type will depend, will be a failure in case of errback
+            or optional bytes in case of cb (see _update_whitelist_cb).
+
+            We just need to cancel the timeout call later and return the first parameter,
+            to continue the cb/errback sequence.
+        """
+        if timeout_call.active():
+            timeout_call.cancel()
+        return param
```
msbrogli, line 358: Why do we have to return it?

hathor/p2p/manager.py
```diff
@@ -317,11 +334,29 @@ def update_whitelist(self) -> Deferred[None]:
...
+        # only a TCP connection timeout. The callLater below is a manual client timeout
+        # that will cancel the deferred in case it's called
+        timeout_call = self.reactor.callLater(10, d.cancel)
+        d.addBoth(self._update_whitelist_cancel_timeout, timeout_call)
         d.addCallback(readBody)
         d.addErrback(self._update_whitelist_err)
         d.addCallback(self._update_whitelist_cb)
         return d
 
+    def _update_whitelist_cancel_timeout(self, param: Union[Failure, Optional[bytes]],
```
msbrogli, line 347: Maybe rename it to `_whitelist_update_timeout(...)`?

msbrogli (approved):

hathor/p2p/manager.py
pedroferreira1: I think I prefer `start_whitelist_reconnect` because the name of the `LoopingCall` variable is `wl_reconnect` and this method starts this looping call.

hathor/p2p/manager.py
pedroferreira1: The same reason as the previous one.

hathor/p2p/manager.py
pedroferreira1: The idea was to give some seconds for the server to recover because this errback might be called when an exception happens when getting the whitelist from the server. I used 30 just because it's the same interval between whitelist calls.

hathor/p2p/manager.py
pedroferreira1: When you have two callbacks in sequence, the return of the first is the param of the second, doesn't it?

hathor/p2p/manager.py
pedroferreira1: I tried to keep a similar name than the other methods

```
update_whitelist
_update_whitelist_cb
_update_whitelist_err
_update_whitelist_cancel_timeout
```

hathor/p2p/manager.py
msbrogli: Makes sense.

hathor/p2p/manager.py
msbrogli: Ok. Makes sense.

hathor/p2p/manager.py
msbrogli: I'd remove the cancel. Maybe just _update_whitelist_timeout?

hathor/p2p/manager.py
msbrogli: Makes sense. It might be good to wait a bit to let the network recover.

hathor/p2p/manager.py
msbrogli: It is. I wasn't considering that we would have another callback. But it is always good to return it anyway. Thanks!

msbrogli (approved):

jansegre (approved):

hathor/p2p/manager.py
pedroferreira1: Created a constant in the file and set as 45s

PR #464, opened 2022-08-18
fix: peers connected metric in Prometheus

luislhl:
design: https://github.com/HathorNetwork/ops-tools/issues/427

### Acceptance Criteria
- We should make the service return to Prometheus the same count returned by /v1a/status
- We should also include in Prometheus the 2 other counters we have in /v1a/status: connecting_peers, handshaking_peers and known_peers.

### Notes

Some Soft Voided tests broke because of the changes in this PR.

The root cause for this was that the changes interfered with the ~pseudo-random number generator in~ `hathor/simulator/tx_generator.py`, making it generate a different blockchain in the tests. See https://github.com/HathorNetwork/hathor-core/pull/464#discussion_r962025563

Since the tests depended on a hardcoded tx hash that we knew beforehand would be generated by the simulator, they couldn't find this tx hash in the newly generated blockchain and broke.

To solve this, I had to print the new blockchains and choose new txs to be marked as soft voided in the tests.

 `test_soft_voided4` also required a increase in the simulator running time, to make it generate a tx that wasn't confirmed by a block yet, which was the only way I could make it work.
 
 I'll upload here the pdfs of the old and new blockchains generated in each soft voided test, to facilitate review
 
 **test_soft_voided**
 Before:
[test_soft_voided_before.pdf](https://github.com/HathorNetwork/hathor-core/files/9481254/test_soft_voided_before.pdf)

 After:
[test_soft_voided_after.pdf](https://github.com/HathorNetwork/hathor-core/files/9481257/test_soft_voided_after.pdf)

**test_soft_voided3**
Before:
[test_soft_voided3_before.pdf](https://github.com/HathorNetwork/hathor-core/files/9481260/test_soft_voided3_before.pdf)

After:
[test_soft_voided3_after.pdf](https://github.com/HathorNetwork/hathor-core/files/9481261/test_soft_voided3_after.pdf)

**test_soft_voided4**
Before:
[test_soft_voided4_before.pdf](https://github.com/HathorNetwork/hathor-core/files/9481262/test_soft_voided4_before.pdf)

After:
[test_soft_voided4_after.pdf](https://github.com/HathorNetwork/hathor-core/files/9481263/test_soft_voided4_after.pdf)


tests/consensus/test_soft_voided4.py
```diff
@@ -43,7 +43,7 @@ def test_soft_voided(self):
 
         gen_tx2 = self.simulator.create_tx_generator(manager2, rate=10 / 60., hashpower=1e6, ignore_no_funds=True)
         gen_tx2.start()
-        self.simulator.run(900)
+        self.simulator.run(950)
```
jansegre, line 46: I'm assuming this had to change because the case changed slightly se it needed to run for longer to produce a similar situation. Right?

jansegre (approved):

hathor/p2p/manager.py
```diff
@@ -158,6 +158,16 @@ def stop(self) -> None:
         if self.lc_reconnect.running:
             self.lc_reconnect.stop()
 
+    def _get_peers_count(self) -> Dict[str, int]:
+        """Get a dict containing the count of peers in each state"""
+
+        return {
```
msbrogli, line 164: I don't like using dicts internally because it messes up with linters. For instance, a linter cannot check whether a given key exists or not. I feel we should use named tuples instead here.

msbrogli (changes requested):
I still have to double check the changes in the soft voided tests. I'm submitting my review now but I'll check all tests before I approve it.

hathor/p2p/manager.py
```diff
@@ -274,7 +300,11 @@ def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
                 # chance it can happen if both connections start at the same time and none of them has
                 # reached READY state while the other is on PEER_ID state
                 self.connected_peers[protocol.peer.id] = existing_protocol
-        self.pubsub.publish(HathorEvents.NETWORK_PEER_DISCONNECTED, protocol=protocol)
+        self.pubsub.publish(
```
msbrogli, line 303: I'm struggling to understand why your changes have affected the simulation of the soft voided tests. Do you understand it?

tests/consensus/test_soft_voided.py
```diff
@@ -19,7 +19,7 @@ def assertNoParentsAreSoftVoided(self, tx):
             self.assertNotIn(settings.SOFT_VOIDED_ID, tx2_voided_by)
 
     def test_soft_voided(self):
-        txA_hash = bytes.fromhex('4586c5428e8d666ea59684c1cd9286d2b9d9e89b4939207db47412eeaabc48b2')
+        txA_hash = bytes.fromhex('1ae4ce163495279dafddca041b7c99abd71af55f29568746f3a20deead15f14d')
```
msbrogli, line 22: The pdfs attached to this PR do not have the labels txA. Why is that? Did you disable it?

tests/consensus/test_soft_voided3.py
```diff
@@ -18,7 +19,7 @@ def assertNoParentsAreSoftVoided(self, tx):
             self.assertNotIn(settings.SOFT_VOIDED_ID, tx2_voided_by)
 
     def test_soft_voided(self):
-        txA_hash = bytes.fromhex('4586c5428e8d666ea59684c1cd9286d2b9d9e89b4939207db47412eeaabc48b2')
+        txA_hash = bytes.fromhex('1ae4ce163495279dafddca041b7c99abd71af55f29568746f3a20deead15f14d')
```
msbrogli, line 22: The same here. I can't see txA in both pdfs. Why is that?

tests/consensus/test_soft_voided4.py
```diff
@@ -12,8 +12,8 @@ class BaseSoftVoidedTestCase(SimulatorTestCase):
     seed_config = 5988775361793628169
 
     def test_soft_voided(self):
-        txA_hash = bytes.fromhex('4586c5428e8d666ea59684c1cd9286d2b9d9e89b4939207db47412eeaabc48b2')
-        txB_hash = bytes.fromhex('d19bee149abdce63bbfd523c90c3cf110563970a34100b2a62583a1eba48dfc8')
+        txA_hash = bytes.fromhex('1ae4ce163495279dafddca041b7c99abd71af55f29568746f3a20deead15f14d')
+        txB_hash = bytes.fromhex('6343d6549f6743bd1718d92919f5dabf6705762953aaec2b14e68cb048b4207a')
```
msbrogli, line 16: Same here. Can't find txA and txB in the pdfs.

msbrogli (commented):
I checked and all soft voided tests seems to be correctly fixed. Good job! The last confirmation for me will be checking the pdfs with the correct labels.

We need to open a KTLO issue to change this tests in a way they do not depend on hard-coded hashes.

hathor/p2p/manager.py
luislhl: I've tried debugging this to find differences when executing the tests with and without my changes.

I know which of my changes specifically triggers the problem. It's the one in line `240` of this file. If I remove it, the tests work without modifications.

I got down to the point where I noticed that the tx value generated in `hathor/simulator/tx_generator.py:99` was the difference that impacted the tx hashes. After 16 generated transactions, this value starts to be different when comparing the test runs with and without my changes.

Initially I thought that my changes could be impacting the random number generator somehow, but I don't think this is the case anymore, because actually the `balance.available` there is also coming different in the 17th generated transaction when comparing the test runs.

So maybe there is some kind of race condition that got impacted because I have added a new `self.clock.callLater` call, which is impacting the wallet balance during the tests. But I didn't went farther than this.

tests/consensus/test_soft_voided4.py
luislhl: Yeah, this sentence in the description talks about this:

> test_soft_voided4 also required a increase in the simulator running time, to make it generate a tx that wasn't confirmed by a block yet, which was the only way I could make it work.

Basically the change was that there were some blocks as the last generated things in the new blockchain, and this was apparently causing the test case to break for any transaction I chose.

After I increase the simulation run a little, I managed to get some additional transactions after these blocks, which were still unconfirmed by any block. Choosing these transactions made the test case work again.

tests/consensus/test_soft_voided.py
luislhl: You're right. I think I may have added partial pdfs (which I generated in the middle of the test) instead of full pdfs.

I'll generate new ones that are full, for both the "before" and "after" and update in the description

tests/consensus/test_soft_voided3.py
luislhl: I think in this case I may have generated the pdfs before adding the labels in the code, since the labels didn't exist before.

I'll generate new ones with the labels.

hathor/p2p/manager.py
luislhl: Fixed in https://github.com/HathorNetwork/hathor-core/pull/464/commits/3c0ab8048a073a89664b618bee4043f0599cef9f

tests/consensus/test_soft_voided4.py
luislhl: labels added in https://github.com/HathorNetwork/hathor-core/pull/464/commits/f6e11d93a012840ad3b007baf5ffa2f22d14430f

hathor/p2p/manager.py

hathor/p2p/manager.py
msbrogli: Got it. Thanks for the explanation.

tests/consensus/test_soft_voided.py
msbrogli: Thanks!

tests/consensus/test_soft_voided3.py
msbrogli: Thanks!

tests/consensus/test_soft_voided4.py
msbrogli: Thanks!

tests/consensus/test_soft_voided.py
msbrogli, line 22: You won't need this change after PR #488 .

msbrogli (changes requested):

tests/consensus/test_soft_voided4.py
msbrogli: You won't need this change after PR https://github.com/HathorNetwork/hathor-core/pull/488 .

msbrogli (approved):

PR #479, opened 2022-09-02
fix(sync): downloader could wrongfully re-order pending transactions

### LINKED ISSUE #465: Node sync can get stuck for a long time receiving transactions that it can't validate
When syncing from zero (or from old snapshots), the node can get stuck on a sync loop where it receives a large number of transactions it cannot validate because it is missing another transaction, it is eventually able to recover but can take a long time, slowing down the sync time significantly.

jansegre:
Fixes #465

### DIFF
--- a/hathor/p2p/downloader.py
+++ b/hathor/p2p/downloader.py
@@ -53,12 +53,40 @@ class TxDetails:
     requested_index: int
 
     def __init__(self, tx_id: bytes, deferred: Deferred, connections: List['NodeSyncTimestamp']):
+        self.log = logger.new()
         self.tx_id = tx_id
         self.deferred = deferred
         self.connections = connections
         self.downloading_deferred = None
         self.requested_index = 0
         self.retry_count = 0
+
+    def drop_connection(self, connection: 'NodeSyncTimestamp') -> bool:
+        """Called by the downloader to remove a connection from the tx_details connections, returns a bool.
+
+        It will return True if after the drop there are no more viable connections to download the tx from, which means
+        the _remove_pending_tx should be called for this tx, returns False otherwise."""
+        try:
+            connection_index = self.connections.index(connection)
+        except ValueError:
+            return False
+        if self.requested_index > connection_index:
+            self.requested_index -= 1
+        self.connections.pop(connection_index)
+        if not self.has_available_connection():
+            if self.downloading_deferred is not None:
+                self.log.debug('cancel downloading deferred', tx=self.tx_id.hex())
+                self.downloading_deferred.cancel()
+                self.downloading_deferred = None
+            return True
+        return False
+
+    def has_available_connection(self) -> bool:
+        """ Whether there is any valid connection which the tx can be downloaded from."""
+        for conn in self.connections:
+            if not conn.protocol.aborting:
+                return True
+        return False
 
     def get_connection(self) -> Optional['NodeSyncTimestamp']:
         """ Get a connection to start the download for this tx detail
@@ -127,6 +155,15 @@ class Downloader:
         self.downloading_buffer = {}
         self.window_size = window_size
 
+    def drop_connection(self, connection: 'NodeSyncTimestamp') -> None:
+        """ Called when a peer is disconnected to remove that connection from all the affected pending transactions."""
+        to_remove: List[bytes] = []
+        for tx_details in self.pending_transactions.values():
+            if tx_details.drop_connection(connection):
+                to_remove.append(tx_details.tx_id)
+        for tx_id in to_remove:
+            self._remove_pending_tx(tx_id)
+
     def get_tx(self, tx_id: bytes, connection: 'NodeSyncTimestamp') -> Deferred:
         """ Add a transaction to be downloaded and add to the DAG.
         """
@@ -185,6 +222,7 @@ class Downloader:
         assert details.downloading_deferred is None
         details.downloading_deferred = connection.request_data(tx_id)
         details.downloading_deferred.addCallback(self.on_new_tx)
+        details.downloading_deferred.addErrback(self.on_error)
 
         # Adding timeout to callback
         fn_timeout = partial(self.on_deferred_timeout, tx_id=tx_id)
@@ -199,6 +237,11 @@ class Downloader:
             It just calls the retry method
         """
         self.retry(tx_id)
+
+    def on_error(self, result: Any) -> None:
+        """ Errback for downloading deferred.
+        """
+        self.log.error('failed to download tx', err=result)
 
     def on_new_tx(self, tx: 'BaseTransaction') -> None:
         """ This is called when a new transaction arrives.
@@ -254,7 +297,8 @@ class Downloader:
         details = self.pending_transactions.get(tx_id, None)
 
         if details is None:
-            # Nothing to retry but should never enter here
+            self.log.error('nothing to retry', tx=tx_id.hex())
+            # XXX: Nothing to retry but should never enter here
             # Maybe a race condition in which the timeout has triggered and the tx has arrived.
             return
 
@@ -284,3 +328,7 @@ class Downloader:
         # also remove from downloading queue
         if tx_id in self.downloading_deque:
             self.downloading_deque.remove(tx_id)
+
+        # also remove from waiting queue
+        if tx_id in self.waiting_deque:
+            self.waiting_deque.remove(tx_id)
--- a/tests/p2p/test_sync.py
+++ b/tests/p2p/test_sync.py
@@ -1,11 +1,16 @@
 import random
 
+from twisted.python.failure import Failure
+
+from hathor.conf import HathorSettings
 from hathor.crypto.util import decode_address
 from hathor.p2p.protocol import PeerIdState
 from hathor.p2p.sync_version import SyncVersion
 from hathor.simulator import FakeConnection
 from hathor.transaction.storage.exceptions import TransactionIsNotABlock
 from tests import unittest
+
+settings = HathorSettings()
 
 
 class BaseHathorSyncMethodsTestCase(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


jansegre:
I'm placing this on "In Progress Done" for a pre-review while I add the unit-tests that cover the issue.

hathor/p2p/downloader.py
```diff
@@ -60,6 +60,22 @@ def __init__(self, tx_id: bytes, deferred: Deferred, connections: List['NodeSync
...
+        """Called by the downloader to remove a connection from the tx_details connections."""
+        if connection not in self.connections:
+            return
+        connection_index = self.connections.index(connection)
+        if self.requested_index > connection_index:
+            self.requested_index -= 1
+        self.connections.pop(connection_index)
+
+    def has_available_connection(self) -> bool:
+        """ Whether this is any valid connection which the tx can be downloaded from."""
```
msbrogli, line 73: Typo: Whether there is...

msbrogli, line 73: Typo: Whether there is...

hathor/p2p/downloader.py
```diff
@@ -60,6 +60,22 @@ def __init__(self, tx_id: bytes, deferred: Deferred, connections: List['NodeSync
         self.requested_index = 0
         self.retry_count = 0
 
+    def drop_connection(self, connection: 'NodeSyncTimestamp') -> None:
+        """Called by the downloader to remove a connection from the tx_details connections."""
+        if connection not in self.connections:
```
msbrogli, line 65: Is it better to go through the list twice or just capture the `ValueError` exception on the `self.connections.index(...)` call?

msbrogli (approved):

hathor/p2p/downloader.py
jansegre: Oh, that makes sense. Apparently there is no `list.find` (like there `str.find`) which would be more convenient. But I'll change to handle the exception instead.

jansegre: (although performance-wise, it's not that obvious what is better, exceptions are usually somewhat expensive)

hathor/p2p/downloader.py
jansegre: Fixed.

tests/p2p/test_sync.py
```diff
@@ -305,6 +310,142 @@ def test_downloader(self):
...
+        - while tx_A and tx_B are in the downloader, these are the key events that trigger the issue:
+          - tx_A is requested for download to peer_X
+          - tx_B is requested for download to peer_X
+          - peer_X disconnects
+          - peer_Y disconnects
+          - download of tx_A timeouts, since there are no nodes to download it from it is removed from the downloader,
+            and it is not retried
+          - peer_X re-connects, sync starts and tx_A and tx_B are added to the downloader
+          - peer_Y re-connects, sync starts and tx_A and tx_B are added to the downloader
+          - download of tx_B timeouts, since there now are peers to download it from, it isn't removed and a retry is
```
msbrogli, line 334: Typo: "since there are peers available to download it from".

tests/p2p/test_sync.py
```diff
@@ -305,6 +310,142 @@ def test_downloader(self):
...
+          - tx_B is downloaded, and now it will be processed before tx_A
+          - tx_A is eventually downloaded (or not, this doesn't have to happen)
+          - tx_B is processed, but it will fail because tx_A has not been added yet
+        """
+        from hathor.p2p.sync_version import SyncVersion
+
+        # ## premise setup
+        #
+        # - peer_X will be self.manager
+        # - peer_Y will be manger2
```
msbrogli, line 345: Typo: manager2

tests/p2p/test_sync.py
```diff
@@ -305,6 +310,142 @@ def test_downloader(self):
...
+
+        # wait for the download of tx_A to timeout but not yet the download of tx_B
+        conn1.disconnect(Failure(Exception('testing')))
+        conn2.disconnect(Failure(Exception('testing')))
+
+        # the previous behavior was to let the tx be removed only when the download request timeouts,
+        self.assertNotIn(tx_A.hash, downloader.pending_transactions)
+        self.assertNotIn(tx_B.hash, downloader.pending_transactions)
+
+        # XXX: the following sequence will trigger the bug when the solution is not being used, but since the behavior
```
msbrogli, line 411: As far as I understood, if you comment out the code below, it would pass since the bug has been fixed. Is it correct? If yes, why don't you let it run?

jansegre: Not really, the sequence fails after the fix because we can't get `details_A` and `details_B`, because after the fix they are immediately removed when the peer disconnects. This could be worked around by not using `details_A`/`details_B` in the following `for`, however the timing for that `for` is somewhat sensitive, we need to advance it enough for the new connection to be added to the list of both tx_details, but not too much that it would download the transactions. The timing could be hardcoded in the number of steps because of the predictability, but if we do that, any change that causes a change in the timing of the sync would potentially make this test meaningless, it's not that it would break the test, but more that it wouldn't, so it would no longer reproduce the problem. Does that make sense?

pedroferreira1 (approved):

tests/p2p/test_sync.py
jansegre: Fixed.

tests/p2p/test_sync.py
jansegre: Fixed.

tests/p2p/test_sync.py
jansegre: I've worked it around and made it so the base setup is a method and there are two tests, one that triggers the bug and checks whether the transaction was added successfully (which only happens when the bug is not triggered), and a second one that tests the change in behavior (that is, if a connection is immediately removed from tx-details when disconnected).

tests/p2p/test_sync.py
```diff
@@ -305,6 +312,145 @@ def test_downloader(self):
...
+        self.assertTrue(self.manager_bug.tx_storage.transaction_exists(self.tx_A.hash))
+
+    def test_downloader_disconnect(self):
+        """ This is related to test_downloader_retry_reorder, but it basically tests the change in behavior instead.
+
+        When a peer disconnects it should be immediately remvoed from the tx-detail's connections list.
+        """
+        self._downloader_bug_setup()
+
+        # disconnect and wait for the download of tx_A to timeout but not yet the download of tx_B
```
pedroferreira1, line 450: I think this comment is not correct here

tests/p2p/test_sync.py
```diff
@@ -305,6 +312,145 @@ def test_downloader(self):
...
+        self.conn1.disconnect(Failure(Exception('testing')))
+        self.conn2.disconnect(Failure(Exception('testing')))
+
```
pedroferreira1, line 453: Should we assert that both conn are removed from the tx details connection list?

jansegre: Yeah, you're right. I think I removed it by mistake when refactoring the setup.

tests/p2p/test_sync.py
jansegre: Yeah, that's for the other test. I'm fixing this too.

tests/p2p/test_sync.py
```diff
@@ -305,6 +312,145 @@ def test_downloader(self):
...
+            self.fail('expected to break out of loop')
+
+        # wait for the download of B to be retried
+        self.clock.advance(11.0)
+
+        # this situation should cause the bug before the fix
+        # just advancing the connection a little bit should be enough to finish syncing without the bug
+        for _ in range(20):
+            self.conn2.run_one_step()
+            # self.clock.advance(0.1)
```
msbrogli, line 438: Can we remove this commented line of code? If not, maybe add a comment explaining what for and when it should be used.

tests/p2p/test_sync.py
msbrogli, line 453: No other asserts here?

msbrogli (approved):

tests/p2p/test_sync.py
jansegre: Fixed.

tests/p2p/test_sync.py
jansegre: Yes, I forgot it commented out from testing a fix. Fixed.

tests/p2p/test_sync.py
jansegre: Fixed.

tests/p2p/test_sync.py
jansegre: Fixed.

msbrogli (approved):

tests/p2p/test_sync.py
msbrogli: Great! Thanks!

tests/p2p/test_sync.py
msbrogli: Great job! Just reviewed the tests and approved the PR.

pedroferreira1 (approved):

hathor/p2p/downloader.py
```diff
@@ -127,6 +144,20 @@ def __init__(self, manager: 'HathorManager', window_size: int = 100):
...
         self.window_size = window_size
 
+    def drop_connection(self, connection: 'NodeSyncTimestamp') -> None:
+        """ Called when a peer is disconnected to remove that connection from all the affected pending transactions."""
+        to_remove: List[bytes] = []
+        for tx_details in self.pending_transactions.values():
+            tx_details.drop_connection(connection)
+            if not tx_details.has_available_connection():
+                to_remove.append(tx_details.tx_id)
+                if tx_details.downloading_deferred is not None:
```
msbrogli, line 154: Should we move this whole logic to the `tx_details.drop_connection(...)`?

msbrogli (approved):

hathor/p2p/downloader.py
jansegre: Yeah, that makes sense.

PR #482, opened 2022-09-06
feat(cli): subcommand to replay json logs as console logs

jansegre:


hathor/cli/replay_logs.py
```diff
@@ -0,0 +1,48 @@
...
+
+    while True:
+        line_with_break = args.input.readline()
+        if not line_with_break:
+            break
+        line = line_with_break.strip()
+        if line.startswith('#'):
+            pass  # ignore
+        elif line.startswith('--'):
+            print('...', file=args.output, flush=True)
```
msbrogli, line 42: When does it happen?

hathor/cli/replay_logs.py
```diff
@@ -0,0 +1,48 @@
...
+        else:
+            try:
+                log_dict = json_loads(line)
+                print(renderer(None, '', log_dict), file=args.output, flush=True)
+            except JSONDecodeError:
+                print('!!!', line, file=args.output, flush=True)
```
msbrogli, line 48: Should we say "parse error"?

msbrogli (approved):

luislhl (approved):

hathor/cli/replay_logs.py
jansegre: I added this to help with printing json logs that have been `grep`'d with context (`-A`, `-B` or `-C`), although it's possible to omit the context separator (`--`) on grep, even if it's possible to omit it it's useful to see when there was a skip.

hathor/cli/replay_logs.py
jansegre: The main reason I added this was to "pass through" any other captured lines that aren't json, in rare cases it might be that an exception isn't captured by the logger and this will make it not too hard to see it. Because this affects multiple lines, if I print "parse error" it will basically be spammed on multiple lines.

PR #483, opened 2022-09-08
feat: p2p readiness healthcheck

luislhl:
design: https://github.com/HathorNetwork/ops-tools/issues/431

### Acceptance Criteria
- We should have a new endpoint `/p2p/readiness` that checks if:
    - The node has blocks with a recent timestamp
    - The node has at least one connected and synced peer

### DIFF
--- a/hathor/p2p/resources/healthcheck.py
+++ b/hathor/p2p/resources/healthcheck.py
@@ -27,7 +27,7 @@ class HealthcheckReadinessResource(Resource):
 
 
 HealthcheckReadinessResource.openapi = {
-    '/p2p/status': {
+    '/p2p/readiness': {
         'x-visibility': 'public',
         'x-rate-limit': {
             'global': [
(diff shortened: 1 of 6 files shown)


hathor/p2p/resources/healthcheck.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if not healthy:
+            request.setResponseCode(503)
+            return json_dumpb({'success': False, 'reason': reason})
+
+        return json_dumpb({'success': True})
+
+
+HealthcheckReadinessResource.openapi = {
+    '/p2p/status': {
```
jansegre, line 30: This should be `/p2p/readiness`.

jansegre (approved):
Looks great to me.

hathor/p2p/resources/healthcheck.py
luislhl:
```suggestion
    '/p2p/readiness': {
```

pedroferreira1 (approved):

PR #484, opened 2022-09-09
feat: support to add a blacklist of peer ids in the netfilter

pedroferreira1:
### Acceptance Criteria

- Support for adding a blacklist of peer ids in the full node initialization (cli).
- Support for listing, adding, and removing netfilter rules through an API.
- The GET/POST APIs were implemented in a way that the user could easily dump -> reload the rules if he needs to restart the node. (this will work for peer id and ip address matches only because the others were more complex to implement and out of the scope of this PR)

#### CLI

Running in the command line:

```
--peer-id-blacklist peerid1 peerid2
```

Using env vars:

```
export HATHOR_PEER_ID_BLACKLIST=[peerid1, peerid2]
```

#### API

GET `/v1a/p2p/netfilter?chain=post_peerid`

Response example:

```
{
    "rules": [
        {
            "uuid": "93095688-8ad5-4b9a-ab6b-c9e7b6d1fab5",
            "chain": {
                "name": "post_peerid",
                "table": {
                    "name": "filter"
                },
                "policy": {
                    "type": "NetfilterAccept",
                    "target_params": {}
                }
            },
            "target": {
                "type": "NetfilterReject",
                "target_params": {}
            },
            "match": {
                "type": "NetfilterMatchPeerId",
                "match_params": {
                    "peer_id": "f7397705bc07aabf6fc3f68de6605d93b560bc832d9ebbdfb0d3bd41e1f9480b"
                }
            }
        }
    ]
}
```

POST `/v1a/p2p/netfilter`

Body example:

```

        {
            "chain": {
                "name": "post_peerid",
            },
            "target": {
                "type": "NetfilterReject",
                "target_params": {}
            },
            "match": {
                "type": "NetfilterMatchPeerId",
                "match_params": {
                    "peer_id": "f7397705bc07aabf6fc3f68de6605d93b560bc832d9ebbdfb0d3bd41e1f9480b"
                }
            }
        }
```

DELETE `/v1a/p2p/netfilter`

Body example:

```
{
    "chain": "post_peerid",
    "rule_uuid": "93095688-8ad5-4b9a-ab6b-c9e7b6d1fab5",
}
```


jansegre:
> * We must have an easy way to configure a blacklist of peer ids from the cli.

But then restarting the node is needed each time we want to update the blacklist. We could add an API instead to add/remove live (like a single resource with POST/DELETE). What do you think?

hathor/cli/run_node.py
```diff
@@ -103,6 +103,8 @@ def create_parser(self) -> ArgumentParser:
         parser.add_argument('--x-rocksdb-indexes', action='store_true', help='Use RocksDB indexes (currently opt-in)')
         parser.add_argument('--x-enable-event-queue', action='store_true', help='Enable event queue mechanism')
         parser.add_argument('--x-retain-events', action='store_true', help='Retain all events in the local database')
+        parser.add_argument('--peer-id-blacklist', action='extend', default=[], nargs='+', type=str,
```
msbrogli, line 106: I feel you can add an example of use in the PR description, both using cmd line and envvars.

jansegre: Yeah, I think using `HATHOR_PEER_ID_BLACKLIST` could be tricky because it's an `action='extend'` argument. I'm not sure how this is dealt with by `configargparse`.

pedroferreira1: The examples were added in the PR description

hathor/p2p/netfilter/rule.py
```diff
@@ -28,6 +28,32 @@ def __init__(self, match: 'NetfilterMatch', target: 'NetfilterTarget'):
         self.match = match
         self.target = target
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, NetfilterRule):
+            return NotImplemented
+
+        self_chain_name = self.chain.name if self.chain else None
```
msbrogli, line 35: I don't feel it's worth accessing attributes of the chain object. Why don't you just do `self.chain == other.chain`?

hathor/p2p/netfilter/rule.py
```diff
@@ -28,6 +28,32 @@ def __init__(self, match: 'NetfilterMatch', target: 'NetfilterTarget'):
...
+        other_chain_name = other.chain.name if other.chain else None
+        return self_chain_name == other_chain_name and self.match == other.match and self.target == other.target
+
+    def to_json(self) -> Dict[str, Any]:
+        data: Dict[str, Any] = {}
+
+        if self.chain:
+            data['chain'] = self.chain.name
+
+        data_target: Dict[str, Any] = {}
```
msbrogli, line 45: Why don't we let the target object serializes itself into a json?

hathor/p2p/netfilter/rule.py
```diff
@@ -28,6 +28,32 @@ def __init__(self, match: 'NetfilterMatch', target: 'NetfilterTarget'):
...
+        data_target['type'] = type(self.target).__name__
+        data_target['parameters'] = self.target.__dict__
+        data['target'] = data_target
+
+        data_match: Dict[str, Any] = {}
```
msbrogli, line 50: Why don't we let the match object serializes itself into a json?

hathor/p2p/netfilter/rule.py
```diff
@@ -28,6 +28,32 @@ def __init__(self, match: 'NetfilterMatch', target: 'NetfilterTarget'):
...
+            return NotImplemented
+
+        self_chain_name = self.chain.name if self.chain else None
+        other_chain_name = other.chain.name if other.chain else None
+        return self_chain_name == other_chain_name and self.match == other.match and self.target == other.target
+
+    def to_json(self) -> Dict[str, Any]:
+        data: Dict[str, Any] = {}
+
+        if self.chain:
```
msbrogli, line 42: Why don't we let the chain object serializes itself into a json?

hathor/p2p/netfilter/targets.py
```diff
@@ -25,6 +25,18 @@ class NetfilterTarget:
     def execute(self, rule: 'NetfilterRule', context: 'NetfilterContext') -> None:
         pass
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, NetfilterTarget):
+            return NotImplemented
+        eq_type = type(self) == type(other)
+        eq_attrs = True
+        for key in self.__dict__.keys():
```
msbrogli, line 33: Targets are very simply objects. I feel we should remove this generic comparison algorithm and just implement the `__eq__` method in NetfilterLog. All other targets don't need any customization.

hathor/p2p/netfilter/matches.py
```diff
@@ -27,6 +27,18 @@ class NetfilterMatch(ABC):
     def match(self, context: 'NetfilterContext') -> bool:
         raise NotImplementedError
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, NetfilterMatch):
+            return NotImplemented
+        eq_type = type(self) == type(other)
+        eq_attrs = True
+        for key in self.__dict__.keys():
```
msbrogli, line 35: I feel this kind of comparison creates bugs that are hard to find. For instance, it wouldn't work for `NetfilterMatchRemoteURL` since it has many internal attributes that will differ.

Why don't you implement a mixin for both "and" and "or" operation. Something like this:

```
class TwoVariablesMixin:
    def __eq__(self, other):
        if notinstance(other, NetfilterMatch):
            return NotImplemented
        if type(self) != type(other):
            return False
        return self.a == other.a and self.b == other.b
```

And you'd also need to implement custom `__eq__` for `NetfilterMatchPeerId` and `NetfilterMatchRemoteURL`.

tests/p2p/netfilter/test_utils.py
```diff
@@ -0,0 +1,26 @@
+from hathor.p2p.netfilter import get_table
+from hathor.p2p.netfilter.utils import add_peer_id_blacklist
+from tests import unittest
+
+
+class NetfilterUtilsTest(unittest.TestCase):
```
msbrogli, line 6: I feel we also need to test the comparison of objects.

hathor/p2p/resources/netfilter.py
```diff
@@ -0,0 +1,402 @@
...
+    POST to add a rule, GET to get rules and DELETE to remove a rule.
+
+    You must run with option `--status <PORT>`.
+    """
+    isLeaf = True
+
+    def __init__(self, manager: 'HathorManager'):
+        self.manager = manager
+
+    def render_GET(self, request: 'Request') -> bytes:
```
msbrogli, line 122: If one saves the return of the GET method and send it to the POST method, would it work?

That's a common task to dump the current rules and restore them when the full node restarts.

hathor/p2p/netfilter/rule.py
pedroferreira1: This code does not exist anymore

hathor/p2p/netfilter/rule.py
pedroferreira1: Added a `to_json` for each class

hathor/p2p/netfilter/targets.py
pedroferreira1: This code does not exist anymore

hathor/p2p/netfilter/matches.py
pedroferreira1: This code does not exist anymore

hathor/p2p/resources/netfilter.py
pedroferreira1: Changed how we expect the POST format, so we can reuse the GET response on it. Can you see again, please?

hathor/p2p/netfilter/rule.py
pedroferreira1: Added a to_json for each class

hathor/p2p/netfilter/rule.py
pedroferreira1: Added a to_json for each class

tests/p2p/netfilter/test_utils.py
pedroferreira1: The comparison methods were removed and the tests were improved

hathor/p2p/netfilter/chain.py
```diff
@@ -30,12 +30,30 @@ def __init__(self, name: str, policy: 'NetfilterTarget'):
...
 
-    def add_rule(self, rule: 'NetfilterRule') -> 'NetfilterChain':
+    def to_json(self) -> Dict[str, Any]:
+        return {
+            'name': self.name,
+            'table': self.table.to_json() if self.table else None,
+            'policy': self.policy.to_json()
+        }
+
+    def add_rule(self, rule: 'NetfilterRule') -> Optional['NetfilterChain']:
```
msbrogli, line 40: Why return type is `Optional`?

msbrogli (approved):

hathor/p2p/netfilter/chain.py
pedroferreira1: It shouldn't, it's a leftover from a previous change that I had to undo.

jansegre (approved):

PR #485, opened 2022-09-10
feat: RocksDB storage monitoring

luislhl:
https://github.com/HathorNetwork/ops-tools/issues/426

### Acceptance Criteria
- Export the storage size of each RocksDB column family to Prometheus
- Make sure we don't break when using other types of storage

Notes:
- I made a change to the proposed solution in the design. Instead of always using timestamps to control when we should update the metrics, I try to first use the best block height. This way, we are able to update the metrics during a sync in a more responsive way. When using just timestamps, we would take 24h to have the real metric after the sync finishes.
- For fullnodes syncing from the ground up, it can take some time for the metrics of some column families to be updated, because RocksDB may keep data in its memtables and not write it to disk for a while. Only when RocksDB writes them to disk, we will get the storage sizes.
- I didn't implement support for monitoring the storage when using a JSON storage or Memory storage because I'm supposing they are only used in testing scenarios and don't really matter


hathor/metrics.py
```diff
@@ -27,6 +28,7 @@
 from hathor.transaction.storage import TransactionStorage
 from hathor.transaction.storage.cache_storage import TransactionCacheStorage
 from hathor.transaction.storage.memory_storage import TransactionMemoryStorage
+from hathor.transaction.storage.rocksdb_storage import TransactionRocksDBStorage
```
jansegre, line 31: This will break tests on Windows because the rocksdb storage can't be imported. Usually we import only when needed inside functions that will not run in Windows. I see that the way it is being used within an `isinstance` will not work on Windows. Maybe it would be better to have an abstract function in `TransactionStorage` that `TransactionRocksDBStorage` implements, so that the rocksdb specific portions aren't leaked. What do you think?

luislhl: Thanks for the tip. I've checked the error logs, and actually this import was not the problem, but the initialization of this class that I was doing in 2 of my tests.

Skipping these tests in Windows seems to have fixes the problem.

hathor/metrics.py
```diff
@@ -322,10 +339,45 @@ def set_cache_data(self) -> None:
             if misses:
                 self.transaction_cache_misses = misses
 
+    def set_tx_storage_data(self) -> None:
+        self.log.info("Set tx storage data")
```
luislhl, line 343: Remove this

hathor/metrics.py
```diff
@@ -322,10 +339,45 @@ def set_cache_data(self) -> None:
...
+        if isinstance(self.tx_storage, TransactionCacheStorage):
+            store = self.tx_storage.store
+
+        if not isinstance(store, TransactionRocksDBStorage):
+            # We currently only collect metrics for RocksDB
+            return
+
+        try:
+            # Try to use the number of blocks to decide when we should update the tx storage data
+            block_count = self.tx_storage.get_height_best_block()
```
msbrogli, line 355: Should we use the best block's height or timestamp? If we use the timestamp, we can easily see the evolution of the storage use.

hathor/metrics.py
```diff
@@ -322,10 +339,45 @@ def set_cache_data(self) -> None:
...
+
+            if self.last_txstorage_data_block and (
+                block_count - self.last_txstorage_data_block
+            ) < self.txstorage_data_block_interval:
+                return
+
+            self.last_txstorage_data_block = block_count
+        except NotImplementedError:
```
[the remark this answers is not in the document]
luislhl: I think this is unreachable currently, it's probably some leftover. I'll remove it.

hathor/metrics.py
luislhl: I'm not sure I understand what you mean. Why couldn't we easily see the evolution of storage usage by using the blocks height as interval for the updates?

luislhl: But just to explain the reasoning behind using the blocks height, I did this specifically to avoid having a wrong metric value in the first 24h of the fullnode's life.

If using solely the timestamp, we would get the first measure still during sync, and the storage size would still be wrong.

Then we would get the right measure only 24h later.

By using blocks height, I was able to keep updating the storage size in smaller intervals during sync until it is finished, and then we would have approximately the same interval of 24h for updating it again.

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -153,6 +153,21 @@ def get_count_tx_blocks(self) -> int:
...
 
+    def get_sst_files_sizes_by_cf(self, cfs: Optional[List[Any]] = None) -> Dict[str, float]:
+        """Get the SST files sizes of each Column Family in bytes
+
+        :param cfs: The list of column families, defaults to None, in which case all of them are returned
+        :return: A dict containing the names of the cfs and their sizes
+        """
+        column_families = cfs or self._db.column_families
+
+        sizes: Dict[str, float] = {}
```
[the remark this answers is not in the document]
luislhl: You're right, they are bytes.

hathor/prometheus.py
```diff
@@ -136,10 +157,18 @@ def set_new_metrics(self) -> None:
         for metric_name in METRIC_INFO.keys():
             self.metric_gauges[metric_name].set(getattr(self.metrics, metric_name))
 
+        self._set_new_tx_storage_metrics()
         self._set_new_peer_connection_metrics()
 
         write_to_textfile(self.filepath, self.registry)
 
+    def _set_new_tx_storage_metrics(self) -> None:
+        if hasattr(self.metrics, 'rocksdb_cfs_sizes'):
```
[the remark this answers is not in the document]
luislhl: I'll remove this, it was a protection needed only in older versions of the code. But the Metrics class in its current form will always have this attribute.

hathor/metrics.py
```diff
@@ -187,6 +189,21 @@ def __init__(
         self._lc_collect_data = LoopingCall(self._collect_data)
         self._lc_collect_data.clock = reactor
 
+        # Dict that stores the sizes of each column-family in RocksDB
+        self.rocksdb_cfs_sizes = {}
```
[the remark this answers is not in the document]
luislhl: Will add it.

hathor/metrics.py
msbrogli: I got it. It's a clever solution. But we might use the block's timestamp instead of the block's height. Both are increasing sequences but the timestamp would allow us to show a date/time in the x-axis (instead of the block number).

hathor/metrics.py
```diff
@@ -322,10 +337,33 @@ def set_cache_data(self) -> None:
...
+
+        if isinstance(self.tx_storage, TransactionCacheStorage):
+            store = self.tx_storage.store
+
+        if not isinstance(store, TransactionRocksDBStorage):
+            # We currently only collect metrics for RocksDB
+            return
+
+        # Try to use the number of blocks to decide when we should update the tx storage data
+        block_count = self.tx_storage.get_height_best_block()
```
msbrogli, line 351: Note that `block_count` might be different from `best_block.height` since we might have orphan blocks. I guess you meant `best_block_count`, right?

msbrogli (approved):

hathor/metrics.py
luislhl: Yes, `best_block_count` would describe it more accurately

hathor/metrics.py
luislhl: Just registering here what we talked externally, actually the x-axis in Prometheus will still contain a timestamp, this is not determined by the use of the block height.

All metrics in Prometheus are time-based and the timestamp assigned to them is the moment Prometheus collects them from its targets.

hathor/metrics.py
```diff
@@ -187,6 +188,20 @@ def __init__(
         self._lc_collect_data = LoopingCall(self._collect_data)
         self._lc_collect_data.clock = reactor
 
+        # Dict that stores the sizes of each column-family in RocksDB, in bytes
+        self.rocksdb_cfs_sizes: Dict[bytes, float] = {}
```
jansegre, line 192: The type is already defined above along with the other ones. I think the comment with details about the units can be there, actually all other properties have descriptions inside the constructor but type annotation outside, we probably haven't been consistent with this.

jansegre: Or they can all be inside the constructor, I think it's fine. Ideally we could use dataclasses more to avoid these constructors, but I wouldn't bother with this here.

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -153,6 +153,21 @@ def get_count_tx_blocks(self) -> int:
         keys_count = int(keys_bcount)
         return keys_count
 
+    def get_sst_files_sizes_by_cf(self, cfs: Optional[List[Any]] = None) -> Dict[bytes, float]:
+        """Get the SST files sizes of each Column Family in bytes
+
+        :param cfs: The list of column families, defaults to None, in which case all of them are returned
+        :return: A dict containing the names of the cfs and their sizes
+        """
+        column_families = cfs or self._db.column_families
```
jansegre, line 162: There's a slight edge case here where passing `cfs=[]` will return all column families instead of no column families. I think it should be `self._db.column_families if cfs is None else cfs` instead.

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -153,6 +153,21 @@ def get_count_tx_blocks(self) -> int:
         keys_count = int(keys_bcount)
         return keys_count
 
+    def get_sst_files_sizes_by_cf(self, cfs: Optional[List[Any]] = None) -> Dict[bytes, float]:
```
jansegre, line 156: I think we could use `cfs: Optional[List['rocksdb.ColumnFamilyHandle']]` here, that's what I used in other places. In practice it will be the same as using `Any` because `rocksdb` isn't typed, but I think it's better than using `Any` directly.

jansegre (approved):

hathor/metrics.py
luislhl: I though it was a good idea indeed to turn the class into a `dataclass`, and since it seemed easy enough, I did it already: https://github.com/HathorNetwork/hathor-core/pull/485/commits/d5f0e12371d7cc6605e508b62ad6aa4194ed7562

hathor/transaction/storage/rocksdb_storage.py
luislhl: Fixed in https://github.com/HathorNetwork/hathor-core/pull/485/commits/bddd8196b7cdf1124cc30af11d8745822c19d47c

hathor/transaction/storage/rocksdb_storage.py
luislhl: Fixed in https://github.com/HathorNetwork/hathor-core/pull/485/commits/bddd8196b7cdf1124cc30af11d8745822c19d47c

hathor/metrics.py
luislhl: Fixed

hathor/prometheus.py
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/transaction/storage/rocksdb_storage.py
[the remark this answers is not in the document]
msbrogli: Ok!

hathor/metrics.py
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/metrics.py
msbrogli: Got it. Thanks for the explanation.

hathor/metrics.py
msbrogli: Thanks!

hathor/prometheus.py
```diff
@@ -136,10 +157,17 @@ def set_new_metrics(self) -> None:
         for metric_name in METRIC_INFO.keys():
             self.metric_gauges[metric_name].set(getattr(self.metrics, metric_name))
 
+        self._set_new_tx_storage_metrics()
         self._set_new_peer_connection_metrics()
 
         write_to_textfile(self.filepath, self.registry)
 
+    def _set_new_tx_storage_metrics(self) -> None:
```
msbrogli, line 165: Maybe rename to `_set_rocksdb_tx_storage_metrics()`?

hathor/metrics.py
```diff
@@ -57,135 +57,80 @@ class PeerConnectionMetrics:
...
-        self.collect_data_interval = settings.METRICS_COLLECT_DATA_INTERVAL
+            self.reactor = twisted_reactor
 
         # A timer to periodically collect data
         self._lc_collect_data = LoopingCall(self._collect_data)
-        self._lc_collect_data.clock = reactor
+        self._lc_collect_data.clock = self.reactor
+
+        # The number of blocks interval to control periodic collection of RocksDB data
+        # We try to use it if possible, instead of the time interval, because it's better to make sure we update the
```
msbrogli, line 131: As far as I understood, we are always using it. Is that right?

msbrogli (approved):

hathor/metrics.py
luislhl: Yes, fixed in https://github.com/HathorNetwork/hathor-core/pull/485/commits/0cc50d54b6cc541c78340b92480db3e64a40b8dc

hathor/prometheus.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/485/commits/2b4f53ebae7134bc81952faae6e65d2f14b1b2a6

jansegre (approved):

hathor/metrics.py
jansegre: That seems good then.

Issue #487, opened 2022-09-13
Database corruption when stopping node during sync with cache enabled


luislhl:
I'm getting an error when testing hathor-core locally.

Basically, when using cache, if I start syncing it without a snapshot then stop the process with a `Ctrl + C`, when starting it again with the same command I would get this error:

```
2022-09-12 22:05:17 [info     ] [hathor.cli.run_node] hathor-core v0.50.2            genesis=3fdff62 hathor=0.50.2 my_peer_id=<redacted> pid=55649 platform=Linux-5.15.0-47-generic-x86_64-with-glibc2.35 python=3.9.13-CPython settings=/home/luislhl/Workspace/hathor/hathor-core/hathor/conf/mainnet.py
2022-09-12 22:05:17 [info     ] [hathor.cli.run_node] with storage                   path=./data storage_class=TransactionRocksDBStorage
2022-09-12 22:05:17 [info     ] [hathor.cli.run_node] with cache                     capacity=1000 interval=5
2022-09-12 22:05:17 [info     ] [hathor.cli.run_node] with indexes                   indexes_class=RocksDBIndexesManager
2022-09-12 22:05:17 [info     ] [hathor.manager] start manager                  network=mainnet
2022-09-12 22:05:17 [info     ] [hathor.p2p.manager] update whitelist
2022-09-12 22:05:17 [info     ] [hathor.manager] initialize
2022-09-12 22:05:17 [error    ] [hathor.cli.main] Uncaught exception:            
Traceback (most recent call last):
  File "/home/luislhl/Workspace/hathor/hathor-core/hathor/cli/main.py", line 157, in main
    sys.exit(CliManager().execute_from_command_line())
  File "/home/luislhl/Workspace/hathor/hathor-core/hathor/cli/main.py", line 152, in execute_from_command_line
    module.main()
  File "/home/luislhl/Workspace/hathor/hathor-core/hathor/cli/run_node.py", line 703, in main
    RunNode().run()
  File "/home/luislhl/Workspace/hathor/hathor-core/hathor/cli/run_node.py", line 692, in __init__
    self.prepare(args)
  File "/home/luislhl/Workspace/hathor/hathor-core/hathor/cli/run_node.py", line 326, in prepare
    self.start_manager(args)
  File "/home/luislhl/Workspace/hathor/hathor-core/hathor/cli/run_node.py", line 352, in start_manager
    self.manager.start()
[...]

PR #489, opened 2022-09-14
fix: database corruption when stopping node with cache enabled

luislhl:
fixes: https://github.com/HathorNetwork/hathor-core/issues/487

### Acceptance Criteria
We shouldn't corrupt the database when stopping the node.

There is a better description of the error and the exact scenario that triggered it in the issue above.

I've only observed this when running a local fullnode and stopping it with `Ctrl + C`, but I'm not sure if this couldn't also happen when running it with Docker in the cloud.

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -36,7 +36,7 @@ from hathor.profiler import get_cpu_profiler
 from hathor.pubsub import HathorEvents, PubSubManager
 from hathor.transaction import BaseTransaction, Block, MergeMinedBlock, Transaction, TxVersion, sum_weights
 from hathor.transaction.exceptions import TxValidationError
-from hathor.transaction.storage import TransactionCacheStorage, TransactionStorage
+from hathor.transaction.storage import TransactionStorage
 from hathor.transaction.storage.exceptions import TransactionDoesNotExist
 from hathor.util import LogDuration, Random, Reactor
 from hathor.wallet import BaseWallet
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -319,8 +319,7 @@ class HathorManager:
             if wait_stratum:
                 waits.append(wait_stratum)
 
-        if isinstance(self.tx_storage, TransactionCacheStorage):
-            self.tx_storage.flush()
+        self.tx_storage.flush()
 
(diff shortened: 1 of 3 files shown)


jansegre (approved):

hathor/manager.py
```diff
@@ -318,6 +318,10 @@ def stop(self) -> Deferred:
             wait_stratum = self.stratum_factory.stop()
             if wait_stratum:
                 waits.append(wait_stratum)
+
+        if isinstance(self.tx_storage, TransactionCacheStorage):
```
msbrogli, line 322: Why don't you add the `flush()` method to all storages? All others would do nothing while the `TransactionCacheStorage` would.

msbrogli (changes requested):

hathor/manager.py
luislhl: Makes sense, will do it.

msbrogli (approved):

hathor/transaction/storage/transaction_storage.py
```diff
@@ -1116,3 +1123,10 @@ def get_blocks_before(self, hash_bytes: bytes, num_blocks: int = 100) -> List[Bl
                     used.add(parent_hash)
                     pending_visits.append(parent_hash)
         return result
+
+    def flush(self) -> None:
+        """Flushes the storage. It's called during shutdown of the node, for instance.
+
+           Should be implemented by storages that provide some kind of in-memory cache
+        """
```
jansegre, line 1131: I think we can remove the docstring here, so it's not duplicated from the base interface method.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -749,6 +749,13 @@ def update_last_started_at(self, timestamp: int) -> None:
                 continue
             self.set_index_last_started_at(index_db_name, timestamp)
 
+    def flush(self) -> None:
```
jansegre, line 752: This should probably be annotated with `@abstractmethod` so we don't forget to implement it.

jansegre (approved):
I think it looks good, all changes I suggested are really minor.

PR #490, opened 2022-09-21
chore: measure sync and load times

luislhl:
design: https://github.com/HathorNetwork/ops-tools/issues/423

### Acceptance Criteria
- We should have a task that will keep checking if the sync is already finished and log the total sync time when it is, along with contextual info about the environment

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -68,7 +68,7 @@ class HathorManager:
         NO_SYNCED_PEER = "Node doesn't have a synced peer"
 
     # This is the interval to be used by the task to check if the node is synced
-    CHECK_SYNC_STATE_INTERVAL: 30  # seconds
+    CHECK_SYNC_STATE_INTERVAL = 30  # seconds
 
     def __init__(self, reactor: Reactor, peer_id: Optional[PeerId] = None, network: Optional[str] = None,
                  hostname: Optional[str] = None, pubsub: Optional[PubSubManager] = None,
@@ -1142,7 +1142,7 @@ class HathorManager:
             total_sync_time = LogDuration(self.first_time_fully_synced - self.start_time)
             vertex_count = self.tx_storage.get_tx_count() + self.tx_storage.get_block_count()
 
-            # Changing the fields in this log could impact log collector that parse them
+            # Changing the fields in this log could impact log collectors that parse them
             self.log.info('has recent activity for the first time', total_sync_time=total_sync_time,
                           vertex_count=vertex_count, **self.environment_info.as_dict())
 
--- a/hathor/util.py
+++ b/hathor/util.py
@@ -19,7 +19,6 @@ import math
 import sys
 import time
 import warnings
-from argparse import Namespace
 from collections import OrderedDict
 from contextlib import AbstractContextManager
 from dataclasses import asdict, dataclass


hathor/manager.py
```diff
@@ -592,7 +615,17 @@ def _initialize_components_new(self) -> None:
...
         self.tx_storage.update_last_started_at(started_at)
         self.state = self.NodeState.READY
-        self.log.info('ready')
+
+        t1 = time.time()
+        tdt = LogDuration(t1 - t0)
+
+        environment_info = self.environment_info.as_dict() if self.environment_info else {}
+
+        vertex_count = self.tx_storage.get_tx_count() + self.tx_storage.get_block_count()
```
luislhl, line 624: I wasn't sure if I should use this or `vertex_count = self.tx_storage.get_count_tx_blocks()`. 

I chose this way because it seems these methods use indexes, while the other one doesn't

hathor/manager.py
```diff
@@ -592,7 +615,17 @@ def _initialize_components_new(self) -> None:
...
+
+        # Changing the field names in this log could impact log collectors that parse them
+        self.log.info('ready', vertex_count=vertex_count,
```
luislhl, line 627: Note that I'm using here a new nomenclature `vertex_count` for the count of `txs` + `blocks`.

If it's better to use some other nomenclature (like `tx_count`), let me know.

jansegre (approved):

hathor/util.py
```diff
@@ -759,3 +763,33 @@ def is_token_uid_valid(token_uid: bytes) -> bool:
...
+
+
+def get_environment_info(args: Namespace, peer_id: Optional[str]) -> EnvironmentInfo:
+    environment_info = EnvironmentInfo(
+        python_implementation=str(sys.implementation),
+        hathor_core_args=str(args),
+        hathor_core_version=get_hathor_core_version(),
+        peer_id=peer_id,
+        # We want to ignore the testnet suffixes here. "testnet-golf" should be reported only as "testnet".
+        network=settings.NETWORK_NAME.split("-")[0]
```
msbrogli, line 788: Shouldn't we add both the suffix (e.g., mainnet and testnet) and the full network?

hathor/util.py
```diff
@@ -759,3 +763,33 @@ def is_token_uid_valid(token_uid: bytes) -> bool:
...
+    hathor_core_args: str
+    hathor_core_version: str
+    peer_id: Optional[str]
+    network: str
+
+    def as_dict(self):
+        return asdict(self)
+
+
+def get_environment_info(args: Namespace, peer_id: Optional[str]) -> EnvironmentInfo:
```
msbrogli, line 781: I feel `Namespace` is more related to argparse. Maybe this method should receive the `args` as a string to make it more adaptable to other cases.

msbrogli, line 781: Does argparse add parameters passed as envvars to the `Namespace` object?

hathor/manager.py
```diff
@@ -221,6 +223,14 @@ def __init__(self, reactor: Reactor, peer_id: Optional[PeerId] = None, network:
         else:
             self.capabilities = DEFAULT_CAPABILITIES
 
+        # This is included in some logs to provide more context
+        self.environment_info = environment_info
+
+        # Task that will count the total sync time
+        self.lc_check_sync_state = LoopingCall(self.check_sync_state)
+        self.lc_check_sync_state.clock = self.reactor
+        self.lc_check_sync_state_interval = 30  # seconds
```
msbrogli, line 232: Move to a constant? It's ok to be a constant in the class.

hathor/manager.py
```diff
@@ -592,7 +615,17 @@ def _initialize_components_new(self) -> None:
         # XXX: last step before actually starting is updating the last started at timestamps
         self.tx_storage.update_last_started_at(started_at)
         self.state = self.NodeState.READY
-        self.log.info('ready')
+
+        t1 = time.time()
+        tdt = LogDuration(t1 - t0)
```
msbrogli, line 620: Rename to `total_load_time`?

hathor/manager.py
```diff
@@ -1094,6 +1127,21 @@ def is_healthy(self) -> Tuple[bool, Optional[str]]:
...
         return True, None
 
+    def check_sync_state(self):
+        now = time.time()
+
+        if self.has_recent_activity():
+            self.first_time_fully_synced = now
+
+            total_sync_time = LogDuration(self.first_time_fully_synced - self.start_time)
+            vertex_count = self.tx_storage.get_tx_count() + self.tx_storage.get_block_count()
```
msbrogli, line 1137: Maybe use `tx_storage.get_count_tx_blocks()` here as well?

hathor/manager.py
```diff
@@ -1094,6 +1127,21 @@ def is_healthy(self) -> Tuple[bool, Optional[str]]:
...
+
+            # Changing the fields in this log could impact log collector that parse them
+            self.log.info('reached synced state for the first time', total_sync_time=total_sync_time,
```
msbrogli, line 1140: Maybe say "has recent activity for the first time"? I feel the log might mislead us to think the full node is synced while the criteria is related to recent activity. Note that each sync algorithm might have a different criteria for what being synced means.

msbrogli (changes requested):

hathor/util.py
jansegre: Yes, `Namespace` is a merge of the options from the cli and env after parsing them.

hathor/manager.py
luislhl: Actually I was doing the same thing as here on the other place where I needed this.

Do you think it's better to use `tx_storage.get_count_tx_blocks()` on both instead?

hathor/manager.py
luislhl: I tested this, and it seems these methods don't always match on their counts.

In a small local DB I have, I was getting 73 from `get_tx_count/get_block_count` vs 79 from `get_count_tx_blocks`.

There is something about this in a [RocksDB FAQ](https://github.com/aayushKumarJarvis/rocks-wiki/blob/master/RocksDB-FAQ.md), and it seems it's actually expected that the result is not 100% accurate:

> Obtaining an accurate number of keys in any LSM databases like RocksDB is a challenging problem as they have duplicate keys and deletion entries (i.e., tombstones) that will require a full compaction in order to get an accurate number of keys. In addition, if the RocksDB database contains merge operators, it will also make the estimated number of keys less accurate.

hathor/util.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/490/commits/cd78536cfae9fc4d4335f4d7838415a06867e428

hathor/manager.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/490/commits/a273bdd4943e18022ce122159dcb88c3e7781903

hathor/util.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/490/commits/33b1b7b113d4cd61dd9f40c9609258f400f3fb8a

hathor/manager.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/490/commits/a834a24942ae222f2d421b187e9a1675dc3c534d

hathor/manager.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/490/commits/0628898e0e618ecc70ea8f92a3ab5fe9f49d58f9

hathor/manager.py
jansegre: Then I think we should phase-out `get_count_tx_blocks` in favor of the other two, or rather implement `get_count_tx_blocks` based on the other two.

hathor/util.py
msbrogli: Thanks!

hathor/util.py
msbrogli: Thanks!

msbrogli: Thanks!

hathor/manager.py

hathor/manager.py

hathor/manager.py

hathor/manager.py
msbrogli: I agree we should use the same method in both cases. Let's discuss which one is better there.

msbrogli (approved):

hathor/manager.py
msbrogli: @luislhl I feel we should add a test to it. As far as I understood, this test will fail for RocksDB, right? In this case, I feel RocksDB should override `get_count_tx_blocks()` to do exactly `get_tx_count() + get_block_count()`.

luislhl: Yes, it should fail for RocksDB. I'll try to develop such test.

luislhl: I developed the test and a fix proposal in this commit: https://github.com/HathorNetwork/hathor-core/pull/490/commits/6d5f9723ecb642823bd5f0f6f2981241178f6efa

The funny thing is that, before the fix, my test was only failing for the `CacheRocksDBStorageTest`, but not for the `TransactionRocksDBStorageTest`.

I'm not sure why this happened. While testing manually, I was able to reproduce the problem both with and without cache.

So I proposed a fix that will use `get_tx_count() + get_block_count()` both when using cache and when not using it.

@msbrogli @jansegre

luislhl: This change broke one test that is not trivial to fix. So I decided to move it to a new PR: https://github.com/HathorNetwork/hathor-core/pull/496

msbrogli (approved):

hathor/manager.py
jansegre: That makes sense, I think it's a good approach.

jansegre (approved):

jansegre:
I approved this even though it isn't rebased yet. I don't think the rebase should affect anything meaningful on this PR, so it's fine.

PR #495, opened 2022-10-06
refactor(init): Use a builder to create objects and inject dependencies

msbrogli:


jansegre (approved):

hathor/exception.py
```diff
@@ -18,6 +18,11 @@ class HathorError(Exception):
     pass
 
 
+class BuilderError(Exception):
+    """Base class for builder."""
```
pedroferreira1, line 22: Maybe update to "Base class for exceptions in the Builder"?

pedroferreira1 (approved):

hathor/exception.py
msbrogli: Done! Thanks!

PR #496, opened 2022-10-10
chore: avoid using RocksDB keys estimation

luislhl:
This was created from the discussion https://github.com/HathorNetwork/hathor-core/pull/490#discussion_r982751561

### Acceptance Criteria

We should prefer to use the InfoIndex when possible to get a count of the number of vertices, instead of relying on specific logic for each kind of storage to count the number of vertices.

Specifically, we notices that the specific logic we had to count the number of vertices in RocksDB was not accurate. See the link above for more info.

This change highlighted a bug we had in the InfoIndex count of txs/blocks when a split-brain happens. It was counting the txs/blocks of the winner chain twice when the nodes connected again and the winner was chosen. We had to fix this bug in this PR as well.

### DIFF
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -233,6 +233,8 @@ class IndexesManager(ABC):
         # XXX: this method is idempotent and has no result
         self.deps.add_tx(tx)
 
+        # We need to check r1 as well to make sure we don't count twice the transactions/blocks that are
+        # just changing from voided to executed or vice-versa
         if r1 and r3:
             self.info.update_counts(tx)
 
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -173,7 +173,7 @@ class IndexesManager(ABC):
             index.init_start()
 
         self.log.debug('indexes init')
-        for tx in progress(tx_storage.topological_iterator(), log=self.log, total=tx_storage.get_count_tx_blocks()):
+        for tx in progress(tx_storage.topological_iterator(), log=self.log, total=tx_storage.get_vertices_count()):
             tx_meta = tx.get_metadata()
             # feed each transaction to the indexes that they are interested in
             for index_filter, index in indexes_to_init:
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -233,7 +233,7 @@ class IndexesManager(ABC):
         # XXX: this method is idempotent and has no result
         self.deps.add_tx(tx)
 
-        if r3:
+        if r1 and r3:
             self.info.update_counts(tx)
 
         return r3
--- a/hathor/transaction/storage/cache_storage.py
+++ b/hathor/transaction/storage/cache_storage.py
@@ -217,6 +217,12 @@ class TransactionCacheStorage(BaseTransactionStorage):
 
     def get_count_tx_blocks(self) -> int:
         self._flush_to_storage(self.dirty_txs.copy())
+
+        if self.store.is_rocksdb_storage():
+            # We know RocksDB estimation for the number of keys is not 100% accurate, so we prefer to
+            # count using the indexes
+            return self.get_tx_count() + self.get_block_count()
+
         return self.store.get_count_tx_blocks()
 
     def add_value(self, key: str, value: str) -> None:
@@ -230,3 +236,6 @@ class TransactionCacheStorage(BaseTransactionStorage):
 
     def flush(self):
         self._flush_to_storage(self.dirty_txs.copy())
+
+    def is_rocksdb_storage(self) -> bool:
+        return self.store.is_rocksdb_storage()
--- a/hathor/transaction/storage/cache_storage.py
+++ b/hathor/transaction/storage/cache_storage.py
@@ -215,9 +215,9 @@ class TransactionCacheStorage(BaseTransactionStorage):
             self._save_to_weakref(tx)
             yield tx
 
-    def get_vertices_count(self) -> int:
+    def is_empty(self) -> bool:
         self._flush_to_storage(self.dirty_txs.copy())
-        return super().get_vertices_count()
+        return self.store.is_empty()
 
     def add_value(self, key: str, value: str) -> None:
         self.store.add_value(key, value)
--- a/hathor/transaction/storage/cache_storage.py
+++ b/hathor/transaction/storage/cache_storage.py
@@ -230,6 +230,3 @@ class TransactionCacheStorage(BaseTransactionStorage):
 
     def flush(self):
         self._flush_to_storage(self.dirty_txs.copy())
-
-    def is_rocksdb_storage(self) -> bool:
-        return self.store.is_rocksdb_storage()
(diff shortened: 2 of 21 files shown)


tests/tx/test_tx_storage.py
```diff
@@ -162,6 +162,29 @@ def test_storage_basic_v2(self):
         self.assertEqual(2, len(tx_parents_hash))
         self.assertEqual(set(tx_parents_hash), {self.genesis_txs[0].hash, self.genesis_txs[1].hash})
 
+    def test_vertices_count(self):
+        _set_test_mode(TestMode.TEST_ALL_WEIGHT)
+
+        _new_blocks_count = 0
```
msbrogli, line 168: Why don't you rename it to `blocks_count` and initializes it with `1`? This way you can remove the hard `+ 1` at the asserts. You can do the same with `_new_txs_count`.

hathor/transaction/storage/cache_storage.py
```diff
@@ -207,6 +207,12 @@ def get_all_transactions(self):
 
     def get_count_tx_blocks(self) -> int:
         self._flush_to_storage(self.dirty_txs.copy())
+
+        if self.store.is_rocksdb_storage():
```
msbrogli, line 211: I don't understand why you need this here. You are already handling things in the `TransactionRocksDBStorage .get_count_tx_blocks(...)` method.

Anyway, I don't feel the cache storage should change its behavior according to the real storage behind it.

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -151,8 +151,16 @@ def get_all_transactions(self) -> Iterator['BaseTransaction']:
             yield tx
 
     def get_count_tx_blocks(self) -> int:
```
msbrogli, line 153: You might add a parameter here `use_rocksdb_estimation: bool = False`. And I feel it is ok to raise `NotImplementedError` if `use_rocksdb_estimation` is false and it has no indexes.

Another way would be to move this estimation method to new method `get_estimated_count_tx_blocks(...)`. What do you think?

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -151,8 +151,16 @@ def get_all_transactions(self) -> Iterator['BaseTransaction']:
...
-        keys_bcount = self._db.get_property(b'rocksdb.estimate-num-keys', self._cf_tx)
-        keys_count = int(keys_bcount)
+        try:
+            # We know RocksDB estimation for the number of keys is not 100% accurate, so we prefer to
+            # count using the indexes, if available
+            keys_count = self.get_tx_count() + self.get_block_count()
+        except NotImplementedError:
+            # This will happen if self.with_index == False, and the only way out is to fallback
+            # to RocksDB key estimation anyway
+            keys_bcount = self._db.get_property(b'rocksdb.estimate-num-keys', self._cf_tx)
```
msbrogli, line 161: As this returns an estimation instead of the exact value, I feel we should really use it as a fallback unless authorized by the caller. Or we can just move it to another method and let the caller decide what's best for their usage.

msbrogli (changes requested):

hathor/transaction/storage/rocksdb_storage.py
jansegre: Couldn't we just move `get_count_tx_blocks` to the index manager and get rid of this method completely? We could even rename it to `get_count_vertices` while we're at it.

jansegre: Or maybe even to the info index, maybe it makes sense to be there.

hathor/transaction/storage/cache_storage.py
luislhl: Because when using cache, the underlying storage has no indices enabled, only the cache storage. So `TransactionRocksDBStorage` would always fallback to using the RocksDB key estimation if cache is enabled.

I wanted to make sure we would still avoid this even when using cache.

But I think @jansegre 's suggestion in https://github.com/HathorNetwork/hathor-core/pull/496/files#r992298106 is  better way to solve this.

What do you think?

I'll give it a try and see if it works.

tests/tx/test_tx_storage.py
luislhl: Done!

hathor/indexes/manager.py
```diff
@@ -233,7 +233,7 @@ def add_tx(self, tx: BaseTransaction) -> bool:
         # XXX: this method is idempotent and has no result
         self.deps.add_tx(tx)
 
-        if r3:
+        if r1 and r3:
```
luislhl, line 236: Add a comment explaining this

hathor/transaction/storage/rocksdb_storage.py
jansegre: I liked it on the info-index, shouldn't we get rid of the `TransactionStorage.get_vertices_count` method then? Or is it still useful to have it?

luislhl: I tried to do this, but a test broke and I had to readd these methods in https://github.com/HathorNetwork/hathor-core/pull/496/commits/6ca5bef1670adcc95d88ad7c88ba0e1079e58ff0

I'll try to find out which test broke. Doing rebases doesn't help with this because we lose the history of tests run for each commit in the PR.

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -150,10 +150,13 @@ def get_all_transactions(self) -> Iterator['BaseTransaction']:
...
             yield tx
 
-    def get_count_tx_blocks(self) -> int:
-        keys_bcount = self._db.get_property(b'rocksdb.estimate-num-keys', self._cf_tx)
-        keys_count = int(keys_bcount)
-        return keys_count
+    def get_vertices_count(self) -> int:
+        try:
+            return super().get_vertices_count()
+        except NotImplementedError:
```
msbrogli, line 156: Why do we need the fallback for each storage type? We might rely on the index only and raise an exception if the index is not there. I feel we will make the index non-optional soon anyway.

msbrogli (approved):

hathor/transaction/storage/rocksdb_storage.py
luislhl: I talked about this in https://github.com/HathorNetwork/hathor-core/pull/496#discussion_r1003423365

Basically one test failed when I did this.

hathor/transaction/storage/rocksdb_storage.py
msbrogli: I guess this discussions went in a different way. I was discussing about the fallback which returns an estimation instead of the actual value.

hathor/transaction/storage/rocksdb_storage.py
msbrogli: It was refactored. Good job. Thanks!

hathor/transaction/storage/rocksdb_storage.py
msbrogli: Thanks!

hathor/transaction/storage/cache_storage.py
msbrogli: Thanks!

jansegre (approved):

msbrogli (approved):

PR #499, opened 2022-10-12
chore(ci): update workflow actions

jansegre:


.github/workflows/main.yml
```diff
@@ -25,7 +25,6 @@ jobs:
           # available OS's: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on
           'os': ['ubuntu-22.04', 'macos-12', 'windows-2022'],
           'include': [
-            {'os': 'ubuntu-20.04', 'python': '3.8'},
```
pedroferreira1, line 28: The `include` array is empty now, can we remove it?

pedroferreira1 (approved):

.github/workflows/main.yml
jansegre: Yes, being empty doesn't cause any problems, but it's a helpful placeholder for when we need to add something again in the future (possibly new Python/PyPy versions that are only available on specific platforms).

luislhl (approved):

PR #500, opened 2022-10-17
feat(events): emit reorg event during consensus

jansegre:

## Acceptance Criteria

- `REORG_STARTED` and `REORG_FINISHED` must be correctly emitted with their corresponding groups

### DIFF
--- a/hathor/consensus.py
+++ b/hathor/consensus.py
@@ -35,21 +35,30 @@ class ConsensusAlgorithmContext:
     """
 
     consensus: 'ConsensusAlgorithm'
+    pubsub: PubSubManager
     block_algorithm: 'BlockConsensusAlgorithm'
     transaction_algorithm: 'TransactionConsensusAlgorithm'
     txs_affected: Set[BaseTransaction]
+    reorg_common_block: Optional[Block]
 
-    def __init__(self, consensus: 'ConsensusAlgorithm') -> None:
+    def __init__(self, consensus: 'ConsensusAlgorithm', pubsub: PubSubManager) -> None:
         self.consensus = consensus
-        self.block_algorithm = consensus.block_algorithm_factory(self)
-        self.transaction_algorithm = consensus.transaction_algorithm_factory(self)
+        self.pubsub = pubsub
+        self.block_algorithm = self.consensus.block_algorithm_factory(self)
+        self.transaction_algorithm = self.consensus.transaction_algorithm_factory(self)
         self.txs_affected = set()
+        self.reorg_common_block = None
 
     def save(self, tx: BaseTransaction) -> None:
         """Only metadata is ever saved in a consensus update."""
         assert tx.storage is not None
         self.txs_affected.add(tx)
         tx.storage.save_transaction(tx, only_metadata=True)
+
+    def mark_as_reorg(self, common_block: Block) -> None:
+        """Must only be called once, will raise an assert error if called twice."""
+        assert self.reorg_common_block is None
+        self.reorg_common_block = common_block
 
 
 class ConsensusAlgorithm:
@@ -79,15 +88,15 @@ class ConsensusAlgorithm:
     """
 
     def __init__(self, soft_voided_tx_ids: Set[bytes], pubsub: PubSubManager) -> None:
-        self.pubsub = pubsub
         self.log = logger.new()
+        self._pubsub = pubsub
         self.soft_voided_tx_ids = frozenset(soft_voided_tx_ids)
         self.block_algorithm_factory = BlockConsensusAlgorithmFactory()
         self.transaction_algorithm_factory = TransactionConsensusAlgorithmFactory()
 
     def create_context(self) -> ConsensusAlgorithmContext:
         """Handy method to create a context that can be used to access block and transaction algorithms."""
-        return ConsensusAlgorithmContext(self)
+        return ConsensusAlgorithmContext(self, self._pubsub)
 
     @cpu.profiler(key=lambda self, base: 'consensus!{}'.format(base.hash.hex()))
     def update(self, base: BaseTransaction) -> None:
@@ -119,14 +128,32 @@ class ConsensusAlgorithm:
                               count=len(to_remove))
                 storage.remove_transactions(to_remove)
                 for tx_removed in to_remove:
-                    self.pubsub.publish(HathorEvents.CONSENSUS_TX_REMOVED, tx_hash=tx_removed.hash)
+                    context.pubsub.publish(HathorEvents.CONSENSUS_TX_REMOVED, tx_hash=tx_removed.hash)
+
+        # emit the reorg started event if needed
+        if context.reorg_common_block is not None:
+            old_best_block = base.storage.get_transaction(best_tip)
+            new_best_block = base.storage.get_transaction(new_best_tip)
+            old_best_block_meta = old_best_block.get_metadata()
+            common_block_meta = context.reorg_common_block.get_metadata()
+            reorg_size = old_best_block_meta.height - common_block_meta.height
+            assert old_best_block != new_best_block
+            assert reorg_size > 0
+            context.pubsub.publish(HathorEvents.REORG_STARTED, old_best_height=best_height,
+                                   old_best_block=old_best_block, new_best_height=new_best_height,
+                                   new_best_block=new_best_block, common_block=context.reorg_common_block,
+                                   reorg_size=reorg_size)
 
         # finally signal an index update for all affected transactions
         for tx_affected in context.txs_affected:
             assert tx_affected.storage is not None
             assert tx_affected.storage.indexes is not None
             tx_affected.storage.indexes.update(tx_affected)
-            self.pubsub.publish(HathorEvents.CONSENSUS_TX_UPDATE, tx=tx_affected)
+            context.pubsub.publish(HathorEvents.CONSENSUS_TX_UPDATE, tx=tx_affected)
+
+        # and also emit the reorg finished event if needed
+        if context.reorg_common_block is not None:
+            context.pubsub.publish(HathorEvents.REORG_FINISHED)
 
     def filter_out_soft_voided_entries(self, tx: BaseTransaction, voided_by: Set[bytes]) -> Set[bytes]:
         if not (self.soft_voided_tx_ids & voided_by):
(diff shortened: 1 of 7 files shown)


hathor/consensus.py
```diff
@@ -30,20 +31,21 @@
 _base_transaction_log = logger.new()
 
 
+@dataclass
```
msbrogli, line 34: As we have multiple fields and the `__init__()` method would have only two parameters, is it worth using a dataclass? I feel it's more complex with no benefits.

hathor/consensus.py
```diff
@@ -119,14 +121,22 @@ def update(self, base: BaseTransaction) -> None:
...
             assert tx_affected.storage.indexes is not None
             tx_affected.storage.indexes.update(tx_affected)
-            self.pubsub.publish(HathorEvents.CONSENSUS_TX_UPDATE, tx=tx_affected)
+            context.pubsub.publish(HathorEvents.CONSENSUS_TX_UPDATE, tx=tx_affected)
+
+        # and also emit the reorg event if needed
+        if context.reorg_started:
+            old_best_block = base.storage.get_transaction(best_tip)
+            new_best_block = base.storage.get_transaction(new_best_tip)
+            context.pubsub.publish(HathorEvents.REORG_FINISHED, old_best_height=best_height,
```
msbrogli, line 137: Should we assert that `old_best_block != new_best_block`?

hathor/consensus.py
```diff
@@ -312,6 +322,11 @@ def update_voided_info(self, block: Block) -> None:
                 # Either we have a single best chain or all chains have already been voided.
                 assert len(valid_heads) <= 1, 'We must never have more than one valid head'
 
+                # Emit reorg start event, only once
+                if not self.context.reorg_started:
```
msbrogli, line 326: This method is just called once per block. Shouldn't we assert that `self.context.reorg_started is False`?

hathor/event/event_manager.py
```diff
@@ -42,9 +42,63 @@
...
+
+def _extract_reorg(data: Dict[str, Any]) -> Dict[str, Any]:
+    return {
+        'reorg_size': data['new_best_height'] - data['old_best_height'],
+        'previous_best_block': data['old_best_block'].hash_hex,
+        'new_best_block': data['new_best_block'].hash_hex,
+    }
+
+
+_EVENT_EXTRACT_MAP: Dict[HathorEvents, Callable[[Dict[str, Any]], Dict[str, Any]]] = {
```
msbrogli, line 74: Shouldn't it be `Dict[HathorEvents, Callable[[EventArguments, Dict[str, Any]]]`?

tests/event/test_event_reorg.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # check events
+        event_count = self.event_storage.get_last_event().id + 1
+        events = []
+        for i in range(event_count):
+            events.append(self.event_storage.get_event(i))
+
+        for ev in events:
+            print(ev.type)
+            # TODO
```
msbrogli, line 46: Should we assert that the correct events were generated? Maybe compare it with a list of expected event types?

msbrogli (changes requested):

hathor/consensus.py
jansegre: These are basically the two versions:

Without dataclasses:

```python
class ConsensusAlgorithmContext:
    """ An instance of this class holds all the relevant information related to a single run of a consensus update.
    """

    consensus: 'ConsensusAlgorithm'
    pubsub: PubSubManager
    block_algorithm: 'BlockConsensusAlgorithm'
    transaction_algorithm: 'TransactionConsensusAlgorithm'
    txs_affected: Set[BaseTransaction] = field(default_factory=set)
    reorg_started: bool = False

    def __init__(self, consensus: 'ConsensusAlgorithm', pubsub: PubSubManager) -> None:
        self.consensus = consensus
        self.pubsub = pubsub
        self.block_algorithm = self.consensus.block_algorithm_factory(self)
        self.transaction_algorithm = self.consensus.transaction_algorithm_factory(self)
        self.txs_affected = set()
        self.reorg_started = False
```

With dataclasses:

```python
@dataclass
class ConsensusAlgorithmContext:
    """ An instance of this class holds all the relevant information related to a single run of a consensus update.
    """

    consensus: 'ConsensusAlgorithm'
    pubsub: PubSubManager
    block_algorithm: 'BlockConsensusAlgorithm' = field(init=False)
    transaction_algorithm: 'TransactionConsensusAlgorithm' = field(init=False)
    txs_affected: Set[BaseTransaction] = field(default_factory=set)
    reorg_started: bool = False

    def __post_init__(self) -> None:
        self.block_algorithm = self.consensus.block_algorithm_factory(self)
        self.transaction_algorithm = self.consensus.transaction_algorithm_factory(self)
```

I honestly much prefer the later, the side-effect being that it can now be constructed with more parameters (explicit txs_affected and reorg_started) which isn't useful, but I don't think it's harmful.

hathor/consensus.py
jansegre: I think that could make sense, but is that a guarantee that we already have?

hathor/consensus.py
jansegre: This is just like the previous one. It makes sense, but I'm not sure if this is a guarantee that we have. Is it not possible to have "nested" reorgs? Externally I can present them as a single reorg, but internally I'm not sure if that's not possible, if it isn't, specially if we have tests that cover that, then I think the assert would be great.

hathor/event/event_manager.py
jansegre: That's a possible alternative, but the annotation matches the current implementation, which feeds a dict to the function, not the event.

tests/event/test_event_reorg.py
jansegre: Yes, exactly, I missed this TODO.

hathor/consensus.py
jansegre: Yeah but I misunderstood our implementation, I thought we had a recursive implementation when updating the scores, where it would be possible to temporarily have a "reorg" but after advancing the updates have another "reorg".

tests/event/test_event_reorg.py
jansegre: Done.

hathor/consensus.py
jansegre: Done.

hathor/event/event_manager.py
```diff
@@ -42,9 +42,63 @@
...
+
+def build_event_data(event: HathorEvents, event_args: EventArguments) -> Dict[str, Any]:
+    """Extract and build event data from event_args for a given event type."""
+    event_extract_fn = _EVENT_EXTRACT_MAP.get(event)
+    if event_extract_fn is None:
+        raise ValueError(f'The given event type ({event}) is not a supported event')
+    try:
+        return event_extract_fn(event_args.__dict__)
+    except NotImplementedError:
+        print(event)
```
luislhl, line 98: Do you intend to keep this print? If so, shouldn't we use a logger?

hathor/event/event_manager.py
```diff
@@ -42,9 +42,63 @@
...
+
+def _empty(data: Dict[str, Any]) -> Dict[str, Any]:
+    return {}
+
+
+def _extract_tx(data: Dict[str, Any]) -> Dict[str, Any]:
+    tx = data['tx']
+    return {
+        'hash': tx.hash_hex,
+        # TODO: the rest
```
luislhl, line 62: Are you leaving this TODO intentionally?

luislhl (approved):

hathor/consensus.py
jansegre: I tried that and several test cases fail, so it doesn't look like we have that guarantee. I think it should be fine to live with this because the alternatives would be buffering events and only emit REORG_STARTED and start a group-id if it turns out there actually was a re-org, _or_ figure out how to change the implementation so that we actually have that guarantee.

hathor/event/event_manager.py
jansegre: I was using this for debugging. I'll remove it, I think any errors should bubble up and not be handled/logged around here.

hathor/event/event_manager.py
jansegre: I'll expand the message, but I think this is out of the scope of this PR and it's perfectly fine to be left as a TODO for a later PR in the project.

hathor/event/event_manager.py
msbrogli: What's the benefit of passing the dict instead of the event itself? I mean, each extractor might access `event_args.__dict__` if they need.

hathor/consensus.py
msbrogli: So we might have zero-length reorgs? Is it the case when we have to handle ties? I'm still figuring out when it happens.

tests/event/test_event_reorg.py
msbrogli: Great! Thanks!

hathor/consensus.py
msbrogli: Great! Thanks!

msbrogli (changes requested):

hathor/event/event_manager.py
jansegre: The only "benefit" is that I expect all extractors to just use the `data`, so they might as take the dict directly.

hathor/consensus.py
jansegre: I'm not sure if it's only related to ties.

Also, I'm considering removing the `reorg_size` field, I don't think it has (1) a clear definition (I chose the height difference, but if there's a reorg from genesis that reaches the same height `reorg_size` will be 0, which might be really misleading), and maybe more importantly (2) an application (if there is any it's not obvious and it wasn't mentioned on the design).

hathor/consensus.py
msbrogli: After a discussion, we decided to keep the original code for now because instances are carrying more than data. We should refactor it later.

hathor/consensus.py
msbrogli: As discussed, `reorg_size` has a clear definition: It is the number of blocks that belonged to the best blockchain and became voided. The only benefit of adding this to the event is that `reorg_size` is a well-known parameters used by exchanges and use cases.

msbrogli (approved):

hathor/consensus.py
msbrogli: Just fixed all issues related to detecting a reorg event. So, I'm marking this conversation as resolved.

hathor/event/event_manager.py
msbrogli: You've already refactored this part. Thanks!

PR #503, opened 2022-11-03
test: Fix too many files open errors

msbrogli:
## Acceptance Criteria

1) Be able to run tests without increasing the maximum number of open files.
2) Properly close the file descriptor after it's been used.

### DIFF
--- a/hathor/storage/rocksdb_storage.py
+++ b/hathor/storage/rocksdb_storage.py
@@ -69,3 +69,6 @@ class RocksDBStorage:
         if cf is None:
             cf = self._db.create_column_family(cf_name, rocksdb.ColumnFamilyOptions())
         return cf
+
+    def close(self) -> None:
+        self._db.close()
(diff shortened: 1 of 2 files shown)


luislhl:
The error we are trying to fix happened only when running tests? Do we have more info about it?

jansegre (approved):

hathor/storage/rocksdb_storage.py
```diff
@@ -69,3 +69,6 @@ def get_or_create_column_family(self, cf_name: bytes) -> 'rocksdb.ColumnFamilyHa
         if cf is None:
             cf = self._db.create_column_family(cf_name, rocksdb.ColumnFamilyOptions())
         return cf
+
+    def close(self) -> None:
+        del self._db
```
jansegre, line 74: This could be `self._db.close()` instead. I'm testing to see if it works for me as well.

jansegre (approved):

hathor/storage/rocksdb_storage.py
msbrogli: Just did it. Thanks!

PR #510, opened 2022-11-07
fix(tests): simple regression from #509 that ended up being merged

jansegre:
The previous PR (#509) was merged with failing tests (that were only noticed post-merge). This PR fixes those.

### DIFF
--- a/hathor/transaction/storage/memory_storage.py
+++ b/hathor/transaction/storage/memory_storage.py
@@ -52,12 +52,12 @@ class TransactionMemoryStorage(BaseTransactionStorage):
             return x
 
     def get_migration_state(self, migration_name: str) -> MigrationState:
-        # XXX: migrations aren't needed for memory storage, this method shouldn't even be called
-        raise NotImplementedError
+        # XXX: it will always return COMPLETED, migrations don't apply to memory storage
+        return MigrationState.COMPLETED
 
     def set_migration_state(self, migration_name: str, state: MigrationState) -> None:
-        # XXX: migrations aren't needed for memory storage, this method shouldn't even be called
-        raise NotImplementedError
+        # XXX: do nothing, migrations have no effect on memory storage
+        pass
 
     def remove_transaction(self, tx: BaseTransaction) -> None:
         assert tx.hash is not None


hathor/transaction/storage/memory_storage.py
```diff
@@ -52,12 +52,12 @@ def _clone(self, x: _Clonable) -> _Clonable:
             return x
 
     def get_migration_state(self, migration_name: str) -> MigrationState:
-        # XXX: migrations aren't needed for memory storage, this method shouldn't even be called
-        raise NotImplementedError
+        # XXX: it will always return COMPLETED, migrations don't apply to memory storage
+        return MigrationState.COMPLETED
```
msbrogli, line 56: #509 has not changed these lines. How has it failed these tests?

msbrogli (approved):

hathor/transaction/storage/memory_storage.py
jansegre: #509 made it so the `pre_init` implementation from `TransactionStorage` on the instance of `CacheStorage` is used instead of on the instance of the internal storage (`self.store`), because the implementation needs to have access to the indexes and `self.store` will always have `self.store.indexes = None`.

Because of that change, the `MemoryStorage` lost its ability to specialize `_check_and_apply_migrations` to do nothing, which is called by `pre_init` as part of the initialization.

With the changes on #509 the `pre_init` implementation from `TransactionStorage` called on the instance of `CacheStorage` via `super().pre_init()` will have access to `self.indexes` _but_ will not use the specialized implementation from `MemoryStorage` and thus will end up calling `get_migration_state` and `set_migration_state`.

So instead of relying on a specialized `_check_and_apply_migrations` to not call those, this PR changes `get_migration_state` and `set_migration_state` to be dummy enough to be able to be called (indirectly) by the default `pre_init` and succeed without issues.

luislhl (approved):

hathor/transaction/storage/memory_storage.py
msbrogli: Got it. Thanks for the explanation.

PR #519, opened 2023-02-02
Implement Pydantic POC

glevco:
**_This PR is not intended to be merged as it is. Please read below._**

### Motivation

Following discussions with @msbrogli and @jansegre, I suggested the addition of the pydantic lib ([docs](https://docs.pydantic.dev/), [GitHub](https://github.com/pydantic/pydantic), [overview article](https://towardsdatascience.com/8-reasons-to-start-using-pydantic-to-improve-data-parsing-and-validation-4f437eae7678)) to the project. From the docs:

> Data validation and settings management using Python type annotations.
> 
> pydantic enforces type hints at runtime, and provides user friendly errors when data is invalid.
> 
> Define how data should be in pure, canonical Python; validate it with pydantic.

It is a well known and established lib in the Python community, and serves as an addition to Python's type hinting system. The main use case for our project would be to enforce types and validate data incoming from external sources, that arrive as JSON objects or dictionaries (REST request bodies, WebSocket request bodies, possibly database records, etc).

With Pydantic, instead of manipulating data with dictionaries and string keys, which is fragile and requires doing a lot validation manually, we create models similarly to Python's data classes, and data is converted and validated automatically. It is also enforced statically and integrates with Mypy, which we already use, keeping the code more structured and less prone to bugs.

### PR Explanation

This PR is simply a POC to demonstrate how pydantic could be use to improve our code. It specifically uses pydantic to structure and validate the `POST` method on the `netfilter` resource. The code runs perfectly for the **success** path, passing the success tests for the `POST` method on the `test_netfilter.py` file. It fails for the **error** paths, because a different validation error is thrown by pydantic, not the custom one that is thrown currently.

I have purposefully not updated the tests, and also not touched other methods (such as `DELETE`), to keep this PR leaner. The intention of this PR is just to validate the installation of pydantic, and not to actually merge the `netfilter` refactor.

What was done here was creating a structure of classes that represent the possible JSON schemas incoming on the `POST` method. All the previous code that was used to manually validate and interpret the JSON, using dictionaries and string keys, could be removed as all of this is done automatically by pydantic, given the new, typed, classes. [Unions](https://docs.pydantic.dev/usage/types/#unions) and [Discriminated Unions](https://docs.pydantic.dev/usage/types/#discriminated-unions-aka-tagged-unions) are specially useful for us.

### Next Steps

Given the installation is approved, only [this PR](https://github.com/HathorNetwork/hathor-core/pull/520) should be merged, _just_ to install pydantic. Then, if prioritized, we can continue working on this refactor (split up test cases, implement pydantic for other methods on this resource, etc). I believe there are even more advantages in creating new code with pydantic in mind, so it's possible that `netfilter` code could benefit form a major refactor in the future, making code simpler and more structured.

For now, I'll keep this as a draft so it can't be merged.


glevco:
As explained in the description, I'm closing this pull request. The PR that actually install Pydantic is https://github.com/HathorNetwork/hathor-core/pull/520.

PR #520, opened 2023-02-02
feat: Install Pydantic

glevco:
This PR simply installs the pydantic lib and sets some configuration. Please read https://github.com/HathorNetwork/hathor-core/pull/519.


pyproject.toml
```diff
@@ -97,7 +98,10 @@ warn_unused_ignores = true
 namespace_packages = true
 show_error_codes = true
 show_error_context = true
-plugins = ["mypy_zope:plugin"]
+plugins = [
+    "mypy_zope:plugin",
+    "pydantic.mypy"
```
glevco, line 103: Fix this, the plugin is not working

hathor/utils/pydantic.py
```diff
@@ -0,0 +1,14 @@
+from pydantic import BaseModel as PydanticBaseModel, Extra
+from pydantic.generics import GenericModel as PydanticGenericModel
+
+
+class BaseModel(PydanticBaseModel):
```
msbrogli, line 5: Missing docstring explaining what it is. Maybe add some guidance for whoever is gonna use it in the future.

hathor/utils/pydantic.py
```diff
@@ -0,0 +1,14 @@
...
+    class Config:
+        allow_mutation = False
+        extra = Extra.forbid
+
+
+class GenericModel(PydanticGenericModel):
```
msbrogli, line 11: Missing docstring explaining what it is. Maybe add some guidance for whoever is gonna use it in the future.

msbrogli (changes requested):

hathor/utils/pydantic.py
glevco: Done!

hathor/utils/pydantic.py
glevco: Done!

pyproject.toml
glevco: For some reason it was conflicting with `mypy_zope:plugin`. I found out that inverting the order works.

jansegre (approved):

msbrogli (approved):

hathor/utils/pydantic.py
msbrogli: Thanks!

hathor/utils/pydantic.py
msbrogli: Thanks!

PR #521, opened 2023-02-06
chore: Test Improvements

glevco:
- ~~Add ability to pass a `HathorSettings` object to `CliBuilder.create_manager()` to help mocking settings on tests.~~
- Change way filepath is constructed on `_load_peer_id_pool()`, so it's calculated. Before, with the hardcoded path, running tests from any directory other then the project root would cause errors. This prevented running specific test files from an IDE.

Depends on https://github.com/HathorNetwork/hathor-core/pull/520.

_EDIT: I have rolled back the first change as I found a better way to do it and it's not necessary anymore._


jansegre (changes requested):
I think this won't work quite as expected. I'll talk to you to see a way around it. Anyway, I the change proposed on this PR shouldn't be merged.

hathor/builder.py
```diff
@@ -26,6 +26,8 @@
 from twisted.web import server
 from twisted.web.resource import Resource
 
+from hathor.conf.get_settings import HathorSettings as DefaultHathorSettings
```
msbrogli, line 29: You don't need the get_settings.

hathor/builder.py
```diff
@@ -362,7 +367,7 @@ def create_resources(self, args: Namespace) -> server.Site:
         )
         from hathor.websocket import HathorAdminWebsocketFactory, WebsocketStatsResource
 
-        settings = HathorSettings()
+        settings = DefaultHathorSettings()
```
msbrogli, line 370: Why are you mixing different settings? Shouldn't all created objects use the same?

tests/unittest.py
```diff
@@ -32,8 +32,14 @@ def shorten_hash(container):
     return container_type(h[-2:].hex() for h in container)
 
 
-def _load_peer_id_pool(file_path: str = 'tests/peer_id_pool.json') -> Iterator[PeerId]:
+def _load_peer_id_pool(file_path: Optional[str] = None) -> Iterator[PeerId]:
     import json
+
+    if file_path is None:
+        this_file_path = os.path.dirname(__file__)
```
msbrogli, line 39: Maybe move it to a `get_default_peer_id_filepath(...)`?

msbrogli (changes requested):

hathor/builder.py
glevco: I have rolled back that change as I found a better way to do it and it's not necessary anymore.

hathor/builder.py
glevco: I have rolled back that change as I found a better way to do it and it's not necessary anymore.

tests/unittest.py
glevco: Done!

glevco:
> I think this won't work quite as expected. I'll talk to you to see a way around it. Anyway, I the change proposed on this PR shouldn't be merged.

@jansegre could you please re-review it? I have rolled back the `HathorSettings` change as I found a better way to do it and it's not necessary anymore.

msbrogli (approved):

hathor/builder.py
msbrogli: Ok!

tests/unittest.py
msbrogli: Thanks!

hathor/builder.py
msbrogli: Ok!

jansegre (approved):

PR #526, opened 2023-02-10
feat(mypy-daemon): implement mypy daemon commands

glevco:
Implement `make dmypy` and `make dcheck` commands, analogous to the existing `make mypy` and `make check`.

[Mypy daemon](https://mypy.readthedocs.io/en/stable/mypy_daemon.html) is a mypy server running in the background. Once started, each subsequent `dmypy` call is 10 or more times faster than the regular `mypy` command, according to the docs. In my setup, `dmypy` runs almost instantaneously while `mypy` can take several minutes sometimes, really slowing down the workflow when you're fighting with type hints.

I set a timeout of `86400` seconds (1 day) in the command, that is, the daemon will shut down itself after 1 day of inactivity. It can also be stopped manually before that with the `dmypy stop` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,10 @@ tests-full:
 mypy:
 	mypy -p hathor -p tests
 
+.PHONY: dmypy
+dmypy:
+	dmypy run --timeout 86400 -- -p hathor -p tests
+
 .PHONY: flake8
 flake8:
 	flake8 $(py_sources)
@@ -71,6 +75,9 @@ check-version:
 
 .PHONY: check
 check: check-version flake8 isort-check mypy
+
+.PHONY: check
+dcheck: check-version flake8 isort-check dmypy
 
 # formatting:
 
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ check-version:
 .PHONY: check
 check: check-version flake8 isort-check mypy
 
-.PHONY: check
+.PHONY: dcheck
 dcheck: check-version flake8 isort-check dmypy
 
 # formatting:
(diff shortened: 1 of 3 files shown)


hathor/conf/get_settings.py
```diff
@@ -15,10 +15,11 @@
 import importlib
 import os
 from types import ModuleType
+from typing import Optional
 
 from hathor.conf.settings import HathorSettings as Settings
 
-_config_file = None
+_config_file: Optional[str] = None
```
glevco, line 22: I had to add this hint because for some reason `dmypy` was complaining about it, even if `mypy` was not.

Makefile
```diff
@@ -72,6 +76,9 @@ check-version:
 .PHONY: check
 check: check-version flake8 isort-check mypy
 
+.PHONY: check
```
msbrogli, line 79: Typo: dcheck

msbrogli (approved):

Makefile
glevco: Fixed!

Makefile
```diff
@@ -72,6 +76,9 @@ check-version:
 .PHONY: check
 check: check-version flake8 isort-check mypy
 
+.PHONY: dcheck
+dcheck: check-version flake8 isort-check dmypy
```
jansegre, line 80: This will make `check-version`, `flake8` and `isort-check` run *once* and `dmypy` run continuously. I think that might be a bit unexpected and less useful, ideally `dcheck` would continuously run `flake8` and `isort-check` too (I'd drop `check-version` since it's only really useful on the CI).

Although running those on any change might not be so easy to do, it could need some tool that we don't have installed. So I'd be OK to not add this command for now.

What do you think?

jansegre: Oh, I got if after our call.

jansegre (approved):

PR #528, opened 2023-02-13
fix(indexes): deps index was not initialized correctly

jansegre:


jansegre:
Closed in favor of #538

PR #534, opened 2023-02-23
feat(load-events): handle full node loading events

glevco:


tests/event/test_event_reorg.py
```diff
@@ -70,9 +71,12 @@ def test_reorg_events(self):
         class unsorted(list):
             pass
         expected_events_grouped = [
+            [
+                (HathorEvents.LOAD_FINISHED, {})
+            ],
             # XXX: the order of the following events can vary depending on which genesis is spent/confirmed first
             unsorted([
-                (HathorEvents.VERTEX_METADATA_CHANGED, {'hash': settings.GENESIS_BLOCK_HASH.hex()}),
+                # (HathorEvents.VERTEX_METADATA_CHANGED, {'hash': settings.GENESIS_BLOCK_HASH.hex()}),
```
glevco, line 79: After disabling `full_verification` above, on the `create_peer()` line in the `setUp()`, I had to remove this event from the expected events for the test to pass. I'm not quite sure this is the actual expected behavior. @jansegre, could you please check?

hathor/manager.py
```diff
@@ -617,6 +620,10 @@ def _initialize_components_new(self) -> None:
         #       complex code
         self.tx_storage.indexes._manually_initialize(self.tx_storage)
 
+        # Publish events for all stored transactions
+        for tx in self.tx_storage.topological_iterator():
+            self.pubsub.publish(HathorEvents.NETWORK_NEW_TX_ACCEPTED, tx=tx)
+
```
glevco, line 626: @jansegre, we talked about emitting those `LOAD` phase events analogously to the index reconstruction (iterating over all txs in topological order). I didn't reuse the same iteration as the index reconstruction on `IndexesManager` because it felt misplaced, since this event publishing has nothing to do with indexes. Then I replicated the iteration here, on `HathorManager` scope, but I'm not sure that is a good idea.

Also, I'm not sure if this could create any unwanted side effects, since the `NETWORK_NEW_TX_ACCEPTED` event is subscribed in multiple parts of the code.

What do you think?

msbrogli: I don't feel it would have unwanted side effects, but it will have an impact on the loading time. Iterating over all transactions is a slow operation and take at least a few minutes.

Another remark: Why would we emit `NETWORK_NEW_TX_ACCEPTED` for all transactions? Shouldn't we filter out partially validated transactions? Why about voided transactions? Should we create a `LOADING_NEW_TX` event type?

hathor/event/event_manager.py
```diff
@@ -36,7 +36,6 @@
     HathorEvents.NETWORK_NEW_TX_ACCEPTED,
     HathorEvents.NETWORK_BEST_BLOCK_FOUND,
     HathorEvents.NETWORK_ORPHAN_BLOCK_FOUND,
-    HathorEvents.LOAD_STARTED,
```
msbrogli, line 39: Why won't we emit `LOAD_STARTED` as well?

tests/event/test_event_reorg.py
msbrogli: Why would we emit `VERTEX_METADATA_CHANGED` for the genesis block? Which metadata were changed?

msbrogli (changes requested):

tests/event/test_event_reorg.py
glevco: After discussing this with @jansegre, we concluded:

In the currently existing code (`dev` branch, before this PR) this event is emitted when `full_verification` is `True`, and is not emitted when it's `False`.

There are two possibilities:

1. If the genesis block's metadata is not being changed during a load (as expected), this event should not be emitted, and then removing this line is actually expected.
2. If the genesis block's metadata is being changed during the load, then we need to investigate further.

I'm testing these scenarios to better understand it. In both cases, it appears that `full_verification` is doing something that it shouldn't.

glevco: About `full_verification`, while talking to @jansegre I also raised the possibility of we actually removing this feature. Jan told me that using it may be even slower than just resetting a full node from a snapshot, and also he's currently unaware of any use cases needing it. I checked and it's still considered experimental (the CLI flag is `--x-full-verification`).

Also, we keep two similar versions of the same method in `HathorManager` just to support this, `_initialize_components()` and `_initialize_components_new()`. This is related to [this TODO](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/manager.py#L634-L636).

What do you think, @msbrogli? Is there anything that requires us to keep supporting this feature?

If we decide to remove it, I would create a separate PR, of course.

hathor/manager.py
glevco: > it will have an impact on the loading time. Iterating over all transactions is a slow operation and take at least a few minutes.

I'll change it so these events are only emitted if `--emit-load-events` is enabled. (This flag already prevents _processing_ load events, but I'll change it to actually prevent this loop since it's not necessary)

> Shouldn't we filter out partially validated transactions?

Partial transactions are not iterated when using the `topological_iterator()`. This was confirmed by @jansegre and should be checked in the Sync V2 related PRs.

> Why about voided transactions?

Those are expected to be emitted, and the related metadata will contain the information that this transaction is actually voided.

>  Should we create a `LOADING_NEW_TX` event type?

It's indeed our idea to change the name of the `NETWORK_NEW_TX_ACCEPTED` event. We thought about something like `NEW_TX_ACCEPTED`, and then adding an attribute to the event body specifying its `source`, that could be for example `NETWORK`, `LOAD`, `API`, etc. This would be done in a separate, future PR.

hathor/event/event_manager.py
glevco: I removed this event since it was supposed to be emitted in the next line of code where the `MANAGER_ON_START` event is emitted, and therefore it looked like an unnecessary duplication.

It could indeed be necessary if we want it to carry some information, but that is currently not foreseen in the design. About full verification, I discussed this with @jansegre and we are not supporting simultaneously enabling the events feature and full verification flags, since this would create problems and new edge cases to be handled during the load phase. This is enforced in this PR. Also, in that subject, please read https://github.com/HathorNetwork/hathor-core/pull/534#discussion_r1128974258.

hathor/manager.py
glevco: > I'll change it so these events are only emitted if --emit-load-events is enabled. (This flag already prevents processing load events, but I'll change it to actually prevent this loop since it's not necessary)

Done in https://github.com/HathorNetwork/hathor-core/pull/534/commits/9d2d55c6f4b961b2d39ff26197a76ea735cb5dee

tests/event/test_event_reorg.py
glevco: I've investigated it and found that on [this line](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/manager.py#L490) in the `_initialize_components()` method of `HathorManager`, the full verification calls the consensus' `update()` method, eventually calling `calculate_score()` also in the consensus.

There, the genesis block's score is updated (from `2.0` to `2.0`, that is, it's not changed), then it's saved and added to `txs_affected`. This causes the `VERTEX_METADATA_CHANGED` event to be emitted.

This event is not emitted when `full_verification` is `False`, becase `_initialize_components_new()` never calls the `update()` method in the consensus.

This is the alternative 1 described [above](https://github.com/HathorNetwork/hathor-core/pull/534#discussion_r1128967870), so it makes sense to just remove this line from the expected events to be emitted in this test.

msbrogli: What if we run a full node with an empty database? Would these events be emitted for the genesis block and transactions? I feel we should discuss it further on the design because it might affect integrations.

Regarding removing the `--x-full-verification`, some remarks:

1) The full verification is not experimental. It was the default choice but it is too slow on practice so we developed faster ways of initializing the full node.

2) How would one re-process the whole database to make sure the consensus is correct? When one uses a snapshot, they are relying on the snapshot.

hathor/manager.py
msbrogli: Just saw your changes on https://github.com/HathorNetwork/hathor-core/pull/534/commits/9d2d55c6f4b961b2d39ff26197a76ea735cb5dee, and it was confusing the use of it.

The way it is implemented, it seems new transactions have arrived during load time. I feel we should only publish events to the integration (and not internally into the full node). My understanding is that the use case would use this feature when they need to re-run all events for specific occasions (e.g., disaster recovery).

Sorry to bring that up now. Maybe you should remove this piece of code to get your PR merged ASAP, and discuss it further in the design or in a separate PR.

hathor/event/event_manager.py
msbrogli: Got it. Ok, we can remove it for now.

hathor/manager.py
msbrogli: I just finished reading all threads. This is the only point blocking me from approving the PR.

tests/event/test_event_reorg.py
glevco: > What if we run a full node with an empty database? Would these events be emitted for the genesis block and transactions? I feel we should discuss it further on the design because it might affect integrations.

I agree we should discuss this further, maybe it's a good idea to prepare a list of all possible scenarios, and the respective current behavior. Then we adapt it from there.

> 1. The full verification is not experimental. It was the default choice but it is too slow on practice so we developed faster ways of initializing the full node.

Ah, I see. Then maybe we could remove the `--x` prefix from the flag?

> 2. How would one re-process the whole database to make sure the consensus is correct? When one uses a snapshot, they are relying on the snapshot.

I had understood that running a full verification could be even slower than just re-downloading all transactions from the network. If that's not the case, then I agree it makes sense to keep the feature. But then I believe we should consider unifying the `_initialize_components()` and `_initialize_components_new()` methods.

hathor/manager.py
glevco: > The way it is implemented, it seems new transactions have arrived during load time.

Even if we change the event name?

> I feel we should only publish events to the integration (and not internally into the full node).

That would be a simple change, we would just need to call the `EventManager`'s handler directly instead of publishing the event to the `PubSub`.

> Sorry to bring that up now. Maybe you should remove this piece of code to get your PR merged ASAP, and discuss it further in the design or in a separate PR.

Okay, I'll do that!

tests/event/test_event_reorg.py
jansegre: > > 1\. The full verification is not experimental. It was the default choice but it is too slow on practice so we developed faster ways of initializing the full node.
> 
> Ah, I see. Then maybe we could remove the `--x` prefix from the flag?

I don't think we need to change anything related to the full verification on this PR other than making sure it interacts correctly with the handling of events.

> > 2\. How would one re-process the whole database to make sure the consensus is correct? When one uses a snapshot, they are relying on the snapshot.
> 
> I had understood that running a full verification could be even slower than just re-downloading all transactions from the network. If that's not the case, then I agree it makes sense to keep the feature. But then I believe we should consider unifying the `_initialize_components()` and `_initialize_components_new()` methods.

Currently full-verification works by: iterating over the whole database (using the DFS iterator, that can be really slow) and re-processing every transaction as if it was just received from the network. Syncing from scratch has the advantage of not having to run the DFS iterator, I haven't compared it recently, but a few months ago it was faster.

Today there is one alternative to the full verification, which is exporting all transactions to a file and importing them again. To export all transactions all we need is a working iterator, so it can work with a "corrupted" database (for certain kinds of corruption, like bad metadata and indexes data, we can force it to use the DFS iterator for exporting).

The export/import implementation is much more simple than the full-verification implementation, which makes it much less error prone and easier to maintain. So if we want that is a direction we can take in the future to remove the `--x-full-verification` (or maybe re-implementing it based off the export/import).

Another direction would be actually implementing a verification mechanism, which to me implies it is read-only (it wouldn't make any writes and if any verification fails it just won't start the node). That would possibly require a separate implementation of the consensus algorithm that only does verification, with no writes. It's not ideal to have separate implementations and this is a very critical code, so we never considered doing this. Maybe there's a way to have a single implementation that can be read-write/read-only as desired, but that could get complicated I'm not sure.

jansegre (approved):
For me it's approved considering that we discussed to leave this provisional and make tweaks later.

msbrogli (approved):

hathor/manager.py
msbrogli: Ok. Thanks!

tests/event/test_event_reorg.py
glevco: > I don't think we need to change anything related to the full verification on this PR other than making sure it interacts correctly with the handling of events.

Agreed, I didn't intend to make these changes on this PR. I have created an issue (https://github.com/HathorNetwork/hathor-core/issues/542) to represent the full verification discussion, so we can comeback to it in the future.

For now, the `--x-full-verification` flag won't be supported when enabling the events flag, as enforced by this PR. I'm resolving this thread, and pending discussions about handling events during node initialization will be continued in the design update. Discussion about full verification can be had in the issue mentioned above.

PR #538, opened 2023-03-06
refactor(indexes): move interest scope to inside the index

jansegre:
## Acceptance Criteria

- replace `hathor.indexes.manager._IndexFilter` with `hathor.indexes.base_index.Scope`, making the indexes themselves report the scope of transactions they're interested in
- do not alter any behavior other than exposing the deps-indexes to partial transactions
- reduce rate of `tests/tx/test_stratum.py` failing because of running in parallel and listening on the same port


hathor/indexes/base_index.py
```diff
@@ -13,12 +13,82 @@
...
+        tx_meta = tx.get_metadata()
+        if tx_meta.voided_by and not self.include_voided:
+            return False
+        if not tx_meta.validation.is_fully_connected() and not self.include_partial:
+            return False
+        # XXX: self.topologial_order doesn't affect self.match()
+        # passed all checks
+        return True
+
+    def get_iterator(self, tx_storage: 'TransactionStorage') -> Iterator[BaseTransaction]:
```
msbrogli, line 70: Missing docstring.

tests/tx/test_stratum.py
```diff
@@ -38,7 +38,8 @@ def setUp(self):
         super().setUp()
         self.manager = self.create_peer('testnet')
         self.manager.allow_mining_without_peers()
-        self.factory = StratumFactory(self.manager, port=8123, reactor=MemoryReactorHeapClock())
+        port = self.rng.randint(8000, 9000)
```
msbrogli, line 41: How is this change related to this PR? Is it necessary to be able to run tests simultaneously?

jansegre: Exactly, it's completely unrelated to this PR, but since we've enabled parallel tests this has been happening a lot locally, and _can_ happen on the CI it's just less probable (because it has n=2, but locally it's like n=10).

I didn't think it was worth to open a separate PR, but I could do it.

msbrogli: Got it. We can do it here. Just add a note about it in the Acceptance Criteria, please.

hathor/indexes/base_index.py
```diff
@@ -13,12 +13,82 @@
...
+                yield tx
+        if self.include_partial and not iterator_covers_partial:
+            # if partial transactions are needed and were not already covered, we use get_all_transactions, which
+            # includes partial transactions, to yield them, skipping all that aren't partial
+            for tx in tx_storage.get_all_transactions():
+                tx_meta = tx.get_metadata()
+                if tx_meta.validation.is_fully_connected():
+                    continue
+                if self.matches(tx):
+                    yield tx
```
glevco, line 91: If `self.topological_order = True` implies requiring topological order, is it correct to yield all partial transactions after the ones from the `topological_iterator`? In other words, are partial transactions always considered last in the topological order?

hathor/transaction/storage/rocksdb_storage.py
```diff
@@ -29,6 +30,7 @@
     from hathor.transaction import BaseTransaction, TransactionMetadata
 
 logger = get_logger()
+settings = HathorSettings()
```
glevco, line 33: Is this used?

hathor/indexes/tips_index.py
```diff
@@ -38,6 +56,20 @@ class TipsIndex(BaseIndex):
...
+    def __init__(self, *, txs: bool = False, blocks: bool = False, all: bool = False):
+        if sum([txs, blocks, all]) != 1:
+            raise TypeError('Exactly one of these: txs, blocks, all, must be set to True')
+        self._scope: Scope
+        if txs:
+            self._scope = SCOPE_TXS
+        elif blocks:
+            self._scope = SCOPE_BLOCKS
+        elif all:
+            self._scope = SCOPE_ALL
```
glevco, line 68: It looks like you could benefit from using an Enum here, since only one option can be chosen:

```python
    class ScopeType(Enum):
        ALL = Scope(
            include_blocks=True,
            include_txs=True,
            include_voided=True,
        )
        
        TXS = Scope(
            include_blocks=False,
            include_txs=True,
            include_voided=False,
        )
        
        BLOCKS = Scope(
            include_blocks=True,
            include_txs=False,
            include_voided=True,
        )
```

And then:

```python
    def __init__(self, *, scope: ScopeType):
        self._scope: Scope = scope.value
```

That would also simplify some subclasses, since `name` just becomes `name = scope.name.lower()`

hathor/indexes/timestamp_index.py
```diff
@@ -32,6 +50,20 @@ class TimestampIndex(BaseIndex):
...
+    def __init__(self, *, txs: bool = False, blocks: bool = False, all: bool = False):
+        if sum([txs, blocks, all]) != 1:
+            raise TypeError('Exactly one of these: txs, blocks, all, must be set to True')
+        self._scope: Scope
+        if txs:
+            self._scope = SCOPE_TXS
+        elif blocks:
+            self._scope = SCOPE_BLOCKS
+        elif all:
+            self._scope = SCOPE_ALL
```
glevco, line 62: Same as in `hathor/indexes/tips_index.py`

tests/tx/test_stratum.py
glevco: Another option is changing the `--dist` flag on `pytest-xdist`: https://pytest-xdist.readthedocs.io/en/stable/distribution.html

By doing that we can make all tests in one file, or module, or with the same `@mark`, run in the same worker. That would prevent tests using the same port to run in parallel.

hathor/indexes/manager.py
```diff
@@ -87,29 +82,24 @@ def __init_checks__(self):
...
+            yield self.mempool_tips
         if self.addresses is not None:
-            yield _IndexFilter.ALL, self.addresses
+            yield self.addresses
         if self.tokens is not None:
-            yield _IndexFilter.ALL, self.tokens
+            yield self.tokens
         if self.utxo is not None:
-            yield _IndexFilter.ALL, self.utxo
+            yield self.utxo
```
glevco, line 102: Maybe could be simplified to something like:

```python
    def iter_all_indexes(self) -> Iterator[BaseIndex]:
        indexes = [
            self.info,
            self.all_tips,
            self.block_tips,
            self.tx_tips,
            self.sorted_all,
            self.sorted_blocks,
            self.sorted_txs,
            self.height,
            self.deps,
            self.mempool_tips,
            self.addresses,
            self.tokens,
            self.utxo
        ]
            
        return (index for index in indexes if index is not None)
```

hathor/indexes/manager.py
```diff
@@ -184,27 +173,20 @@ def _manually_initialize(self, tx_storage: 'TransactionStorage') -> None:
         for index in self.iter_all_indexes():
             index.init_start(self)
 
-        self.log.debug('indexes init')
         if indexes_to_init:
-            tx_iter = progress(tx_storage.topological_iterator(), log=self.log, total=tx_storage.get_vertices_count())
+            overall_scope = reduce(operator.__or__, map(lambda i: i.get_scope(), indexes_to_init))
```
glevco, line 177: I think it's a bit more readable if you don't use that extra `map`, and also by doing this you get the advantage of using the `__or__` operator that was implemented:

```suggestion
            overall_scope = reduce(operator.__or__, map(lambda i: i.get_scope(), indexes_to_init))
            overall_scope = reduce(lambda i1, i2: i1.get_scope() or i2.get_scope(), indexes_to_init)
```

What do you think?

glevco (changes requested):

tests/tx/test_stratum.py
jansegre: > Another option is changing the `--dist` flag on `pytest-xdist`: https://pytest-xdist.readthedocs.io/en/stable/distribution.html
> 
> By doing that we can make all tests in one file, or module, or with the same `@mark`, run in the same worker. That would prevent tests using the same port to run in parallel.

That would be good, I'll just add a note though. I haven't _really_ fixed the problem, just made it much less likely to happen (now a collision needs for both two workers run the test at the same time and get the same port from randint). The `xdist_group` is nice, but it might need a bit of consideration, I think it's still useful for those tests to run in parallel.

hathor/indexes/tips_index.py
jansegre: Oh, that's a good point. It adds a symbol to be imported (`ScopeType`), which I'll make a little longer because I don't want to tie the specific values between different indexes (timestamp and tips indexes only incidentally have the same values, we might refactor one of them in a way that the scope changes). So I think I'll make it `TipsScopeType` and `TimestampScopeType`, or at least import them as that.

hathor/indexes/manager.py
jansegre: Yeah, that'd be better, I'll refactor it to `return filter(None, [self.info, self.all_tips, ...])`.

hathor/indexes/tips_index.py
glevco: Sounds good!

hathor/indexes/manager.py
jansegre: Although it would be more readable, it doesn't quite work, `lambda i1, i2: i1.get_scope() or i2.get_scope()` takes two indexes and returns one scope, but reduce works by chaining the result of the previous call as first argument of the next call, so `s0 = fun(i[0], i[1])` and then `s1 = fun(s0, i[2])` would break it.

The `or` operator is still being used `operator.__or__` is just the function version to use with functional stuff, so `a or b` is equivalent to `operator.__or__(a, b)`. But, yeah, it's less readable.

hathor/transaction/storage/rocksdb_storage.py
jansegre: You're right, it's not, I'll remove it.

hathor/indexes/manager.py
glevco: That's true! Sorry, my mistake. Then maybe:
```suggestion
            all_scopes = map(BaseIndex.get_scope, indexes_to_init)
            overall_scope = reduce(operator.__or__, all_scopes)
```

hathor/indexes/base_index.py
jansegre: Exactly, partial transactions will always come after all the other transactions that were yielded in topological order first, mainly because it's the only straightforward way to implement it (otherwise we'd need to index them, and having indexes depend on each other adds complexity that we don't need to add right now, besides the deps-index already handles knowing all partially validated transactions). There's no other option, since there can't be a partially validated transaction between two fully validated transactions (or rather, there _could_ be, but there must always be a topological order where there isn't one, since fully validated transactions will always point to other fully validated transactions). Also partially validated transactions don't _really_ need to be sorted, specially because it's not guaranteed that we'll be able to sort them.

hathor/indexes/base_index.py
```diff
@@ -13,12 +13,82 @@
...
+from typing import TYPE_CHECKING, Iterator, NamedTuple, Optional
 
 from hathor.transaction.base_transaction import BaseTransaction
 
 if TYPE_CHECKING:  # pragma: no cover
     from hathor.indexes.manager import IndexesManager
+    from hathor.transaction.storage import TransactionStorage
+
+
+class Scope(NamedTuple):
```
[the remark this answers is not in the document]
jansegre: Makes sense.

tests/tx/test_stratum.py
jansegre: Also, I suspect that it might be the case for other tests too, any test that listens on a fixed port, or writes to a fixed path is subject to this (I think I got this problem with some resource tests). So we could fix all of those in one go consistently. This was the only one that was happening way too much here.

hathor/indexes/manager.py

hathor/indexes/timestamp_index.py
jansegre: Done.

hathor/indexes/tips_index.py
jansegre: Done.

hathor/transaction/storage/rocksdb_storage.py
jansegre: Done.

hathor/indexes/base_index.py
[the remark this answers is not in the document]
jansegre: Done.

hathor/indexes/base_index.py
jansegre: Done.

hathor/indexes/base_index.py
glevco: Got it. Thanks!

glevco (approved):

PR #544, opened 2023-03-21
Update static-typing related packages

glevco:
Acceptance Criteria:

- Update `mypy`, `mypy-zope`, and `pydantic` dependencies
- Include pydantic's `init_typed` config to improve type checking robustness
- Fix in `base_transaction.py` (alerted by the new mypy version)
- Fix in `factory.py` (alerted by the `init_typed` config)

### DIFF
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -1185,7 +1185,7 @@ class TxOutput:
 
     def __str__(self) -> str:
         cls_name = type(self).__name__
-        value_str = hex(self.value) if self.is_token_authority else str(self.value)
+        value_str = hex(self.value) if self.is_token_authority() else str(self.value)
         if self.token_data:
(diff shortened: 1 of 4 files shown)


hathor/transaction/base_transaction.py
```diff
@@ -1185,7 +1185,7 @@ def __repr__(self) -> str:
 
     def __str__(self) -> str:
         cls_name = type(self).__name__
-        value_str = hex(self.value) if self.is_token_authority else str(self.value)
+        value_str = hex(self.value) if self.is_token_authority() else str(self.value)
```
msbrogli, line 1188: Was it a bug?

glevco: Yes, it looks like it. Previous versions of mypy didn't catch it. I guess it wasn't a problem because this is the `__str__` method, so probably only used for logging

msbrogli: Got it. Thanks!

PR #546, opened 2023-03-28
feat: allow version overwriting by creating a BUILD_VERSION file

luislhl:
### Acceptance Criteria

1. We should be able to overwrite the version reported by hathor-core by creating a BUILD_VERSION file with the new version.
2. We should allow an optional arg in our `check_version.sh` script containing an additional version against which we should check

These mechanisms will be used mainly by CI/CD to inject the built version into Docker images and validate it.

### DIFF
--- a/extras/check_version.sh
+++ b/extras/check_version.sh
@@ -3,7 +3,6 @@
 OPENAPI_FILE="hathor/cli/openapi_files/openapi_base.json"
 SRC_FILE="hathor/version.py"
 PACKAGE_FILE="pyproject.toml"
-BUILD_VERSION_FILE="BUILD_VERSION"
 
 OPENAPI_VERSION=`grep "version\":" ${OPENAPI_FILE} | cut -d'"' -f4`
 SRC_VERSION=`grep "BASE_VERSION =" ${SRC_FILE} | cut -d "'" -f2`
@@ -26,14 +25,12 @@ if [[ x${PACKAGE_VERSION}x != x${OPENAPI_VERSION}x ]]; then
 	EXITCODE=-1
 fi
 
-if [[ -f "$BUILD_VERSION_FILE" ]]; then
-	# Get the build version and ignore the suffix
-    BUILD_VERSION=$(cat "$BUILD_VERSION_FILE" | cut -d"-" -f1)
-
-	if [[ x${PACKAGE_VERSION}x != x${BUILD_VERSION}x ]]; then
-		echo "Version different in ${PACKAGE_FILE} and ${BUILD_VERSION_FILE}"
-		EXITCODE=-1
-	fi
+# We expect an optional argument containing a version string to be checked against the others
+if [[ $# -eq 1 ]]; then
+    if [[ x${PACKAGE_VERSION}x != x$1x ]]; then
+        echo "Version different in ${PACKAGE_FILE} and passed argument"
+        EXITCODE=-1
+    fi
 fi
 
 exit $EXITCODE
--- a/extras/check_version.sh
+++ b/extras/check_version.sh
@@ -4,8 +4,11 @@
 # This script will check all source files containing the project version and exit with an error code -1 in case
 # they don't match.
 #
-# An optional argument is also accepted containing a version string to be checked agaisnt the others.
-# E.g: ./extras/check_version.sh 0.52.1
+# usage: ./extras/check_version.sh [version]
+#
+# example: ./extras/check_version.sh 0.52.1
+#
+# When a version is provided, it is checked against the package version.
 ###
 
 OPENAPI_FILE="hathor/cli/openapi_files/openapi_base.json"
--- a/extras/check_version.sh
+++ b/extras/check_version.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+###
+# This script will check all source files containing the project version and exit with an error code -1 in case
+# they don't match.
+#
+# An optional argument is also accepted containing a version string to be checked agaisnt the others.
+# E.g: ./extras/check_version.sh 0.52.1
+###
+
 OPENAPI_FILE="hathor/cli/openapi_files/openapi_base.json"
 SRC_FILE="hathor/version.py"
 PACKAGE_FILE="pyproject.toml"
--- a/extras/check_version.sh
+++ b/extras/check_version.sh
@@ -3,9 +3,10 @@
 OPENAPI_FILE="hathor/cli/openapi_files/openapi_base.json"
 SRC_FILE="hathor/version.py"
 PACKAGE_FILE="pyproject.toml"
+BUILD_VERSION_FILE="BUILD_VERSION"
 
 OPENAPI_VERSION=`grep "version\":" ${OPENAPI_FILE} | cut -d'"' -f4`
-SRC_VERSION=`grep "__version__" ${SRC_FILE} | cut -d "'" -f2`
+SRC_VERSION=`grep "BASE_VERSION =" ${SRC_FILE} | cut -d "'" -f2`
 PACKAGE_VERSION=`grep '^version' ${PACKAGE_FILE} | cut -d '"' -f2`
 
 # For debugging:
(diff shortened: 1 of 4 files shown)


extras/check_version.sh
```diff
@@ -25,4 +26,14 @@ if [[ x${PACKAGE_VERSION}x != x${OPENAPI_VERSION}x ]]; then
 	EXITCODE=-1
 fi
 
+if [[ -f "$BUILD_VERSION_FILE" ]]; then
+	# Get the build version and ignore the suffix
+    BUILD_VERSION=$(cat "$BUILD_VERSION_FILE" | cut -d"-" -f1)
+
+	if [[ x${PACKAGE_VERSION}x != x${BUILD_VERSION}x ]]; then
+		echo "Version different in ${PACKAGE_FILE} and ${BUILD_VERSION_FILE}"
```
luislhl, line 34: The BUILD_VERSION shoudl'nt always be equal to the other versions. This may fail sometimes.

hathor/version.py
```diff
@@ -12,4 +12,27 @@
...
-__version__ = '0.52.1'
+import os
+
+BASE_VERSION = '0.52.1'
+
+DEFAULT_VERSION_SUFFIX = "-local"
+BUILD_VERSION_FILE_PATH = "./BUILD_VERSION"
+
+
+def _get_build_version():
```
jansegre, line 23: missing type annotation `-> Optional[str]`

hathor/version.py
```diff
@@ -12,4 +12,27 @@
...
+    if not os.path.isfile(BUILD_VERSION_FILE_PATH):
+        return None
+
+    with open(BUILD_VERSION_FILE_PATH, 'r') as f:
+        build_version = f.readline()
+
+        return build_version
+
+
+def _get_version():
```
jansegre, line 33: missing type annotation `-> str`

jansegre (changes requested):

hathor/version.py
luislhl: Humm, shouldn't we make mypy error out for missing types?

luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/546/commits/4f30e623a4080384d19f9979ed838bef19432602

hathor/version.py
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/546/commits/4f30e623a4080384d19f9979ed838bef19432602

hathor/version.py
```diff
@@ -12,4 +12,27 @@
...
+DEFAULT_VERSION_SUFFIX = "-local"
+BUILD_VERSION_FILE_PATH = "./BUILD_VERSION"
+
+
+def _get_build_version():
+    if not os.path.isfile(BUILD_VERSION_FILE_PATH):
+        return None
+
+    with open(BUILD_VERSION_FILE_PATH, 'r') as f:
+        build_version = f.readline()
```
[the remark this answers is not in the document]
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/546/commits/4f30e623a4080384d19f9979ed838bef19432602

extras/check_version.sh
```diff
@@ -25,4 +25,12 @@ if [[ x${PACKAGE_VERSION}x != x${OPENAPI_VERSION}x ]]; then
 	EXITCODE=-1
 fi
 
+# We expect an optional argument containing a version string to be checked against the others
```
msbrogli, line 28: Is it optional? I feel we should add some piece of documentation at the beginning of this file.

msbrogli (changes requested):

hathor/version.py
```diff
@@ -12,4 +12,50 @@
...
+            logger.warn("A build version with an invalid format was found. Ignoring it.", build_version=build_version)
+            return None
+
+
+def _get_version() -> str:
+    """Get the current hathor-core version from the build version or the default one with a local suffix
+
+    :return: The current hathor-core version
+    """
+    local_version = BASE_VERSION + DEFAULT_VERSION_SUFFIX
```
luislhl, line 57: We need to incluide the commit hash here as well, as accorded in the thread https://github.com/HathorNetwork/ops-tools/pull/627#discussion_r1157673133

tests/resources/test_version.py
```diff
@@ -1,10 +1,17 @@
...
+
 from twisted.internet.defer import inlineCallbacks
 
 import hathor
+from hathor.version import BASE_VERSION, DEFAULT_VERSION_SUFFIX, _get_version
 from hathor.version_resource import VersionResource
 from tests import unittest
 from tests.resources.base_resource import StubSite, _BaseResourceTest
 
+TMP_DIR = tempfile.gettempdir()
```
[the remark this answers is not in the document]
luislhl: This is different from `tempfile.mkdtemp()` in that it doesn't create a directory, it only returns the default os temp dir, so we shouldn't delete it.

But I could move this to the `setUp()` method, or use `tempfile.mkdtemp()` instead and delete the folder in the end.

extras/check_version.sh
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/546/commits/966ac4ebb52dae171ab5ff1cc80ffdf650821554

extras/check_version.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/bash
 
+###
+# This script will check all source files containing the project version and exit with an error code -1 in case
+# they don't match.
+#
+# An optional argument is also accepted containing a version string to be checked agaisnt the others.
```
msbrogli, line 7: Typo: ... against.

extras/check_version.sh
msbrogli: Thanks!

hathor/version.py
```diff
@@ -12,4 +12,70 @@
...
+            "Will report local version without any git info."
+        ))
+        return None
+
+
+def _get_local_version() -> str:
+    git_head = _get_git_revision_short_hash()
+
+    if git_head:
+        return f"{BASE_VERSION}-{_get_git_revision_short_hash()}-{DEFAULT_VERSION_SUFFIX}"
```
msbrogli, line 68: Why don't you use `{git_head}`?

tests/resources/test_version.py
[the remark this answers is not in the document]
msbrogli: My point that the test will leave an orphan file in the tempdir if the test fails before the clean up is executed. Moving the clean up code to the `tearDown()` method avoids it.

One downside of using the system temp dir is that there might be a race condition if tests are executed in parallel and two or more use the same file name there. Creating a separate directory for each test fixes this race condition.

msbrogli (changes requested):

hathor/version.py
luislhl: You're right, my bad

tests/resources/test_version.py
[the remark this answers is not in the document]
luislhl: Got it! Fixed in https://github.com/HathorNetwork/hathor-core/pull/546/commits/9a5aa2ed108746f4cd0ed8755c12971e1e8d8ef5

extras/check_version.sh
```diff
@@ -1,11 +1,19 @@
...
+# E.g: ./extras/check_version.sh 0.52.1
```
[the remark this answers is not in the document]
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/546/commits/41d947719f42e32e6bb67b42ceb0b87e3608c732

extras/check_version.sh
luislhl: Removed in https://github.com/HathorNetwork/hathor-core/commit/41d947719f42e32e6bb67b42ceb0b87e3608c732

hathor/version.py
luislhl: Fixed in https://github.com/HathorNetwork/hathor-core/pull/546/commits/9a5aa2ed108746f4cd0ed8755c12971e1e8d8ef5

tests/resources/test_version.py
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

msbrogli (approved):

extras/check_version.sh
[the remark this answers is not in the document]
msbrogli: Thanks!

extras/check_version.sh
msbrogli: Thanks!

jansegre (approved):

PR #547, opened 2023-03-29
ci: adapt to the new release process

luislhl:
### Acceptance Criteria
- Disable caching of Docker layers when building `nightly` builds
- Write the BUILD_VERSION file with the version that was just built. This should be done only for `nigthly`, `release-candidate` and `release` builds.
- For the `release-candidate`, we should only build a Docker image for the default Python version (currently 3.9), and it should have only the main tag on it, not the tags for specific python versions/implementations)
- We should send messages to Slack for builds of `release-candidate` and `release` images, but only for the default python version in our version matrix. Otherwise we would be sending many alerts at once.
- Fix a previous error in the workflow where we were apparently making comparisons like `if: steps.prep.outputs.login-dockerhub`, but we were assign the value `'false'` to these variables, which is a truthy value.
- We should have a guide about how to use https://github.com/nektos/act to test our Github Actions workflow

### Notes

I'll be proposing a refactoring on this in a new PR (https://github.com/HathorNetwork/hathor-core/pull/548) to extract these Python script into actual Python files, because they are becoming too hard to maintain.

Then we will just call the scripts from the workflow and we will be able to have unit tests for them.

### TODO
- [ ] Create the Slack secrets in the repo

### DIFF
--- a/.github/workflows/TESTING.md
+++ b/.github/workflows/TESTING.md
@@ -1,0 +1,39 @@
+# Testing Github Action workflows
+
+It's possible to test the Github Action workflows locally by using https://github.com/nektos/act
+
+You can start by listing all available jobs in our workflows:
+
+```bash
+act -l
+```
+
+We have prepared examples on how to test some types of triggers in the `docker.yml` workflow, but it shouldn't be hard to adapt these examples to test other combinations of jobs and triggers.
+
+## Testing a Tag Push
+
+To simulate the workflow being trigger by the push of a tag, first generate an event file like this:
+
+```bash
+cat <<EOF > event.json
+{
+  "ref": "refs/tags/v0.53.0-rc.1"
+}
+EOF
+```
+
+You can change the tag in this event to simulate different types of tags.
+
+Then, run the `buildx` job with a `push` event providing the event context and a secret called `DOCKERHUB_IMAGE`:
+
+```bash
+act push -e event.json -j buildx -s DOCKERHUB_IMAGE=testing_locally
+```
+
+## Testing a Scheduled run
+
+Simulating a scheduled run is similar, just change the type of event:
+
+```bash
+act schedule -j buildx -s DOCKERHUB_IMAGE=testing_locally
+```
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -36,26 +36,44 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
-    - name: Prepare tags
-      id: prep
+    - name: Prepare base version
+      id: prep_base_version
       shell: python
       run: |
-        import datetime
         import re
-        def extract_pyver(filename):
-            for line in open(filename).readlines():
-                if line.startswith('ARG PYTHON'):
-                    return line.split('=')[1].strip()
+        import os
+
         ref = '${{ github.ref }}'
-        dockerfile_cpython = 'Dockerfile'
-        dockerfile_pypy = 'Dockerfile.pypy'
-        default_python = 'python' + extract_pyver(dockerfile_cpython)
-        default_pypy = 'pypy' + extract_pyver(dockerfile_pypy)
+
+        # Set base_version according to the github ref type
+        is_release_candidate = False
+        is_release = False
+        is_nightly = False
+
+        overwrite_hathor_core_version = False
+
         if '${{ github.event_name }}' == 'schedule':
             commit_short_sha = '${{ github.sha }}'[:8]
             base_version = 'nightly-' + commit_short_sha
+            is_nightly = True
         elif ref.startswith('refs/tags/'):
-            base_version = ref[10:].split('-', 1)[0]
+            git_tag = ref[10:]
+            base_version = git_tag.split('-', 1)[0]
+
+            pre_release = (git_tag.split('-', 1)[1:] or [None])[0]
+            # This will be used to check against the versions in our source files
+            check_version = base_version[1:]
+            print('::set-output name=check-version::' + check_version)
+
+            # Check if this is a release-candidate
+            if pre_release:
+                if re.match(r'^rc\.[0-9]{1,3}$', pre_release):
+                    base_version = base_version + '-' + pre_release
+                    is_release_candidate = True
+                else:
+                    raise ValueError(f'Invalid Tag Value: {git_tag}')
+            else:
+                is_release = True
         elif ref.startswith('refs/heads/'):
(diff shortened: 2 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -64,25 +79,77 @@ jobs:
...
+        print('::set-output name=enable-docker-cache::' + ('true' if enable_docker_cache else 'false'))
+        print('::set-output name=overwrite-hathor-core-version::' + ('true' if overwrite_hathor_core_version else 'false'))
+    - name: Overwrite version
+      if: steps.prep_base_version.outputs.overwrite-hathor-core-version == 'true'
+      shell: python
+      run: |
+        base_version = '${{ steps.prep_base_version.outputs.base-version }}'
+
+        with open('BUILD_VERSION', 'w') as file:
+            file.write(base_version + '\n')
```
luislhl, line 94: We should remove the starting `v` character.

.github/workflows/docker.yml
```diff
@@ -36,26 +36,43 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
-    - name: Prepare tags
-      id: prep
+    - name: Prepare base version
+      id: prep_base_version
       shell: python
       run: |
```
jansegre, line 42: This script was already big, but it's huge now. Is there a simple way to have it as a file?

.github/workflows/docker.yml
```diff
@@ -170,3 +252,13 @@ jobs:
           org.opencontainers.image.licenses=${{ github.event.repository.license.spdx_id }}
         cache-from: type=local,src=/tmp/.buildx-cache
         cache-to: type=local,dest=/tmp/.buildx-cache
+    - name: Slack Notification
+      if: ${{ steps.prep.outputs.slack-notification-version && steps.prep_base_version.outputs.should-notify-slack == 'true' && job.status == 'success' }}
```
jansegre, line 256: On `main.yml` I used `if github_repository.lower() == 'hathornetwork/hathor-core':` to check test if this is the main repo or a fork. Should we do something similar to decide whether to enable this step? Otherwise if it's running on a fork this step will just fail.

.github/workflows/main.yml
```diff
@@ -7,8 +7,8 @@ on:
     tags:
     - v*
   pull_request:
-    branches:
-    - dev
+    branches-ignore:
+    - master
```
jansegre, line 11: The `master` will be the branch where we merge most PRs to, right? Is this change correct?

.github/workflows/docker.yml
luislhl: Yes, but I decided to do it in a different PR that is still WIP: https://github.com/HathorNetwork/hathor-core/pull/548

Otherwise I think it would become harder for the reviewers to know what changed exactly in the logic

.github/workflows/main.yml
luislhl: You're right, it's wrong.

But I was actually trying to make sure we run the tests for PRs that are based in feature-brances (like this one which is based in `feat/allow-version-override`), with the previous config the tests weren't running for these cases.

So I think we should just remove any branch restriction on the `pull_request` event, what do you think?

.github/workflows/docker.yml
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/547/commits/e018be00b4c0152945686c6de7890b54a56bf157 (This link will eventually break since I need to rebase the branch frequently, which is one of the reasons I would rather use merges inside of PRs)

.github/workflows/main.yml
jansegre: I think that makes sense. We can always add "[skip ci]" on the commit message if we don't want it to run.

.github/workflows/main.yml
```diff
@@ -7,8 +7,6 @@ on:
     tags:
     - v*
   pull_request:
-    branches:
-    - dev
```
msbrogli, line 11: Shouldn't we use `master` here?

.github/workflows/docker.yml
msbrogli, line 1: Why don't we move these scripts to separate files, just like you did [here](https://github.com/HathorNetwork/hathor-core/pull/548)?

.github/workflows/TESTING.md
```diff
@@ -0,0 +1,39 @@
+# Testing Github Action workflows
+
+It's possible to test the Github Action workflows locally by using https://github.com/nektos/act
+
+You can start by listing all available jobs in out workflows:
```
msbrogli, line 5: typo: ... in out workflows

msbrogli (changes requested):

.github/workflows/docker.yml
luislhl: If I had done this in this PR, I think it would get hard for you reviewers to know what changed exactly in the CI logic.

.github/workflows/main.yml
luislhl: If we use it here, it means only PRs based on the `master` branches will have the tests run.

I know that you will be suggesting some processual changes that could include a suggestion to stop stacking up PRs. but I would say that, although we can surely try to avoid stacking them up, we might not be able to in some situations, unless we accept that the PRs could get bigger in order to not be stacked up.

So, I would say the rules here should still support running tests for stacked PRs, even if our process at some point changes to try to prohibit this practice.

.github/workflows/docker.yml
msbrogli: We can split this change in two PRs. One refactor that moves the code to a separate file, and then your changes. What do you think?

.github/workflows/TESTING.md
msbrogli: Thanks!

.github/workflows/main.yml
msbrogli: Got it. Ok. We can let it run for all PRs, regardless of the destination branch.

.github/workflows/docker.yml
luislhl: No problem, I can do this if you think it's worth the time.

But I would say it's not, since we actually already have it split in two PRs. I just did things in a different order: first perform the CI changes in a first PR (this one), then move the script to separate files in https://github.com/HathorNetwork/hathor-core/pull/548

But I could surely create 2 new PRs to first move the script and then perform the changes.

msbrogli: Oh! Got it. My bad. Ok. Let's get this one merged then.

msbrogli (approved):

PR #549, opened 2023-04-03
feat: add best_block and best_block_tips to dag on status

alexruzenhack:
### Motivation

This addition will allow the hathor-p2p-monitor to monitor synchronization with sync-v2 peers in a simplified way.

Ref.: https://github.com/HathorNetwork/hathor-p2p-monitor/pull/69

### Acceptance criteria

- add `best_block` and `best_block_tips` to dag property of `/status`

### DIFF
--- a/hathor/p2p/resources/status.py
+++ b/hathor/p2p/resources/status.py
@@ -187,12 +187,12 @@ StatusResource.openapi = {
                                             'latest_timestamp': 1539271483,
                                             'best_block_tips': [
                                                 {
-                                                    'hash': '000006c',
+                                                    'hash': '000007eb968a6cdf0499e2d033faf1e163e0dc9cf41876acad4d421836972038',  # noqa: E501
                                                     'height': 0
                                                 }
                                             ],
                                             'best_block': {
-                                                'hash': '000006c',
+                                                'hash': '000007eb968a6cdf0499e2d033faf1e163e0dc9cf41876acad4d421836972038',  # noqa: E501
                                                 'height': 0
                                             }
                                         }
--- a/hathor/p2p/resources/status.py
+++ b/hathor/p2p/resources/status.py
@@ -81,6 +81,15 @@ class StatusResource(Resource):
             })
 
         app = 'Hathor v{}'.format(hathor.__version__)
+
+        best_block_tips = []
+        for tip in self.manager.tx_storage.get_best_block_tips():
+            tx = self.manager.tx_storage.get_transaction(tip)
+            meta = tx.get_metadata()
+            best_block_tips.append({'hash': tx.hash_hex, 'height': meta.height})
+
+        best_block = self.manager.tx_storage.get_best_block()
+
         data = {
             'server': {
                 'id': self.manager.connections.my_peer.id,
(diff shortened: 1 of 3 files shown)


tests/resources/p2p/test_status.py
```diff
@@ -22,11 +22,28 @@ def setUp(self):
...
         server_data = data.get('server')
         self.assertEqual(server_data['app_version'], 'Hathor v{}'.format(hathor.__version__))
         self.assertEqual(server_data['network'], 'testnet')
         self.assertGreater(server_data['uptime'], 0)
 
+        dag_data = data.get('dag')
+        # We have the genesis block
+        self.assertEqual(len(dag_data['best_block_tips']), 1)
+        self.assertIsNotNone(dag_data['best_block_tips'][0])
+        # As we don't have a type, we most check if the keys are there,
```
luislhl, line 35:
```suggestion
        # As we don't have a type, we must check if the keys are there,
```

tests/resources/p2p/test_status.py
```diff
@@ -22,11 +22,28 @@ def setUp(self):
...
+        # and the types are correct
+        self.assertIn('hash', dag_data['best_block_tips'][0])
+        self.assertIn('height', dag_data['best_block_tips'][0])
+        self.assertIsInstance(dag_data['best_block_tips'][0]['hash'], str)
+        self.assertIsInstance(dag_data['best_block_tips'][0]['height'], int)
+        self.assertIsNotNone(dag_data['best_block'])
+        self.assertIn('hash', dag_data['best_block'])
+        self.assertIn('height', dag_data['best_block'])
+        self.assertIsInstance(dag_data['best_block']['hash'], str)
+        self.assertIsInstance(dag_data['best_block']['height'], int)
```
luislhl, line 45: Maybe we should assert if the values returned are correct as well. We are only checking their types and their presence here, but not the values themselves.

alexruzenhack: Mmm... the `height` value is ok to test, it is always `0`. But the `hash` is dynamic and extracting its value would expose the internal process of the endpoint, which is not recommended for this type of test. Maybe testing the length of the `hash` could be an alternative.

luislhl: Since this is the Genesis block, the hash is not dynamic.

The genesis block hash used in the tests is defined in https://github.com/HathorNetwork/hathor-core/blob/master/hathor/conf/unittests.py#L28

hathor/p2p/resources/status.py
```diff
@@ -170,7 +184,17 @@ def render_GET(self, request):
...
-                                            'latest_timestamp': 1539271483
+                                            'latest_timestamp': 1539271483,
+                                            'best_block_tips': [
+                                                {
+                                                    'hash': '000006c',
+                                                    'height': 0
+                                                }
+                                            ],
+                                            'best_block': {
+                                                'hash': '000006c',
```
jansegre, line 195: I'd put the full hash of the genesis block, just to make it really clear how it works, that is, it is not the first few characters of the hash.

jansegre (approved):

hathor/p2p/resources/status.py
alexruzenhack: Done at https://github.com/HathorNetwork/hathor-core/pull/549/commits/fc94a87f00db8fc15384ea407db3b1dd384d70dc

PR #550, opened 2023-04-03
bugfix: fix initialization order bugs

glevco:
Acceptance Criteria:

- Fix bug that was preventing node initialization due to incorrect import order
  - The Event Simulator indirectly imported code that used `get_settings` before it was called by `run_node` with the correct configuration, therefore generating a `loading config twice with a different file` error.
- Fix bug that was preventing node initialization with the events feature due to incorrect start order
  - The `HathorManager` was sending events relevant to the `EventManager` before starting it.

### DIFF
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -54,6 +54,7 @@ class EventManager:
     """
 
     _peer_id: str
+    _is_running: bool = False
     _load_finished: bool = False
 
     @property
@@ -85,9 +86,11 @@ class EventManager:
     def start(self, peer_id: str) -> None:
         self._peer_id = peer_id
         self._event_ws_factory.start()
+        self._is_running = True
 
     def stop(self):
         self._event_ws_factory.stop()
+        self._is_running = False
 
     def _assert_closed_event_group(self):
         # XXX: we must check that the last event either does not belong to an event group or that it just closed an
@@ -110,6 +113,8 @@ class EventManager:
             self._pubsub.subscribe(event, self._handle_event)
 
     def _handle_event(self, event_type: HathorEvents, event_args: EventArguments) -> None:
+        assert self._is_running, 'Cannot handle event, EventManager is not started.'
+
         event_type = _EVENT_CONVERTERS.get(event_type, event_type)
         event_specific_handlers = {
             HathorEvents.LOAD_FINISHED: self._handle_load_finished
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -84,11 +84,15 @@ class EventManager:
         self._subscribe_events()
 
     def start(self, peer_id: str) -> None:
+        assert self._is_running is False, 'Cannot start, EventManager is already running'
+
         self._peer_id = peer_id
         self._event_ws_factory.start()
         self._is_running = True
 
     def stop(self):
+        assert self._is_running is True, 'Cannot stop, EventManager is not running'
+
         self._event_ws_factory.stop()
         self._is_running = False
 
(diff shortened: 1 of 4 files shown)


jansegre (approved):
The `cli` import fix is good as is, about the other one I wonder if we should make a test for it.

glevco:
> The `cli` import fix is good as is, about the other one I wonder if we should make a test for it.

@jansegre I was investigating this and I found out that existing tests in `test_event_manager.py` were supposed to have caught this. The code in `PubSub` for publishing events is:

```python
    def publish(self, key: HathorEvents, **kwargs: Any) -> None:
        reactor_thread = ReactorThread.get_current_thread(self.reactor)

        args = EventArguments(**kwargs)
        for fn in self._subscribers[key]:
            if reactor_thread == ReactorThread.NOT_RUNNING:
                fn(key, args)
            else:
                is_empty = bool(not self.queue)
                self.queue.append((fn, key, args))
                if is_empty:
                    self._schedule_call_next()
```

The reason the tests passed is because tests use the `Clock` class instead of the reactor, so incidentally things are called in the correct order: the `else` branch is called and the event handler function call is delayed to after the `EventManager` is started. In the real case, when running the full node, the `if reactor_thread == ReactorThread.NOT_RUNNING:` branch is called, and therefore the event handler function is called before the `EventManager` is started.

Also, `ReactorThread` is where the `MAIN_THREAD` value is returned for `Clock` (for testing):

```python
    def get_current_thread(cls, reactor: Reactor) -> 'ReactorThread':
        """ Returns if the code is being run on the reactor thread, if it's running already.
        """
        running = getattr(reactor, 'running', None)
        if running is not None:
            if running:
                return cls.MAIN_THREAD if isInIOThread() else cls.NOT_MAIN_THREAD
            else:
                # if reactor is not running yet, there's no threading
                return cls.NOT_RUNNING
        else:
            # on tests, we use Clock instead of a real Reactor, so there's
            # no threading. We consider that the reactor is running
            return cls.MAIN_THREAD
```

I feel that we should somehow either change `ReactorThread` or `PubSub` themselves, to prevent this from happening again in the future. Code logic is getting executed differently in production and in tests for this reason.

For example, changing the above last line from `return cls.MAIN_THREAD` to `return cls.NOT_RUNNING` would fix the problem, but I'm not sure of other unintended consequences. What do you think?

jansegre:
> What do you think?

I think we should open an issue and investigate this separately and independently from this fix. I'd say this PR can me merged as is.

glevco:
> I think we should open an issue and investigate this separately and independently from this fix

I have created an issue: https://github.com/HathorNetwork/hathor-core/issues/552

hathor/event/event_manager.py
```diff
@@ -85,9 +86,11 @@ def __init__(
     def start(self, peer_id: str) -> None:
         self._peer_id = peer_id
         self._event_ws_factory.start()
+        self._is_running = True
```
msbrogli, line 89: Should we assert `self._is_running is False`?

hathor/event/event_manager.py
```diff
@@ -85,9 +86,11 @@ def __init__(
...
 
     def stop(self):
         self._event_ws_factory.stop()
+        self._is_running = False
```
msbrogli, line 93: Should we assert `self._is_running is True`?

msbrogli (approved):

hathor/event/event_manager.py
glevco: It wouldn't be a problem since the operation is idempotent, but I've added it to prevent human errors.

hathor/event/event_manager.py
glevco: It wouldn't be a problem since the operation is idempotent, but I've added it to prevent human errors.

hathor/event/event_manager.py
msbrogli: Got it. Thanks!

PR #555, opened 2023-04-13
feat(events): realiable integration during the load phase

glevco:
Implement changes described in https://github.com/HathorNetwork/rfcs/pull/52

Acceptance Criteria:

- Implement new `EventStorage` functionality
  - Storage and retrieval of `NodeState`
  - Storage and retrieval of event queue feature state
  - Clearing events
- Update `HathorManager`
  - Remove `event_storage` arg, make `event_manager` required, and add `enable_event_queue`
  - Initialization fails if it starts without the event queue feature, and it was enabled in a previous start
  - Handling of LOAD phase events
- Update `EventManager` to support handling of LOAD phase events
- Remove `--x-emit-load-events` CLI flag, as it's not necessary anymore in the way events are handled in the LOAD phase
- New `bytes_to_int()` function in `transaction.util`

### DIFF
--- a/hathor/builder/cli_builder.py
+++ b/hathor/builder/cli_builder.py
@@ -143,12 +143,8 @@ class CliBuilder:
                 event_storage=event_storage,
                 event_ws_factory=self.event_ws_factory,
                 pubsub=pubsub,
-                reactor=reactor,
-                emit_load_events=args.x_emit_load_events
+                reactor=reactor
             )
-        else:
-            self.check_or_raise(not args.x_emit_load_events, '--x-emit-load-events cannot be used without '
-                                                             '--x-enable-event-queue')
 
         if args.wallet_index and tx_storage.indexes is not None:
             self.log.debug('enable wallet indexes')
@@ -232,7 +228,7 @@ class CliBuilder:
 
             self.manager.enable_event_queue = True
             self.log.info('--x-enable-event-queue flag provided. '
-                          'The events detected by the full node will be stored and retrieved to clients')
+                          'The events detected by the full node will be stored and can be retrieved by clients')
 
         for description in args.listen:
             self.manager.add_listen_address(description)
--- a/hathor/builder/cli_builder.py
+++ b/hathor/builder/cli_builder.py
@@ -136,15 +136,15 @@ class CliBuilder:
 
         pubsub = PubSubManager(reactor)
 
-        event_manager: Optional[EventManager] = None
         if args.x_enable_event_queue:
             self.event_ws_factory = EventWebsocketFactory(reactor, event_storage)
-            event_manager = EventManager(
-                event_storage=event_storage,
-                event_ws_factory=self.event_ws_factory,
-                pubsub=pubsub,
-                reactor=reactor
-            )
+
+        event_manager = EventManager(
+            event_storage=event_storage,
+            event_ws_factory=self.event_ws_factory,
+            pubsub=pubsub,
+            reactor=reactor
+        )
 
         if args.wallet_index and tx_storage.indexes is not None:
             self.log.debug('enable wallet indexes')
@@ -170,7 +170,6 @@ class CliBuilder:
             network=network,
             hostname=hostname,
             tx_storage=tx_storage,
-            event_storage=event_storage,
             event_manager=event_manager,
             wallet=self.wallet,
             stratum_port=args.stratum,
@@ -181,7 +180,8 @@ class CliBuilder:
             enable_sync_v2=enable_sync_v2,
             consensus_algorithm=consensus_algorithm,
             environment_info=get_environment_info(args=str(args), peer_id=peer_id.id),
-            full_verification=full_verification
+            full_verification=full_verification,
+            enable_event_queue=bool(args.x_enable_event_queue)
         )
 
         if args.data:
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -214,7 +214,7 @@ class EventManager:
     def _should_reload_events(self) -> bool:
         return self._previous_node_state in [None, NodeState.LOAD]
 
-    def handle_load_phase_events(self, topological_iterator: Iterator[BaseTransaction]) -> None:
+    def handle_load_phase_vertices(self, topological_iterator: Iterator[BaseTransaction]) -> None:
         """
         Either generates load phase events or not, depending on previous node state.
         Does so asynchronously so events generated here are not processed before normal event handling.
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -12,15 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Callable, Optional
+from typing import Callable, Iterator, Optional
 
 from structlog import get_logger
 
 from hathor.event.model.base_event import BaseEvent
 from hathor.event.model.event_type import EventType
+from hathor.event.model.node_state import NodeState
 from hathor.event.storage import EventStorage
 from hathor.event.websocket import EventWebsocketFactory
 from hathor.pubsub import EventArguments, HathorEvents, PubSubManager
+from hathor.transaction import BaseTransaction
 from hathor.util import Reactor
 
 logger = get_logger()
@@ -51,7 +53,7 @@ class EventManager:
 
     _peer_id: str
     _is_running: bool = False
-    _load_finished: bool = False
+    _previous_node_state: Optional[NodeState] = None
 
     @property
     def event_storage(self) -> EventStorage:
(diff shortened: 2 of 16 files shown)


hathor/event/model/node_state.py
```diff
@@ -0,0 +1,20 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from enum import Enum, auto
+
+
+class NodeState(Enum):
+    LOAD = auto()
```
msbrogli, line 19: Maybe avoid using `auto()` since the value might change if we add intermediate states in the future.

hathor/transaction/util.py
```diff
@@ -28,6 +28,10 @@ def int_to_bytes(number: int, size: int, signed: bool = False) -> bytes:
     return number.to_bytes(size, byteorder='big', signed=signed)
 
 
+def bytes_to_int(data: bytes, signed: bool = False) -> int:
```
msbrogli, line 31: Missing docstring.

msbrogli, line 31: Maybe change it to `bytes_to_int(data: bytes, *, signed: bool = False)` to force signed to be a named argument.

hathor/event/event_manager.py
```diff
@@ -193,8 +199,35 @@ def _create_event(
...
+        """
+        Either generates load phase events or not, depending on previous node state.
+        Does so asynchronously so events generated here are not processed before normal event handling.
+        """
+        assert self._is_running, 'Cannot handle load phase events, EventManager is not started.'
+
+        if not self._should_reload_events():
+            return
+
+        for vertex in topological_iterator:
```
msbrogli, line 227: I'm still uncomfortable with iterating over the whole database, as the full node already has to do it to start itself. My discomfort happens because it is a very slow operation for large blockchains. Couldn't the iterator on HathorManager call this method while it iterates over the vertices?

msbrogli (changes requested):

hathor/event/model/node_state.py
glevco: Makes sense. Changed in https://github.com/HathorNetwork/hathor-core/pull/555/commits/0a4cb8cabcfac6f176d3b1556dc6997393c6259c

hathor/transaction/util.py
glevco: Added in https://github.com/HathorNetwork/hathor-core/pull/555/commits/0a4cb8cabcfac6f176d3b1556dc6997393c6259c

glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/555/commits/0a4cb8cabcfac6f176d3b1556dc6997393c6259c

hathor/event/event_manager.py
jansegre: I'm actually OK with this being a separate iterator loop, as long as the loop only happens when there are events to be emitted and this happens after the indexes initialization. My rationale is that if these events do need to be emitted it will be O(N) anyway (for N events), and if the indexes are initialized, an optimal iterator with no warmup will be able to be used, so at the end of the day it does seem like it will follow `N * (A + B)` -> `N * A + N * B` for A being a indexes initialization step and B the event emitting step. What do you think? @msbrogli @glevco

hathor/event/storage/event_storage.py
```diff
@@ -43,3 +44,18 @@ def get_last_group_id(self) -> Optional[int]:
...
         raise NotImplementedError
+
+    @abstractmethod
+    def clear_events(self) -> None:
+        """Clear all stored events and related metadata."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def save_node_state(self, state: NodeState) -> None:
+        """Save a node state in the storage"""
```
jansegre, line 55: What happens in this situation when using the same database:

1. node is initialized with `--x-enable-event-queue`: load events are emitted for the first time
2. node is stopped and initialized without `--x-enable-event-queue`: new transactions and blocks are received, which would have resulted in new events in case the event queue was enabled
3. node is stopped and initialized with `--x-enable-event-queue`: ...

Should the events be silently reset and the new load events be emitted? Should we try to emit load events just for the new transactions that were not emitted before? Should we prevent the node from being initialized in this state?

The indexes have a similar problem, and the solution I chose on 3. was to silently reset the index and re-initialize it, but IMO the equivalent would not be a good option for the case of the event queue.

Also, on the indexes, I included the initialization time to the saved state to be able to identify if the node is in this situation. I think we need at the very least a similar mechanism to identify this on the reliable integration case and be able to abort the node initialization. Possibly even detect it on 2. to prevent new transactions from being added that would not result in new events (in that case we would have to have a way to delete the event queue in order to safely start a node on 2. and 3. would be indistinguishable from 1.)

Does this make sense? What do you think? @msbrogli @glevco

glevco: Yes, it makes sense. Nicely spotted. What about always saving a state of whether `--x-enable-event-queue` was used or not? Something like `event_queue_enabled: bool` that would also be stored as metadata in RocksDB, every time the full node is initialized.

That would allow us to check if events were used in the previous full node initialization, when a full node is restarted. If `--x-enable-event-queue` is used, and `event_queue_enabled = False` (meaning we are in situation 3), then we have to regenerate events from the beginning, just like when `node_state = LOAD`.

| Previous NodeState | Previous event_queue_enabled | Action |
| - | - | - |
| `LOAD` | `False` | send events from the beginning |
| `LOAD` | `True` | send events from the beginning |
| `SYNC` | `False` | send events from the beginning |
| `SYNC` | `True` | send events from where it stopped |

In the last case, assuming we use the UUID idea to represent the "event stream version", the client would know it has to reprocess all events.

hathor/event/event_manager.py
glevco: I agree it works for now. I think in the future when we decide to refactor the iterator out of `IndexesManager`, we should take this into account and optimize it into a single loop. But considering your big O comments, I think it's fine for now.

> as long as the loop only happens when there are events to be emitted and this happens after the indexes initialization

Yep, that is the case

hathor/event/model/node_state.py
msbrogli: Thanks!

hathor/transaction/util.py
msbrogli: Thanks!

hathor/event/event_manager.py
msbrogli: I partially agree with the complexity analysis. Even though it will be N * (A + B), the iterator must load the transactions twice, and loading might be slower than emitting the events.

Anyway, I agree that we do like this for now. If the full node loading time increases too much, we can change it later.

hathor/event/storage/event_storage.py
msbrogli: Nice catch, @jansegre . I feel that, once the database was initialized with `event_queue` enabled, it should force it to be enabled. We might add a command to clear the event queue, allowing the full node to run with the event queue disabled. What do you think?

glevco: I'm fine going either way. What's your preference @jansegre?

jansegre (dismissed):

hathor/event/storage/event_storage.py
glevco: Ok! I'll implement it like this, then. Thanks!

glevco: Done! I have updated the Acceptance Criteria in the description to address all changes.

Tests using the simulator will be created to test the scenario described by Jan. They will be implemented in a separate PR, after https://github.com/HathorNetwork/hathor-core/pull/565 is merged.

jansegre (dismissed):

hathor/builder/cli_builder.py
```diff
@@ -143,19 +143,15 @@ def create_manager(self, reactor: PosixReactorBase, args: Namespace) -> HathorMa
...
-        else:
-            self.check_or_raise(not args.x_emit_load_events, '--x-emit-load-events cannot be used without '
-                                                             '--x-enable-event-queue')
+
+        event_manager = EventManager(
+            event_storage=event_storage,
+            event_ws_factory=self.event_ws_factory,
+            pubsub=pubsub,
+            reactor=reactor
+        )
```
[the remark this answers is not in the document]
glevco: Exactly. The `EventManager` is still used to retrieve the last event queue feature state (either enabled or disabled) from when the full node was previously initialized, but it can only be used for the actual event queue mechanism if `_enable_event_queue` is `True`

hathor/event/storage/event_storage.py
glevco: Sure! I have created an issue: https://github.com/HathorNetwork/hathor-core/issues/580

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/582

msbrogli (approved):

msbrogli (approved):

jansegre (approved):

hathor/event/event_manager.py
msbrogli: Agreed. Thanks!

PR #557, opened 2023-04-17
fix: handle missing git command when getting local version

luislhl:


hathor/version.py
```diff
@@ -53,7 +53,7 @@ def _get_build_version() -> Optional[str]:
 def _get_git_revision_short_hash() -> Optional[str]:
     try:
         return subprocess.check_output(['git', 'rev-parse', '--short', 'HEAD']).decode('ascii').strip()
-    except subprocess.CalledProcessError:
+    except (subprocess.CalledProcessError, FileNotFoundError):
         logger.warn((
             "Error while trying to get local git head. We are probably not in a git repo. "
```
jansegre, line 58: Should we rephrase this as something like "git not available"? Since it can happen that we are on a git repo but there's no git command in the current PATH (like it happens on the CI).

jansegre (approved):

glevco (approved):

PR #559, opened 2023-04-20
refactor(builder): Change HathorManager.__init__ and change tests to use TestBuilder

msbrogli:
## Acceptance criteria

1) Make parameters `peer_id`, `tx_storage`, and `network` required on `HathorManager.__init__(...)`.
2) Create a `TestBuilder` to build all objects for tests.
3) Refactor tests to use `TestBuilder`.

Note: I was considering to refactor `CliBuilder` to use `Builder` but I felt it was too much for now. Besides that, I don't need to change the `CliBuilder` to move forward with my project.

### DIFF
--- a/hathor/builder/__init__.py
+++ b/hathor/builder/__init__.py
@@ -1,0 +1,22 @@
+# Copyright 2023 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from hathor.builder.builder import BuildArtifacts, Builder
+from hathor.builder.cli_builder import CliBuilder
+
+__all__ = [
+    'BuildArtifacts',
+    'Builder',
+    'CliBuilder',
+]
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -1,0 +1,346 @@
+# Copyright 2021 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+from typing import Any, Dict, List, NamedTuple, Optional
+
+from structlog import get_logger
+
+from hathor.checkpoint import Checkpoint
+from hathor.conf import HathorSettings
+from hathor.conf.settings import HathorSettings as HathorSettingsType
+from hathor.consensus import ConsensusAlgorithm
+from hathor.event import EventManager
+from hathor.indexes import IndexesManager
+from hathor.manager import HathorManager
+from hathor.p2p.peer_id import PeerId
+from hathor.pubsub import PubSubManager
+from hathor.storage import RocksDBStorage
+from hathor.transaction.storage import TransactionMemoryStorage, TransactionRocksDBStorage, TransactionStorage
+from hathor.util import Random, Reactor, get_environment_info
+from hathor.wallet import BaseWallet, Wallet
+
+logger = get_logger()
+
+
+class StorageType(Enum):
+    MEMORY = 'memory'
+    ROCKSDB = 'rocksdb'
+
+
+class BuildArtifacts(NamedTuple):
+    """Artifacts created by a builder."""
+    settings: HathorSettingsType
+    rng: Random
+    reactor: Reactor
+    manager: HathorManager
+    pubsub: PubSubManager
+    consensus: ConsensusAlgorithm
+    tx_storage: TransactionStorage
+    indexes: Optional[IndexesManager]
+    wallet: Optional[BaseWallet]
+    rocksdb_storage: Optional[RocksDBStorage]
+
+
+class Builder:
+    """Builder builds the core objects to run a full node.
+
+    Example:
+
+        builder = Builder()
+        builder.use_memory()
+        artifacts = builder.build()
+    """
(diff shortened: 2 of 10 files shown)


hathor/builder.py
```diff
@@ -36,11 +40,266 @@
...
+        self._enable_utxo_index: bool = False
+
+        self._enable_sync_v1: Optional[bool] = None
+        self._enable_sync_v2: Optional[bool] = None
+
+        self._stratum_port: Optional[int] = None
+
+    def build(self) -> BuildArtifacts:
+        if self.artifacts is not None:
+            return self.artifacts
```
jansegre, line 110: Is there any application currently for being able to call `builder.build()` multiple times? Otherwise I'd rather have it only build once and fail on next calls, and if the user of `builder` needs it can persist `build_artifacts` instead of the `builder`.

Otherwise I think it opens up to this kind of silent error:

```python
builder = Builder()
builder.set_rng(...)  # or any other set method
... = builder.build()
# this can happen on another interface
builder.set_rng(...)  # this will silently fail, that is it will not have any effect on the next call
... = builder.build()
```

Maybe it's worth to block any `set_*` method after `build` is called (in Rust for example this would be prevented by making the `build` method "consume" the builder object and prevent it from being used after that, I'm not sure if it's possible to make `del self` inside of `build` right before returning, which would have a similar effect). Either way, IMO it would be enough to just `raise ValueError('cannot call build twice')` here and leave other improvements for the future.

What do you think?

hathor/builder.py
```diff
@@ -36,11 +40,266 @@
...
+        self._enable_sync_v1 = False
+
+    def enable_sync_v2(self) -> None:
+        self._enable_sync_v2 = True
+
+    def disable_sync_v2(self) -> None:
+        self._enable_sync_v2 = False
+
+
 class CliBuilder:
```
jansegre, line 303: Should we add a TODO to port `CliBuilder` to use the `Builder` interface? I think it could even be ported to a `Builder.from_args` classmethod (and the reactor could be a `set_*` method).

hathor/builder.py
```diff
@@ -36,11 +40,266 @@
...
+
+        return self.artifacts
+
+    def set_event_manager(self, event_manager: EventManager) -> None:
+        self._event_manager = event_manager
+
+    def _get_peer_id(self) -> PeerId:
+        if self._peer_id is not None:
+            return self._peer_id
+        raise NotImplementedError
```
jansegre, line 188: I think `raise ValueError('property not set')` would be better than `raise NotImplementedError`, for this and all `_get_*` methods that do this.

hathor/builder.py
```diff
@@ -36,11 +40,266 @@
...
+    def set_capabilities(self, capabilities: List[str]) -> None:
+        self._capabilities = capabilities
+
+    def set_peer_id(self, peer_id: PeerId) -> None:
+        self._peer_id = peer_id
+
+    def _get_settings(self) -> HathorSettingsType:
+        return HathorSettings()
+
+    def _create_pubsub(self, reactor: Reactor) -> PubSubManager:
```
jansegre, line 205: Should we call it `_get_or_create_pubsub` and persist the created pubsub like it is done with `_get_rocksdb_storage`.

Also, should we unify `_create_*` and `_get_*` methods all as `_get_or_create_*`? To me the semantics are already sort of mixed and "get or create" would fit better (specially when we want to override them like we do with `_get_peer_id` in `TestBuilder`.

hathor/builder.py
```diff
@@ -36,11 +40,266 @@
...
+            rocksdb_storage = self._get_rocksdb_storage()
+            use_memory_index = self._force_memory_index
+            return TransactionRocksDBStorage(rocksdb_storage, use_memory_indexes=use_memory_index)
+        raise NotImplementedError
+
+    def use_memory(self) -> None:
+        self._storage_type = StorageType.MEMORY
+
+    def use_rocksdb(self, path: str) -> None:
+        self._storage_type = StorageType.ROCKSDB
```
jansegre, line 236: Should it `assert self._storage_type is None` or if not an `assert` check and raise a `ValueError`?

jansegre (approved):
Functionally I'm more than OK with this PR, I really like the direction we're going with the Builder. I only have some smaller naming and behavior suggestions.

hathor/builder.py
msbrogli: Agreed. Done!

hathor/builder.py
msbrogli: Done!

hathor/builder.py
msbrogli: Done!

hathor/builder.py
msbrogli: Done!

hathor/builder.py
msbrogli: It might. But I don't see a problem with the following calls:

```python
builder = Builder()
builder.use_memory()
builder.use_rocksdb(path)
builder.use_memory()
artifacts = builder.build()
```

In this case, it would use memory objects.

hathor/builder.py
```diff
@@ -36,12 +40,308 @@
...
+    manager: HathorManager
+    pubsub: PubSubManager
+    consensus: ConsensusAlgorithm
+    tx_storage: TransactionStorage
+    indexes: Optional[IndexesManager]
+    wallet: Optional[BaseWallet]
+    rocksdb_storage: Optional[RocksDBStorage]
+
+
+class Builder:
```
glevco, line 70: What do you think about moving this and other new related classes to their own file?

hathor/builder.py
jansegre: I think it makes sense, that is more consistent with how all the other methods work.

hathor/builder.py
```diff
@@ -36,12 +40,308 @@
...
+
+    def set_checkpoints(self, checkpoints: List[Checkpoint]) -> None:
+        self.check_if_can_modify()
+        self._checkpoints = checkpoints
+
+    def set_capabilities(self, capabilities: List[str]) -> None:
+        self.check_if_can_modify()
+        self._capabilities = capabilities
+
+    def set_peer_id(self, peer_id: PeerId) -> None:
```
glevco, line 212: A common builder pattern is to `return self` on every config method, allowing for method chaining and a less verbose api, for example:

```python
artifacts = (
    Builder()
        .set_event_manager(event_manager)
        .set_rng(rng)
        .set_checkpoints(checkpoints)
        .set_capabilities(capabilities)
        .set_peer_id(peer_id)
)
```

What do you think?

glevco (approved):
Like Jan, I like this direction and I only have minor comments. Moving `CliBuilder` to this solution soon would be great.

hathor/builder.py
msbrogli: Nice catch! Thanks! Done!

hathor/builder.py
msbrogli: Sure! It makes a lot of sense. Thanks! Done!

PR #561, opened 2023-04-20
refactor: use new Builder class on Simulator

glevco:
Acceptance Criteria:

- Refactor `Simulator` to use `Builder`
- Changes to `Builder`
  - Add `full_verification` setting
  - Add `soft_voided_tx_ids` setting
  - Add `set_enable_sync_v1` and `set_enable_sync_v2` helper methods
- Add `full_verification` argument to `HathorManager.__init__()`
  - Update `CliBuilder` to use it
  - Update `TestCase` to use it

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -101,7 +101,7 @@ class Builder:
 
         self._stratum_port: Optional[int] = None
 
-        self._full_verification: bool = False
+        self._full_verification: Optional[bool] = None
 
         self._soft_voided_tx_ids: Optional[Set[bytes]] = None
 
@@ -149,6 +149,9 @@ class Builder:
         if self._event_manager is not None:
             kwargs['event_manager'] = self._event_manager
 
+        if self._full_verification is not None:
+            kwargs['full_verification'] = self._full_verification
+
         manager = HathorManager(
             reactor,
             pubsub=pubsub,
@@ -161,7 +164,6 @@ class Builder:
             checkpoints=self._checkpoints,
             capabilities=self._capabilities,
             environment_info=get_environment_info(self._cmdline, peer_id.id),
-            full_verification=self._full_verification,
             **kwargs
         )
 
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from enum import Enum
-from typing import Any, Dict, List, NamedTuple, Optional
+from typing import Any, Dict, List, NamedTuple, Optional, Set
 
 from structlog import get_logger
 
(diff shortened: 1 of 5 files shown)


hathor/builder/builder.py
```diff
@@ -157,6 +161,7 @@ def build(self) -> BuildArtifacts:
             checkpoints=self._checkpoints,
             capabilities=self._capabilities,
             environment_info=get_environment_info(self._cmdline, peer_id.id),
+            full_verification=self._full_verification,
```
msbrogli, line 164: If you pass this parameter this way, you might pass `None` but the `HathorManager.__init__(...)` expects a bool. I think you can easily fix it by doing the following:

```python
if self._full_verification is not None:
    kwargs['full_verification'] = self._full_verification
```

hathor/simulator/simulator.py
```diff
@@ -127,52 +126,45 @@ def stop(self) -> None:
...
         self.run_to_completion()
 
         # Don't use it anywhere else. It is unsafe to generate mnemonic words like this.
         # It should be used only for testing purposes.
         m = Mnemonic('english')
         words = m.to_mnemonic(self.rng.randbytes(32))
         self.log.debug('randomized step: generate wallet', words=words)
-        wallet.unlock(words=words, tx_storage=manager.tx_storage)
-        return manager
+        wallet.unlock(words=words, tx_storage=artifacts.manager.tx_storage)
```
msbrogli, line 165: You can use `artifacts.tx_storage` instead of `artifacts.manager.tx_storage`.

msbrogli (approved):

hathor/builder/builder.py
glevco: But this parameter was defined as `bool`, it could not be `None`:

```python
self._full_verification: bool = False
```

hathor/simulator/simulator.py
glevco: Thanks! Done in https://github.com/HathorNetwork/hathor-core/pull/561/commits/bcd0b3e99b73fa96f64ca92141d54d63732b6795

hathor/builder/builder.py
msbrogli: Got it. My only consideration here would be that, if we change it to `Optional[bool]`, we can skip passing the parameter which would let `HathorManager` decide the best default value for it since it's optional. If we pass `False`, we would be overriding `HathorManager` default value.

hathor/simulator/simulator.py
msbrogli: Thanks!

msbrogli (dismissed):

hathor/builder/builder.py
glevco: Ahh I see. Makes sense. Changed in https://github.com/HathorNetwork/hathor-core/pull/561/commits/d8a18d2d3759bb2f674b840618cf85063236077e, thanks!

msbrogli: Great! Thanks!

msbrogli (approved):

jansegre (approved):

jansegre (approved):

PR #563, opened 2023-04-22
feat(simulator): Add a trigger to stops the simulator at selected times

msbrogli:


hathor/simulator/trigger.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def reset(self) -> None:
+        """Reset the counter, so this trigger can be reused."""
+        self.initial_blocks_found = self.miner.blocks_found
+
+    def should_stop(self) -> bool:
+        diff = self.miner.blocks_found - self.initial_blocks_found
+        if diff >= self.quantity:
+            return True
+        return False
```
glevco, line 44: Why not just:

```suggestion
        return diff >= self.quantity
```

hathor/simulator/simulator.py
```diff
@@ -250,7 +256,13 @@ def run_until_complete(self,
     def run(self,
             interval: float,
             step: float = DEFAULT_STEP_INTERVAL,
-            status_interval: float = DEFAULT_STATUS_INTERVAL) -> None:
+            status_interval: float = DEFAULT_STATUS_INTERVAL,
+            *,
+            trigger: Optional[Trigger] = None) -> bool:
```
glevco, line 261: What about receiving a list here to support multiple triggers? It could stop if any of them stops it

hathor/simulator/simulator.py
```diff
@@ -250,7 +256,13 @@ def run_until_complete(self,
...
         assert self._started
         for _ in self._run(interval, step, status_interval):
-            pass
+            if trigger is not None and trigger.should_stop():
+                return True
+        if trigger is not None:
+            return False
+        return True
```
glevco, line 268: What does `True` mean in this case?

glevco (approved):

hathor/simulator/trigger.py
msbrogli: I usually prefer the most explicitly form. But I agree that the readability is better for your suggestion. So I'll do it.

hathor/simulator/simulator.py
msbrogli: I thought it would be confusing about the stop criteria. It's quite simple to develop a trigger that has a list an apply an OR or AND criteria. What do you think?

hathor/simulator/simulator.py
msbrogli: I'll add a docstring for this method.

hathor/simulator/trigger.py
msbrogli: Done!

hathor/simulator/simulator.py
msbrogli: Done! What do you think?

glevco: By reading the code, I had understood that `True` meant "stopped by trigger". With the docstring it makes sense, thanks!

hathor/simulator/simulator.py
glevco: Ok, you're right, a composed trigger could be implemented if necessary

glevco (dismissed):

jansegre (dismissed):

jansegre (approved):

glevco (approved):

PR #568, opened 2023-04-25
chore: add RocksDB and EventManager builder configs

glevco:
Acceptance Criteria:

- Make `HathorManager` receive a required `EventStorage`
  - This will be necessary on https://github.com/HathorNetwork/hathor-core/pull/555, to check if the event queue feature was enabled in the previous run
- `Builder` changes
  - Add `peer_id` to builder artifacts
  - Make memory be the default storage type
  - Add RockDB's `cache_capacity` and `with_index` configs
  - Add `enable_event_manager()` auxiliary method
- Update usages of `Builder`, (`CliBuilder` and `Simulator`) to use event related configs

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -78,10 +78,11 @@ class Builder:
         self._network: Optional[str] = None
         self._cmdline: str = ''
 
-        self._storage_type: Optional[StorageType] = None
+        self._storage_type: StorageType = StorageType.MEMORY
         self._force_memory_index: bool = False
 
         self._event_manager: Optional[EventManager] = None
+        self._event_ws_factory: Optional[EventWebsocketFactory] = None
 
         self._rocksdb_path: Optional[str] = None
         self._rocksdb_storage: Optional[RocksDBStorage] = None
@@ -89,6 +90,7 @@ class Builder:
         self._with_rocksdb_index: Optional[bool] = None
 
         self._tx_storage: Optional[TransactionStorage] = None
+        self._event_storage: Optional[EventStorage] = None
 
         self._reactor: Optional[Reactor] = None
         self._pubsub: Optional[PubSubManager] = None
@@ -124,6 +126,8 @@ class Builder:
         consensus_algorithm = ConsensusAlgorithm(soft_voided_tx_ids, pubsub)
 
         wallet = self._get_or_create_wallet()
+        event_storage = self._get_or_create_event_storage()
+        event_manager = self._get_or_create_event_manager()
         tx_storage = self._get_or_create_tx_storage()
         indexes = tx_storage.indexes
         assert indexes is not None
@@ -151,9 +155,6 @@ class Builder:
         if self._network is None:
             raise TypeError('you must set a network')
 
-        if self._event_manager is not None:
-            kwargs['event_manager'] = self._event_manager
-
         if self._full_verification is not None:
             kwargs['full_verification'] = self._full_verification
 
@@ -163,12 +164,14 @@ class Builder:
             consensus_algorithm=consensus_algorithm,
             peer_id=peer_id,
             tx_storage=tx_storage,
+            event_storage=event_storage,
             network=self._network,
             wallet=wallet,
             rng=self._rng,
             checkpoints=self._checkpoints,
             capabilities=self._capabilities,
             environment_info=get_environment_info(self._cmdline, peer_id.id),
+            event_manager=event_manager,
             **kwargs
         )
 
@@ -283,26 +286,33 @@ class Builder:
 
         raise NotImplementedError
 
-    def _get_event_storage(self) -> EventStorage:
-        if self._storage_type == StorageType.MEMORY:
-            return EventMemoryStorage()
-
-        if self._storage_type == StorageType.ROCKSDB:
+    def _get_or_create_event_storage(self) -> EventStorage:
+        if self._event_storage is not None:
+            pass
+        elif self._storage_type == StorageType.MEMORY:
+            self._event_storage = EventMemoryStorage()
+        elif self._storage_type == StorageType.ROCKSDB:
             rocksdb_storage = self._get_or_create_rocksdb_storage()
-            return EventRocksDBStorage(rocksdb_storage)
+            self._event_storage = EventRocksDBStorage(rocksdb_storage)
+        else:
+            raise NotImplementedError
 
-        raise NotImplementedError
+        return self._event_storage
+
+    def _get_or_create_event_manager(self) -> Optional[EventManager]:
+        if self._event_manager is None and self._event_ws_factory is not None:
+            self._event_manager = EventManager(
+                reactor=self._get_reactor(),
+                pubsub=self._get_or_create_pubsub(),
+                event_storage=self._get_or_create_event_storage(),
+                event_ws_factory=self._event_ws_factory
+            )
+
+        return self._event_manager
 
     def with_event_manager(self, *, event_ws_factory: EventWebsocketFactory) -> 'Builder':
         self.check_if_can_modify()
-
-        self._event_manager = EventManager(
-            reactor=self._get_reactor(),
-            pubsub=self._get_or_create_pubsub(),
-            event_storage=self._get_event_storage(),
-            event_ws_factory=event_ws_factory
-        )
-
+        self._event_ws_factory = event_ws_factory
         return self
 
     def use_memory(self) -> 'Builder':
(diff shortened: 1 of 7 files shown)


jansegre (approved):

jansegre (dismissed):

hathor/builder/builder.py
```diff
@@ -239,30 +245,82 @@ def _get_or_create_pubsub(self) -> PubSubManager:
...
+    def use_rocksdb(
+        self,
+        path: str,
+        with_index: Optional[bool] = None,
+        cache_capacity: Optional[int] = None
+    ) -> 'Builder':
         self.check_if_can_modify()
         self._storage_type = StorageType.ROCKSDB
         self._rocksdb_path = path
+        self._with_rocksdb_index = with_index
```
msbrogli, line 322: Isn't be better to standardize with the `_rocksdb_` prefix? For instance, this one would be `_rocksdb_with_index`.

hathor/builder/builder.py
```diff
@@ -239,30 +245,82 @@ def _get_or_create_pubsub(self) -> PubSubManager:
...
+
+            return TransactionRocksDBStorage(
+                rocksdb_storage,
+                use_memory_indexes=use_memory_index,
+                **kwargs
+            )
+
         raise NotImplementedError
 
+    def _get_event_storage(self) -> EventStorage:
```
msbrogli, line 286: We must be careful because it's not safe to call this method twice. I feel it is the same for many other methods in the builder.

hathor/builder/builder.py
```diff
@@ -239,30 +245,82 @@ def _get_or_create_pubsub(self) -> PubSubManager:
...
+        if self._storage_type == StorageType.ROCKSDB:
+            rocksdb_storage = self._get_or_create_rocksdb_storage()
+            return EventRocksDBStorage(rocksdb_storage)
+
+        raise NotImplementedError
+
+    def with_event_manager(self, *, event_ws_factory: EventWebsocketFactory) -> 'Builder':
+        self.check_if_can_modify()
+
+        self._event_manager = EventManager(
```
msbrogli, line 299: Why are you building while the builder is being configured?

hathor/builder/builder.py
```diff
@@ -239,30 +245,82 @@ def _get_or_create_pubsub(self) -> PubSubManager:
...
+        if self._storage_type == StorageType.MEMORY:
+            return EventMemoryStorage()
+
+        if self._storage_type == StorageType.ROCKSDB:
+            rocksdb_storage = self._get_or_create_rocksdb_storage()
+            return EventRocksDBStorage(rocksdb_storage)
+
+        raise NotImplementedError
+
+    def with_event_manager(self, *, event_ws_factory: EventWebsocketFactory) -> 'Builder':
```
msbrogli, line 296: Why `with_` instead of `set_`? From the code, we might rename it to `enable_event_manager(...)`. What do you think?

msbrogli, line 296: Why aren't you passing the event manager to the `HathorManager` in the `build()` method?

msbrogli (changes requested):

hathor/builder/builder.py
glevco: Agreed. Done in https://github.com/HathorNetwork/hathor-core/pull/568/commits/33a1b40d435bb62afe13891b4c418ab5667bb54e

glevco:
I have moved some changes from https://github.com/HathorNetwork/hathor-core/pull/555 to here in https://github.com/HathorNetwork/hathor-core/pull/568/commits/f1d9c96e5cff430a28b68943ada2700e54ac9c3b, as they resolve most of @msbrogli's review discussions. I have updated the Acceptance Criteria accordingly

hathor/builder/builder.py
glevco: Changed as explained in https://github.com/HathorNetwork/hathor-core/pull/568#issuecomment-1526563055

hathor/builder/builder.py
glevco: Changed as explained in https://github.com/HathorNetwork/hathor-core/pull/568#issuecomment-1526563055

hathor/builder/builder.py
glevco: Changed as explained in https://github.com/HathorNetwork/hathor-core/pull/568#issuecomment-1526563055

glevco: I didn't use `set_` because this method doesn't receive an `EventManager`, it receives an `EventWebsocketFactory` that is used (and required) to build the `EventManager` later.

I agree and changed it to `enable_event_manager()` in https://github.com/HathorNetwork/hathor-core/pull/568/commits/c10b4428a83ee376a7b68ccae29aeabf8a99137e

hathor/builder/builder.py
msbrogli: Thanks!

hathor/builder/builder.py
msbrogli: Great! Thanks!

hathor/manager.py
```diff
@@ -87,6 +88,7 @@ def __init__(self,
                  consensus_algorithm: ConsensusAlgorithm,
                  peer_id: PeerId,
                  tx_storage: TransactionStorage,
+                 event_storage: EventStorage,
```
msbrogli, line 91: Why would `HathorManager.__ini__(...)` receive an `EventStorage` object?

hathor/builder/builder.py
```diff
@@ -158,16 +164,19 @@ def build(self) -> BuildArtifacts:
...
             peer_id=peer_id,
             tx_storage=tx_storage,
+            event_storage=event_storage,
             network=self._network,
             wallet=wallet,
             rng=self._rng,
             checkpoints=self._checkpoints,
             capabilities=self._capabilities,
             environment_info=get_environment_info(self._cmdline, peer_id.id),
+            event_manager=event_manager,
```
msbrogli, line 174: Are we always instantiating an event manager? I thought it could be disabled.

hathor/builder/builder.py
```diff
@@ -239,30 +248,84 @@ def _get_or_create_pubsub(self) -> PubSubManager:
...
+    def _get_or_create_event_manager(self) -> Optional[EventManager]:
+        if self._event_manager is None and self._event_ws_factory is not None:
+            self._event_manager = EventManager(
+                reactor=self._get_reactor(),
+                pubsub=self._get_or_create_pubsub(),
+                event_storage=self._get_or_create_event_storage(),
+                event_ws_factory=self._event_ws_factory
+            )
+
+        return self._event_manager
```
msbrogli, line 311: Shouldn't we assert that `self._event_ws_factory` is None?

msbrogli (changes requested):

hathor/builder/builder.py
glevco: No, it can be disabled. The `_get_or_create_event_manager()` method in the `Builder` may return `None` if the `EventManager` was not set with `set_event_manager()` neither enabled with `enable_event_manager()`. The `HathorManager` also receives an `Optional[EventManager]`. Should it be passed in the kwargs instead?

hathor/builder/builder.py
glevco: We could, but I left it like this in the same vein as letting the user set `use_memory()` and then `use_rocksdb()`. An user could use `enable_event_manager()` and then `set_event_manager()`.

hathor/manager.py
msbrogli: Why don't you add methods to the EventManager to handle it? I mean, not sure we need to surface EventStorage to do this.

hathor/builder/builder.py
msbrogli: I guess it is fine. Thanks!

hathor/builder/builder.py
msbrogli: Got it. Thanks!

hathor/builder/builder.py
msbrogli: Thanks!

hathor/builder/builder.py
msbrogli: Thanks!

hathor/builder/builder.py
msbrogli: Thanks!

msbrogli (approved):

hathor/manager.py
msbrogli: Perfect. Let's do it in a separate PR so you can move forward with your other PRs.

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/555

PR #569, opened 2023-04-25
fix: EventData model

glevco:
Acceptance Criteria:

- Fix model bugs on `EventData`
- Move `zip_chunkify()` function to `tests/utils`

### DIFF
--- a/hathor/event/model/event_data.py
+++ b/hathor/event/model/event_data.py
@@ -14,7 +14,7 @@
 
 from typing import List, Optional, Union
 
-from pydantic import Extra
+from pydantic import Extra, validator
 
 from hathor.pubsub import EventArguments
 from hathor.utils.pydantic import BaseModel
@@ -23,7 +23,7 @@ from hathor.utils.pydantic import BaseModel
 class TxInput(BaseModel):
     tx_id: str
     index: int
-    data: int
+    data: str
 
 
 class TxOutput(BaseModel):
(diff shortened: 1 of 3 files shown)


jansegre (dismissed):

hathor/event/model/event_data.py
```diff
@@ -55,6 +51,15 @@ class TxMetadata(BaseModel, extra=Extra.ignore):
     height: int
     validation: str
 
+    @validator('spent_outputs', pre=True, each_item=True)
+    def parse_spent_outputs(cls, spent_output):
```
msbrogli, line 55: What's the expected type for `spent_output`?

msbrogli (approved):

hathor/event/model/event_data.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/569/commits/0012d991fdd21dd7dacb689758ff0fb9a61f2320

I've also added a docstring to make it clear that this method is supposed to only be called by pydantic

msbrogli: Thanks! Just checked the changes. I understand that json always return a list. In this case, it would be a `Tuple[int, List[str]]`, right? Maybe add this to the docstring and give an example to illustrate.

msbrogli (approved):

hathor/event/model/event_data.py
glevco: It looks like a tuple, but it's actually a list [in the code](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/transaction/transaction_metadata.py#L267) (that makes the type hint more complicated):

```python
data['spent_outputs'].append([idx, [h_bytes.hex() for h_bytes in hashes]])
```

I've added docstring tests (examples) in https://github.com/HathorNetwork/hathor-core/pull/569/commits/a9be780d9623922d732d562e8a6b941db113cbba

msbrogli: Got it. Thanks!

msbrogli (dismissed):

jansegre (approved):

msbrogli (approved):

PR #572, opened 2023-04-27
feat(consensus): Change meta.score and meta.accumulated_weight to int type

msbrogli:
## Motivation

The metadata `accumulated_weight` and `score` are calculated summing up the vertices' weight. Even though both metadata can be integer numbers, they are calculated as the logarithm of their integer value. This might lead to minor issues caused by precision error. We currently use the `calculate_min_significant_weight()` to prevent precision error on mining activity.

This PR converts both metadata to integer values, simplifying parts of the code that had to take float precision into consideration.

## Acceptance criteria

1. Change `meta.score` and `meta.accumulated_weight` to type `int`.
2. Change the calculation of `meta.score` to use `weight_to_work()`.
3. Change the calculation of `meta.accumulated_weight` to use `weight_to_work()`.
4. Change the consensus algorithm to do direct comparisons (instead of comparisons using `WEIGHT_TOL`).
5. Replace `sum_weights(a, b)` to `weight_to_work(a) + weight_to_work(b)`.
6. Replace `sum_weights(score, a)` to `score + weight_to_work(a)`.
7. Replace `sum_weights(accumulated_weight, a)` to `accumulated_weight + weight_to_work(a)`.
8. Remove `sum_weights()`.
9. Serialize `meta.score` and `meta.accumulated_weight` as strings to prevent precision issues with JSON parsers.


jansegre:
I haven't reviewed the code yet, but I think we should keep `calculate_min_significant_weight` for at least one release, in order to prevent an updated miner to propagate a block that would cause an issue with an outdated node.

hathor/consensus/block_consensus.py
glevco, line 1: Wouldn't this `int()` rounding used throughout the code create some problems? For example, considering the minimum tx weight of 14, we have `int(2**14) == int(2**14.00001)`. And for blocks, `int(2**21) == int(2**21.0000001)`.

Those wouldn't be considered ties in the current code, since `WEIGHT_TOL: float = 1e-10`.

hathor/consensus/block_consensus.py
```diff
@@ -171,15 +171,15 @@ def update_voided_info(self, block: Block) -> None:
...
+                    assert best_score == head_meta.score
+            assert isinstance(best_score, int)
 
             # Calculate the score.
             # We cannot calculate score before getting the heads.
             score = self.calculate_score(block)
 
             # Finally, check who the winner is.
-            if score <= best_score - self._settings.WEIGHT_TOL:
+            if score < best_score:
```
glevco, line 182: Why was the `=` removed? Same question for other checks below.

hathor/consensus/block_consensus.py
```diff
@@ -289,28 +289,30 @@ def update_score_and_mark_as_the_best_chain_if_possible(self, block: Block) -> N
...
+                    assert best_score == head_meta.score
                     best_heads.append(head)
-            assert isinstance(best_score, (int, float)) and best_score > 0
+            assert isinstance(best_score, int) and best_score > 0
 
             assert len(best_heads) > 0
             first_block = self._find_first_parent_in_best_chain(best_heads[0])
             self.add_voided_by_to_multiple_chains(best_heads[0], [block], first_block)
             if len(best_heads) == 1:
+                assert best_heads[0].hash != block.hash
```
glevco, line 315: Why was this assert removed?

glevco (changes requested):

hathor/consensus/block_consensus.py
jansegre: Maybe the motivation could use some examples.

The problem this change aims to solve is in the resolution of the `score` when considering new blocks. For example, take a recent block on the mainnet: [b1 = 0..d5698a7f2 (with height 4163578)](https://explorer.hathor.network/transaction/00000000000000026f5b7be2c6b07d1ec2cadfb68a395f19b8bd174d5698a7f2) and [b2 = 0...4e716e83 (with height 4163579)](https://explorer.hathor.network/transaction/0000000000000003d42946b85fa5aa11ad202c9e337aa29de61628df4e716e83). I'll break down some relevant values:

- b1's score: 88.17024203159183
- b2's weight: 61.80130451022343
- b2's score: 88.17024204823873

The `score` is basically a measure of how much work was performed since the genesis up until (and, very importantly, including) a given block. So `b2.score = b1.score + b2.weight`, so in this case, using the previous method:

```python
import math
b1_score = 88.17024203159183
b2_weight = 61.80130451022343
b2_score = b2_weight + math.log2(1 + 2 ** (b1_score - b2_weight))  # 88.17024204823872
```

You can see that the increment in score is very small (`b2_score - b1_score = 1.6646893641336646e-08`), dangerously close to the `WEIGHT_TOL = 1e-10`.

If b2's weight were to be a little bit smaller (we have a function to calculate this: `calculate_min_significant_weight`, `54.422195918736655` in this case), it would mean that the score wouldn't change. We implemented this function in the fullnode to aid miners, but it isn't ideal, because it means that if the hashrate drops too much, miners will force themselves to mine harders blocks (higher weight) solely because the implementation of the score will not pick their blocks otherwise.

Using the new method, these values would look something like this:

```python
b1_intscore = 348247154450026521432162304  # rough estimate
b2_weight = 61.80130451022343
b2_intscore = b1_intscore + int(2 ** b2_weight)  # 348247158468364191876180992
```

In this case the score will need more space to be stored (we're using string and python's seemless bigint, so unlimited, but we could probably cap it to 256 bits, instead of 64 bits using a double (python's float)). However even a weight of 0 (which is impossible, since the minimum for blocks is 21) would cause the score to increase. Meaning blocks' weights will _always_ make a significant increase to the score.

Now, for the accumulated weight, we don' have this problem, and since we stop updating after a certain cap, it likely doesn't make any difference, but since it works very symmetrically to the score (except going to the "future", instead of the "past"), it makes sense to use that too, but I don't have a preference to keep the current method or use the new one, at least using the new method will allow for some cleanup.

hathor/manager.py
```diff
@@ -795,14 +795,7 @@ def _make_block_template(self, parent_block: Block, parent_txs: 'ParentTxs', cur
...
         parent_block_metadata = parent_block.get_metadata()
-        # this is the min weight to cause an increase of twice the WEIGHT_TOL, we make sure to generate a template with
-        # at least this weight (note that the user of the API can set its own weight, the block sumit API will also
-        # protect agains a weight that is too small but using WEIGHT_TOL instead of 2*WEIGHT_TOL)
-        min_significant_weight = calculate_min_significant_weight(
-            parent_block_metadata.score,
-            2 * self._settings.WEIGHT_TOL
-        )
-        weight = max(self.daa.calculate_next_weight(parent_block, timestamp), min_significant_weight)
+        weight = self.daa.calculate_next_weight(parent_block, timestamp)
```
jansegre, line 798: Keep `calculate_min_significant_weight` for at least one more release. Maybe something like `TODO: remove calculate_min_significant_weight in the future`.

hathor/manager.py
```diff
@@ -863,15 +857,7 @@ def submit_block(self, blk: Block, fails_silently: bool = True) -> bool:
...
             return False
-        parent_block = self.tx_storage.get_transaction(parent_hash)
-        parent_block_metadata = parent_block.get_metadata()
         # this is the smallest weight that won't cause the score to increase, anything equal or smaller is bad
-        min_insignificant_weight = calculate_min_significant_weight(
-            parent_block_metadata.score,
-            self._settings.WEIGHT_TOL
-        )
-        if blk.weight <= min_insignificant_weight:
-            self.log.warn('submit_block(): insignificant weight? accepted anyway', blk=blk.hash_hex, weight=blk.weight)
```
jansegre, line 874: Keep `calculate_min_significant_weight` for at least one more release. Maybe something like `TODO: remove calculate_min_significant_weight in the future`.

hathor/consensus/block_consensus.py
glevco: Thank you for the detailed example, indeed it would be useful to have this in the motivation. Why did you say `b1_intscore` is a "rough estimate"?

hathor/utils/weight.py
```diff
@@ -0,0 +1,23 @@
...
+
+from math import exp, log
+
+_LN_2 = log(2)
+
+
+def weight_to_work(weight: float) -> int:
+    """Convert weight to work rounding to the nearest integer."""
+    # return int(0.5 + 2**weight)
+    return int(0.5 + exp(weight * _LN_2))
```
msbrogli, line 23: Which one has greater precision?

hathor/transaction/base_transaction.py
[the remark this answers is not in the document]
msbrogli: Done!

hathor/consensus/block_consensus.py
msbrogli: Please open one thread for each case so we double check I'm doing it right. In this case, we are checking if our score is smaller than the best_score (but > and == cases are handled in another part of the code).

hathor/manager.py
msbrogli: Done!

hathor/manager.py
msbrogli: Done!

msbrogli:
> I haven't reviewed the code yet, but I think we should keep `calculate_min_significant_weight` for at least one release, in order to prevent an updated miner to propagate a block that would cause an issue with an outdated node.

Nice catch! I just did it. Thanks!

hathor/transaction/storage/migrations/change_score_acc_weight_metadata.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Migration(BaseMigration):
+    def skip_empty_db(self) -> bool:
+        return True
+
+    def get_db_name(self) -> str:
+        return 'change_score_acc_weight_metadata'
+
+    def run(self, storage: 'TransactionStorage') -> None:
+        raise Exception('Cannot migrate your database. Please, delete your data folder and sync again.')
```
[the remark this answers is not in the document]
msbrogli: Done!

hathor/transaction/resources/transaction.py
```diff
@@ -405,8 +405,8 @@ def get_list_tx(self, request):
                                             'conflict_with': [],
                                             'voided_by': [],
                                             'twins': [],
-                                            'accumulated_weight': 10,
-                                            'score': 12,
+                                            'accumulated_weight': '1024',
+                                            'score': '4096',
```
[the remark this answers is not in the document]
msbrogli: Nice catch. Fixed there too.

hathor/consensus/block_consensus.py
msbrogli: I think it's a rough estimate because it just `int(2**score)` (and a more precise value would be calculated as `sum(int(2**w) for w in previous_blocks_and_confirmed_transactions)`.

msbrogli: Short answer:

- No issues for blocks since we will compare scores (and not weights). Besides that, this new approach is more precise than the current tolerance (1e-10) for weights greater than 34.
- No issues for confirmed transactions because they will have at least the block weight added to their accumulated weights (e.g., `acc_weight = weight_to_work(tx.weight) + weight_to_work(meta.first_block.weight)`.
- No major issues for transactions in the mempool because other transactions confirming them will also be added to the accumulated weight.

I'm preparing a short rfc to give more details about these remarks.

hathor/transaction/base_transaction.py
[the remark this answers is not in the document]
glevco: Thanks!

hathor/consensus/block_consensus.py
glevco: Ok!

hathor/consensus/block_consensus.py
```diff
@@ -200,7 +201,7 @@ def update_voided_info(self, block: Block) -> None:
                 common_block = self._find_first_parent_in_best_chain(block)
                 self.add_voided_by_to_multiple_chains(block, heads, common_block)
 
-                if score >= best_score + self._settings.WEIGHT_TOL:
+                if score > best_score:
```
glevco, line 204: Why was the `=` removed?

hathor/consensus/block_consensus.py
```diff
@@ -289,28 +290,30 @@ def update_score_and_mark_as_the_best_chain_if_possible(self, block: Block) -> N
...
         if self.update_voided_by_from_parents(block):
             storage = block.storage
             heads = [cast(Block, storage.get_transaction(h)) for h in storage.get_best_block_tips()]
-            best_score = 0.0
+            best_score = 0
             best_heads: list[Block]
             for head in heads:
                 head_meta = head.get_metadata(force_reload=True)
-                if head_meta.score <= best_score - self._settings.WEIGHT_TOL:
+                if head_meta.score < best_score:
```
glevco, line 301: Why was the `=` removed?

hathor/consensus/block_consensus.py
```diff
@@ -289,28 +290,30 @@ def update_score_and_mark_as_the_best_chain_if_possible(self, block: Block) -> N
...
                     continue
 
-                if head_meta.score >= best_score + self._settings.WEIGHT_TOL:
+                if head_meta.score > best_score:
```
glevco, line 304: Why was the `=` removed?

hathor/consensus/block_consensus.py
msbrogli: @glevco and @jansegre , confirm that you agree that I did the right thing. I'll wait for your confirmation before resolving this conversation.

hathor/consensus/block_consensus.py
msbrogli: Here we are checking if the block has the highest score. So we must check if `score > best_score`.

@glevco and @jansegre , confirm that you agree that I did the right thing. I'll wait for your confirmation before resolving this conversation.

hathor/consensus/block_consensus.py
msbrogli: This loop is selecting the heads with highest score. As it's possible to have more than one head with the same score, it returns a list of blocks. This comparison is skipping heads that have lower score so they will no be selected at all. So we have to check if `head_meta.score < best_score`.

@glevco and @jansegre , confirm that you agree that I did the right thing. I'll wait for your confirmation before resolving this conversation.

hathor/consensus/block_consensus.py
msbrogli: This case is checking whether the `head_meta.score > best_score`. If it is, all previously seen heads are eliminated and the current head becomes the only selected head so far.

hathor/consensus/block_consensus.py
jansegre: > Thank you for the detailed example, indeed it would be useful to have this in the motivation. Why did you say `b1_intscore` is a "rough estimate"?

> I think it's a rough estimate because it just `int(2**score)` (and a more precise value would be calculated as `sum(int(2**w) for w in previous_blocks_and_confirmed_transactions)`.

Exactly, repeatedly doing `sum_weights` on float weights will cause precision loss compared to repeatedly adding int weights.

glevco: Makes sense, thanks! I'm closing this thread, we can continue the discussion on the RFC @msbrogli is going to release.

hathor/consensus/block_consensus.py
```diff
@@ -171,15 +172,15 @@ def update_voided_info(self, block: Block) -> None:
                     best_score = head_meta.score
                 else:
                     # All heads must have the same score.
-                    assert abs(best_score - head_meta.score) < 1e-10
-            assert isinstance(best_score, (int, float))
+                    assert best_score == head_meta.score
+            assert isinstance(best_score, int)
```
glevco, line 176: Maybe set the type of `best_score: int | None = None` above, and then this could be `assert best_score is not None`

hathor/transaction/base_transaction.py
```diff
@@ -60,26 +61,6 @@
 _ONE_BYTE = 0xFF
 
 
-def sum_weights(w1: float, w2: float) -> float:
```
msbrogli, line 63: I think I won't be able to remove the `sum_weights()` in this PR because it is still used in 3 places:

- `metrics.py`: Seems easy to remove it.
- `difficulty.py`: This is used by merged mining and maybe Jan can help me understand the consequences of changing it. Anyway, it seems feasible.
- `stratum/straum.py`: Seems easy but we need to better assess its impact. I think it's used in the hash rate estimator only. And, to be honest, I don't know if we still use this code.
- `daa.py`: I think we should not change this one, as it might affect the minimum weight of the blocks and invalidate blocks that are on the mainnet. To confirm the consequence, we would need to check it against the mainnet but it seems not worth the risk.

My impression is that, in the end, we will remove the `sum_weights()` from (1), (2), and (3) and then we will move its code to the `daa.py` and continue using it there.

What do you think?

hathor/consensus/block_consensus.py
glevco: I agree.

hathor/consensus/block_consensus.py
glevco: I agree.

hathor/consensus/block_consensus.py
glevco: I agree.

hathor/consensus/block_consensus.py
glevco: I agree.

hathor/transaction/base_transaction.py
jansegre: In `difficulty` it is only used to implement a `logsum` for function reuse. The following patch ports the exact same math:

```
diff --git a/hathor/difficulty.py b/hathor/difficulty.py
index 4a8467ef..fa80c124 100644
--- a/hathor/difficulty.py
+++ b/hathor/difficulty.py
@@ -787,8 +787,8 @@ class Weight(float):
 
         Currently is just a proxy to `hathor.transaction.sum_weights`.
         """
-        from hathor.transaction import sum_weights
-        return Weight(sum_weights(self, other))
+        a, b = self, other if self > other else other, self
+        return Weight(a + log2(1 + 2 ** (b - a)))
 
     def to_u256(self) -> U256:
         """ Convert to U256

```

Since `hathor.difficulty` is intended to abstract that kind of values, it can be used in places like `hathor.metrics`, `hathor.stratum.stratum` and `hathor.daa`. That should preserve the exact same operations as before. I can see how `hathor.daa` could maybe be refactored but I agree it's not worth to touch it now.

hathor/manager.py
glevco, line 1: We have to update the calls to `calculate_min_significant_weight()`, or rather, the function itself, right? It's expecting to receive a `log_score`, but we're passing the new `score`.

jansegre (dismissed):

hathor/transaction/storage/migrations/change_score_acc_weight_metadata.py
```diff
@@ -0,0 +1,33 @@
...
+    def skip_empty_db(self) -> bool:
+        return True
+
+    def get_db_name(self) -> str:
+        return 'change_score_acc_weight_metadata'
+
+    def run(self, storage: 'TransactionStorage') -> None:
+        raise Exception('Cannot migrate your database due to an incompatible change in the metadata. '
+                        'Please, delete your data folder and use the latest available snapshot or sync '
+                        'from beginning.')
```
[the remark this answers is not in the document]
glevco: I agree, and I have two open PRs (https://github.com/HathorNetwork/hathor-core/pull/927, https://github.com/HathorNetwork/hathor-core/pull/931) that have migrations. It would be good if we could merge them before this one.

hathor/consensus/block_consensus.py
glevco: Resolving this, then

hathor/consensus/block_consensus.py
glevco: Resolving this, then.

hathor/consensus/block_consensus.py
glevco: Resolving this, then.

hathor/transaction/storage/migrations/change_score_acc_weight_metadata.py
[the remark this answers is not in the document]
glevco: I just released https://github.com/HathorNetwork/hathor-core/pull/933 which also has a migration, so it would be nice to wait for it too.

hathor/manager.py
jansegre: Yes, I've updated this function.

hathor/transaction/base_transaction.py
jansegre: I've decided to keep `sum_weights` so we don't have to refactor `hathor.metrics`, `hathor.stratum.stratum`, etc.

hathor/utils/weight.py
jansegre: I've decided to go with `math.ceil(2**weight)`, it's exactly the same as method1, but doesn't need the `0.5 +` operation.

hathor/event/model/event_data.py
```diff
@@ -52,8 +52,8 @@ class TxMetadata(BaseModel, extra=Extra.ignore):
     received_by: list[int]
     children: list[str]
     twins: list[str]
-    accumulated_weight: float
-    score: float
+    accumulated_weight: str
```
msbrogli, line 55: Can this have impact on the wallet service and other use case using the reliable integrations? Should we calculate `log2(...)` to keep the model consistent? If yes, should we do it on the other APIs too? I mean, is there any benefit on using the whole integer instead of its more friendly log2 representation?

hathor/util.py
```diff
@@ -805,10 +805,10 @@ def get_hathor_core_version():
     return hathor.__version__
 
 
-def calculate_min_significant_weight(score: float, tol: float) -> float:
+def calculate_min_significant_weight(score: int, tol: float) -> float:
     """ This function will return the min significant weight to increase score by tol.
     """
-    return score + math.log2(2 ** tol - 1)
+    return math.log2(score) + math.log2(2 ** tol - 1)
```
msbrogli, line 811: Maybe create a method `work_to_weight(...)`?

hathor/utils/weight.py
```diff
@@ -0,0 +1,24 @@
...
+#  limitations under the License.
+
+import math
+
+# _LN_2 = math.log(2)
+
+
+def weight_to_work(weight: float) -> int:
+    """Convert weight to work rounding to the nearest integer."""
+    # return int(0.5 + 2**weight)
```
msbrogli, line 22: I guess we should remove the commented lines because I guess it makes no sense to change it in the future.

hathor/event/model/event_data.py
jansegre: I think the event data and the API data should be consistent, so I'll consider this for how we serialize the API data too. This would be the time to split the representation of the data for API and for the storage (currently it's the same method).

I do prefer to not change any existing field, so we would keep `accumulated_weight: float` and `score: float`.

However we could still expose the actual raw value somehow. Maybe we can do this on a different PR and explore the options there. For example we could add `accumulated_weight_raw: str` and `score_raw: str` if a `raw=true` parameter is given, although this would only work for APIs, event data wouldn't have a simple toggle.

hathor/consensus/block_consensus.py
jansegre: It seems correct to me so I've kept it.

hathor/consensus/block_consensus.py
jansegre: Done.

hathor/event/model/event_data.py
jansegre: @msbrogli @glevco What do you think about keeping `accumulated_weight: float` and `score: float` in this PR by converting them from the internal `int` values using `work_to_weight` when serializing for APIs and event-data?

And we can discuss if it's worth to somehow expose the raw int values in a future PR.

msbrogli: I agree to keep the event data and the API data consistent. I guess we can use this PR to add the raw values too as suggested. So use cases can pick whatever is best for them.

hathor/util.py
jansegre: Done.

hathor/utils/weight.py
jansegre: Done.

hathor/wallet/resources/send_tokens.py
```diff
@@ -309,7 +309,7 @@ def render_OPTIONS(self, request):
                                             'inputs': [],
                                             'outputs': [],
                                             'tokens': [],
-                                            'accumulated_weight': 14
+                                            'accumulated_weight': '16384'
```
msbrogli, line 312: I guess we won't do it anymore, right? I mean, it will return `accumulated_weight: 14` and `accumulated_weight_raw: '16384'`, right?

hathor/wallet/resources/thin_wallet/send_tokens.py
```diff
@@ -464,7 +464,7 @@ def render_OPTIONS(self, request):
                                             'inputs': [],
                                             'outputs': [],
                                             'tokens': [],
-                                            'accumulated_weight': 14
+                                            'accumulated_weight': '16384'
```
msbrogli, line 467: Same here.

hathor/utils/weight.py
msbrogli: Thanks!

hathor/util.py
msbrogli: Thanks!

hathor/event/model/event_data.py
msbrogli: You did it! Thanks!

hathor/transaction/resources/transaction_confirmation.py
```diff
@@ -125,7 +125,7 @@ def render_GET(self, request):
                                 'success': {
                                     'summary': 'Success',
                                     'value': {
-                                        'accumulated_weight': 15.4,
+                                        'accumulated_weight': 43237,
```
glevco, line 128: Same as Marcelo mentioned above, this value is actually a string, right? And `accumulated_weight_raw`?

hathor/utils/weight.py
```diff
@@ -0,0 +1,37 @@
...
+        return 0.0
+    return math.log2(work)
+
+
+def calculate_min_significant_weight(score: int, tol: float) -> float:
+    """ This function will return the min significant weight to increase score by tol.
+
+    When most peers are updated to store work as integers for their internal score and accumulated weight metadata,
+    this function will not be needed anymore. It's only use currently is to make sure miner nodes will produce blocks
+    with weight that are high enough for outdated nodes to be able to observe the score increasing.
```
glevco, line 35: Should we use Feature Activation to remove this?

hathor/transaction/base_transaction.py
glevco, line 1: What do you think about creating a `@property work` that returns `weight_to_work(self.weight)`? This would prevent having to call `weight_to_work` all over the codebase.

glevco (approved):

hathor/manager.py
glevco: Thanks!

hathor/wallet/resources/send_tokens.py
jansegre: Fixed.

hathor/wallet/resources/thin_wallet/send_tokens.py
jansegre: Fixed.

hathor/transaction/resources/transaction_confirmation.py
jansegre: Yes, it's fixed now.

hathor/utils/weight.py
jansegre: I don't think it will be too useful. Because the peers that would "suffer any consequences" would be the peers that aren't mining. And removing the "minimum significant weight" is not technically a breaking change, because it is not a rule that must be followed, it's just the default generated weight that miners will typically use.

Also, we don't have any rush to remove it, so in the future when we observe that most peers are running `>v0.63.0` we could prepare to remove the code for the minimum significant weight.

hathor/transaction/base_transaction.py
jansegre: For now, I'd still prefer to leave it as a function. I'd rather avoid accidentally implying that they can be converted back and forth. I think we could refactor it in the future to use our own `Weight` and `Work` types and maybe when we convert we can use a sub-class like `WorkFromWeight` that keeps the original `Weight` so if it is converted back there is no math and no loss.

glevco (approved):

hathor/wallet/resources/send_tokens.py
msbrogli: Thanks!

hathor/wallet/resources/thin_wallet/send_tokens.py
msbrogli: Thanks!

msbrogli (commented):
Approved!

jansegre (approved):

jansegre:
Just as a report, last night I synced a node from scratch using this branch, it synced correctly up to the same block as the network, and this is a comparison of the score of the best block (which accumulates tiny errors in the current release):

```
# my node
❯ curl localhost:9080/v1a/transaction\?id=0000000000000000046147a92a42b4e2f50ef587ecb17775efda3628150f2c32 -s | jq .meta.score
88.71139817081715
```

```
# public node
❯ curl https://node2.mainnet.hathor.network/v1a/transaction\?id=0000000000000000046147a92a42b4e2f50ef587ecb17775efda3628150f2c32 -s | jq .meta.score
88.71139817083098
```

Even after the tiny accumulated errors the difference is currently minimal. I'm going forward with the merge.

PR #576, opened 2023-04-27
fix(version): copy the BUILD_VERSION file into the Docker image

luislhl:
### Acceptance Criteria

Our built Docker images should contain the BUILD_VERSION file, so they are able to report the injected version when running.

### How to test

1. Create a BUILD_VERSION file in your local repo:
```
cat <<EOF > BUILD_VERSION                         
1.2.3
EOF
```

2. Build the image with `docker build -t local/hathor-core .`

3. Run it with `docker run -it local/hathor-core run_node`

The first log should contain the right version.

4. Remove the BUILD_VERSION file and build and run the image again. The version should now contain the `-local` prefix.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,5 +29,8 @@ ARG PYTHON
 RUN apt-get -qy update
 RUN apt-get -qy install libssl1.1 graphviz librocksdb6.11
 COPY --from=stage-0 /app/.venv/lib/python${PYTHON}/site-packages/ /usr/local/lib/python${PYTHON}/site-packages/
+# XXX: copy optional BUILD_VERSION file using ...VERSIO[N] instead of ...VERSION* to ensure only one file will be copied
+# XXX: also copying the README.md because we need at least one existing file
+COPY README.md BUILD_VERSIO[N] /
 EXPOSE 40403 8080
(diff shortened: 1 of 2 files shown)


jansegre (approved):

msbrogli (approved):

Dockerfile
```diff
@@ -18,7 +18,8 @@ WORKDIR /app/
 COPY pyproject.toml poetry.lock  ./
 RUN poetry install -n -E sentry --no-root --only=main
 COPY hathor ./hathor
-COPY README.md ./
+# BUILD_VERSION is optional to exist
+COPY README.md BUILD_VERSIO[N] ./
```
jansegre, line 22: - Does the stage-0 need this for any thing? The next stage could copy BUILD_VERSION directly, right?
- Also the `[N]` is just to force the completer to not complete and allow it to be optional, right?

jansegre (approved):

Dockerfile
luislhl: Yes, the next stage could copy directly.

I can remove this from stage-0.

The `[N]` is just to allow it to be optional, otherwise the build would fail if the file doesn't exist.

jansegre: I fixed it.

PR #581, opened 2023-05-04
feat(sync): Limit number of sync running simultaneously

msbrogli:
The reasoning behind this change is described in https://github.com/HathorNetwork/on-call-incidents/issues/109#issuecomment-1535409697

Acceptance Criteria:

- Limit the number of syncs that can happen simultaneously
  - Ability to configure that number via `HathorSettings`
- Rotate the group of peers that are syncing every 10 minutes
- Connections should not be changed. Only whether the connected peer is syncing or not.
- Set a group of peers to always run the sync
- Create a `cmd_path`, a place to put files to communicate with the full node.
    - Use `force_sync_rotate` file to force a sync rotation.
    - Use `always_enable_sync.txt` file to set peers to always have sync enabled.
- Add a global rate limiter to the number of calls per second to `NodeSyncTimestamp.send_tips()`.

### DIFF
--- a/hathor/conf/settings.py
+++ b/hathor/conf/settings.py
@@ -373,3 +373,9 @@ class HathorSettings(NamedTuple):
     EVENT_API_DEFAULT_BATCH_SIZE: int = 100
 
     EVENT_API_MAX_BATCH_SIZE: int = 1000
+
+    # Maximum number of sync running simultaneously.
+    MAX_ENABLED_SYNC: int = 16
+
+    # Time to update the peers that are running sync.
+    SYNC_UPDATE_INTERVAL: int = 10 * 60  # seconds
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -127,6 +127,8 @@ class HathorManager:
         self._enable_sync_v1 = enable_sync_v1
         self._enable_sync_v2 = enable_sync_v2
 
+        self._cmd_path: Optional[str] = None
+
         self.log = logger.new()
 
         if rng is None:
@@ -1222,6 +1224,14 @@ class HathorManager:
                           vertex_count=vertex_count, **self.environment_info.as_dict())
 
             self.lc_check_sync_state.stop()
+
+    def set_cmd_path(self, path: str) -> None:
+        """Set the cmd path, where sysadmins can place files to communicate with the full node."""
+        self._cmd_path = path
+
+    def get_cmd_path(self) -> Optional[str]:
+        """Return the cmd path. If no cmd path is set, returns None."""
+        return self._cmd_path
 
 
 class ParentTxs(NamedTuple):
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import TYPE_CHECKING, Any, Dict, Iterable, NamedTuple, Optional, Set, Union
+import os
+from typing import TYPE_CHECKING, Any, Dict, Iterable, List, NamedTuple, Optional, Set, Tuple, Union
 
 from structlog import get_logger
 from twisted.internet import endpoints
@@ -27,6 +28,7 @@ from hathor.p2p.netfilter.factory import NetfilterFactory
 from hathor.p2p.peer_id import PeerId
 from hathor.p2p.peer_storage import PeerStorage
 from hathor.p2p.protocol import HathorProtocol
+from hathor.p2p.rate_limiter import RateLimiter
 from hathor.p2p.states.ready import ReadyState
 from hathor.p2p.sync_factory import SyncManagerFactory
 from hathor.p2p.sync_version import SyncVersion
@@ -48,6 +50,26 @@ settings = HathorSettings()
 WHITELIST_REQUEST_TIMEOUT = 45
 
 
+def parse_text(text: str) -> List[str]:
+    ret: List[str] = []
+    for line in text.splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        if line.startswith('#'):
+            continue
+        ret.append(line)
+    return ret
+
+
+class _SyncRotateInfo(NamedTuple):
+    candidates: List[str]
+    old: Set[str]
+    new: Set[str]
+    to_disable: Set[str]
+    to_enable: Set[str]
+
+
 class _ConnectingPeer(NamedTuple):
     connection_string: str
     endpoint_deferred: Deferred
@@ -63,6 +85,11 @@ class PeerConnectionsMetrics(NamedTuple):
 class ConnectionsManager:
     """ It manages all peer-to-peer connections and events related to control messages.
     """
+    MAX_ENABLED_SYNC = settings.MAX_ENABLED_SYNC
+    SYNC_UPDATE_INTERVAL = settings.SYNC_UPDATE_INTERVAL
+
+    class GlobalRateLimiter:
+        SEND_TIPS = 'NodeSyncTimestamp.send_tips'
 
     connections: Set[HathorProtocol]
     connected_peers: Dict[str, HathorProtocol]
@@ -70,6 +97,8 @@ class ConnectionsManager:
     handshaking_peers: Set[HathorProtocol]
     whitelist_only: bool
     _sync_factories: Dict[SyncVersion, SyncManagerFactory]
+
+    rate_limiter: RateLimiter
 
     def __init__(self, reactor: Reactor, my_peer: PeerId, server_factory: 'HathorServerFactory',
                  client_factory: 'HathorClientFactory', pubsub: PubSubManager, manager: 'HathorManager',
@@ -98,6 +127,10 @@ class ConnectionsManager:
 
         self.max_connections: int = settings.PEER_MAX_CONNECTIONS
 
+        # Global rate limiter for all connections.
+        self.rate_limiter = RateLimiter(self.reactor)
+        self.enable_rate_limiter()
+
         # All connections.
         self.connections = set()
 
@@ -121,6 +154,17 @@ class ConnectionsManager:
         self.lc_reconnect = LoopingCall(self.reconnect_to_all)
         self.lc_reconnect.clock = self.reactor
 
+        # A timer to update sync of all peers.
+        self.lc_sync_update = LoopingCall(self.sync_update)
+        self.lc_sync_update.clock = self.reactor
+        self.lc_sync_update_interval: float = 5  # seconds
+
+        # Peers that always have sync enabled.
+        self.always_enable_sync: Set[str] = set()
+
+        # Timestamp of the last time sync was updated.
+        self._last_sync_rotate: float = 0.
+
         # A timer to try to reconnect to the disconnect known peers.
         if settings.ENABLE_PEER_WHITELIST:
             self.wl_reconnect = LoopingCall(self.update_whitelist)
@@ -141,8 +185,21 @@ class ConnectionsManager:
         if enable_sync_v2:
             self._sync_factories[SyncVersion.V2] = SyncV1Factory(self)
 
+    def disable_rate_limiter(self) -> None:
+        """Disable global rate limiter."""
+        self.rate_limiter.unset_limit(self.GlobalRateLimiter.SEND_TIPS)
+
+    def enable_rate_limiter(self, max_hits: int = 16, window_seconds: float = 1) -> None:
+        """Enable global rate limiter. This method can be called to change the current rate limit."""
+        self.rate_limiter.set_limit(
+            self.GlobalRateLimiter.SEND_TIPS,
+            max_hits,
+            window_seconds
+        )
+
     def start(self) -> None:
         self.lc_reconnect.start(5, now=False)
+        self.lc_sync_update.start(self.lc_sync_update_interval, now=False)
         if settings.ENABLE_PEER_WHITELIST:
             self._start_whitelist_reconnect()
 
@@ -164,6 +221,9 @@ class ConnectionsManager:
     def stop(self) -> None:
         if self.lc_reconnect.running:
             self.lc_reconnect.stop()
+
+        if self.lc_sync_update.running:
+            self.lc_sync_update.stop()
 
     def _get_peers_count(self) -> PeerConnectionsMetrics:
         """Get a dict containing the count of peers in each state"""
@@ -279,6 +339,12 @@ class ConnectionsManager:
 
         # In case it was a retry, we must reset the data only here, after it gets ready
         protocol.peer.reset_retry_timestamp()
+
+        if len(self.connected_peers) <= self.MAX_ENABLED_SYNC:
+            protocol.enable_sync()
+
+        if protocol.peer.id in self.always_enable_sync:
+            protocol.enable_sync()
 
         # Notify other peers about this new peer connection.
         for conn in self.iter_ready_connections():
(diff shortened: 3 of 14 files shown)


hathor/p2p/manager.py
```diff
@@ -63,6 +63,8 @@ class PeerConnectionsMetrics(NamedTuple):
 class ConnectionsManager:
     """ It manages all peer-to-peer connections and events related to control messages.
     """
+    # Interval to update which peers are syncing.
+    SYNC_UPDATE_INTERVAL: int = 10 * 60  # in seconds
```
obiyankenobi, line 67: Why not a parameter in config?

obiyankenobi (dismissed):

obiyankenobi:
How are we planning to test this?

hathor/conf/settings.py
```diff
@@ -373,3 +373,6 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
     EVENT_API_DEFAULT_BATCH_SIZE: int = 100
 
     EVENT_API_MAX_BATCH_SIZE: int = 1000
+
+    # Maximum number of sync running simultaneously.
+    MAX_ENABLED_SYNC: int = 6
```
pedroferreira1, line 378: Is there any reasoning behind the number 6?

hathor/p2p/states/ready.py
```diff
@@ -180,3 +180,9 @@ def handle_pong(self, payload: str) -> None:
         self.ping_min_rtt = min(self.ping_min_rtt, self.ping_rtt)
         self.ping_start_time = None
         self.log.debug('rtt updated', rtt=self.ping_rtt, min_rtt=self.ping_min_rtt)
+
+    def enable_sync(self) -> None:
```
pedroferreira1, line 184: Missing docstring in both methods

hathor/p2p/node_sync.py
```diff
@@ -492,6 +495,10 @@ def next_step(self) -> Generator[Deferred, Any, None]:
             self.log.debug('already running')
             return
 
+        if not self.is_enabled:
```
pedroferreira1, line 498: Just to fully understand this change, with this `return` this node (node A) won't be trying to find the synced timestamp with the other node (node B) but if B is still in the old version, it will continue working as it works today, right? Finding the synced timestamp and we will reply the messages even with `sync_enabled = False`. The high CPU usage happens in this method (next_step), so the new nodes will be fine with CPU use, is that correct?

hathor/p2p/protocol.py
```diff
@@ -470,3 +470,19 @@ def print_stats(self, prefix: str = '') -> None:
             100.0 * self.discarded_txs / (self.received_txs + self.discarded_txs)
         ))
         print('----')
+
+    def is_sync_enabled(self) -> bool:
```
pedroferreira1, line 474: Missing docstrings

hathor/p2p/manager.py
```diff
@@ -560,3 +573,24 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
         protocol = self.connected_peers.get(peer_id)
         if protocol:
             self.drop_connection(protocol)
+
+    def sync_update(self) -> None:
```
pedroferreira1, line 577: Missing docstring

hathor/p2p/manager.py
```diff
@@ -143,6 +149,7 @@ def __init__(self, reactor: Reactor, my_peer: PeerId, server_factory: 'HathorSer
 
     def start(self) -> None:
         self.lc_reconnect.start(5, now=False)
+        self.lc_sync_update.start(self.SYNC_UPDATE_INTERVAL, now=False)
```
pedroferreira1, line 152: If all sync managers start with `sync_enabled = False`, does this means that the first 10 minutes you won't be connected to anyone?

hathor/p2p/manager.py
```diff
@@ -560,3 +573,24 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
             self.drop_connection(protocol)
+
+    def sync_update(self) -> None:
+        current_enabled: Set[str] = set(
+            conn.peer.id
+            for conn in self.connected_peers.values()
+            if conn.is_sync_enabled()
+        )
+        k = settings.MAX_ENABLED_SYNC
+        selected_peers: Set[str] = set(self.rng.choices(self.connected_peers.keys(), k=k))
```
pedroferreira1, line 584: I ran some tests with python `Random` (I know we use a modified class) but can this `choices` select the same peer more than once instead of choosing other peers?

Check this example:

```Python
In [5]: rng = Random()

In [6]: rng.choices([1,2,3], k=4)
Out[6]: [2, 1, 2, 2]
```

It selected the number `2` three times and number `3` wasn't selected. Not sure if something is different in my example

msbrogli: Great catch! I'll change the sample method.

hathor/p2p/states/ready.py
msbrogli: Fixed! Thanks!

hathor/p2p/node_sync.py
msbrogli: Your understand of the impact of this line is correct. But you're right that the other node might run the sync algorithm and it might have a high impact on the CPU. I'll fix it in a separate commit.

hathor/p2p/manager.py
msbrogli: We enable the sync for all `MAX_ENABLED_SYNC` first connections.

hathor/conf/settings.py
```diff
@@ -373,3 +373,9 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
     EVENT_API_DEFAULT_BATCH_SIZE: int = 100
 
     EVENT_API_MAX_BATCH_SIZE: int = 1000
+
+    # Maximum number of sync running simultaneously.
+    MAX_ENABLED_SYNC: int = 8
+
+    # Peers to always enable sync.
+    ALWAYS_ENABLE_SYNC: Set[str] = set()
```
glevco, line 381: Are we going to set peer IDs here? Shouldn't we use an approach like the whitelist, downloading it from an URL?

hathor/p2p/manager.py
jansegre: I agree that would be good. I'm testing a node with a custom interval time and I would prefer if I could do that without carrying a change in code (which would make it harder to do that on a docker image, for example).

msbrogli: I'll do this right away.

hathor/p2p/manager.py
msbrogli: Fixed!

hathor/conf/settings.py
msbrogli: You're right. Let me change this part.

hathor/p2p/manager.py
msbrogli: Done.

hathor/conf/settings.py
msbrogli: I changed it to 8 earlier today. It is inspired by Bitcoin's default config.

hathor/p2p/protocol.py
msbrogli: Fixed. Thanks!

jansegre (approved):

hathor/p2p/manager.py
```diff
@@ -560,3 +596,114 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
+        self.log.info('update always_enable_sync', new=new, to_enable=to_enable, to_disable=to_disable)
+
+        for peer_id in new:
+            if peer_id not in self.connected_peers:
+                continue
+            self.connected_peers[peer_id].enable_sync()
+
+        for peer_id in to_disable:
+            if peer_id not in self.connected_peers:
+                continue
```
glevco, line 652: Maybe add a warning log in the `continue`? It's not an error, but probably indicates a human mistake when setting the cmd file

tests/p2p/test_sync_rate_limiter.py
```diff
@@ -0,0 +1,54 @@
...
+        for _ in range(8):
+            sync2.send_tips()
+            self.assertTrue(sync2._send_tips.called)
+            sync2._send_tips.reset_mock()
+
+        sync2.send_tips()
+        self.assertFalse(sync2._send_tips.called)
+
+        self.simulator._clock.advance(2000)
+        self.assertTrue(sync2._send_tips.called)
```
glevco, line 41: It's better to assert `call_count` instead of `called`, so you don't get false successes. By changing the range to `100`, we can see that `send_tips` is actually called 100 times, catching a bug:

```suggestion
        for _ in range(100):
            sync2.send_tips()

        self.assertEqual(sync2._send_tips.call_count, 8)

        sync2.send_tips()
        self.assertEqual(sync2._send_tips.call_count, 8)

        self.simulator._clock.advance(2000)
        self.assertEqual(sync2._send_tips.call_count, 9)
```

glevco (changes requested):

hathor/p2p/node_sync.py
```diff
@@ -223,6 +228,10 @@ def __init__(self, protocol: 'HathorProtocol', downloader: Downloader, reactor:
         # Indicate whether the sync manager has been started.
         self._started: bool = False
 
+        # Indicate whether the synchronization is enabled.
+        # When the sync is disabled, it will keep the last synced_timestamp.
```
pedroferreira1, line 232: How will the explorer handle this? As if they are out of sync? Are we going to trigger the alerts we have nowadays for important nodes out of sync?

hathor/p2p/manager.py
```diff
@@ -560,3 +596,114 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
+        self._last_sync_rotate = now
+
+        current_enabled: Set[str] = set()
+        for peer_id, conn in self.connected_peers.items():
+            if conn.is_sync_enabled():
+                current_enabled.add(peer_id)
+
+        candidates = list(self.connected_peers.keys())
+        self.rng.shuffle(candidates)
+        selected_peers: Set[str] = set(candidates[:self.MAX_ENABLED_SYNC])
```
pedroferreira1, line 688: With the current code we might select here only peers in the `always_enable_sync` list, is this a problem somehow? I don't see any bugs or problems that might happen because of this.

hathor/p2p/manager.py
```diff
@@ -560,3 +596,114 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
+                line = line.strip()
+                if not line:
+                    continue
+                if line.startswith('#'):
+                    continue
+                new.add(line)
+
+        old = self.always_enable_sync
+        if new == old:
+            return
```
pedroferreira1, line 638: The whitelist parse has a code similar to this, right? Does it make sense to refactor it?

pedroferreira1 (dismissed):

hathor/p2p/node_sync.py
luislhl: I think we will going to trigger alerts for nodes out of sync, because the alerts are based on the `synced_timestamp` property of this class, which will not be updated anymore when the sync is disabled.

This can be seen in https://github.com/HathorNetwork/hathor-p2p-monitor/blob/main/hathor_p2p_monitor/p2p_monitor_important_peers/p2p_monitor_peer_id.py#L91

Maybe we will need a patch there to check the new `is_enabled` property on the list of connected_peers, and don't alert for the ones which have the sync disabled.

hathor/manager.py
```diff
@@ -1223,6 +1225,14 @@ def check_sync_state(self):
 
             self.lc_check_sync_state.stop()
 
+    def set_cmd_path(self, path: str) -> None:
+        """Set the cmd path, where sysadmins can place files to communicate with the full node."""
```
luislhl, line 1229: I think we will need to update the hathor-core SOP with this information.

hathor/p2p/node_sync.py
```diff
@@ -586,7 +605,17 @@ def handle_get_tips(self, payload: str) -> None:
             args = GetTipsPayload(**data)
             self.send_tips(args.timestamp, args.include_hashes, args.offset)
 
-    def send_tips(self, timestamp: Optional[int] = None, include_hashes: bool = False, offset: int = 0) -> None:
+    def send_tips(self, timestamp: Optional[int] = None, include_hashes: bool = False, offset: int = 0,
+                  *, rate_limiter_weight: int = 1) -> None:
+        """Try to send a TIPS message. If rate limit has been reached, it schedules to send it later."""
+        if not self.global_rate_limiter.add_hit('NodeSyncTimestamp.send_tips', weight=rate_limiter_weight):
```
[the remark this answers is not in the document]
msbrogli: Done! Thanks!

hathor/p2p/manager.py
msbrogli: Done! Thanks!

hathor/p2p/manager.py
msbrogli: Closing this conversation. Thanks!

hathor/p2p/manager.py
msbrogli: Fixed! Thanks!

hathor/conf/settings.py
msbrogli: Done! Thanks!

hathor/p2p/manager.py
```diff
@@ -560,3 +596,114 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
+            self._sync_update_cmds()
+        except Exception:
+            self.log.error('_sync_update_cmds failed', exc_info=True)
+
+        try:
+            self._sync_rotate_if_needed()
+        except Exception:
+            self.log.error('_sync_rotate_if_needed failed', exc_info=True)
+
+    def _sync_update_cmds(self) -> None:
```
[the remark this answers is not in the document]
msbrogli: I'll do this refactor and more here: https://github.com/HathorNetwork/hathor-core/pull/587

hathor/p2p/manager.py
msbrogli: Not really. We might add a list of predefined peer-ids where we are not connected to all of them. One can easily check the status of a peer using the HTTP API `/v1a/status`.

hathor/p2p/manager.py
```diff
@@ -560,3 +596,114 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
+            new_sample=selected_peers,
+            to_disable=to_disable,
+            to_enable=to_enable
+        )
+
+        for peer_id in to_disable:
+            self.connected_peers[peer_id].disable_sync()
+
+        for peer_id in to_enable:
+            self.connected_peers[peer_id].enable_sync()
```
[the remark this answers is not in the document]
msbrogli: Done! Thanks!

hathor/p2p/node_sync.py
```diff
@@ -586,7 +605,17 @@ def handle_get_tips(self, payload: str) -> None:
...
+            self.log.debug('send_tips throttled')
+            self.reactor.callLater(1, self.send_tips, timestamp, include_hashes, offset,
+                                   rate_limiter_weight=0)
```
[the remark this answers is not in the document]
msbrogli: You're right. Fixed! Nice catch! Thanks!

tests/p2p/test_sync_rate_limiter.py
msbrogli: Much better. Done! Thanks!

hathor/manager.py
msbrogli: @luislhl You are right. And it will change considerably more after https://github.com/HathorNetwork/hathor-core/pull/587.

hathor/p2p/node_sync.py
msbrogli: Should we add them to the `always_enable_sync` list?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -34,6 +35,7 @@
 from hathor.util import not_none
 
 settings = HathorSettings()
+cpu = get_cpu_profiler()
```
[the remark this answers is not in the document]
msbrogli: It's being used to add this method to the `top` profiler. I added this method because it is an important method to be profiled. Shouldn't we do it?

hathor/p2p/node_sync.py
msbrogli: Answering @pedroferreira1 's question: The explorer will show as not synced. We can use the new `is_enabled` to show that we are not syncing with that node at that time.

hathor/p2p/manager.py
msbrogli: I don't think so. The idea is that we must keep the sync enabled for a set of peers. If some of these peers are selected from the candidates, they will be enabled anyway. What do you think?

hathor/p2p/manager.py
msbrogli: The whitelist parser is specific for its needs. But I have refactored and moved this code to a separate function. Can you review it again, please?

hathor/p2p/node_sync.py
pedroferreira1: > Should we add them to the always_enable_sync list?

I don't think so, the list is not small.

I agree that we should patch both the p2p alert monitor and explorer to consider the `is_enabled` property now.

pedroferreira1 (dismissed):

glevco (approved):

glevco (dismissed):

pedroferreira1 (dismissed):

glevco (approved):

pedroferreira1 (approved):

PR #582, opened 2023-05-08
feat(events): reset event queue command

### LINKED ISSUE #580: Add reset event queue command
Can you create a task/issue for a command to reset the event queue? I think that should be required before we stabilize the event queue system, otherwise it will not be practical to restore a working database. Other than that the current implementation is OK for me.

_Originally posted by @jansegre in https://github.com/HathorNetwork/hathor-core/pull/555#discussion_r1182814067_

glevco:
### Acceptance Criteria

- Implement `--x-reset-event-queue` CLI flag to reset the event queue, allowing the start of a full node without the event feature after it was previously enabled
- Implement ability to reset all `EventStorage` data
- Fix support for merged mining block

Closes #580


glevco:
Moved to https://github.com/HathorNetwork/hathor-core/pull/608

PR #584, opened 2023-05-09
feat(start): Run the event loop during full node initialization

msbrogli:


hathor/indexes/base_index.py
```diff
@@ -29,12 +32,14 @@ class BaseIndex(ABC):
     created to generalize how we initialize indexes and keep track of which ones are up-to-date.
     """
 
-    def init_start(self, indexes_manager: 'IndexesManager') -> None:
+    @inlineCallbacks
+    def init_start(self, reactor: 'Reactor', indexes_manager: 'IndexesManager') -> Generator[Any, Any, None]:
```
jansegre, line 36: This method doesn't really need to use `@inlineCallback`, it can simply indicate that is an "async" method by using `-> Deferred` and the implementations can choose to use `@inlineCallback` to simplify making a function that returns a `Deferred`. Also it can always return `defer.succeed(None)` to have a similar effect and maintain the interface of returning a `Deferred`.

hathor/indexes/memory_info_index.py
```diff
@@ -28,8 +31,11 @@ def __init__(self):
         self._first_timestamp = 0
         self._latest_timestamp = 0
 
-    def init_start(self, indexes_manager: 'IndexesManager') -> None:
+    @inlineCallbacks
+    def init_start(self, reactor: 'Reactor', indexes_manager: 'IndexesManager') -> Generator[Any, Any, None]:
```
jansegre, line 35: Same here, it doesn't need to use `@inlineCallbacks`.

hathor/indexes/rocksdb_info_index.py
```diff
@@ -42,10 +44,13 @@ def __init__(self, db: 'rocksdb.DB', *, cf_name: Optional[bytes] = None) -> None
         RocksDBIndexUtils.__init__(self, db, cf_name or _CF_NAME_ADDRESS_INDEX)
         MemoryInfoIndex.__init__(self)
 
-    def init_start(self, indexes_manager: 'IndexesManager') -> None:
+    @inlineCallbacks
+    def init_start(self, reactor: 'Reactor', indexes_manager: 'IndexesManager') -> Generator[Any, Any, None]:
```
jansegre, line 48: Same here.

jansegre (dismissed):

hathor/indexes/base_index.py
msbrogli: Great tip. Thanks! Done!

hathor/indexes/memory_info_index.py
msbrogli: Done! Thanks!

jansegre (approved):

glevco:
@msbrogli just to make it easier for you, I think you'll probably need to change the following lines in this PR:

- https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1261437771
- https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1261439394

PR #587, opened 2023-05-10
feat(core): Add a sysctl subsystem to modify full node parameters at runtime

msbrogli:
### `sysctl` subsystem

The goal of the sysctl subsystem is to allow sysadmins to modify full node parameters at runtime. This PR provides the initial implementation for the sysctl subsystem and also add few parameters from the `hathor.p2p.manager.ConnectionsManager`. More parameters will be added in the future.

Each parameter optionally has a getter and a setter. The getter is used to read the state of the parameter, while the setter is used to modify the state of the parameter. The value provided to the setter is always the same as the value returned by the getter.

Parameters are described using a dotted set of components, such as `p2p.max_enabled_sync` and `p2p.sync_update_interval`.

### Example

Note the syntax `[name]` to get the value and `[name]=[value]` to set the value. In this case, `p2p.force_sync_rotate=` does not expect arguments because it just forces rotating the enabled sync.

```
❯ nc -U sysctl.sock
p2p.max_enabled_sync
8
p2p.max_enabled_sync=10
p2p.max_enabled_sync
10
p2p.force_sync_rotate=
p2p.rate_limit.global.send_tips
8, 2
p2p.rate_limit.global.send_tips=20, 4
p2p.rate_limit.global.send_tips
20, 4
```

### Acceptance criteria

1. Allow sysctl endpoint to support either `unix:/path/sysctl.sock` or `tcp:5000`.
2. Add a `--sysctl=[description]` to the `run_node` cli.
3. Add a builder for the sysctl tree.
4. Add a simple protocol to run getters and setters from the sysctl tree.
5. Replace the approach from `hathor.p2p.manager.ConnectionsManager` to modify parameters at runtime using files.


hathor/sysctl/protocol.py
```diff
@@ -0,0 +1,111 @@
...
+    def _deserialize(self, value_str: str) -> Any:
+        """Deserialize a value sent by the client."""
+        if len(value_str) == 0:
+            return ()
+        if value_str[0] == '(':
+            if value_str[-1] != ')':
+                raise ValueError('not closing (')
+            parts = [x.strip() for x in value_str[1:-1].split(',')]
+            return tuple(json.loads(x) for x in parts)
+        return json.loads(value_str)
```
glevco, line 111: What if an user sends `my.values=1, 2, 3)`? Maybe you could just remove support for parenthesis, it would be simpler to implement and similar to use:

```
p2p.rate_limit.global.send_tips=20, 4
p2p.rate_limit.global.send_tips
20, 4
```

hathor/sysctl/p2p/manager.py
```diff
@@ -0,0 +1,143 @@
...
+        self.register(
+            'always_enable_sync',
+            self.get_always_enable_sync,
+            self.set_always_enable_sync,
+        )
+        self.register(
+            'always_enable_sync.readtxt',
+            None,
+            self.set_always_enable_sync_readtxt,
+        )
```
glevco, line 70: In the end this is very similar to just overriding `Sysctl`'s attributes, but in an imperative and not declarative way. You could do something like:

```python
class ConnectionsManagerSysctl(Sysctl):
    commands: Dict[str, Tuple[Getter, Setter] = {
        'max_enabled_sync': (get_max_enabled_sync, set_max_enabled_sync),
        'rate_limit.global.send_tips': (get_global_send_tips_rate_limit, set_global_send_tips_rate_limit)
        ...
    }
```

You could also combine this with the `SysctlCommand` named tuple suggested above.

glevco (approved):

hathor/sysctl/sysctl.py
```diff
@@ -0,0 +1,103 @@
...
+        self._setters: Dict[str, Setter] = {}
+        self._paths: Set[str] = set()
+
+    def put_child(self, path: str, sysctl: 'Sysctl') -> None:
+        """Add a child to the tree."""
+        assert path not in self._children
+        assert path not in self._paths
+        self._children[path] = sysctl
+
+    def register(self, path: str, getter: Optional[Getter], setter: Optional[Setter]) -> None:
```
pedroferreira1, line 43: I might have misunderstood how the `children` attribute works but you have in the manager this code:

```python
self.register('rate_limit.global.send_tips', ...)
```

Shouldn't you handle this path registration with dots in this method, in order to create the children for each partition?

hathor/sysctl/sysctl.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class Sysctl:
+    """A node in the sysctl tree."""
+
+    def __init__(self) -> None:
+        self._children: Dict[str, 'Sysctl'] = {}
+        self._getters: Dict[str, Getter] = {}
+        self._setters: Dict[str, Setter] = {}
+        self._paths: Set[str] = set()
```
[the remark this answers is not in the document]
msbrogli: Even the code got better after the refactor. Thanks! Done here: https://github.com/HathorNetwork/hathor-core/pull/587/commits/d411a5d5b095a8c7e6d41a8f2a9019852bf3e09f

hathor/sysctl/p2p/manager.py
```diff
@@ -0,0 +1,143 @@
...
+
+    #############
+
+    def get_max_enabled_sync(self) -> int:
+        """Return the maximum number of peers running sync simultaneously."""
+        return self.connections.MAX_ENABLED_SYNC
+
+    def set_max_enabled_sync(self, value: int) -> None:
+        """Change the maximum number of peers running sync simultaneously."""
+        if value < 0:
```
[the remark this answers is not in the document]
msbrogli: It would only connect to the always enabled peers.

hathor/sysctl/p2p/manager.py
```diff
@@ -0,0 +1,143 @@
...
+
+    #############
+
+    def get_lc_sync_update_interval(self) -> float:
+        """Return the interval to rotate sync (in seconds)."""
+        return self.connections.lc_sync_update_interval
+
+    def set_lc_sync_update_interval(self, value: float) -> None:
+        """Change the interval to rotate sync (in seconds)."""
+        if value < 0:
```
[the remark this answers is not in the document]
msbrogli: It does. Fixed! And I also added other sanitizers. Done here: https://github.com/HathorNetwork/hathor-core/pull/587/commits/a7d22235a62e2de889a3b1765e011f2bc0341453

hathor/cli/run_node.py
```diff
@@ -45,6 +45,7 @@ def create_parser(cls) -> ArgumentParser:
                             help='Reduces tx weight to 1 for testing purposes')
         parser.add_argument('--dns', action='append', help='Seed DNS')
         parser.add_argument('--peer', help='json file with peer info')
+        parser.add_argument('--sysctl', help='socket file path')
```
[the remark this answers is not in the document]
msbrogli: The helper was outdated. Fixed! https://github.com/HathorNetwork/hathor-core/pull/587/commits/a7d22235a62e2de889a3b1765e011f2bc0341453

hathor/builder/cli_builder.py
```diff
@@ -98,7 +98,7 @@ def create_manager(self, reactor: PosixReactorBase, args: Namespace) -> HathorMa
 
         tx_storage: TransactionStorage
         event_storage: EventStorage
-        rocksdb_storage: RocksDBStorage
+        self.rocksdb_storage: RocksDBStorage
```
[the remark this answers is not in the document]
msbrogli: Fixed! Thanks! https://github.com/HathorNetwork/hathor-core/pull/587/commits/a7d22235a62e2de889a3b1765e011f2bc0341453

hathor/sysctl/protocol.py
msbrogli: I agree. Done here: https://github.com/HathorNetwork/hathor-core/pull/587/commits/a7d22235a62e2de889a3b1765e011f2bc0341453

hathor/sysctl/p2p/manager.py
msbrogli: I tested a few different syntaxes but I feel the current one is the easiest to read. I kept the `register()` method because we might refactor it to something else in the near future.

hathor/p2p/manager.py
```diff
@@ -633,109 +620,16 @@ def drop_connection_by_peer_id(self, peer_id: str) -> None:
...
-            return
-        self.log.warn(f'{key} changed', old=limit, new=(max_hits, window_seconds))
-        if window_seconds == 0:
-            self.disable_rate_limiter()
-        else:
-            self.enable_rate_limiter(max_hits, window_seconds)
-
-    def _sync_update_cmd_always_enable_sync(self) -> None:
+    def set_always_enable_sync(self, values: List[str]) -> None:
         """`always_enable_sync.txt` should contain a list of peer ids that will always have sync enabled.
```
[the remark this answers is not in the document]
msbrogli: Nice catch! Thanks! Fixed! https://github.com/HathorNetwork/hathor-core/pull/587/commits/1df86bb27a2e449f71cf74802b034f15482194f2

hathor/sysctl/p2p/manager.py
glevco: Ok, I'm fine going how it is today

glevco (approved):

hathor/sysctl/sysctl.py
msbrogli: I don't think it's worth it. I left to the coder to decide whether it's better to add a branch in the tree or just add the remaining path to the current node. As an example, in the p2p part, we have the commands `always_enable_sync` and `always_enable_sync.readtxt`, both in the same node. What do you think of this design?

pedroferreira1: I think it's good

pedroferreira1 (dismissed):

pedroferreira1 (approved):

glevco (approved):

PR #593, opened 2023-05-10
refactor: move hathor settings to yaml configuration files [part 1/2]

glevco:
### Acceptance Criteria

- Install `pyyaml` dependency (the go to, canonical yaml implementation for python, [repo](https://github.com/yaml/pyyaml))
- Add utility methods to `HathorSettings` named tuple to create it from a dict or a yaml file, leveraging pydantic for validation
- Create `yaml` files to substitute python config files for all network settings (`mainnet`, `testnet`, and `unitttests`).
  - The python configuration files are removed in a separate PR: https://github.com/HathorNetwork/hathor-core/pull/597
- Update `get_settings.HathorSettings()` to get settings from yaml files instead of python modules.
  - First, it tries to get a python module settings, and logs a deprecation warning if it's found. Then, it gets the yaml
  - This change should be backwards compatible with our current code. A breaking change removing support for settings via python modules will be introduced in a future release.
- Implement tests to make sure the `yaml` versions of settings are equal to their python versions, during the migration.


tests/others/test_hathor_settings.py
```diff
@@ -0,0 +1,88 @@
...
+        BLOCK_DIFFICULTY_N_BLOCKS=20,
+    )
+
+
+# TODO: Tests below are temporary while settings via python coexist with settings via yaml, just to make sure
+#  the conversion was made correctly. After python settings are removed, this file can be removed too.
+
+
+def test_mainnet_settings_migration():
+    assert MAINNET_SETTINGS == HathorSettings.from_yaml(filepath=MAINNET_SETTINGS_FILEPATH)
```
msbrogli, line 80: How does this `==` go? I'm asking because the left side has a different type then the right side, right?

msbrogli (approved):

hathor/conf/settings.py
```diff
@@ -14,17 +14,20 @@
...
+from hathor.utils import pydantic
 
 DECIMAL_PLACES = 2
 
 GENESIS_TOKEN_UNITS = 1 * (10**9)  # 1B
 GENESIS_TOKENS = GENESIS_TOKEN_UNITS * (10**DECIMAL_PLACES)  # 100B
 
 
-class HathorSettings(NamedTuple):
+class HathorSettings(pydantic.BaseModel):
```
msbrogli, line 30: What if the yaml file does have a required attribute? What if it has a typo that leads to an unknown attribute?

msbrogli, line 30: Is it immutable? Should we add a test to make sure?

msbrogli (changes requested):

tests/others/test_hathor_settings.py
glevco: Changed as explained in https://github.com/HathorNetwork/hathor-core/pull/593#issuecomment-1546102674

hathor/conf/settings.py
glevco: Changed as explained in https://github.com/HathorNetwork/hathor-core/pull/593#issuecomment-1546102674

glevco: Changed as explained in https://github.com/HathorNetwork/hathor-core/pull/593#issuecomment-1546102674

hathor/utils/yaml.py
```diff
@@ -0,0 +1,23 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import Dict
+
+import yaml
+
+
+def dict_from(*, filepath: str) -> Dict:
```
msbrogli, line 20: Maybe `Dict[str, Any]`?

msbrogli (approved):

hathor/conf/settings.py
msbrogli: Great! Thanks! Should we add a test with an invalid yaml to confirm the validation is raising an exception?

tests/others/test_hathor_settings.py
msbrogli: Now we are comparing two named tuples. Good job! Thanks!

hathor/conf/settings.py
msbrogli: Great! Thanks!

glevco: Indeed, I'll add it

hathor/utils/yaml.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/593/commits/48715e3d8a37dcb59ff1cbb2c3ce848906b8c42a

hathor/conf/settings.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/593/commits/48715e3d8a37dcb59ff1cbb2c3ce848906b8c42a

hathor/conf/testnet.yml
```diff
@@ -0,0 +1,37 @@
...
+  700_000: 0000008fda01c9e5fd6f99a5461e6dbf1039cba38cc8d0fc738a097d71caa968
+  800_000: 000000397af32fcc4eeb6985d96326c1ff4644792631872a00394688b1782af5
+  900_000: 00000097ae405036614f4335ad0e631df8fc5f7434e82c3421627e2fea4e1830
+  1_000_000: 000000145ba662cdee0d72034658f93a0a3a4568d5ba5083ff09013ca1e6556c
+  1_100_000: 000000404e6ff6a23695a6ffe712ce1c4efc02e75bbc11c3129f4c2377b07743
+  1_200_000: 0000003be5fae5bb2c9ceaed589d172bcd9e74ca6c8d7d2ca06567f65cea7c9b
+  1_300_000: 0000000000007d39de6e781c377bc202213b0b5b60db14c13d0b16e06d6fd5ac
+  1_400_000: 000000000df9cb786c68a643a52a67c22ab54e8b8e41cbe9b761133f6c8abbfe
+  1_500_000: 000000000c3591805f4748480b59ac1788f754fc004930985a487580e2b5de8f
+  1_600_000: 00000000060adfdfd7d488d4d510b5779cf35a3c50df7bcff941fbb6957be4d2
```
jansegre, line 37: Just to confirm if my understanding is correct. On the YAML side the `_` character will make the key be a string (or maybe even keys are always strings), but because of pydantic (not even `_parse_checkpoints`, since it already expects a `Dict[int, str]`) the keys will somehow become `int`s. Can we double check that?

jansegre (changes requested):
I like the proposed new format. But for the sake of minimizing migration issues, I would rather have this PR still try to load a settings module using the old format and emit a deprecation warning.

Also I'd rather that we keep `HATHOR_CONFIG_FILE` _only_ for the old format (even if it is not a good name for an env var that points to a module, not a file) and use a new name for the new config, some suggestions on the new name:

- `HATHOR_CONFIG`
- `HATHOR_CONFIG_YAML`
- `HATHOR_CONFIG_PATH`

This way there is no immediate breakage and we can have a transition release.

Even if we decide that a breakage is acceptable, I would rather still have a new name for the env var and exit with an explanation if the old var is found, that way there wouldn't be any silent (or hard to debug, like a parsing error) breakage.

Another point is that most of the time I customize the settings locally, and also when we customize it for our deploys, we inherit a specific config (mainnet/testnet) and alter only a few properties. This is particularly convenient for when there are new parameters, such that the customized settings don't need any change. I think we should consider supporting some settings inherit mechanism before deprecating the old settings (it doesn't have to be on this PR).

What do you think?

glevco:
Good points @jansegre , I'll work on them

hathor/conf/testnet.yml
glevco: Actually the `_` padding in numbers is part of the [YAML specification](https://yaml.org/type/int.html), and is not related to python (or at least, it shouldn't be).

The `pyyaml` lib we use for yaml parsing implements that, so it's not something done by pydantic. It works for numbers both in keys and values, in any part of the yaml file.

hathor/conf/settings.py

hathor/utils/yaml.py
msbrogli: Thanks!

hathor/conf/testnet.yml
jansegre: Ooh, got it, thank you.

hathor/conf/get_settings.py
```diff
@@ -12,38 +12,29 @@
...
-        Get the file from environment variable 'HATHOR_CONFIG_FILE'
-        If not set we return the config file of the mainnet
-    """
-    settings_module = get_settings_module()
-    settings = getattr(settings_module, 'SETTINGS')
-    assert isinstance(settings, Settings)
-    return settings
-
-
-def get_settings_module() -> ModuleType:
```
jansegre, line 36: A simple way to support the old settings scheme would be to keep this function, but make it return an `Optional[ModuleType]` so that `HathorSettings()` will try that before trying `get_settings_filepath()`. That way compatibility is not broken. That may also be where we log a warning that this mechanism is deprecated.

glevco: Indeed, I followed your suggestion. I just had to change `get_settings_module()` so it wouldn't get the `hathor.conf.mainnet` module by default if `HATHOR_CONFIG_FILE` is not set. Instead, it returns `None` and it's `get_settings_filepath()`'s job to return `hathor/conf/mainnet.yml` as a default.

By doing this, we can still remove all python module files from our code, but keep compatibility for any third party using the old mechanism.

Done in https://github.com/HathorNetwork/hathor-core/pull/593/commits/e0c35ad26929f965144d4485c029e646cc6657b0

glevco:
@jansegre 

> I like the proposed new format. But for the sake of minimizing migration issues, I would rather have this PR still try to load a settings module using the old format and emit a deprecation warning.

Makes sense, done in https://github.com/HathorNetwork/hathor-core/pull/593/commits/e0c35ad26929f965144d4485c029e646cc6657b0. I used `HATHOR_CONFIG_YAML` as you suggested, and also added an equivalent CLI option for it.

> Another point is that most of the time I customize the settings locally, and also when we customize it for our deploys, we inherit a specific config (mainnet/testnet) and alter only a few properties. This is particularly convenient for when there are new parameters, such that the customized settings don't need any change. I think we should consider supporting some settings inherit mechanism before deprecating the old settings (it doesn't have to be on this PR).

I created an issue for that in https://github.com/HathorNetwork/hathor-core/issues/614. I left it in the To Do column of our project I'll do it as soon as I close some of the current tasks.

hathor/conf/get_settings.py
jansegre: That's great.

jansegre (dismissed):

hathor/conf/get_settings.py
msbrogli: I guess we can't remove because most custom settings inherit from mainnet. What do you think?

glevco: Indeed, I talked with Jan about this in the part 2 PR: https://github.com/HathorNetwork/hathor-core/pull/597#issuecomment-1551961013

I guess this doesn't prevent this PR, part 1, from being merged, right? We just cannot merge part 2

msbrogli: That's right. As we have the tests to confirm the yaml files are the same as the python modules, we are good.

msbrogli (dismissed):

jansegre (approved):

msbrogli (approved):

PR #597, opened 2023-05-11
refactor: move hathor settings to yaml configuration files [part 2/2]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/593

### Acceptance Criteria

- Remove unused python configuration files


msbrogli (approved):

jansegre (commented):
The changes seem correct. But I would prefer to either hold them off for `0.55.0` (instead of `0.54.0` which is on track to have YAML settings support), or have a better mechanism for steering people to the correct way to use settings if they are using the old mechanism ([mentioned here](https://github.com/HathorNetwork/hathor-core/pull/593#pullrequestreview-1427014910)).

glevco:
@jansegre 

> The changes seem correct. But I would prefer to either hold them off for `0.55.0` (instead of `0.54.0` which is on track to have YAML settings support), or have a better mechanism for steering people to the correct way to use settings if they are using the old mechanism ([mentioned here](https://github.com/HathorNetwork/hathor-core/pull/593#pullrequestreview-1427014910)).

Backwards compatibility has been added as explained in https://github.com/HathorNetwork/hathor-core/pull/593#issuecomment-1551952259. But if people are inheriting from our python modules, I agree we should not delete them in a transitioning release. What do you think should be done with this PR? Should I close it and reopen when it's time, or do we just leave it here, waiting?

glevco:
I'm closing this PR and moving it to the Parking Lot. We'll release the backwards compatible version in #593 , and eventually reopen this one when we are sure the python modules can be removed.

PR #598, opened 2023-05-11
chore: add yamllint to all files

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/593

### Acceptance Criteria

- Configure `yamllint` to lint all yaml files in the project, and update files accordingly


Makefile
```diff
@@ -76,7 +76,7 @@ isort-check:
 
 .PHONY: yamllint
 yamllint:
```
msbrogli, line 78: Does it have a command to fix in case the linter fails? Maybe add another command here so it fixes all yaml files automatically.

msbrogli (changes requested):

Makefile
glevco: Unfortunately [it doesn't support](https://github.com/adrienverge/yamllint/issues/62#issuecomment-322393129) auto fix :/

We could install another tool to do this, but I'm not sure it's worth it. We rarely change the yml files anyway

.codecov.yml
```diff
@@ -7,7 +7,7 @@ coverage:
   # https://docs.codecov.io/docs/commit-status
   status:
     # TODO: re-enable patch in the future
-    patch: off
+    patch: false
```
luislhl, line 10: Is `false` really accepted by Codecov?

I see their docs use `off` as well: https://docs.codecov.com/docs/commit-status#disabling-a-status

glevco: It should, as `off` is part of the [YAML specification](https://yaml.org/type/bool.html) and should be interpreted as a `false` bool, not a `'off'` string. But indeed, their documentation don't make it clear. If you think it's safer, we could leave it as `off` and add a disable-line comment for the linter in this line. What do you think?

luislhl: I guess we can try with `false`, this is not a critical config

luislhl (dismissed):

msbrogli (approved):

Makefile
msbrogli: Got it. I agree that it's so rare that it is not worth adding another tool for auto fix.

msbrogli (approved):

msbrogli (approved):

luislhl (approved):

PR #604, opened 2023-05-12
fix(p2p): Fix v1.1 not enforcing whitelist

msbrogli:
### Acceptance criteria

1) Consider `SyncVersion.v1_1` as v1.

### DIFF
--- a/hathor/p2p/sync_version.py
+++ b/hathor/p2p/sync_version.py
@@ -47,6 +47,10 @@ class SyncVersion(Enum):
         else:
             raise ValueError('value is either invalid for this enum or not implemented')
 
+    def is_v1(self) -> bool:
+        """Return True for V1 and V1_1."""
+        return self.get_priority() < 20
+
     # XXX: total_ordering decorator will implement the other methods: __le__, __gt__, and __ge__
     def __lt__(self, other):
         """Used to sort versions by considering the value on get_priority."""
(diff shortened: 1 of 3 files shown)


hathor/p2p/sync_version.py
```diff
@@ -47,6 +47,14 @@ def get_priority(self) -> int:
         else:
             raise ValueError('value is either invalid for this enum or not implemented')
 
+    def is_v1(self) -> bool:
+        """Return True for V1 and V1_1."""
+        if self is SyncVersion.V1:
+            return True
+        if self is SyncVersion.V1_1:
+            return True
+        return False
```
glevco, line 56: Maybe it's a non-issue, but if for any reason we need to release something like a sync-v1.2 before sync-v2 is out, we would need to remember to update this code. Maybe negating the condition would be safer, something like `if not sync-v2`.

glevco (approved):

hathor/p2p/sync_version.py
msbrogli: Good point. I just changed it to use `get_priority()`. What do you think?

pedroferreira1 (approved):

glevco (approved):

hathor/p2p/sync_version.py
glevco: Great, that priority method is perfect for this. Thanks!

PR #608, opened 2023-05-13
feat(events): reset event queue command

### LINKED ISSUE #580: Add reset event queue command

glevco:
### Acceptance Criteria

- Implement `--x-reset-event-queue` CLI flag to reset the event queue, allowing the start of a full node without the event feature after it was previously enabled
- Implement ability to reset all `EventStorage` data
- Fix support for merged mining block

Closes #580


hathor/builder/cli_builder.py
```diff
@@ -163,6 +164,19 @@ def create_manager(self, reactor: PosixReactorBase, args: Namespace) -> HathorMa
         soft_voided_tx_ids = set(settings.SOFT_VOIDED_TX_IDS)
         consensus_algorithm = ConsensusAlgorithm(soft_voided_tx_ids, pubsub=pubsub)
 
+        if args.x_enable_event_queue:
+            if not settings.ENABLE_EVENT_QUEUE_FEATURE:
+                self.log.error('The event queue feature is not available yet')
+                sys.exit(-1)
+
+            self.log.info('--x-enable-event-queue flag provided. '
+                          'The events detected by the full node will be stored and can be retrieved by clients')
```
msbrogli, line 173: Should we say that "previous events were cleared out" when `args.x_reset_event_queue is True`?

hathor/cli/run_node.py
```diff
@@ -100,6 +100,7 @@ def create_parser(cls) -> ArgumentParser:
         parser.add_argument('--x-localhost-only', action='store_true', help='Only connect to peers on localhost')
         parser.add_argument('--x-rocksdb-indexes', action='store_true', help=SUPPRESS)
         parser.add_argument('--x-enable-event-queue', action='store_true', help='Enable event queue mechanism')
+        parser.add_argument('--x-reset-event-queue', action='store_true', help='Reset the event queue')
```
msbrogli, line 103: I'm not sure I like doing it this way. It seems too easy to forget this flag in the `Dockerfile` and wipe events out by mistake. Maybe a cli command (e.g., `./hathor-cli reset-event-queue`) is safer. What do you think?

hathor/event/model/EventQueueOptions.py
```diff
@@ -0,0 +1,20 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from hathor.utils.pydantic import BaseModel
+
+
+class EventQueueOptions(BaseModel):
```
msbrogli, line 18: Missing documentation for each parameter.

hathor/event/storage/event_storage.py
```diff
@@ -46,8 +46,13 @@ def iter_from_event(self, key: int) -> Iterator[BaseEvent]:
...
     @abstractmethod
-    def clear_events(self) -> None:
-        """Clear all stored events and related metadata."""
+    def reset_events(self) -> None:
+        """Reset stored events and related metadata."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def reset_all(self) -> None:
+        """Reset all data."""
```
msbrogli, line 55: According to the memory storage implementation, this would also disable the event queue. Is that right? If yes, add it to the docstring.

msbrogli, line 55: What's the difference between `reset_events()` and `reset_all()`? Maybe add some notes to each docstring such as "if you want to do XXX, you should use `reset_all()`".

hathor/event/storage/rocksdb_storage.py
```diff
@@ -91,7 +91,7 @@ def get_last_event(self) -> Optional[BaseEvent]:
     def get_last_group_id(self) -> Optional[int]:
         return self._last_group_id
 
-    def clear_events(self) -> None:
+    def reset_events(self) -> None:
         self._last_event = None
```
msbrogli, line 95: Should you reset these parameters in RocksDB as well?

msbrogli (changes requested):

hathor/builder/cli_builder.py
glevco: This CLI arg was removed in https://github.com/HathorNetwork/hathor-core/pull/608/commits/79f2773a773adcbef4f5d853344e51bb66cfb4d2

hathor/cli/run_node.py
glevco: Oh, i hadn't thought about that. Great suggestion. Changed in https://github.com/HathorNetwork/hathor-core/pull/608/commits/79f2773a773adcbef4f5d853344e51bb66cfb4d2

hathor/event/model/EventQueueOptions.py
glevco: Class removed in https://github.com/HathorNetwork/hathor-core/pull/608/commits/79f2773a773adcbef4f5d853344e51bb66cfb4d2

hathor/event/storage/event_storage.py
glevco: > this would also disable the event queue

What do you mean?

Anyway, I've improved the docstrings in  https://github.com/HathorNetwork/hathor-core/pull/608/commits/79f2773a773adcbef4f5d853344e51bb66cfb4d2

glevco: Docstrings improved in https://github.com/HathorNetwork/hathor-core/commit/79f2773a773adcbef4f5d853344e51bb66cfb4d2

hathor/event/storage/rocksdb_storage.py
glevco: `self._last_event` is not stored explicitly in RocksDB, it's calculated dynamically from the events column family, and then stored on memory in `self._last_event`

`self._last_group_id` is stored explicitly in RocksDB, so it's deleted from there: `self._db.delete((self._cf_meta, _KEY_LAST_GROUP_ID))`

hathor/builder/cli_builder.py
msbrogli: Ok. Thanks!

hathor/cli/run_node.py
msbrogli: Great! Thanks!

hathor/event/model/EventQueueOptions.py
msbrogli: Ok. Thanks!

hathor/event/storage/event_storage.py
msbrogli: Great! Thanks!

hathor/event/storage/rocksdb_storage.py
msbrogli: Got it. Thanks!

hathor/event/storage/event_storage.py
msbrogli: Never mind. My comment doesn't make sense for the new implementation. Thanks!

msbrogli (approved):

jansegre (approved):

PR #611, opened 2023-05-15
feat(feature-activation): implement basic structures

glevco:
### Acceptance Criteria

- Implement basic classes to be used by Feature Activation
  - `FeatureState` enum
  - `Feature` enum, validation, and tests
  - `Criteria` model, validation, and tests
- Add Feature Activation related attributes to `HathorSettings`, validation, and tests

Here's an example of how Feature Activation settings are configured in the yaml:

```yaml
# testnet.yml
...
FEATURE_ACTIVATION:
  evaluation_interval: 1000
  default_threshold: 900
  features:
    NOP_FEATURE_1:
      bit: 0
      start_height: 0
      timeout_height: 1000
      version: 0.0.0
    NOP_FEATURE_2:
      bit: 1
      start_height: 0
      timeout_height: 1000
      version: 0.0.0
```


hathor/conf/settings.py
```diff
@@ -384,6 +386,14 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
     # Time to update the peers that are running sync.
     SYNC_UPDATE_INTERVAL: int = 10 * 60  # seconds
 
+    # List of criteria for all features that participate in the feature activation process for this network,
+    # past or future, activated or not. Features should NOT be removed from this list, to preserve history.
+    FEATURE_ACTIVATION: List[Criteria] = []
+
+    # Specifies the default minimum number of blocks per evaluation interval required to activate a feature.
+    # Usually calculated from a percentage of feature_activation.constants.EVALUATION_INTERVAL.
+    DEFAULT_FEATURE_ACTIVATION_THRESHOLD: int = 36288  # 36288 = 90% of 40320
```
msbrogli, line 395: Shouldn't we let this parameter be part of the criteria?

hathor/feature_activation/constants.py
```diff
@@ -0,0 +1,20 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+# The number of blocks in the feature activation evaluation interval.
+# Equivalent to 14 days (40320 * 30 seconds = 14 days)
+EVALUATION_INTERVAL = 40320
+
+# The number of bits used in the first byte of a block's version field. The 4 left-most bits are not used.
+MAX_SIGNAL_BITS = 4
```
msbrogli, line 20: This might be different for each network. I feel we should move it to the settings.

hathor/feature_activation/exception.py
```diff
@@ -0,0 +1,20 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+class FeatureActivationException(Exception):
+    pass
+
+
+class InvalidFeaturesConfigurationException(FeatureActivationException):
```
msbrogli, line 19: I'd say it should stop the full node completely. So, I feel we should add a validation step at the beginning of the full node initialization (and maybe this exception should inherit from a `InitializationFailed` or `PreInitializationFailed` exception).

hathor/feature_activation/feature.py
```diff
@@ -0,0 +1,24 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from enum import Enum
+
+
+class Feature(Enum):
+    """
+    An enum containing all features that participate in the feature activation process, past or future,
+    activated or not, for all networks. Features should NOT be removed from this enum, to preserve history.
```
msbrogli, line 21: Maybe add that their numbers should never be changed as well.

hathor/feature_activation/model/criteria.py
```diff
@@ -0,0 +1,81 @@
...
+
+from pydantic import Field, validator
+
+from hathor import version
+from hathor.feature_activation import constants
+from hathor.feature_activation.feature import Feature
+from hathor.utils.pydantic import BaseModel
+
+
+class Criteria(BaseModel):
```
msbrogli, line 25: I like this class. But I also think that each criteria should inherit from this class. This way, they can have their own attributes and methods according to their needs.

hathor/feature_activation/feature.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+
+    NOP_FEATURE = 0
```
msbrogli, line 24: Shouldn't it have a `name` attribute? Maybe it's better than an integer.

hathor/feature_activation/model/criteria.py
```diff
@@ -0,0 +1,81 @@
...
+    timeout_height: int = Field(ge=0, multiple_of=constants.EVALUATION_INTERVAL)
+    threshold: Optional[int] = Field(ge=0, le=constants.EVALUATION_INTERVAL, default=None)
+    minimum_activation_height: int = Field(ge=0, multiple_of=constants.EVALUATION_INTERVAL, default=0)
+    activate_on_timeout: bool = False
+    version: str = Field(regex=version.BUILD_VERSION_REGEX)
+
+    @validator('name')
+    def _validate_name(cls, name: str) -> str:
+        """Validates that the name exists."""
+        valid_names = [feature.name for feature in Feature]
```
msbrogli, line 60: Couldn't find the attribute name in the `Feature` enum.

hathor/feature_activation/validation.py
```diff
@@ -0,0 +1,87 @@
...
+
+    for criteria in features:
+        intervals = intervals_by_bit[criteria.bit]
+        intervals.append((criteria.start_height, criteria.timeout_height))
+
+    return intervals_by_bit
+
+
+def _has_any_overlap(intervals: List[Tuple[int, int]]) -> bool:
+    """Takes a list of intervals and returns whether any intervals overlap.
```
msbrogli, line 65: Open intervals or closed intervals? I mean, are the boundaries included or not?

hathor/feature_activation/validation.py
```diff
@@ -0,0 +1,87 @@
...
+    >>> _has_any_overlap([(0, 10)])
+    False
+    >>> _has_any_overlap([(0, 10), (11, 20)])
+    False
+    >>> _has_any_overlap([(0, 10), (10, 20)])
+    True
+    >>> _has_any_overlap([(0, 10), (20, 30), (15, 25)])
+    True
+    """
+    sorted_intervals = sorted(intervals, key=lambda interval: interval[0])
```
msbrogli, line 78: You don't need the key here. As they are tuples, they will be sorted by interval[0] and, when they are the same, interval[1]. I feel the native comparator should be faster than your custom method.

hathor/feature_activation/validation.py
```diff
@@ -0,0 +1,87 @@
...
+        for intervals in intervals_by_bit.values()
+    ]
+
+    if any(intervals_has_overlaps):
+        raise InvalidFeaturesConfigurationException(
+            'One or more Features have the same bit configured for an overlapping interval'
+        )
+
+
+def _get_intervals_by_bit(features: List[Criteria]) -> Dict[int, List[Tuple[int, int]]]:
```
msbrogli, line 53: Should we create a `IntervalByBit` namedtuple as below? In this case, we would be able to add the feature and improve the error message in case of overlap.

```python
class IntervalByBit(NamedTuple):
    begin: int
    end: int
    feature: Feature
```

msbrogli (changes requested):

hathor/conf/settings.py
glevco: It is! The threshold is configurable per-feature in the Criteria class: `threshold: Optional[int]`

However, it's optional, and we do provide a default value for it. That settings attribute is the default value. Since the default value is different per network, I cannot assign it at class definition, but rather get it in runtime. If `criteria.threshold` is `None`, than I'll get `settings.DEFAULT_FEATURE_ACTIVATION_THRESHOLD`

hathor/feature_activation/constants.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/611/commits/a6acaa337a02a51e89436c97723bb5ff95e66a62

hathor/feature_activation/model/criteria.py
glevco: As discussed in meeting, the `Criteria` class will only contain Feature Activation parameters and will not be subclassed

hathor/feature_activation/feature.py
glevco: I changed it to a string enum in https://github.com/HathorNetwork/hathor-core/pull/611/commits/85ba2615408565af9abb032ac41992c4531da6ec

hathor/feature_activation/validation.py
glevco: They're closed, there's a doctest for that:

```python
>>> _has_any_overlap([(0, 10), (10, 20)])
True
```

I added it to the docstring as well. This makes it invalid to configure a feature using the same bit as a previous one, starting at the exact block the previous one ends. A full evaluation interval is necessary between them, before the bit can be reused.

hathor/feature_activation/exception.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/611/commits/b141815b3655383aa17f7df5c33cba323cb58c19

hathor/feature_activation/feature.py
glevco: I changed it to a string enum, but done in https://github.com/HathorNetwork/hathor-core/pull/611/commits/b141815b3655383aa17f7df5c33cba323cb58c19

hathor/feature_activation/validation.py
glevco: Nice suggestion, I had changed it, but since I'm now using the NamedTuple suggested below, I still have to pass the key, as `Feature` does not implement a comparator

hathor/feature_activation/validation.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/611/commits/b141815b3655383aa17f7df5c33cba323cb58c19

hathor/feature_activation/feature.py
msbrogli: Thanks!

hathor/feature_activation/settings.py
```diff
@@ -0,0 +1,120 @@
...
+from hathor.feature_activation.model.criteria import Criteria
+from hathor.utils.pydantic import BaseModel
+
+
+class Settings(BaseModel, validate_all=True):
+    """Feature Activation settings."""
+
+    # The number of blocks in the feature activation evaluation interval.
+    # Equivalent to 14 days (40320 * 30 seconds = 14 days)
+    evaluation_interval: PositiveInt = 40320
```
msbrogli, line 30: Is it a default value for all features?

msbrogli (changes requested):

hathor/conf/settings.py
msbrogli: I don't see any value to add a default value for a feature. I feel that features are so important that we should specify all its settings without default values. We can have a recommended value as a comment in the `Feature` class but they should be explicit put in the config files.

hathor/feature_activation/constants.py
msbrogli: Thanks!

hathor/feature_activation/exception.py
msbrogli: Great! Thanks!

hathor/feature_activation/model/criteria.py
msbrogli: Right. Thanks!

hathor/feature_activation/feature.py
msbrogli: Ok. Thanks!

hathor/feature_activation/model/criteria.py
msbrogli: My bad. Thanks!

hathor/feature_activation/validation.py
msbrogli: Thanks!

hathor/feature_activation/validation.py
msbrogli: You're right. Ok. Thanks!

hathor/feature_activation/validation.py
msbrogli: Thanks!

hathor/feature_activation/settings.py
glevco: No, this value is actually a constant per network, not a variable per feature. It has to be the same for all features. There's a lost thread about it here: https://github.com/HathorNetwork/rfcs/commit/1c874a535200492260e8b5deb5c837df61629c11#r112715956

hathor/conf/settings.py
glevco: In my understanding this is more a per network rule, like: all features in the mainnet should have at least 90% of miners support, but we could set a different number when we want some specific feature to behave differently, on a case by case basis.

I see there are three levels of flexibility for this, from the least flexible to the most:

1. The `threshold` is a constant per network, that is, every feature has the same `threshold`. In Bitcoin, it's like this. They have 90% for the mainnet and 75% for the testnet. That is, it's not possible to activate any feature in the mainnet if it's not supported by 90%+ of the miners.
2. The `threshold` is variable per feature, but there's a default for each network. That's what's currently implemented. It allows for flexibility, but gives preference to using the default value for the network.
3. Lastly, your suggestion, the `threshold` is variable per feature, and there's no default. Every time a new feature is configured, the threshold must be set.

I don't see a problem going either way, actually. The only point I think we have to have in mind is that every time a new feature is activated, it'll create a fork in the network, so the higher the `threshold` is, the more we are sure that most miners are not going to be forked away.

In other words, I set a default value for each network to provide a high default threshold, to minimize misuse of the feature activation process: although possible, maybe we shouldn't have lots of features getting activated with low thresholds.

That being said, our context is very different than Bitcoin's, as we are close to our partners and community, so it's very unlikely that our features are going to be rejected. Feature thresholds are still going to be discussed and reviewed every time we deploy a new feature.

So, I don't see a problem having a default, and that's what I would vote for. But if you think we should explicitly decide on a threshold for each feature, I also think that's a valid argument. I just think we should try to policy ourselves on not releasing many features with low thresholds (actually, we should do this anyway, even if there's a default).

msbrogli: Got it. You convinced me to keep it as you did. Thanks for the clarification.

hathor/feature_activation/settings.py
msbrogli: Got it. Thanks!

msbrogli (approved):

hathor/cli/run_node.py
```diff
@@ -324,6 +326,11 @@ def __init__(self, *, argv=None):
         elif args.testnet:
             os.environ['HATHOR_CONFIG_YAML'] = TESTNET_SETTINGS_FILEPATH
 
+        try:
+            HathorSettings()
+        except ValidationError as e:
+            raise PreInitializationError(e)
```
jansegre, line 332: An alternative is doing `raise PreInitializationError from e`, which still makes `e` show in the traceback as what caused PreInitializationError, this is called [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining).

hathor/feature_activation/feature.py
```diff
@@ -0,0 +1,27 @@
...
+
+from enum import Enum, unique
+
+
+@unique
+class Feature(Enum):
+    """
+    An enum containing all features that participate in the feature activation process, past or future,
+    activated or not, for all networks. Features should NOT be removed from this enum, to preserve history.
+    Their values should NOT be changed either.
```
jansegre, line 23: Why is this the case? Could you include the reason in this description?

Like, is it used by the storage so it should be preserved so databases are compatible between node versions? Or are the values used inside blocks? Or even on the p2p protocol (which I don't think it's the case)?

jansegre (dismissed):

hathor/cli/run_node.py
glevco: Cool, I didn't know about this, thanks. Changed in https://github.com/HathorNetwork/hathor-core/pull/611/commits/44615a281d8be68479e90e818e774ca1887c2a83. I also added `TypeError` in the `except` block, as it's raised when an extra, invalid attribute is set in `HathorSettings`.

hathor/feature_activation/feature.py
glevco: > Like, is it used by the storage so it should be preserved so databases are compatible between node versions? Or are the values used inside blocks? Or even on the p2p protocol (which I don't think it's the case)?

No, none of those are the case. The reason is just the the value is used to parse criteria from settings. Changing them could mess up with configuration. Ideally I would use `StrEnum` and each option would use `enum.auto()`, so each name and value are equal strings. But it's only available from Python 3.11

Added more docstring info in https://github.com/HathorNetwork/hathor-core/pull/611/commits/44615a281d8be68479e90e818e774ca1887c2a83

jansegre (approved):

msbrogli (approved):

PR #617, opened 2023-05-18
feat: add safeguards for partially validated transactions [part 2/2]

jansegre:
This change was split out from #577. The main objective is to be able to do this:

```python
tx.set_validation(ValidationState.BASIC)  # this is a partially validated transaction
tx_storage.save_transaction(tx)  # this fails with an assertion error
with stx_storage.allow_partially_validated_context():
    tx_storage.save_transaction(tx)  # but this succeeds
tx_storage.get_transaction(tx.hash)  # this also fails, but with an error that inherits from "not found"
with stx_storage.allow_partially_validated_context():
    tx_storage.get_transaction(tx.hash)  # but this succeeds
```

This change is only exists to reduce the scope where the storage APIs accept reading/writing partially validated transactions so that the changes needed by sync-v2 can have a very reduced an limited impact on the rest of the node. The previous iteration of this protection was made using a `allow_partially_validated: bool` argument on most APIs, that ended up being not only verbose, but more error prone and needed to carry this `allow_partially_validated` in a lot of places that shouldn't have to know about that.

## Acceptance Criteria

- TransactionStorage must have a scope that dictates which validation level on accept at save/get methods
- TransactionStorage must have methods that return contextmanagers which temporarily modify that scope

### DIFF
--- a/hathor/consensus/consensus.py
+++ b/hathor/consensus/consensus.py
@@ -70,6 +70,8 @@ class ConsensusAlgorithm:
 
     @cpu.profiler(key=lambda self, base: 'consensus!{}'.format(base.hash.hex()))
     def update(self, base: BaseTransaction) -> None:
+        assert base.storage is not None
+        assert base.storage.is_only_valid_allowed()
         try:
             self._unsafe_update(base)
         except Exception:
@@ -107,11 +109,16 @@ class ConsensusAlgorithm:
         if new_best_height < best_height:
             self.log.warn('height decreased, re-checking mempool', prev_height=best_height, new_height=new_best_height,
                           prev_block_tip=best_tip.hex(), new_block_tip=new_best_tip.hex())
-            to_remove = storage.get_transactions_that_became_invalid()
+            # XXX: this method will mark as INVALID all transactions in the mempool that became invalid because of a
+            #      reward lock
+            to_remove = storage.compute_transactions_that_became_invalid()
             if to_remove:
                 self.log.warn('some transactions on the mempool became invalid and will be removed',
                               count=len(to_remove))
-                storage.remove_transactions(to_remove)
+                # XXX: because transactions in `to_remove` are marked as invalid, we need this context to be able to
+                #      remove them
+                with storage.allow_invalid_context():
+                    storage.remove_transactions(to_remove)
                 for tx_removed in to_remove:
                     context.pubsub.publish(HathorEvents.CONSENSUS_TX_REMOVED, tx_hash=tx_removed.hash)
 
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -413,8 +413,9 @@ class HathorManager:
         # self.start_profiler()
         if self._full_verification:
             self.log.debug('reset all metadata')
-            for tx in self.tx_storage.get_all_transactions():
-                tx.reset_metadata()
+            with self.tx_storage.allow_partially_validated_context():
+                for tx in self.tx_storage.get_all_transactions():
+                    tx.reset_metadata()
 
         self.log.debug('load blocks and transactions')
         for tx in self.tx_storage._topological_sort_dfs():
@@ -459,9 +460,11 @@ class HathorManager:
                             self.tx_storage.indexes.mempool_tips.update(tx)  # XXX: move to indexes.update
                         if self.tx_storage.indexes.deps is not None:
                             self.sync_v2_step_validations([tx])
+                        self.tx_storage.save_transaction(tx, only_metadata=True)
                     else:
                         assert tx.validate_basic(skip_block_weight_verification=skip_block_weight_verification)
-                    self.tx_storage.save_transaction(tx, only_metadata=True)
+                        with self.tx_storage.allow_partially_validated_context():
+                            self.tx_storage.save_transaction(tx, only_metadata=True)
                 else:
                     # TODO: deal with invalid tx
                     if not tx_meta.validation.is_final():
--- a/hathor/transaction/storage/transaction_storage.py
+++ b/hathor/transaction/storage/transaction_storage.py
@@ -15,6 +15,7 @@
 import hashlib
 from abc import ABC, abstractmethod, abstractproperty
 from collections import defaultdict, deque
+from contextlib import AbstractContextManager
 from threading import Lock
 from typing import Any, Dict, Iterator, List, NamedTuple, Optional, Set, Tuple, Type, cast
 from weakref import WeakValueDictionary
@@ -28,8 +29,13 @@ from hathor.profiler import get_cpu_profiler
 from hathor.pubsub import PubSubManager
 from hathor.transaction.base_transaction import BaseTransaction
 from hathor.transaction.block import Block
-from hathor.transaction.storage.exceptions import TransactionDoesNotExist, TransactionIsNotABlock
+from hathor.transaction.storage.exceptions import (
+    TransactionDoesNotExist,
+    TransactionIsNotABlock,
+    TransactionNotInAllowedScopeError,
+)
 from hathor.transaction.storage.migrations import BaseMigration, MigrationState, add_min_height_metadata
+from hathor.transaction.storage.tx_allow_scope import TxAllowScope, tx_allow_context
 from hathor.transaction.transaction import Transaction
 from hathor.transaction.transaction_metadata import TransactionMetadata
 from hathor.util import not_none
@@ -93,6 +99,9 @@ class TransactionStorage(ABC):
 
         # This is a global lock used to prevent concurrent access when getting the tx lock in the dict above
         self._weakref_lock: Lock = Lock()
+
+        # Flag to allow/disallow partially validated vertices.
+        self.allow_scope: TxAllowScope = TxAllowScope.VALID
 
         # Cache for the best block tips
         # This cache is updated in the consensus algorithm.
@@ -330,6 +339,43 @@ class TransactionStorage(ABC):
             return None
         return self._tx_weakref.get(hash_bytes, None)
 
+    # TODO: check if the method bellow is currently needed
+    def allow_only_valid_context(self) -> AbstractContextManager[None]:
+        """This method is used to temporarily reset the storage back to only allow valid transactions.
+
+        The implementation will OVERRIDE the current scope to allowing only valid transactions on the observed
+        storage.
+        """
+        return tx_allow_context(self, allow_scope=TxAllowScope.VALID)
+
+    def allow_partially_validated_context(self) -> AbstractContextManager[None]:
+        """This method is used to temporarily make the storage allow partially validated transactions.
+
+        The implementation will INCLUDE allowing partially valid transactions to the current allow scope.
+        """
+        new_allow_scope = self.allow_scope | TxAllowScope.PARTIAL
+        return tx_allow_context(self, allow_scope=new_allow_scope)
+
+    def allow_invalid_context(self) -> AbstractContextManager[None]:
+        """This method is used to temporarily make the storage allow invalid transactions.
+
+        The implementation will INCLUDE allowing invalid transactions to the current allow scope.
+        """
+        new_allow_scope = self.allow_scope | TxAllowScope.INVALID
+        return tx_allow_context(self, allow_scope=new_allow_scope)
+
+    def is_only_valid_allowed(self) -> bool:
+        """Whether only valid transactions are allowed to be returned/accepted by the storage, the default state."""
+        return self.allow_scope is TxAllowScope.VALID
+
+    def is_partially_validated_allowed(self) -> bool:
+        """Whether partially validated transactions are allowed to be returned/accepted by the storage."""
+        return TxAllowScope.PARTIAL in self.allow_scope
+
+    def is_invalid_allowed(self) -> bool:
+        """Whether invalid transactions are allowed to be returned/accepted by the storage."""
+        return TxAllowScope.INVALID in self.allow_scope
+
     def _enable_weakref(self) -> None:
         """ Weakref should never be disabled unless you know exactly what you are doing.
         """
@@ -354,7 +400,7 @@ class TransactionStorage(ABC):
         self.pre_save_validation(tx, meta)
 
     def pre_save_validation(self, tx: BaseTransaction, tx_meta: TransactionMetadata) -> None:
-        """ Must be run before every save, only raises AssertionError.
+        """ Must be run before every save, will raise AssertionError or TransactionNotInAllowedScopeError
 
         A failure means there is a bug in the code that allowed the condition to reach the "save" code. This is a last
         second measure to prevent persisting a bad transaction/metadata.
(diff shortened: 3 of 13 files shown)


hathor/transaction/storage/tx_allow_scope.py
```diff
@@ -0,0 +1,68 @@
...
+    purpose of filtering it can be any combination of the supported types.
+    """
+    VALID = auto()
+    PARTIAL = auto()
+    INVALID = auto()
+
+    def is_allowed(self, tx: BaseTransaction) -> bool:
+        """True means it is allowed to be used in the storage (as argument or as return), False means not allowed."""
+        tx_meta = tx.get_metadata()
+        is_partial = tx_meta.voided_by is not None and settings.PARTIALLY_VALIDATED_ID in tx_meta.voided_by
```
msbrogli, line 42: Why don't we use `tx_meta.validation`?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -330,6 +339,43 @@ def get_transaction_from_weakref(self, hash_bytes: bytes) -> Optional[BaseTransa
...
+        """This method is used to temporarily make the storage allow invalid transactions.
+
+        The implementation will INCLUDE allowing invalid transactions to the current allow scope.
+        """
+        new_allow_scope = self.allow_scope | TxAllowScope.INVALID
+        return tx_allow_context(self, allow_scope=new_allow_scope)
+
+    def is_only_valid_allowed(self) -> bool:
+        """Whether only valid transactions are allowed to be returned/accepted by the storage, the default state."""
+        return TxAllowScope.INVALID in self.allow_scope
```
msbrogli, line 369: Bug here.

msbrogli (changes requested):

hathor/transaction/storage/transaction_storage.py
jansegre: Fixed.

hathor/transaction/storage/tx_allow_scope.py
jansegre: Done.

hathor/transaction/storage/tx_allow_scope.py
```diff
@@ -0,0 +1,68 @@
...
+    VALID = auto()
+    PARTIAL = auto()
+    INVALID = auto()
+
+    def is_allowed(self, tx: BaseTransaction) -> bool:
+        """True means it is allowed to be used in the storage (as argument or as return), False means not allowed."""
+        tx_meta = tx.get_metadata()
+        # XXX: initial/partial/invalid/fully_connected never overlap and cover all possible validation states
+        #      see hathor.transaction.transaction_metadata.ValidationState for more details
+        validation = tx_meta.validation
```
msbrogli, line 44: Should we add `assert validation in list_with_all_possible_states`?

msbrogli (changes requested):

hathor/transaction/storage/transaction_storage.py

hathor/transaction/storage/tx_allow_scope.py
msbrogli: Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli, line 1: Should `get_metadata()` assert that the validation state and `meta.voided_by` are coherent?

msbrogli, line 1: Shouldn't `save_transaction(...)` follow the scope? For clarity, if one is trying to save an invalid transaction, this transaction must be allowed by the current scope.

hathor/transaction/storage/tx_allow_scope.py
jansegre: In this case `list_with_all_possible_states = list(ValidationState)`. Asserting that in itself is pretty redundant with the whole Enum class.

What maybe wouldn't be redundant is `assert list(ValidationState) = [ValidationState.INITIAL, ValidationState.BASIC, ...]`, which would basically assert what the expected enum variants are. I'm not sure if it makes sense to do this here though. Maybe it would make more sense to have a simple test for that which could point to the known code locations that make certain assumptions based on the enum variants. I'm not too sure how much this helps though.

What do you think?

hathor/transaction/storage/transaction_storage.py
jansegre: I think that makes sense, I'll do that.

jansegre: Yes. I'll do that.

jansegre: That is already the case now since `post_get_validation` happens when `get_metadata` calls `get_transaction`.

jansegre: Done. Also refactored the tests to better cover this.

msbrogli (approved):

hathor/transaction/storage/transaction_storage.py
msbrogli: Great job! Thanks!

msbrogli: Great job! Thanks!

hathor/transaction/storage/tx_allow_scope.py
msbrogli: `assert list(ValidationState) == [ValidationState.INITIAL, ...]` is exactly what I meant. But I guess we can pass it. It would be great to have a mechanism that guarantees we have covered all possible cases otherwise a linter would fail or an exception would be raised.

jansegre: I'm actually tweaking this a little bit because of some edge cases (when using full validations we temporarily save transactions with INITIAL state). I'll add a test to assert that assumption (in the XXX comment) and the state list.

jansegre: Done. I've added those tests.

hathor/manager.py
```diff
@@ -441,13 +441,15 @@ def _initialize_components(self) -> None:
         # self.start_profiler()
         if self._full_verification:
             self.log.debug('reset all metadata')
-            for tx in self.tx_storage.get_all_transactions():
-                tx.reset_metadata()
+            with self.tx_storage.allow_partially_validated_context():
+                for tx in self.tx_storage.get_all_transactions():
```
msbrogli, line 445: Should `get_all_transactions()` only return transactions in the scope?

hathor/manager.py
```diff
@@ -441,13 +441,15 @@ def _initialize_components(self) -> None:
...
+                    tx.reset_metadata()
 
         self.log.debug('load blocks and transactions')
         for tx in self.tx_storage._topological_sort_dfs():
```
msbrogli, line 449: Same thing here: should it go through only transactions in the scope?

hathor/manager.py
```diff
@@ -441,13 +441,15 @@ def _initialize_components(self) -> None:
...
             if self._full_verification:
-                tx.update_initial_metadata()
+                with self.tx_storage.allow_partially_validated_context():
+                    tx.update_initial_metadata()
```
msbrogli, line 452: Why would this access need a broader scope?

hathor/manager.py
```diff
@@ -480,16 +482,19 @@ def _initialize_components(self) -> None:
                     # TODO: deal with invalid tx
                     if tx.can_validate_full():
                         self.tx_storage.add_to_indexes(tx)
-                        assert tx.validate_full(skip_block_weight_verification=skip_block_weight_verification)
+                        assert tx.validate_full(skip_block_weight_verification=skip_block_weight_verification,
+                                                reject_locked_reward=False)
```
msbrogli, line 486: Why are we adding `reject_locked_reward=False` here?

hathor/manager.py
```diff
@@ -480,16 +482,19 @@ def _initialize_components(self) -> None:
...
                         self.consensus_algorithm.update(tx)
                         self.tx_storage.indexes.update(tx)
                         if self.tx_storage.indexes.mempool_tips is not None:
                             self.tx_storage.indexes.mempool_tips.update(tx)  # XXX: move to indexes.update
                         if self.tx_storage.indexes.deps is not None:
                             self.sync_v2_step_validations([tx])
+                        self.tx_storage.save_transaction(tx, only_metadata=True)
```
msbrogli, line 493: Why save here?

hathor/manager.py
```diff
@@ -480,16 +482,19 @@ def _initialize_components(self) -> None:
...
                     else:
-                        assert tx.validate_basic(skip_block_weight_verification=skip_block_weight_verification)
-                    self.tx_storage.save_transaction(tx, only_metadata=True)
+                        with self.tx_storage.allow_partially_validated_context():
+                            assert tx.validate_basic(skip_block_weight_verification=skip_block_weight_verification)
+                            self.tx_storage.save_transaction(tx, only_metadata=True)
```
msbrogli, line 497: Why save here?

hathor/manager.py
```diff
@@ -480,16 +482,19 @@ def _initialize_components(self) -> None:
...
                         self.tx_storage.indexes.update(tx)
                         if self.tx_storage.indexes.mempool_tips is not None:
                             self.tx_storage.indexes.mempool_tips.update(tx)  # XXX: move to indexes.update
                         if self.tx_storage.indexes.deps is not None:
                             self.sync_v2_step_validations([tx])
+                        self.tx_storage.save_transaction(tx, only_metadata=True)
                     else:
-                        assert tx.validate_basic(skip_block_weight_verification=skip_block_weight_verification)
-                    self.tx_storage.save_transaction(tx, only_metadata=True)
+                        with self.tx_storage.allow_partially_validated_context():
```
msbrogli, line 495: Why do we need access to partially validated transactions to run a basic validation?

hathor/consensus/consensus.py
```diff
@@ -111,7 +113,8 @@ def _unsafe_update(self, base: BaseTransaction) -> None:
             if to_remove:
                 self.log.warn('some transactions on the mempool became invalid and will be removed',
                               count=len(to_remove))
-                storage.remove_transactions(to_remove)
+                with storage.allow_invalid_context():
```
msbrogli, line 116: Add a note explaining that the mempool cleaning mechanism invalidates all transactions so we need to allow invalid transactions here.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -958,6 +1024,7 @@ def get_transactions_that_became_invalid(self) -> List[BaseTransaction]:
         for tx in self.iter_mempool_from_best_index():
             if tx.is_spent_reward_locked():
                 tx.get_metadata().validation = ValidationState.INVALID
```
msbrogli, line 1026: Should we refactor it to `tx.set_validation(ValidationState.INVALID)`? I feel we've been calling `mark_partially_validated()` in several places.

hathor/transaction/storage/tx_allow_scope.py
msbrogli: Great! Thanks!

msbrogli (changes requested):

hathor/manager.py
jansegre: I think so.

hathor/manager.py
jansegre: That makes sense, I'm almost sure that it will be implicit as long as `get_all_transaction` respects the `tx_storage.allow_scope`.

hathor/manager.py
jansegre: Because `tx.update_initial_metadata()` does a `tx_storage.save_transaction(tx)` but we haven't validated `tx` yet. I don't like too much that some methods will implicitly save, but a lot of times it's hard to get around that. I think using `allow_partially_validated_context()` is correct here.

hathor/manager.py
jansegre: This is a bug fix. I'm not too sure why it was passing before, but `reject_locked_reward=False` should be correct because we aren't guaranteed to have a "good" sort when initialized (a good sort is a sort that validates the block that unlocks the reward before the transaction that spends the reward). But it is actually missing a final pass on the mempool to check for transactions that could have become invalid (those are the ones where `reject_locked_reward=True` must be used).

hathor/manager.py
jansegre: The structure was:

```
if can do a full validation of tx:
    do full validation of tx
else:
    do partial validation of tx
save tx
```

I refactored it to:

```
if can do a full validation of tx:
    do full validation of tx
    save tx
else:
    with allow partial context:
        do partial validation of tx
        save tx
```

hathor/manager.py
jansegre: Because it might need access to partially validated parents.

hathor/manager.py
jansegre: Because that transaction is only partially validated now, so I have to save inside an "allow partially validated" context.

hathor/transaction/storage/transaction_storage.py
jansegre: That would be good.

hathor/manager.py
msbrogli: Thanks for the explanation!

hathor/manager.py
msbrogli: Got it. Thanks!

hathor/consensus/consensus.py
msbrogli: Great! Thanks!

hathor/manager.py
msbrogli: Great! Thanks!

hathor/manager.py
msbrogli: Great! Thanks!

hathor/manager.py
msbrogli: I agree. Thanks for the explanation.

hathor/transaction/storage/transaction_storage.py
msbrogli: Great refactor. Thanks!

hathor/manager.py
jansegre: Since I fixed the cache clearing issue in the test it only needs the context to make the save now, and not the basic validation.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -563,10 +563,16 @@ def get_metadata(self, hash_bytes: bytes) -> Optional[TransactionMetadata]:
         except TransactionDoesNotExist:
             return None
 
+    def get_all_transactions(self) -> Iterator[BaseTransaction]:
+        """Return all transactions/blocks.
```
msbrogli, line 567: Suggested docstring: "Return all vertices (transactions and blocks) within the allowed scope."

hathor/manager.py
msbrogli: Great!

msbrogli (changes requested):

hathor/manager.py
jansegre: I'll do that, but ideally I think we should remove this code altogether possibly replacing it with two different approaches: 1) actually and explicitly rebuilding the database, 2) actually just verifying (no write operations) and only the last few blocks/transactions.

jansegre: Done. Opened #627

msbrogli (approved):

hathor/manager.py
msbrogli: Great! Thanks!

hathor/transaction/storage/transaction_storage.py

glevco (approved):

PR #620, opened 2023-05-19
feat(sysctl): Add helper command

msbrogli:
### Acceptance criteria

1) Add `!help` command, which shows all commands available.
2) Add `!help [cmd]` command, which shows a help message for the command.

### Example

```
!help
p2p.always_enable_sync
p2p.always_enable_sync.readtxt
p2p.force_sync_rotate
p2p.max_enabled_sync
p2p.rate_limit.global.send_tips
p2p.sync_update_interval
!help p2p.always_enable_sync
getter() -> List[str]:
    Return the list of sync-always-enabled peers.

setter(values: List[str]) -> None:
    Change the list of sync-always-enabled peers.
!help p2p.rate_limit.global.send_tips
getter() -> Tuple[int, float]:
    Return the global rate limiter for SEND_TIPS.

setter(max_hits: int, window_seconds: float) -> None:
    Change the global rate limiter for SEND_TIPS.

    The rate limiter is disabled when `window_seconds == 0`.
```

### DIFF
--- a/hathor/sysctl/protocol.py
+++ b/hathor/sysctl/protocol.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import inspect
 import json
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Callable, List, Optional
 
 from pydantic import ValidationError
 from twisted.protocols.basic import LineReceiver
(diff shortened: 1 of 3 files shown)


hathor/sysctl/protocol.py
```diff
@@ -107,3 +134,16 @@ def _deserialize(self, value_str: str) -> Any:
         if len(parts) > 1:
             return tuple(json.loads(x) for x in parts)
         return json.loads(value_str)
+
+    def _get_method_help(self, method_name: str, method: Optional[Callable]) -> List[str]:
+        """Return a list of strings with the help for `method`."""
```
jansegre, line 139: Is the list of strings the lines?

jansegre (approved):

hathor/sysctl/protocol.py
msbrogli: Yes, it is.

glevco (approved):

PR #621, opened 2023-05-19
tests: Fail tests if they do not stop all simulators

msbrogli:


tests/p2p/test_sync_enabled.py
```diff
@@ -52,13 +52,20 @@ def test_sync_rotate(self):
 
         self.simulator.run(600)
 
+        ready = set(conn for conn in connections if conn.proto1.is_state(conn.proto1.PeerState.READY))
+        self.assertTrue(len(ready), len(other_managers))
```
glevco, line 56: Is this supposed to be `assertEqual`?

tests/p2p/test_sync_enabled.py
```diff
@@ -52,13 +52,20 @@ def test_sync_rotate(self):
...
         self.assertTrue(len(enabled), 3)
 
         manager1.connections._sync_rotate_if_needed(force=True)
         enabled2 = set(conn for conn in connections if conn.proto1.is_sync_enabled())
         self.assertTrue(len(enabled2), 3)
-        # Chance of false positive: 1/comb(20, 3) = 0.0008771929824561404
+        if enabled == enabled2:
+            manager1.connections._sync_rotate_if_needed(force=True)
+            enabled2 = set(conn for conn in connections if conn.proto1.is_sync_enabled())
+            self.assertTrue(len(enabled2), 3)
```
glevco, line 67: Same here, should be `assertEqual`?

tests/unittest.py
```diff
@@ -102,7 +103,14 @@ class TestCase(unittest.TestCase):
     use_memory_storage: bool = USE_MEMORY_STORAGE
     seed_config: Optional[int] = None
 
+    def _runFixturesAndTest(self, result):
+        try:
+            return super()._runFixturesAndTest(result)
+        finally:
+            self.assertEqual(0, Simulator._patches_rc)
+
```
glevco, line 111: Is this necessary? Isn't it the same as asserting in the `tearDown()`?

glevco (commented):
Question, all those `@pytest.mark.flaky` that were removed, is it because the fixes make them not flaky anymore?

tests/unittest.py
msbrogli: The difference is that any class can override the `tearDown()` and prevent calling `super().tearDown()`. Besides that, it's exactly the same. Maybe I should keep only one of them.

tests/p2p/test_sync_enabled.py
msbrogli: Sure. My bad! I just fixed it.

tests/p2p/test_sync_enabled.py
msbrogli: Sure. My bad! I just fixed it.

tests/unittest.py
msbrogli: Another difference is that, even if `super.tearDown()` method is called, it must be called after all simulators have been stopped. Otherwise the assert will fail.

glevco (approved):

tests/unittest.py
glevco: Makes sense. I'm fine keeping both, then. I'm resolving this

jansegre (commented):
I like the changes and I'd like to have them in `master`, but I don't like that the CI is failing in all runs. I think that we should deal with it before merging or make it a top priority to fix the CI if we are to merge it as is.

PR #622, opened 2023-05-19
refactor(p2p): Build ConnectionsManager in builder

msbrogli:
### Acceptance criteria

1) Build `ConnectionsManager` in builders.
2) Move `ssl: bool`, `enable_sync_v1`, `enable_sync_v1_1`, and `enable_sync_v2` to `ConnectionsManager`.
3) Remove `HathorManager.server_factory` and `HathorManager.client_factory`.

### Note

There are still some network-related attributes and methods on `HathorManager`. I'll move those on a separate refactor to keep this one small for reviewers. Some attributes and methods yet-to-be moved out of `HathorManager` are:

- `listen_addresses: List[str]`
- `capabilities`
- `has_sync_version_capability()`
- `add_peer_to_whitelist()`
- `remove_peer_from_whitelist_and_disconnect()`
- `listen()`
- `do_discovery()`

### DIFF
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -40,7 +40,6 @@ if TYPE_CHECKING:
     from twisted.internet.interfaces import IDelayedCall
 
     from hathor.manager import HathorManager
-    from hathor.p2p.factory import HathorClientFactory, HathorServerFactory
 
 logger = get_logger()
 settings = HathorSettings()
@@ -78,6 +77,7 @@ class ConnectionsManager:
     class GlobalRateLimiter:
         SEND_TIPS = 'NodeSyncTimestamp.send_tips'
 
+    manager: Optional['HathorManager']
     connections: Set[HathorProtocol]
     connected_peers: Dict[str, HathorProtocol]
     connecting_peers: Dict[IStreamClientEndpoint, _ConnectingPeer]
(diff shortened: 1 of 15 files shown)


hathor/p2p/manager.py
```diff
@@ -176,6 +187,16 @@ def __init__(self, reactor: Reactor, my_peer: PeerId, server_factory: 'HathorSer
         if enable_sync_v2:
             self._sync_factories[SyncVersion.V2] = SyncV1Factory(self)
 
+    def set_manager(self, manager: 'HathorManager') -> None:
+        """Set the manager. This method must be called before start()."""
```
glevco, line 191: Add an `assert self.manager is not None` in `start()`?

glevco: Or maybe even change the `start()` method to receive the `HathorManager` and set it, instead of having a separate `set_manager()` method

glevco (approved):

jansegre (approved):

hathor/p2p/manager.py
msbrogli: Even though I think it is an interesting idea, we usually leave the `start()` and `stop()` method without parameters. This way, we can run them anywhere without requiring access to the dependencies. That's why I used a `set_manager()` method. What do you think?

glevco: Ok, makes sense!

PR #623, opened 2023-05-22
feat(feature-activation): implement signal_bits field in txs

glevco:
### Acceptance Criteria

- Change `TxVersion` enum so it fails instantiation for values larger than one byte, instead of ignoring it.
- Create new `signal_bits` attribute in `BaseTransaction`, derived from the first byte of the `version` field.

**ATTENTION**: this PR touches important parts of the code

### DIFF
--- a/hathor/client.py
+++ b/hathor/client.py
@@ -392,8 +392,7 @@ def create_tx_from_dict(data: Dict[str, Any], update_hash: bool = False,
     if storage:
         data['storage'] = storage
 
-    _, tx_version = TxVersion.extract_from_bytes(data['version'])
-    cls = tx_version.get_cls()
+    cls = TxVersion(data['version']).get_cls()
     metadata = data.pop('metadata', None)
     tx = cls(**data)
     if update_hash:
--- a/hathor/client.py
+++ b/hathor/client.py
@@ -392,7 +392,8 @@ def create_tx_from_dict(data: Dict[str, Any], update_hash: bool = False,
     if storage:
         data['storage'] = storage
 
-    cls = TxVersion(data['version']).get_cls()
+    version = data['version'] & 0xFF
+    cls = TxVersion(version).get_cls()
     metadata = data.pop('metadata', None)
     tx = cls(**data)
     if update_hash:
--- a/hathor/client.py
+++ b/hathor/client.py
@@ -392,8 +392,8 @@ def create_tx_from_dict(data: Dict[str, Any], update_hash: bool = False,
     if storage:
         data['storage'] = storage
 
-    version = data['version'] & 0xFF
-    cls = TxVersion(version).get_cls()
+    _, tx_version = TxVersion.extract_from_bytes(data['version'])
+    cls = tx_version.get_cls()
     metadata = data.pop('metadata', None)
     tx = cls(**data)
     if update_hash:
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -121,8 +121,8 @@ class TxVersion(IntEnum):
         """
         Takes a 2-byte number and returns signal bits and a TxVersion from it.
 
-        Signal bits, extracted from the first byte, carry information about Feature Activation
-        bits and also extra bits reserved for future use, depending on the configuration.
+        Signal bits, extracted from the first byte, carry extra information that
+        may be interpreted differently by each subclass of BaseTransaction.
 
         Args:
             value: a 2-byte number
@@ -170,8 +170,10 @@ class BaseTransaction(ABC):
 
     _metadata: Optional[TransactionMetadata]
 
-    # The first byte extracted from the version field. Carries information about Feature Activation
-    # bits and also extra bits reserved for future use, depending on the configuration.
+    # Bits extracted from the first byte of the version field. They carry extra information that may be interpreted
+    # differently by each subclass of BaseTransaction.
+    # Currently only the Block subclass uses it, carrying information about Feature Activation bits and also extra
+    # bits reserved for future use, depending on the configuration.
     signal_bits: int
 
     def __init__(self,
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -81,6 +81,9 @@ _TX_PARENTS_BLOCKS = 0
 _BLOCK_PARENTS_TXS = 2
 _BLOCK_PARENTS_BLOCKS = 1
 
+# The int value of one byte
+_ONE_BYTE = 0xFF
+
 
 def sum_weights(w1: float, w2: float) -> float:
     return aux_calc_weight(w1, w2, 1)
@@ -112,28 +115,9 @@ class TxVersion(IntEnum):
     @classmethod
     def _missing_(cls, value: Any) -> None:
         assert isinstance(value, int), f"Value '{value}' must be an integer"
-        assert value <= 0xFF, f'Value {hex(value)} must not be larger than one byte'
+        assert value <= _ONE_BYTE, f'Value {hex(value)} must not be larger than one byte'
 
         raise ValueError(f'Invalid version: {value}')
-
-    @classmethod
-    def extract_from_bytes(cls, value: int) -> Tuple[int, 'TxVersion']:
-        """
-        Takes a 2-byte number and returns signal bits and a TxVersion from it.
-
-        Signal bits, extracted from the first byte, carry extra information that
-        may be interpreted differently by each subclass of BaseTransaction.
-
-        Args:
-            value: a 2-byte number
-
-        Returns: a tuple in the format (signal_bits, tx_version)
-        """
-        assert value <= 0xFFFF, f'Value {hex(value)} must not be larger than two bytes'
-
-        signal_bits, version = int_to_bytes(value, 2)
-
-        return signal_bits, TxVersion(version)
 
     def get_cls(self) -> Type['BaseTransaction']:
         from hathor.transaction.block import Block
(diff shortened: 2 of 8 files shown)


hathor/transaction/base_transaction.py
```diff
@@ -66,9 +66,8 @@
 # H = unsigned short (2 bytes), d = double(8), f = float(4), I = unsigned int (4),
 # Q = unsigned long long int (64), B = unsigned char (1 byte)
 
-# Version (H), inputs len (B), and outputs len (B), token uids len (B).
-# H = unsigned short (2 bytes)
-_SIGHASH_ALL_FORMAT_STRING = '!HBBB'
+# Signal bits (B), version (B), inputs len (B), and outputs len (B), token uids len (B).
+_SIGHASH_ALL_FORMAT_STRING = '!BBBBB'
```
glevco, line 70: This is not used. Can we remove it?

hathor/transaction/block.py
```diff
@@ -38,11 +38,11 @@
...
 
-# Version (H), outputs len (B)
-_FUNDS_FORMAT_STRING = '!HB'
+# Signal bits (B), version (B), outputs len (B)
+_FUNDS_FORMAT_STRING = '!BBB'
 
-# Version (H), inputs len (B) and outputs len (B)
-_SIGHASH_ALL_FORMAT_STRING = '!HBB'
+# Signal bits (B), version (B), inputs len (B) and outputs len (B)
+_SIGHASH_ALL_FORMAT_STRING = '!BBBB'
```
glevco, line 45: This is not used. Can we remove it?

hathor/client.py
```diff
@@ -392,7 +392,8 @@ def create_tx_from_dict(data: Dict[str, Any], update_hash: bool = False,
     if storage:
         data['storage'] = storage
 
-    cls = TxVersion(data['version']).get_cls()
+    version = data['version'] & 0xFF
```
msbrogli, line 395: Maybe use a constant here? `TxVersion.VERSION_MASK` and `TxVersion.SIGNAL_BITS_MASK`?

hathor/transaction/base_transaction.py
```diff
@@ -172,9 +168,12 @@ def __init__(self,
             Outputs: all outputs that are being created
             Parents: transactions you are confirming (2 transactions and 1 block - in case of a block only)
         """
+        assert version <= 0xFFFF, f'Version {hex(version)} must not be larger than two bytes'
+
         self.nonce = nonce
         self.timestamp = timestamp or int(time.time())
-        self.version = version
+        self.signal_bits = (version >> 8) & 0xFF
```
msbrogli, line 175: Maybe move to a separate method (e.g., `get_signal_bits_from_version_byte()`)? This method might be on the `TxVersion` enum.

hathor/transaction/base_transaction.py
```diff
@@ -172,9 +168,12 @@ def __init__(self,
...
+        self.version = version & 0xFF
```
msbrogli, line 176: Use mask here.

hathor/transaction/base_transaction.py
msbrogli, line 175: What's the best type for `signal_bits`? We also should explicitly add a `signal_bits: int` at the beginning of the class.

msbrogli (changes requested):

hathor/transaction/base_transaction.py
glevco: Of course, I should have done it from the beginning. I implemented a new `TxVersion.extract_from_bytes()` in https://github.com/HathorNetwork/hathor-core/pull/623/commits/80814b32d58cd5c0441aec3cb9b10cf124458a78

hathor/transaction/base_transaction.py
glevco: With the change in https://github.com/HathorNetwork/hathor-core/pull/623/commits/80814b32d58cd5c0441aec3cb9b10cf124458a78, masking is not necessary anymore

hathor/transaction/base_transaction.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/623/commits/80814b32d58cd5c0441aec3cb9b10cf124458a78

msbrogli: Great! Thanks!

msbrogli: Great! Thanks!

hathor/client.py
msbrogli: Great! Thanks!

hathor/transaction/base_transaction.py
msbrogli: I don't think we should do it in this PR.

hathor/transaction/base_transaction.py
```diff
@@ -154,6 +169,10 @@ class BaseTransaction(ABC):
 
     _metadata: Optional[TransactionMetadata]
 
+    # The first byte extracted from the version field. Carries information about Feature Activation
+    # bits and also extra bits reserved for future use, depending on the configuration.
+    signal_bits: int
```
msbrogli, line 174: Should we add it to `BaseTransaction` or just to `Block`? I mean, does it make sense for transactions to signal supported features?

msbrogli (changes requested):

hathor/transaction/block.py
msbrogli: I don't think we should do it in this PR.

hathor/transaction/base_transaction.py
glevco: I added it to `BaseTransaction` because it makes sense to extract the `version` field's first byte as a separate field in all cases. I also named it `signal_bits` so it's generic enough, instead of something like `feature_activation_bits`.

It's just an "extra" byte that may be interpreted differently by each subclass of `BaseTransaction`. `Transaction`s may not use them now, but they may in the future (not for feature activation, but for something else).

I've changed the docstrings to make this clearer: https://github.com/HathorNetwork/hathor-core/pull/623/commits/92faeaa168c471e39e7cac465e0b38ec1f4a1a1d

jansegre: If we're changing `TxVersion` we'd need at least to make the implementation of `Transaction` ignore the first byte somehow. I think it's OK to just do it in `BaseTransaction`, since `signal_bits` should be completely ignored for transactions for now, it shouldn't matter much, we can change it later (we could even end up needing it for the next part of Feature Activation design).

hathor/transaction/base_transaction.py
```diff
@@ -174,7 +195,7 @@ def __init__(self,
         """
         self.nonce = nonce
         self.timestamp = timestamp or int(time.time())
-        self.version = version
+        self.signal_bits, self.version = TxVersion.extract_from_bytes(version)
```
jansegre, line 198: And instead of using a "version" with possibly 2 bytes here, we'd use a "normal" version with no signal bits included, and add a `signal_bits` argument to `__init__` arguments.

jansegre (changes requested):

hathor/transaction/base_transaction.py
glevco: I did it like this to more or less keep the same interface we have now, and to keep the historical fact that this extra byte is part of the version: for users of this class, nothing is changed. For example, in the `hathor/client.py` code, it could be that `data['version']` is being sent with two bytes. And I'm not too sure that there aren't other parts of the code that could be relying on the fact that `version` occupies two bytes.

If you think there's no risk is doing it like you suggested, I agree with you that it would be better to have `signal_bits` completely separated from `version` and `TxVersion`. What does @msbrogli think, too?

hathor/transaction/base_transaction.py
```diff
@@ -110,14 +109,30 @@ class TxVersion(IntEnum):
...
-            raise ValueError('Invalid version.')
-        return cls(version)
+    def _missing_(cls, value: Any) -> None:
+        assert isinstance(value, int), f"Value '{value}' must be an integer"
+        assert value <= 0xFF, f'Value {hex(value)} must not be larger than one byte'
+
+        raise ValueError(f'Invalid version: {value}')
+
+    @classmethod
+    def extract_from_bytes(cls, value: int) -> Tuple[int, 'TxVersion']:
```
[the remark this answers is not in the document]
glevco: Answered in https://github.com/HathorNetwork/hathor-core/pull/623#discussion_r1203234442

hathor/transaction/base_transaction.py
msbrogli: I think that the internal representation may be not related to the serialization/deserialization. I mean, we can look at `bytes(tx)` as having a 2-byte `version` but also as having a 1-byte version and a 1-byte `signal_bits`. In this sense, I feel it makes sense to split it in another parameter.

Even if the version really has two bytes (e.g., we might use only 4 bits of the second byte for bit signaling), I feel that they are different animals that are serialized together for historical reasons.

glevco: Ok, makes sense. I'll implement @jansegre's suggestion then

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/623/commits/64baf0c73c0f872c060bed69ac4f5f624bb9ea8d

hathor/transaction/base_transaction.py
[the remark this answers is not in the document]
glevco: Method removed in https://github.com/HathorNetwork/hathor-core/pull/623/commits/64baf0c73c0f872c060bed69ac4f5f624bb9ea8d

msbrogli (dismissed):

jansegre (dismissed):

jansegre (approved):

msbrogli (approved):

PR #626, opened 2023-05-23
refactor(settings): make HathorSettings always return the same instance

jansegre:
This wasn't _exactly_ the case before this commit. On one side when using the deprecated module settings the same module would always be loaded, and the same `SETTINGS` property would be accessed, but reloading the module (or at least leaving that up to importlib) unnecessary. On the new YAML system the file is reloaded everytime, this is also unnecessary.

This commit will make it the last HathorSettings instantiated will be returned, as long as the source is the same, otherwise an exception is raised.

There is a little small caveat that is fixed in this commit is that in the YAML file could in theory be altered between different calls to HathorSettings, which would lead to different content being loaded silently. What happens now is that the first time it is loaded is what is used for every call, any change in the file won't have any effect.

## Acceptance Criteria

- always return the same `hathor.conf.HathorSettings` instance when calling `hathor.conf.get_settings.HathorSettings`
- replace `get_settings_module` with `get_settings_module_path` that returns the loaded path instead of the loaded module

### DIFF
--- a/hathor/conf/get_settings.py
+++ b/hathor/conf/get_settings.py
@@ -14,8 +14,7 @@
 
 import importlib
 import os
-from types import ModuleType
-from typing import Optional
+from typing import NamedTuple, Optional
 
 from structlog import get_logger
 
@@ -24,8 +23,14 @@ from hathor.conf.settings import HathorSettings as Settings
 
 logger = get_logger()
 
-_settings_filepath: Optional[str] = None
-_config_file: Optional[str] = None
+
+class _SettingsMetadata(NamedTuple):
+    source: str
+    is_yaml: bool
+    settings: Settings
+
+
+_settings_singleton: Optional[_SettingsMetadata] = None
 
 
 def HathorSettings() -> Settings:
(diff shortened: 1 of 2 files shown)


hathor/conf/get_settings.py
```diff
@@ -37,52 +37,82 @@ def HathorSettings() -> Settings:
...
+    if _settings_singleton is not None:
+        if _settings_filepath != filepath:
+            raise Exception('loading config twice with a different file')
+        if _config_file is not None:
+            raise Exception('loading config twice with a different file type')
+        return _settings_singleton
+    settings = _load_yaml_settings(filepath)
+    _settings_filepath = filepath
+    _settings_singleton = settings
+    return settings
```
glevco, line 114: I found it a bit weird that we needed 3 global vars, and also that `_load_yaml_settings_with_cache()` is very similar to `_load_module_settings_with_cache()`.

If we implement a structure like `SettingsMetadata` below, we can have a single global var and eliminate duplicate code:

```python
class SettingsMetadata(NamedTuple):
    source: str
    is_yaml: bool
    settings: Settings

_settings_metadata_singleton: Optional[SettingsMetadata] = None

def _load_settings_with_cache(source: str, *, is_yaml: bool) -> Settings:
    global _settings_metadata_singleton

    if _settings_metadata_singleton is not None:
        if _settings_metadata_singleton.is_yaml != is_yaml:
            raise Exception('loading config twice with a different file type')
        if _settings_metadata_singleton.source != source:
            raise Exception('loading config twice with a different file')

        return _settings_metadata_singleton.settings

    settings_loader = _load_yaml_settings if is_yaml else _load_module_settings
    _settings_metadata_singleton = SettingsMetadata(
        source=source,
        is_yaml=is_yaml,
        settings=settings_loader(source)
    )

    return _settings_metadata_singleton.settings
```

Or something like this. What do you think?

glevco (changes requested):

hathor/conf/get_settings.py
jansegre: That seems good, I'll do that.

msbrogli (dismissed):

hathor/conf/get_settings.py
```diff
@@ -37,52 +42,71 @@ def HathorSettings() -> Settings:
...
+    """ Returns the path of the YAML settings loaded, None if no YAML file was loaded.
 
-    if not config_file:
+    XXX: Will raise an assertion error if HathorSettings() wasn't used before.
+    """
+    global _settings_singleton
+    assert _settings_singleton is not None, 'HathorSettings() not called before'
+    if not _settings_singleton.is_yaml:
         return None
+    return _settings_singleton.source
```
glevco, line 75: Maybe now we could just replace those two functions with:

```python
def get_settings_source() -> str:
    assert _settings_singleton is not None, 'HathorSettings() not called before'
    return _settings_singleton.source
```

As their only use is in `CliBuilder`:

`settings_source = get_settings_module_path() or get_settings_filepath()` could become just
`settings_source = get_settings_source()`

glevco (changes requested):

hathor/conf/get_settings.py
glevco: Great, thanks!

hathor/conf/get_settings.py
jansegre: Yeah, makes sense.

jansegre: Done.

PR #633, opened 2023-05-26
feat(feature-activation): bit count optimization

glevco:
Implements the suggestion in https://github.com/HathorNetwork/hathor-core/pull/613#discussion_r1205747507

### Acceptance Criteria

- Implement new `get_bit_list()` utils function
- Create new `TransactionMetadata.feature_activation_bit_counts` attribute
- Implement new `BaseTransaction.calculate_feature_activation_bit_counts()` abstract method and concrete implementations on `Block` and `Transaction`
- Add `feature_activation_bit_counts` to `BaseTransaction.get_metadata()` and `BaseTransaction.update_initial_metadata()`
- Update `FeatureService` to use the `feature_activation_bit_counts` metadata attribute instead of calculating the bit count itself

### DIFF
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -301,24 +301,6 @@ class BaseTransaction(ABC):
     def calculate_min_height(self) -> int:
         raise NotImplementedError
 
-    def calculate_feature_activation_bit_counts(self) -> Optional[list[int]]:
-        """
-        Calculates the feature_activation_bit_counts metadata attribute.
-        It's only calculated by Blocks, and returns None otherwise.
-        """
-        if not self.is_block:
-            return None
-
-        return self._calculate_feature_activation_bit_counts()
-
-    @abstractmethod
-    def _calculate_feature_activation_bit_counts(self) -> Optional[list[int]]:
-        """
-        Calculates the feature_activation_bit_counts metadata attribute.
-        Must only be implemented by Blocks.
-        """
-        raise NotImplementedError
-
     @property
     def hash_hex(self) -> str:
         """Return the current stored hash in hex string format"""
@@ -893,7 +875,12 @@ class BaseTransaction(ABC):
             #        happens include generating new mining blocks and some tests
             height = self.calculate_height() if self.storage else 0
             score = self.weight if self.is_genesis else 0
-            feature_activation_bit_counts = self.calculate_feature_activation_bit_counts()
+            kwargs: dict[str, Any] = {}
+
+            if self.is_block:
+                from hathor.transaction import Block
+                assert isinstance(self, Block)
+                kwargs['feature_activation_bit_counts'] = self.calculate_feature_activation_bit_counts()
 
             metadata = TransactionMetadata(
                 hash=self.hash,
@@ -901,7 +888,7 @@ class BaseTransaction(ABC):
                 height=height,
                 score=score,
                 min_height=0,
-                feature_activation_bit_counts=feature_activation_bit_counts
+                **kwargs
             )
             self._metadata = metadata
         if not metadata.hash:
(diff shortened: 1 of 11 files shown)


hathor/transaction/transaction.py
```diff
@@ -140,6 +140,10 @@ def calculate_min_height(self) -> int:
             self._calculate_my_min_height(),
         )
 
+    def calculate_feature_activation_bit_counts(self) -> Optional[list[int]]:
+        """Returns None, as Transactions do not have the feature_activation_bit_counts metadata attribute."""
+        return None
```
msbrogli, line 145: Should we `raise NotImplementedError`?

hathor/utils/int.py
```diff
@@ -0,0 +1,49 @@
...
+    [1, 0, 0]
+    >>> get_bit_list(0b10, min_size=1)
+    [0, 1]
+    >>> get_bit_list(0b111001010, min_size=10)
+    [0, 1, 0, 1, 0, 0, 1, 1, 1, 0]
+    """
+    assert n >= 0
+
+    shifts = range(n.bit_length())
+    bits = [n >> shift & 1 for shift in shifts]
```
msbrogli, line 46: Maybe the following code is easier to understand and faster:

```python
assert n >=0
bits = []
while n > 0:
    bits.append(n & 1)
    n >>= 1
if min_size is not None:
    while len(bits) < min_size:
        bits.append(0)
return bits
```

msbrogli (dismissed):

hathor/utils/int.py
glevco: Sure, changed in https://github.com/HathorNetwork/hathor-core/commit/8a97f4805e9aab445d6a0f5c99df44dcf335f97f

hathor/transaction/transaction.py
glevco: Yes. I had to change the structure a bit to be able to support this, I created an internal, abstract `BaseTransaction._calculate_feature_activation_bit_counts()` that is only implemented by `Block`, and a public `BaseTransaction.calculate_feature_activation_bit_counts()` that is actually implemented in `BaseTransaction` to calls the block's implementation.

Done in https://github.com/HathorNetwork/hathor-core/commit/8a97f4805e9aab445d6a0f5c99df44dcf335f97f

hathor/utils/int.py
msbrogli: Great! Thanks!

hathor/transaction/base_transaction.py
```diff
@@ -301,6 +301,24 @@ def calculate_height(self) -> int:
     def calculate_min_height(self) -> int:
         raise NotImplementedError
 
+    def calculate_feature_activation_bit_counts(self) -> Optional[list[int]]:
```
msbrogli, line 304: Why do we need it on `BaseTransaction`? Why not have it only on `Block`?

msbrogli (approved):

hathor/transaction/base_transaction.py
glevco: We don't, I just wanted to prevent `if self.is_block` conditionals in `BaseTransaction`, but I guess it's fine, I saw that it's used in other methods. Changed in https://github.com/HathorNetwork/hathor-core/pull/633/commits/3008ef370205679764c275db0ad2ba3087a9170c

msbrogli (dismissed):

hathor/transaction/transaction.py
msbrogli: Ok. Thanks!

hathor/transaction/base_transaction.py
msbrogli: Ok. Thanks!

msbrogli (approved):

PR #636, opened 2023-05-29
refactor: implement typed RunNodeArgs

glevco:
### Acceptance Criteria

- Implement new `from_namespace()` utils function to instantiate typed CLI args from `argparse.Namespace`
- Implement new `RunNodeArgs` named tuple to provide typed `run_node` CLI args

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -20,6 +20,7 @@ from typing import Any, Callable
 from pydantic import ValidationError
 from structlog import get_logger
 
+from hathor.cli.run_node_args import RunNodeArgs
 from hathor.conf import TESTNET_SETTINGS_FILEPATH, HathorSettings
 from hathor.exception import PreInitializationError
 
@@ -28,7 +29,7 @@ logger = get_logger()
 
 
 class RunNode:
-    UNSAFE_ARGUMENTS: list[tuple[str, Callable[[Namespace], bool]]] = [
+    UNSAFE_ARGUMENTS: list[tuple[str, Callable[[RunNodeArgs], bool]]] = [
         ('--test-mode-tx-weight', lambda args: bool(args.test_mode_tx_weight)),
         ('--enable-crash-api', lambda args: bool(args.enable_crash_api)),
         ('--x-sync-bridge', lambda args: bool(args.x_sync_bridge)),
@@ -38,6 +39,10 @@ class RunNode:
 
     @classmethod
     def create_parser(cls) -> ArgumentParser:
+        """
+        Create a new parser with the run_node CLI arguments.
+        Arguments must also be added to hathor.cli.run_node_args.RunNodeArgs
+        """
         from hathor.cli.util import create_parser
         parser = create_parser()
 
@@ -109,12 +114,12 @@ class RunNode:
         parser.add_argument('--config-yaml', type=str, help='Configuration yaml filepath')
         return parser
 
-    def prepare(self, args: Namespace, *, register_resources: bool = True) -> None:
+    def prepare(self, *, register_resources: bool = True) -> None:
         from setproctitle import setproctitle
-        setproctitle('{}hathor-core'.format(args.procname_prefix))
+        setproctitle('{}hathor-core'.format(self._args.procname_prefix))
 
-        if args.recursion_limit:
-            sys.setrecursionlimit(args.recursion_limit)
+        if self._args.recursion_limit:
+            sys.setrecursionlimit(self._args.recursion_limit)
         else:
             sys.setrecursionlimit(5000)
 
@@ -125,7 +130,7 @@ class RunNode:
                 print('Maximum number of open file descriptors is too low. Minimum required is 256.')
                 sys.exit(-2)
 
-        self.check_unsafe_arguments(args)
+        self.check_unsafe_arguments()
         self.check_python_version()
 
         from hathor.util import reactor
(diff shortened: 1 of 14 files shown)


hathor/cli/run_node.py
```diff
@@ -109,12 +110,12 @@ def create_parser(cls) -> ArgumentParser:
         parser.add_argument('--config-yaml', type=str, help='Configuration yaml filepath')
         return parser
```
jansegre, line 111: Could you add a comment here pointing to `RunNodeArgs` so we don't forget to add new relevant arguments there?

hathor/cli/run_node_args.py
```diff
@@ -0,0 +1,64 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import List, NamedTuple, Optional
+
+
+class RunNodeArgs(NamedTuple):
```
jansegre, line 18: Could you add a description here pointing to `RunNode.create_parser`, where the arguments are defined. So it's easier to see where to add them?

hathor/utils/arg_parse.py
```diff
@@ -0,0 +1,24 @@
...
+from argparse import Namespace
+from typing import NamedTuple, Type, TypeVar
+
+from hathor.utils.named_tuple import validated_named_tuple_from_dict
+
+T = TypeVar('T', bound=NamedTuple)
+
+
+def from_namespace(named_tuple_type: Type[T], namespace: Namespace) -> T:
+    return validated_named_tuple_from_dict(named_tuple_type, vars(namespace))
```
jansegre, line 24: Is this in preparation for a generic use in the future? Otherwise wouldn't it be simpler to put this directly in `RunNodeArgs`? Like:

```python
class RunNodeArgs(NamedTuple):
    # ...
    @classmethod
    def from_namespace(cls, namespace: Namespace) -> 'RunNodeArgs':
        return validated_named_tuple_from_dict(cls, vars(namespace))
```

jansegre, line 24: I'm assuming `validated_named_tuple_from_dict` will ignore extra vars. Could you add a comment to highlight that it will silently ignore extra vars/parameters, so if you add arguments to `namespace` they will be ignored unless explicitly added to `RunNodeArgs`. I agree it's the best behavior though, because many subcommands have extra arguments that only make sense to be ignored here.

jansegre (changes requested):

hathor/cli/run_node.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/636/commits/1b1978be1ca0983ea1284e6298c1ea288f9964f9

hathor/cli/run_node_args.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/636/commits/1b1978be1ca0983ea1284e6298c1ea288f9964f9

hathor/utils/arg_parse.py
glevco: Yes, I put it there thinking about migrating all CLI commands to this new approach. Do you think it's overkill?

jansegre: It makes sense, but also, they could inherit from `RunNodeArgs` and their custom parameters there. So the `from_namespace` would be re-used. Or for using it even on non-"run-node" subcommands, we could even do a ` class BaseArgs(NamedTuple)` class with just the `from_namespace` classmethod. But I'm OK leaving it as is for now and changing it when we get to implement it in the future.

glevco: You're right, sub commands that inherit from `RunNode` are not working. I'll implement tests to assert this, and then change the code to support them.

Also, I just noticed that the error I sent above is not from pydantic, but from `NamedTuple`. So I'll have to find another way to fix it.

glevco: In https://github.com/HathorNetwork/hathor-core/pull/636/commits/483db339af90daafb6a8202d8274138b34449e59 I changed the implementation to use pydantic's `BaseModel` directly, instead of using a `NamedTuple` and then validating it with pydantic. By doing this, we have more flexibility.

The problem is fixed, and I've added tests to assert them

jansegre (approved):

jansegre (approved):

tests/cli/test_shell.py
```diff
@@ -11,7 +11,7 @@ class ShellTest(unittest.TestCase):
     # In this case we just want to go through the code to see if it's okay
 
     def test_shell_execution_memory_storage(self):
-        shell = Shell(argv=['--memory-storage'])
+        shell = Shell(argv=['--memory-storage', '--', '--extra-arg'])
```
msbrogli, line 14: Should we allow extra unknown args?

hathor/cli/run_node_args.py
```diff
@@ -0,0 +1,72 @@
...
+#  limitations under the License.
+
+from typing import List, Optional
+
+from pydantic import Extra
+
+from hathor.utils.pydantic import BaseModel
+
+
+class RunNodeArgs(BaseModel, extra=Extra.allow):
```
msbrogli, line 22: I understand we want to allow extra arguments for some commands. But how can we enforce that there is no unknown arguments?

tests/cli/test_shell.py
glevco: Current `Shell` behavior allows any kind of extra argument, so that's what I reflected in tests. But read https://github.com/HathorNetwork/hathor-core/pull/636#discussion_r1229700949

hathor/cli/run_node_args.py
msbrogli: Thanks for the explanation.

tests/cli/test_shell.py
msbrogli: Got it. Thanks!

msbrogli (approved):

PR #640, opened 2023-05-30
refactor(builder): Build IndexesManager on builders

msbrogli:
### Acceptance criteria

1. Build `IndexesManager` on builders.
2. Remove default value for `Metrics.tx_storage`.

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -24,14 +24,19 @@ from hathor.consensus import ConsensusAlgorithm
 from hathor.event import EventManager
 from hathor.event.storage import EventMemoryStorage, EventRocksDBStorage, EventStorage
 from hathor.event.websocket import EventWebsocketFactory
-from hathor.indexes import IndexesManager
+from hathor.indexes import IndexesManager, MemoryIndexesManager, RocksDBIndexesManager
 from hathor.manager import HathorManager
 from hathor.p2p.manager import ConnectionsManager
 from hathor.p2p.peer_id import PeerId
 from hathor.pubsub import PubSubManager
 from hathor.storage import RocksDBStorage
 from hathor.stratum import StratumFactory
-from hathor.transaction.storage import TransactionMemoryStorage, TransactionRocksDBStorage, TransactionStorage
+from hathor.transaction.storage import (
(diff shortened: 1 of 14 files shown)


hathor/builder/builder.py
```diff
@@ -299,28 +306,51 @@ def _get_p2p_manager(self) -> ConnectionsManager:
...
-                rocksdb_storage,
-                use_memory_indexes=use_memory_index,
-                **kwargs
-            )
+        if self._tx_storage_cache:
+            reactor = self._get_reactor()
+            kwargs: Dict[str, Any] = {}
+            if self._tx_storage_cache_capacity is not None:
+                kwargs['capacity'] = self._tx_storage_cache_capacity
+            self._tx_storage = TransactionCacheStorage(self._tx_storage, reactor, indexes=cache_indexes, **kwargs)
```
glevco, line 351: It looks like you don't need the `cache_indexes` variable, you could just use `indexes` and remove it from line 333. Also, you could use `dict` instead of `Dict` for the `kwargs` typing

hathor/builder/cli_builder.py
```diff
@@ -112,15 +114,25 @@ def create_manager(self, reactor: PosixReactorBase, args: Namespace) -> HathorMa
...
+
+            # Initialize indexes manager.
+            if args.memory_indexes:
+                indexes = MemoryIndexesManager()
+            else:
+                indexes = RocksDBIndexesManager(self.rocksdb_storage)
+
+            kwargs = {}
+            if not args.cache:
+                # We should only pass indexes if cache is disable. Otherwise,
```
glevco, line 126: Typo: `disabled`

glevco (dismissed):

jansegre (approved):

hathor/builder/cli_builder.py
msbrogli: Fixed. Thanks!

hathor/builder/builder.py
msbrogli: Nice catch! Fixed. Thanks!

jansegre (approved):

glevco (approved):

PR #641, opened 2023-06-01
fix(simulator): Patch BaseTransaction.resolve() to prevent updating timestamp

msbrogli:
## Acceptance criteria

1. Patch `BaseTransaction.resolve()` to prevent updating timestamp.

Notice that `BaseTransaction.resolve()` calls `BaseTransaction.start_mining()`, which updates the timestamp using `time.time()` if mining takes more than two seconds. This might be causing tests to randomly fail.

The reasons for tests to randomly fail is that `Transaction.start_mining()` is using `time.time()` to get the timestamp (instead of `reactor.seconds()`). As the reactor time is not related to the current timestamp, the transaction timestamp can be too much in the future raising a `InvalidNewTransaction` exception.

For more details, check out this part of the code: https://github.com/HathorNetwork/hathor-core/blob/master/hathor/manager.py#L976-L982

### DIFF
--- a/hathor/simulator/simulator.py
+++ b/hathor/simulator/simulator.py
@@ -56,6 +56,7 @@ class Simulator:
         Patches:
 
         - disable pow verification
+        - disable Transaction.resolve method
         - set DAA test-mode to DISABLED (will actually run the pow function, that won't actually verify the pow)
         - override AVG_TIME_BETWEEN_BLOCKS to 64
         """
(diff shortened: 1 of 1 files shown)


hathor/simulator/simulator.py
```diff
@@ -65,9 +65,17 @@ def verify_pow(self: BaseTransaction, *args: Any, **kwargs: Any) -> None:
             assert self.hash is not None
             logger.new().debug('Skipping BaseTransaction.verify_pow() for simulator')
 
+        def resolve(self: BaseTransaction, update_time: bool = True) -> bool:
```
glevco, line 68: Missing update on `_apply_patches()` docstring

glevco (approved):
Just curious, why would the timestamp update make tests fail?

msbrogli:
> Just curious, why would the timestamp update make tests fail?

Because `Transaction.start_mining()` is using `time.time()` to get the timestamp (instead of `reactor.seconds()`). As the reactor time is not related to the current timestamp, the transaction timestamp can be too much in the future raising a `InvalidNewTransaction` exception.

Check out this part of the code: https://github.com/HathorNetwork/hathor-core/blob/master/hathor/manager.py#L976-L982

hathor/simulator/simulator.py
msbrogli: Fixed! Thanks!

glevco (approved):

jansegre (commented):
Do we still need the timestamp update? Couldn't we just remove it? Or default it to not doing an update and requiring a "timestamp source", like a reactor to do it?

Mostly I'd prefer to avoid doing patches.

PR #644, opened 2023-06-01
feat(feature-activation): implement block ancestor optimization

glevco:
### Acceptance Criteria

- Optimize `FeatureService._get_ancestor_at_height()` by using the height index if the block is in the best blockchain.
- Implement `create_peer_artifacts()` method on simulator
- Implement `TransactionStorage.get_transaction_by_height()`
- Implement feature activation simulation test

### DIFF
--- a/hathor/feature_activation/feature_service.py
+++ b/hathor/feature_activation/feature_service.py
@@ -18,13 +18,15 @@ from hathor.feature_activation.model.feature_description import FeatureDescripti
 from hathor.feature_activation.model.feature_state import FeatureState
 from hathor.feature_activation.settings import Settings as FeatureSettings
 from hathor.transaction import Block
+from hathor.transaction.storage import TransactionStorage
 
 
 class FeatureService:
-    __slots__ = ('_feature_settings',)
+    __slots__ = ('_feature_settings', '_tx_storage')
 
-    def __init__(self, *, feature_settings: FeatureSettings) -> None:
+    def __init__(self, *, feature_settings: FeatureSettings, tx_storage: TransactionStorage) -> None:
         self._feature_settings = feature_settings
+        self._tx_storage = tx_storage
 
     def is_feature_active(self, *, block: Block, feature: Feature) -> bool:
         """Returns whether a Feature is active at a certain block."""
@@ -45,7 +47,7 @@ class FeatureService:
         offset_to_boundary = height % self._feature_settings.evaluation_interval
         offset_to_previous_boundary = offset_to_boundary or self._feature_settings.evaluation_interval
         previous_boundary_height = height - offset_to_previous_boundary
-        previous_boundary_block = _get_ancestor_at_height(block=block, height=previous_boundary_height)
+        previous_boundary_block = self._get_ancestor_at_height(block=block, height=previous_boundary_height)
         previous_state = self.get_state(block=previous_boundary_block, feature=feature)
 
         if offset_to_boundary != 0:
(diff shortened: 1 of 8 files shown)


hathor/feature_activation/feature_service.py
```diff
@@ -133,17 +135,30 @@ def get_bits_description(self, *, block: Block) -> dict[Feature, FeatureDescript
...
+            f"ancestor height must be lower than the block's height: {height} >= {block.get_height()}"
+        )
+
+        metadata = block.get_metadata()
+
+        if not metadata.voided_by and (ancestor := self._tx_storage.get_transaction_by_height(height)):
+            assert isinstance(ancestor, Block)
+            return ancestor
+
+        return _get_ancestor_iteratively(block=block, ancestor_height=height)
```
msbrogli, line 153: This might be slow. When in the vertex life cycle this method is called? I feel that this call should be triggered by the consensus so we can skip it in specific cases.

msbrogli (approved):

hathor/feature_activation/feature_service.py
glevco: > When in the vertex life cycle this method is called?

Whenever some code checks for the state of a feature at a certain block, that is, calls `FeatureService.is_feature_active()`, this method will be called _if_ that block is not in the best blockchain at the moment.

Just to be clear, this won't be called by any code that is running now, only when we start using the feature activation for some feature. It is indeed a bit slow, but since it's called at most once per block, and only for voided blocks, it may not have a big impact.

There's still at least one known optimization to be done though, described in the TODO.

jansegre (dismissed):

tests/feature_activation/test_feature_simulation.py
```diff
@@ -0,0 +1,194 @@
...
+            timeout_height=60,
+            activate_on_timeout=True,
+            version='0.0.0'
+        )
+    }
+)
+
+
+class BaseTestFeatureSimulation(SimulatorTestCase):
+    seed_config = 5604003716498505253
```
msbrogli, line 45: Why are you fixing the seed?

msbrogli (changes requested):

hathor/feature_activation/feature_service.py
msbrogli: I'm wondering whether it can be exploited in a DoS attack, for instance. But, if I got it right, it will be used only when a feature is checked and the block height is within its window, right?

tests/feature_activation/test_feature_simulation.py
glevco: If I don't, the hash of mined blocks becomes different, and the hash is asserted in the tests

hathor/feature_activation/feature_service.py
glevco: > it will be used only when a feature is checked

Correct, and also only if the block is voided. If it's not voided, the height index is used instead.

> and the block height is within its window

What do you mean?

tests/feature_activation/test_feature_simulation.py
msbrogli: I guess you can't count on that because future changes on the full node might have a similar impact. We've had the same issue on other tests. Check them out for ways to workaround a fixed seed:

- https://github.com/HathorNetwork/hathor-core/blob/master/tests/consensus/test_soft_voided.py
- https://github.com/HathorNetwork/hathor-core/blob/master/tests/consensus/test_soft_voided3.py
- https://github.com/HathorNetwork/hathor-core/blob/master/tests/consensus/test_soft_voided4.py

hathor/feature_activation/feature_service.py
msbrogli: I meant that this code won't be executed for blocks outside the feature activation window, right?

tests/feature_activation/test_feature_simulation.py
glevco: Oh, ok. Then maybe I'll change it so hashes are not asserted, only the height (which is what is actually important to test)

hathor/feature_activation/feature_service.py
glevco: Correct, because feature states will be cached for the previous boundary blocks, so this won't be called for any block before the previous boundary (except on the first run).

jansegre (dismissed):

tests/feature_activation/test_feature_simulation.py
msbrogli: Thanks!

msbrogli (approved):

hathor/feature_activation/feature_service.py
glevco: I created an issue to review this point: https://github.com/HathorNetwork/hathor-core/issues/670

PR #647, opened 2023-06-02
fix(events): fix bugs in the event stream and improve tests

glevco:
### Acceptance Criteria

- Add missing docstrings and typings
- Fix parsing bug in `EventData` and add tests for it
- Fix edge cases bugs in `EventWebsocketProtocol` and add tests for them
- Implement `BaseEventSimulationTester` to help writing event queue feature tests, and new tests using it

PR #646, opened 2023-06-01
tests: Skip flaky event tests

msbrogli:
These flaky event tests have been failing a lot recently. So I think it's best to skip them for now while they are fixed in another PR.

## Acceptance criteria

1. Skip flaky event tests.

PR #649, opened 2023-06-04
tests(events): migrate flaky event simulation tests to new format

glevco:
Some event tests were skipped in https://github.com/HathorNetwork/hathor-core/pull/646 for not working correctly. This PR reimplements those tests using the new format introduced by https://github.com/HathorNetwork/hathor-core/pull/647.

### Acceptance Criteria

- Remove previous event scenarios simulation tests and create new ones
- Remove `DummyMiner`, as it was probably the cause for the faulty tests, and using simulator triggers is a much better approach


hathor/simulator/miner/dummy_miner.py
glevco, line 1: @msbrogli I remember that when I created the `DummyMiner`, I implemented the `AbstractMiner` base class for it, abstracting methods from the existing miner simulator, and you raised questions about its necessity. Now that `DummyMiner` is removed, there's only one miner simulator that inherits from `AbstractMiner`, so we could remove it too. What do you think?

msbrogli: I'm ok either removing or keeping it. I'm more curious on why the `DummyMiner` was causing random test failures.

msbrogli (approved):

jansegre (approved):

hathor/simulator/miner/dummy_miner.py
glevco: I'm also curious, but unfortunately I can't pinpoint the exact reason. It was some weird behavior related to how the code interacted with twisted, that I can only assume was originated on the DummyMiner. When I noticed that using triggers was a much better option, I tested and it worked, so I didn't invest time investigating the DummyMiner. There's nothing it does that can't be done with triggers, so there's no reason to keep it as it has been proven faulty, and I don't see a lot of benefit on investigating it further

msbrogli: I agree it's not worth diving deep. Let's just remove it to prevent others from using it.

msbrogli (approved):

jansegre (approved):

msbrogli (approved):

jansegre (approved):

PR #652, opened 2023-06-05
refactor(events): improve events during the load phase

glevco:
### Acceptance Criteria

- Change handling of events during the load phase, so it's not async anymore
  - Remove `HathorEvents.LOAD_FINISHED` as it's no used anymore
- Add progress tracker for handling of load phase events
- Implement `EventStorage.save_events()`


hathor/event/event_manager.py
```diff
@@ -251,10 +258,20 @@ def handle_load_phase_vertices(self, topological_iterator: Iterator[BaseTransact
...
-                delay=0,
-                callable=self._handle_event,
-                event_type=EventType.NEW_VERTEX_ACCEPTED,
-                event_args=EventArguments(tx=vertex)
-            )
+        def create_event_batch() -> Iterable[BaseEvent]:
+            assert self._event_ws_factory is not None
+            self.log.info('Starting creating events from existing database...')
+
+            for vertex in progress(topological_iterator, log=self.log, total=total_vertices):
```
msbrogli, line 265: Why have you moved away from an async operation to a synchronous one? Letting the event loop run is important to enable debugging and profiling APIs.

msbrogli (changes requested):

hathor/event/event_manager.py
glevco: Actually, I believe it should have been like this from the beginning. This happens during the initialization phase of the full node, just after indexes initialization, which is a very similar operation (an iteration over all vertices), that is also _not_ async. The reactor loop is not running at this moment. So the go to implementation should have been synchronous.

The reasoning behind implementing it previously as async, was the following:

1. The event order must be `LOAD_STARTED`, then all events for loaded vertices, then `LOAD_FINISHED`
2. Before, the `LOAD_STARTED` event was simply mapped from the `PubSub` event `MANAGER_ON_START`
3. If the implementation was synchronous, the event order would be incorrect: the `MANAGER_ON_EVENT` would be emitted only _after_ all vertices from the load phase. Then the order would be: `[load vertex events...], LOAD_STARTED, LOAD_FINISHED`

So that's the only reason this was made async. But being async caused a couple of problems:

- The load events initialization would happen _after_ the full node signaled readiness, which was weird, and could be seen in logs.
- The reactor loop is actually not running during full node initialization, `reactor.run()` is only called after it. So using `callLater` during that phase was a bit weird, it worked, it looked like it held the call for when the reactor starts running, but honestly I'm not 100% confident about this behavior. I _think_ some problems I've had with event ordering in tests were caused by this (that's why I also had to change test fixtures in this PR. Hopefully the order is now fixed and won't have to change again).

Looking at the reasoning steps, they way I fixed this was simply stop relying on the `PubSub` for the `LOAD_STARTED` event. Instead of mapping it from the existing `MANAGER_ON_START`, I simply trigger it synchronously with the new `EventManager.load_started()` method.

msbrogli: As discussed, we are about to change the whole initialization process to be async. But it's ok to merge the way it is and we change it to async later.

msbrogli (approved):

jansegre (approved):

PR #653, opened 2023-06-06
refactor: improve manager building on simulator

glevco:


hathor/simulator/simulator.py
```diff
@@ -141,38 +139,41 @@ def stop(self) -> None:
...
+        artifacts = self.create_artifacts(builder)
+        return artifacts.manager
+
+    def create_artifacts(self, builder: Builder) -> BuildArtifacts:
+        """
+        Returns build artifacts from a builder, after configuring it for simulator use.
+        You may get a builder from get_default_builder() for convenience.
+        """
         assert self._started, 'Simulator is not started.'
-        assert peer_id is not None  # XXX: temporary, for checking that tests are using the peer_id
```
glevco, line 155: @msbrogli do you know why we had this? I removed it because the behavior is unchanged: if the user of this method doesn't `set_peer_id()` on the builder he's passing, an exception will be raised.

If we were to allow a default `peer_id`, we could `set_peer_id(PeerId())` on `get_default_builder()`, and then the `builder` argument on `create_peer()` and `create_artifacts()` could be optional, simplifying its use for basic use cases:

```python
def create_artifacts(self, builder: Optional[Builder] = None) -> BuildArtifacts:
    builder = builder or self.get_default_builder()
    ...
```

msbrogli: I guess it was an old change to prevent using default peer ids. I don't feel we need it anymore.

msbrogli (approved):

jansegre (approved):

PR #663, opened 2023-06-14
fix(storage): incorrect validation store when using cache storage

jansegre:
## Acceptance criteria

- Refactor `TransactionStorage.allow_scope` to use getters/setters and have the actual value private;
- Make `TransactionCacheStorage` override the getters/setters so the external storage and internal storage have the same scope.

### DIFF
--- a/hathor/transaction/storage/transaction_storage.py
+++ b/hathor/transaction/storage/transaction_storage.py
@@ -101,7 +101,7 @@ class TransactionStorage(ABC):
         self._weakref_lock: Lock = Lock()
 
         # Flag to allow/disallow partially validated vertices.
-        self.allow_scope: TxAllowScope = TxAllowScope.VALID
+        self._allow_scope: TxAllowScope = TxAllowScope.VALID
 
         # Cache for the best block tips
         # This cache is updated in the consensus algorithm.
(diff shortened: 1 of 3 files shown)


hathor/transaction/storage/transaction_storage.py
```diff
@@ -137,6 +137,12 @@ def __init__(self) -> None:
                 raise ValueError(f'Duplicate migration name "{migration_name}"')
             migration_names.add(migration_name)
 
+    def set_allow_scope(self, allow_scope: TxAllowScope) -> None:
```
msbrogli, line 140: Missing docstring.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -137,6 +137,12 @@ def __init__(self) -> None:
...
+        self._allow_scope = allow_scope
+
+    def get_allow_scope(self) -> TxAllowScope:
```
msbrogli, line 143: Missing docstring.

msbrogli (approved):

hathor/transaction/storage/transaction_storage.py
msbrogli: Great! Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Great! Thanks!

msbrogli (approved):

PR #665, opened 2023-06-14
refactor: use None by default for height metadata

jansegre:
## Acceptance criteria

- `height` metadata type should change from `int` to `Optional[int]`
- initial value should go from `0` to `None`
- any code accessing the `height` metadata directly should preferably use `Block.get_height` method (for both ensuring that the vertex in question is a block and asserting that it is not None)

### DIFF
--- a/hathor/util.py
+++ b/hathor/util.py
@@ -520,8 +520,10 @@ def _tx_progress(iter_tx: Iterator['BaseTransaction'], *, log: 'structlog.stdlib
             log.warn('iterator was slow to yield', took_sec=dt_next)
 
         assert tx.hash is not None
-        tx_meta = tx.get_metadata()
-        h = max(h, tx_meta.height)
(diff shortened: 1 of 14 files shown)


hathor/util.py
```diff
@@ -520,8 +520,10 @@ def _tx_progress(iter_tx: Iterator['BaseTransaction'], *, log: 'structlog.stdlib
             log.warn('iterator was slow to yield', took_sec=dt_next)
 
         assert tx.hash is not None
-        tx_meta = tx.get_metadata()
-        h = max(h, tx_meta.height)
+        # XXX: this is informative and made to work with eithe partially/fully validated blocks/transactions
```
msbrogli, line 523: typo: either

msbrogli (approved):

msbrogli (approved):

hathor/util.py
msbrogli: Fixed! Thanks!

PR #672, opened 2023-06-16
feat(traversal): add BFS order traversal

jansegre:
## Acceptance criteria

- `BFSWalk` should be renamed to `BFSTimestampWalk` to reflect that it prioritizes by timestamp
- add `BFSOrderWalk` which uses actual neighbor order to yield

### DIFF
--- a/hathor/transaction/storage/traversal.py
+++ b/hathor/transaction/storage/traversal.py
@@ -15,12 +15,14 @@
 
 import heapq
 from abc import ABC, abstractmethod
+from collections import deque
 from itertools import chain
-from typing import TYPE_CHECKING, Any, Iterable, Iterator, Optional, Union
+from typing import TYPE_CHECKING, Iterable, Iterator, Optional, Union
 
 if TYPE_CHECKING:
     from hathor.transaction import BaseTransaction  # noqa: F401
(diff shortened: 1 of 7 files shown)


msbrogli (approved):

hathor/transaction/storage/traversal.py
```diff
@@ -160,13 +177,39 @@ def _pop_visit(self) -> 'BaseTransaction':
...
     """
-    to_visit: list['BaseTransaction']
+    _to_visit: list['BaseTransaction']
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._to_visit = []
+
+    def _is_empty(self) -> bool:
+        return not self._to_visit
```
glevco, line 209: It looks like this method is the same in all subclasses, what about moving it to the superclass?

glevco (approved):

hathor/transaction/storage/traversal.py
jansegre: The superclass doesn't know about `self._to_visit`, which has different types depending on the subclass, and in theory the subclass could use something else entirely. I'd say it's mostly a coincidence that this works out exactly the same on the three of them.

glevco: Ok!

PR #675, opened 2023-06-27
feat(feature-activation): new MUST_SIGNAL and LOCKED_IN states

glevco:
WIP


jansegre (dismissed):

hathor/feature_activation/feature_service.py
```diff
@@ -92,32 +91,33 @@ def _calculate_new_state(
...
+            if height < criteria.timeout_height and count >= threshold:
+                return FeatureState.LOCKED_IN
+
+            if (height + evaluation_interval >= criteria.timeout_height) and criteria.lock_in_on_timeout:
+                return FeatureState.MUST_SIGNAL
 
             return FeatureState.STARTED
 
+        if previous_state is FeatureState.MUST_SIGNAL:
+            return FeatureState.LOCKED_IN
```
msbrogli, line 113: Shouldn't we check the block height to change to LOCKED_IN?

msbrogli (changes requested):

hathor/feature_activation/feature_service.py
glevco: No, because the `MUST_SIGNAL` state is defined to always take exactly one evaluation interval. Since this method is only called for boundary blocks, it is guaranteed that after exactly one evaluation interval in `MUST_SIGNAL`, the feature will transition to `LOCKED_IN`, as it should. Any blocks _inside_ that evaluation interval, that is, between the boundary blocks, will continue to be in `MUST_SIGNAL` since they will not call this method and just propagate their parent's state instead.

But just to be clear, what was your thought reasoning?

msbrogli: I was concerned that this method might be called multiple times. I feel it's worth adding a comment about what you said here to clarify for future code readers.

msbrogli: I just checked the assert for boundary blocks. I feel it's worth adding a note about it on the docstring.

msbrogli (approved):

hathor/feature_activation/feature_service.py
glevco: Sure! I updated the docstring and added a comment in https://github.com/HathorNetwork/hathor-core/pull/675/commits/a91c88ca00cc933c38d653849c346c73466e8906

msbrogli: Great! Thanks!

msbrogli (approved):

jansegre (approved):

Issue #679, opened 2023-06-28
Test GET-BEST-BLOCKCHAIN capability


alexruzenhack:
Design: https://github.com/HathorNetwork/internal-rfcs/pull/2

alexruzenhack:
Ref: https://github.com/HathorNetwork/hathor-core/pull/686

Issue #678, opened 2023-06-28
Implement GET-BEST-BLOCKCHAIN in the ReadyState


alexruzenhack:
Design: https://github.com/HathorNetwork/internal-rfcs/pull/2

alexruzenhack:
Implemented at: https://github.com/HathorNetwork/hathor-core/pull/686

Issue #677, opened 2023-06-28
Add GET-BEST-BLOCKCHAIN capability


alexruzenhack:
- save capability in the protocol level

Design: https://github.com/HathorNetwork/internal-rfcs/pull/2

alexruzenhack:
Implemented at: https://github.com/HathorNetwork/hathor-core/pull/686

PR #686, opened 2023-06-29
feat: add get-best-blockchain capability

alexruzenhack:
### Acceptance criteria
- Add capability `GET-BEST-BLOCKCHAIN`
- Make the new capability default
- Add `capabilities` to `HathorProtocol` in `HelloState`
- Implement commands `GET-BEST-BLOCKCHAIN` and `BEST-BLOCKCHAIN` in `ReadyState`
- Test all together

Closes:
- https://github.com/HathorNetwork/hathor-core/issues/677
- https://github.com/HathorNetwork/hathor-core/issues/678
- https://github.com/HathorNetwork/hathor-core/issues/679


hathor/p2p/states/ready.py
```diff
@@ -180,3 +206,32 @@ def handle_pong(self, payload: str) -> None:
...
+        self.send_message(ProtocolMessages.GET_BLOCKS, '10')
+
+    def handle_get_blocks(self, payload: str) -> None:
+        """ Executed when a GET-BLOCKS command is received. It just responds with
+        a list of the latest N blocks from the best blockchain.
+        """
+        # TODO: implement self.get_best_blockchain_blocks(payload) 
+        n = int(payload)
+        assert n > 0
+        assert n < 100
```
jansegre, line 222: because `n` comes from the remote request, `assert` is not correct, we should handle and respond appropriately, something like:

    if not (0 < n < 100):
        self.send_and_close('invalid N')
        return

hathor/p2p/states/ready.py
```diff
@@ -180,3 +206,32 @@ def handle_pong(self, payload: str) -> None:
...
+
+        block_tips = self.protocol.connections.manager.tx_storage.get_block_tips();
```
jansegre, line 224: The tips will technically return a list, because we still accept best block ties. I think we shouldn't expose block ties and instead maybe use the height index, which is free from ties, maybe the even add a `get_best_n_blocks` to the height index and use that.

hathor/p2p/states/ready.py
alexruzenhack: Done.

hathor/p2p/states/ready.py
alexruzenhack: Improved.

hathor/transaction/block.py
```diff
@@ -147,7 +147,18 @@ def _get_previous_feature_activation_bit_counts(self) -> list[int]:
         return parent_block.get_feature_activation_bit_counts()
 
     def get_next_block_best_chain_hash(self) -> Optional[bytes]:
```
alexruzenhack, line 149: In earlier implementation I have used this method hoping to traverse the blockchain from higher height to lower hight and it gave my the opposite. I found the description I little bit counter intuitive and decided to improve its description to let clear the in which direction a traverse would go.

hathor/p2p/protocol.py
```diff
@@ -90,6 +90,7 @@ class WarningFlags(str, Enum):
     diff_timestamp: Optional[int]
     idle_timeout: int
     sync_version: Optional[SyncVersion]  # version chosen to be used on this connection
+    capabilities: list[str]  # capabilities received from the peer in HelloState
```
msbrogli, line 93: Should we use `set[str]`?

hathor/p2p/states/ready.py
```diff
@@ -30,6 +32,11 @@
 
 logger = get_logger()
 
+settings = HathorSettings()
+
+
+BlockInfo = namedtuple("BlockInfo", "hash height weight")
```
msbrogli, line 38: Suggestion:

```python
class BlockInfo(NamedTuple):
    hash: VertexId
    height: int
    weight: float
```

hathor/p2p/states/ready.py
```diff
@@ -57,16 +64,29 @@ def __init__(self, protocol: 'HathorProtocol') -> None:
...
             ProtocolMessages.GET_PEERS: self.handle_get_peers,
             ProtocolMessages.PEERS: self.handle_peers,
-
             # Other messages are added by the sync manager.
         })
 
+        # if the peer has the GET-BEST-BLOCKCHAIN capability
+        if settings.CAPABILITY_GET_BEST_BLOCKCHAIN in protocol.capabilities:
+            # set the loop to get the best blockchain from the peer
+            self.lc_get_best_blockchain = LoopingCall(self.send_get_best_blockchain)
```
msbrogli, line 82: Set `self.lc_get_best_blockchain: Optional[LoopingCall] = None` before the if block. So we can check whether it's none or not.

hathor/p2p/states/ready.py
```diff
@@ -180,3 +200,48 @@ def handle_pong(self, payload: str) -> None:
         self.ping_min_rtt = min(self.ping_min_rtt, self.ping_rtt)
         self.ping_start_time = None
         self.log.debug('rtt updated', rtt=self.ping_rtt, min_rtt=self.ping_min_rtt)
+
+    def send_get_best_blockchain(self, nBlocks: str = '10') -> None:
+        """ Send a GET-BEST-BLOCKCHAIN command, requesting a list of latest
```
msbrogli, line 205: Typo: ... requesting a list of the latest N blocks of the best blockchain."

hathor/p2p/states/ready.py
```diff
@@ -180,3 +200,48 @@ def handle_pong(self, payload: str) -> None:
...
+        N blocks from the best blockchain.
+        """
+        self.send_message(ProtocolMessages.GET_BEST_BLOCKCHAIN, nBlocks)
+
+    def handle_get_best_blockchain(self, payload: str) -> None:
+        """ Executed when a GET-BEST-BLOCKCHAIN command is received.
+        It just responds with a list N blocks from the best blockchain.
```
msbrogli, line 212: Ordered by height? Ascending or descending?

hathor/p2p/states/ready.py
```diff
@@ -180,3 +200,48 @@ def handle_pong(self, payload: str) -> None:
...
+        """
+        nBlocks = int(payload)
```
msbrogli, line 214: Capture the exception and close the connection if payload is not an int, otherwise it will generate an unhandled exception.

hathor/p2p/states/ready.py
```diff
@@ -180,3 +200,48 @@ def handle_pong(self, payload: str) -> None:
...
+        if not (0 < nBlocks < 101):
```
msbrogli, line 215: Add `MAX_BEST_BLOCKCHAIN_BLOCKS` to the class. It allows us to change this parameter through sysctl in the future.

hathor/p2p/states/ready.py
msbrogli, line 82: I couldn't find where you are starting this looping call. I was looking for the interview between calls.

hathor/p2p/states/ready.py
```diff
@@ -180,3 +200,48 @@ def handle_pong(self, payload: str) -> None:
...
+        """ Send a BEST-BLOCKCHAIN command with a best blockchain of N blocks.
+        """
+        self.send_message(ProtocolMessages.BEST_BLOCKCHAIN, json_dumps(best_blockchain))
+
+    def handle_best_blockchain(self, payload: str) -> None:
+        """ Executed when a BEST-BLOCKCHAIN command is received. It updates
+        the best blockchain.
+        """
+        restored_blocks = json_loads(payload)
+        best_blockchain = [BlockInfo(*block) for block in restored_blocks]
```
msbrogli, line 246: Maybe rename `block` to `block_info_raw`? Don't you have to validate the types are correct?

hathor/p2p/states/ready.py
```diff
@@ -180,3 +200,48 @@ def handle_pong(self, payload: str) -> None:
         self.ping_min_rtt = min(self.ping_min_rtt, self.ping_rtt)
         self.ping_start_time = None
         self.log.debug('rtt updated', rtt=self.ping_rtt, min_rtt=self.ping_min_rtt)
+
+    def send_get_best_blockchain(self, nBlocks: str = '10') -> None:
```
msbrogli, line 204: Why `nBlocks` is a string? Should it be an integer?

tests/p2p/test_get_best_blockchain.py
```diff
@@ -0,0 +1,137 @@
...
+from hathor.p2p.states.ready import BlockInfo
+from hathor.simulator import FakeConnection
+from hathor.simulator.trigger import StopAfterNMinedBlocks
+from tests import unittest
+from tests.simulation.base import SimulatorTestCase
+
+settings = HathorSettings()
+
+
+class BaseGetBestBlockchainTestCase(SimulatorTestCase):
```
msbrogli, line 13: I missed a test when one full node doesn't have the best blockchain capability.

msbrogli (changes requested):

hathor/p2p/states/ready.py
```diff
@@ -180,3 +237,61 @@ def handle_pong(self, payload: str) -> None:
...
+        """
+        self.send_message(ProtocolMessages.GET_BEST_BLOCKCHAIN, str(nBlocks))
+
+    def handle_get_best_blockchain(self, payload: str) -> None:
+        """ Executed when a GET-BEST-BLOCKCHAIN command is received.
+        It just responds with a list with N blocks from the best blockchain
+        in descending order.
+        """
+        try:
+            nBlocks = int(payload)
```
msbrogli, line 253: `n_blocks`

hathor/p2p/states/ready.py
```diff
@@ -180,3 +237,61 @@ def handle_pong(self, payload: str) -> None:
...
+        except ValueError:
+            param_type = type(payload).__name__
+            self.protocol.send_error_and_close_connection(
+                f'Invalid param type. \'payload\' should be an int but we got {param_type}.'
+            )
+            return
+
+        if not (0 < nBlocks < self.MAX_BEST_BLOCKCHAIN_BLOCKS):
+            self.protocol.send_error_and_close_connection('Invalid N to get best blockchain.')
```
msbrogli, line 262: `N out of bounds. Valid range: [1, self.MAX_BEST_BLOCKCHAIN_BLOCKS]`.

hathor/p2p/protocol.py
```diff
@@ -90,6 +90,7 @@ class WarningFlags(str, Enum):
     diff_timestamp: Optional[int]
     idle_timeout: int
     sync_version: Optional[SyncVersion]  # version chosen to be used on this connection
+    capabilities: set[str]  # capabilities received from the peer in HelloState
```
msbrogli, line 93: You must initialize it on `__init__()`.

hathor/p2p/states/ready.py
```diff
@@ -30,10 +32,47 @@
...
+settings = HathorSettings()
+
+
+class BlockInfo(NamedTuple):
+    hash_hex: str
+    height: int
+    weight: float
+
+    @staticmethod
+    def from_raw(block_info_raw):
```
msbrogli, line 44: Missing types.

hathor/p2p/states/ready.py
```diff
@@ -30,10 +32,47 @@
...
+        """ Instantiate BlockInfo from a string raw.
```
msbrogli, line 45: I feel this docstring is wrong since `block_info_raw` must be a `tuple[str, int, float]`.

hathor/p2p/states/hello.py
```diff
@@ -108,6 +108,9 @@ def handle_hello(self, payload: str) -> None:
             protocol.send_error_and_close_connection('Must have whitelist capability.')
             return
 
+        # another status can use the informed capabilities
+        protocol.capabilities = data['capabilities']
```
msbrogli, line 112: `set(data['capabilities'])`?

msbrogli (changes requested):

hathor/p2p/protocol.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Thanks!

msbrogli: Done!

hathor/p2p/states/ready.py
```diff
@@ -180,3 +238,64 @@ def handle_pong(self, payload: str) -> None:
...
+        block = tx_storage.get_best_block()
+
+        while len(best_blockchain) < n_blocks:
+            block_info = BlockInfo(
+                block.hash_hex,
+                block.get_height(),
+                block.weight)
+            best_blockchain.append(block_info)
+            # halt if genesis block
+            if block_info.height == 0:
```
msbrogli, line 278: Maybe `block.is_genesis`?

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/p2p/states/hello.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Done!

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/p2p/states/ready.py
msbrogli: Thanks!

hathor/types.py
```diff
@@ -21,3 +24,34 @@
 Timestamp = int         # NewType('Timestamp', int)
 TokenUid = VertexId     # NewType('TokenUid', VertexId)
 Amount = int            # NewType('Amount', int)
+
+
+class BlockInfo(NamedTuple):
```
alexruzenhack, line 29: Moved to this class to avoid polute the hathor manager.

hathor/manager.py
```diff
@@ -703,6 +707,29 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
         can_include = [i.data for i in can_include_intervals]
         return ParentTxs(max_timestamp, can_include, must_include)
 
+    def get_best_blockchain(self, n_blocks: int) -> list['BlockInfo']:
```
alexruzenhack, line 710: Extracted from read state.

hathor/conf/settings.py
```diff
@@ -393,6 +394,12 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
     # Maximum number of GET_TIPS delayed calls per connection while running sync.
     MAX_GET_TIPS_DELAYED_CALLS: int = 5
 
+    # Maximum number of blocks in the best blockchain list.
+    MAX_BEST_BLOCKCHAIN_BLOCKS: int = 100
+
+    # Time in seconds to request the best blockchain from peers.
+    BEST_BLOCKCHAIN_INTERVAL: int = 5  # seconds
```
alexruzenhack, line 401: Moved from ready state, this way the hathor manager can also use these values.

hathor/manager.py
```diff
@@ -703,6 +710,35 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
...
+        if not (0 < n_blocks <= settings.MAX_BEST_BLOCKCHAIN_BLOCKS):
+            raise ValueError(
+                f'N out of bounds. Valid range: [1, {settings.MAX_BEST_BLOCKCHAIN_BLOCKS}].'
+            )
+
+        block = self.tx_storage.get_best_block()
+        if self._latest_best_blockchain:
+            best_blockchain_tip = self._latest_best_blockchain[0]
+            if block.to_blockinfo() == best_blockchain_tip:
+                return self._latest_best_blockchain
```
alexruzenhack, line 726: Implemented the cache.

hathor/p2p/states/ready.py
```diff
@@ -180,3 +207,50 @@ def handle_pong(self, payload: str) -> None:
         self.ping_min_rtt = min(self.ping_min_rtt, self.ping_rtt)
         self.ping_start_time = None
         self.log.debug('rtt updated', rtt=self.ping_rtt, min_rtt=self.ping_min_rtt)
+
+    def send_get_best_blockchain(
+            self, n_blocks: int = settings.DEFAULT_BEST_BLOCKCHAIN_BLOCKS) -> None:
```
alexruzenhack, line 212: Added the default value as a setting.

hathor/types.py
```diff
@@ -21,3 +24,41 @@
...
+            raise ValueError(f"weight must be greater than 0. We got {weight}.")
+
+        return BlockInfo(hash_hex, height, weight)
+
+    def __eq__(self, other):
+        if isinstance(other, BlockInfo):
+            return (self.hash_hex == other.hash_hex and
+                    self.height == other.height and
+                    self.weight == other.weight)
+        return False
```
alexruzenhack, line 64: Implemented this equality operator to improve code readability.

hathor/transaction/block.py
```diff
@@ -284,6 +307,9 @@ def to_json_extended(self) -> dict[str, Any]:
 
         return json
 
+    def to_blockinfo(self) -> BlockInfo:
+        return BlockInfo(self.hash_hex, self.get_height(), self.weight)
+
```
alexruzenhack, line 312: Implemented this conversion to improve code readability.

hathor/manager.py
```diff
@@ -703,6 +710,35 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
         can_include = [i.data for i in can_include_intervals]
         return ParentTxs(max_timestamp, can_include, must_include)
 
+    def get_best_blockchain(self, n_blocks: int) -> list['BlockInfo']:
+        """ Get a list with N blocks from the best blockchain
+        in descending order.
+        """
+        if not (0 < n_blocks <= settings.MAX_BEST_BLOCKCHAIN_BLOCKS):
```
msbrogli, line 717: This is an internal method. We can add this protection on the p2p protocol and let this internal API receive any value for `n_blocks`. Maybe just add `assert n_blocks > 0`.

hathor/manager.py
msbrogli: Ok!

hathor/types.py
```diff
@@ -21,3 +24,41 @@
...
+        if height < 0:
+            raise ValueError(f"height must greater than or equal to 0. We got {height}.")
+        if not isinstance(weight, (float, int)):
+            raise ValueError(f"weight must be a float. We got {weight}.")
+        if not weight > 0:
+            raise ValueError(f"weight must be greater than 0. We got {weight}.")
+
+        return BlockInfo(hash_hex, height, weight)
+
+    def __eq__(self, other):
```
msbrogli, line 59: Why do you need it? As it is a named tuple, I feel this is exactly what it does by default.

msbrogli (changes requested):

hathor/manager.py
alexruzenhack: Ok. Resolved.

hathor/manager.py
```diff
@@ -703,6 +710,35 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
...
+            block_info = BlockInfo(
+                block.hash_hex,
+                block.get_height(),
+                block.weight)
+            best_blockchain.append(block_info)
+            if block.is_genesis:
+                break
+            block = block.get_block_parent()
+
+        self._latest_best_blockchain = best_blockchain
```
jansegre, line 739: A small observation: this cache isn't guaranteed to have a consistent length (calls with different `n_blocks` will lead to caches of different sizes). I don't thinks this is really an issue _here_, but it might be worth adding a comment saying that this isn't a problem.

hathor/p2p/messages.py
```diff
@@ -77,6 +77,12 @@ class ProtocolMessages(Enum):
     # Pong is a response to a PING command.
     PONG = 'PONG'
 
+    # Request a list of blocks from the best blockchain
+    GET_BEST_BLOCKCHAIN = 'GET_BEST_BLOCKCHAIN'
+
+    # Send back the blockchain requested
+    BEST_BLOCKCHAIN = 'BEST_BLOCKCHAIN'
```
jansegre, line 84: Please use a dash `-` instead of an underscore `_` for the string values, to keep consistency with the other commands.

hathor/manager.py
```diff
@@ -703,6 +710,35 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
...
+            best_blockchain_tip = self._latest_best_blockchain[0]
+            if block.to_blockinfo() == best_blockchain_tip:
+                return self._latest_best_blockchain
+
+        best_blockchain: list[BlockInfo] = []
+        while len(best_blockchain) < n_blocks:
+            block_info = BlockInfo(
+                block.hash_hex,
+                block.get_height(),
+                block.weight)
```
jansegre, line 733: You could use the `to_blockinfo()` conversion here.

hathor/manager.py
```diff
@@ -703,6 +710,35 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
...
+        if not (n_blocks > 0):
+            raise ValueError(
+                'Invalid N. N must be greater than 0.'
+            )
+
+        block = self.tx_storage.get_best_block()
+        if self._latest_best_blockchain:
+            best_blockchain_tip = self._latest_best_blockchain[0]
+            if block.to_blockinfo() == best_blockchain_tip:
+                return self._latest_best_blockchain
```
jansegre, line 726: There's a bug here, if `len(self._latest_best_blockchain)` is not `n_blocks` (because of a previous call with a different value) then it will return a list with the wrong length.

But since you can still make use of the cache when the memoized length is higher, you could do something like:

```python
best_blockchain_tip = self._latest_best_blockchain[0]
if block.hash_hex == best_block and n_blocks <= len(self._latest_best_blockchain):
    return self._latest_best_blockchain[:n_blocks]
```

This also has the benefit of always returning a copy of the cache and thus two different calls wouldn't return the same list object which could accidentally have been used in between. Like this is what would happen if you don't return a copy:

```
x = manager.get_best_blockchain(5)  # returns a list with 5 elements
x.pop()  # the list only has 4 elements now
y = manager.get_best_blockchain(5)  # returns a list with 4 elements
```

hathor/p2p/messages.py
alexruzenhack: Done.

hathor/manager.py
alexruzenhack: Done.

jansegre (approved):

jansegre (approved):

hathor/types.py
alexruzenhack, line 1: I removed the `__eq__` method.

jansegre (dismissed):

hathor/manager.py
```diff
@@ -703,6 +710,31 @@ def generate_parent_txs(self, timestamp: Optional[float]) -> 'ParentTxs':
...
+        """
+        if not (n_blocks > 0):
+            raise ValueError(
+                'Invalid N. N must be greater than 0.'
+            )
+
+        block = self.tx_storage.get_best_block()
+        if self._latest_best_blockchain:
+            best_block = self._latest_best_blockchain[0]
+            if block.hash_hex == best_block.hash_hex and n_blocks <= len(self._latest_best_blockchain):
```
msbrogli, line 725: Why are you using `hash_hex`? I feel that `block.hash` is the way to compare.

msbrogli (approved):

hathor/manager.py
alexruzenhack: Sure, binary comparison is better in this case. I will change.

alexruzenhack: Oh. I saw now. I have used the `hash_hex` because the `BlockInfo` only has `hash_hex`.

tests/p2p/test_protocol.py
```diff
@@ -409,12 +410,33 @@ def test_two_connections(self):
...
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
+        self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
```
alexruzenhack, line 431: @jansegre @msbrogli This is a little strange.

jansegre: This is a diff from your PR, you're free to change it.

hathor/transaction/block.py
```diff
@@ -284,6 +307,9 @@ def to_json_extended(self) -> dict[str, Any]:
 
         return json
 
+    def to_blockinfo(self) -> BlockInfo:
```
msbrogli, line 310: Missing docstring.

tests/p2p/test_protocol.py
jansegre: I'm not sure why it happens and if it's a bad thing, but at least I'd rather use:

```python
        for _ in range(19):
            self.assertAndStepConn(self.conn, b'^GET-BEST-BLOCKCHAIN')
```

hathor/p2p/states/ready.py
```diff
@@ -180,3 +207,55 @@ def handle_pong(self, payload: str) -> None:
...
+            self.protocol.send_error_and_close_connection(
+                f'N out of bounds. Valid range: [1, {settings.MAX_BEST_BLOCKCHAIN_BLOCKS}].'
+            )
+            return
+
+        try:
+            best_blockchain = self.protocol.node.get_best_blockchain(n_blocks)
+        except ValueError as e:
+            self.protocol.send_error_and_close_connection(msg=str(e))
+            return
```
msbrogli, line 240: `raise`?

hathor/p2p/states/ready.py
```diff
@@ -180,3 +207,55 @@ def handle_pong(self, payload: str) -> None:
...
+
+        if not (0 < n_blocks <= settings.MAX_BEST_BLOCKCHAIN_BLOCKS):
+            self.protocol.send_error_and_close_connection(
+                f'N out of bounds. Valid range: [1, {settings.MAX_BEST_BLOCKCHAIN_BLOCKS}].'
+            )
+            return
+
+        try:
+            best_blockchain = self.protocol.node.get_best_blockchain(n_blocks)
+        except ValueError as e:
```
msbrogli, line 238: We don't have to handle this exception.

tests/p2p/test_get_best_blockchain.py
```diff
@@ -0,0 +1,357 @@
...
+        protocol1 = connected_peers2[0]
+        state1 = protocol1.state
+        self.assertIsInstance(state1, ReadyState)
+
+        # assert compliance with N blocks inside the boundaries
+        state1.send_get_best_blockchain(n_blocks='1')
+        self.simulator.run(60)
+        self.assertFalse(conn12.tr1.disconnecting)
+
+        state2.send_get_best_blockchain(n_blocks='100')
```
msbrogli, line 114: `'100'`? There are other parts of this test that are still using `str` instead of `int`.

hathor/manager.py
```diff
@@ -213,6 +217,9 @@ def __init__(self,
         # This is included in some logs to provide more context
         self.environment_info = environment_info
 
+        # Memoize latest best blockchain
+        self._latest_best_blockchain: list[BlockInfo] = []
```
msbrogli, line 221: It seems we are not using it anymore, right? Maybe delete it.

msbrogli (changes requested):

hathor/p2p/states/ready.py
```diff
@@ -57,16 +63,32 @@ def __init__(self, protocol: 'HathorProtocol') -> None:
         # Minimum round-trip time among PING/PONG.
         self.ping_min_rtt: float = inf
 
+        # The last blocks from the best blockchain in the peer
+        self.best_blockchain: list[HeightInfo] = []
```
jansegre, line 67: I prefer for this to have 'remote' in the name.

jansegre (approved):

hathor/p2p/states/ready.py
msbrogli: We usually use the `peer_` prefix. So, in this case, it would be `peer_best_blockchain`.

alexruzenhack: Pronto:
513af820bfd27578f345d4cee37f823f23ae2a08

tests/p2p/test_get_best_blockchain.py
alexruzenhack: Done

hathor/manager.py
alexruzenhack: Removed.

jansegre (approved):

msbrogli (approved):

PR #688, opened 2023-07-03
feat(feature-activation): bit signaling POC

glevco:
This is a POC for the RFC described in https://github.com/HathorNetwork/rfcs/pull/59.

After the RFC approval, new PRs will be created for the actual complete implementation. **This PR is not intended to be merged.**

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -39,6 +39,7 @@ from hathor.exception import (
     RewardLockedError,
     SpendingVoidedError,
 )
+from hathor.feature_activation.bit_signaling_service import BitSignalingService
 from hathor.mining import BlockTemplate, BlockTemplates
 from hathor.p2p.manager import ConnectionsManager
 from hathor.p2p.peer_discovery import PeerDiscovery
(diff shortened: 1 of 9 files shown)


hathor/manager.py
```diff
@@ -822,12 +828,13 @@ def _make_block_template(self, parent_block: Block, parent_txs: 'ParentTxs', cur
             parents_any=parents_any,
             height=height,
             score=sum_weights(parent_block_metadata.score, weight),
+            signal_bits=self._bit_signaling_service.get_signal_bits()
```
jansegre, line 831: Make this generic over the height, so that the signal bits aren't always for the best block. A warning could be good if this isn't using the best height (or even a settings toggle to assert or log a warning, so that tests log a warning, and mainnet does an assert). The exception for when this isn't used for generating blocks at the best height is `make_custom_block_template` being used in re-org tests, where `parent_block` will not be a block head.

glevco: Nice catch, I've updated it in https://github.com/HathorNetwork/hathor-core/pull/688/commits/f6ec17bd1e573875f21a9e8f0913cd44a6734503 (the POC). During the actual implementation I'll check what's the best option for warning/asserting etc.

Issue #552, opened 2023-04-04
Investigate different message handling order on PubSub for production and tests


glevco:
> The `cli` import fix is good as is, about the other one I wonder if we should make a test for it.

@jansegre I was investigating this and I found out that existing tests in `test_event_manager.py` were supposed to have caught this. The code in `PubSub` for publishing events is:

```python
    def publish(self, key: HathorEvents, **kwargs: Any) -> None:
        reactor_thread = ReactorThread.get_current_thread(self.reactor)

        args = EventArguments(**kwargs)
        for fn in self._subscribers[key]:
            if reactor_thread == ReactorThread.NOT_RUNNING:
                fn(key, args)
            else:
                is_empty = bool(not self.queue)
                self.queue.append((fn, key, args))
                if is_empty:
                    self._schedule_call_next()
```

The reason the tests passed is because tests use the `Clock` class instead of the reactor, so incidentally things are called in the correct order: the `else` branch is called and the event handler function call is delayed to after the `EventManager` is started. In the real case, when running the full node, the `if reactor_thread == ReactorThread.NOT_RUNNING:` branch is called, and therefore the event handler function is called before the `EventManager` is started.

Also, `ReactorThread` is where the `MAIN_THREAD` value is returned for `Clock` (for testing):

```python
    def get_current_thread(cls, reactor: Reactor) -> 'ReactorThread':
        """ Returns if the code is being run on the reactor thread, if it's running already.
        """
        running = getattr(reactor, 'running', None)
        if running is not None:
            if running:
                return cls.MAIN_THREAD if isInIOThread() else cls.NOT_MAIN_THREAD
            else:
                # if reactor is not running yet, there's no threading
                return cls.NOT_RUNNING
        else:
            # on tests, we use Clock instead of a real Reactor, so there's
            # no threading. We consider that the reactor is running
            return cls.MAIN_THREAD
```

I feel that we should somehow either change `ReactorThread` or `PubSub` themselves, to prevent this from happening again in the future. Code logic is getting executed differently in production and in tests for this reason.

~~For example, changing the above last line from `return cls.MAIN_THREAD` to `return cls.NOT_RUNNING` would fix the problem, but I'm not sure of other unintended consequences.~~ _Edit: I have tested this, and it breaks a lot of other tests. Which may be a bit weird because maybe it indicates that there's code that relies on the specific implementation of PubSub and the order it is going to call message handlers (either at the time of a publish, or the postponed call)..._

What do you think?

[...]

PR #694, opened 2023-07-11
chore: improve pubsub

glevco:
### Motivation

This PR resolves https://github.com/HathorNetwork/hathor-core/issues/552. There are 3 goals:

1. Improve typing of code using Twisted reactors
2. Change tests to use a `MemoryReactorClock` instead of a `Clock`, so it's compatible with the reactor used in production (that is, it implements the `IReactorCore`, `IReactorTime`, and `IReactorTCP` Twisted interfaces. `Clock` only implements `IReactorTime`)
3. Remove PubSub support for reactors that don't implement `IReactorCore`

This was done by leveraging [Python protocols](https://peps.python.org/pep-0544/) to represent an intersection of Twisted interfaces, which allows us to better represent the reactor type. By doing this, we were able to change tests to use a reactor that is more similar to the one used in production, as previously we were using a `Clock`, that behaved differently than the production reactor during full node initialization. This also allows for simplifying `PubSub`, as it had a specifc implementation for dealing with `Clock`.

### Acceptance Criteria

- Create new `ReactorCoreProtocol`, `ReactorTimeProtocol`, and `ReactorTCPProtocol` classes stubbing the respective Twisted interfaces, `IReactorCore`, `IReactorTime`, and `IReactorTCP`
- Create new `ReactorProtocol` that implements the 3 protocols above, effectively creating a type that represents an intersection of Twisted interfaces, and cast the Twisted reactor to that protocol (this resolves a `XXX` comment in `hathor/util.py`)
- Update `PubSub` to improve its code, also removing support for "non-running" reactors (`Clock`)
- Remove the `ReactorThread` enum as it's not used anymore
- Change the `Clock` to `MemoryReactorClock` in tests (this resolves a `XXX` comment)
- Change `HeapClock` to `MemoryReactorHeapClock` on simulator
- Improve `verified_cast()` utils function and move it to its own file

### Checklist

### DIFF
--- a/hathor/debug_resources.py
+++ b/hathor/debug_resources.py
@@ -17,6 +17,7 @@ import sys
 
 from structlog import get_logger
 from twisted.internet import defer
+from twisted.internet.interfaces import IReactorFromThreads
 from twisted.web.http import Request
 
 from hathor.api_util import Resource, get_arg_default, get_args
@@ -24,6 +25,7 @@ from hathor.cli.openapi_files.register import register_resource
 from hathor.exception import HathorError
 from hathor.manager import HathorManager
 from hathor.util import reactor
+from hathor.utils.zope import asserted_cast
 
 logger = get_logger()
 
@@ -61,7 +63,8 @@ class DebugRaiseResource(Resource):
         assert exc_cls_name in self.exc_class_map
         exc_cls = self.exc_class_map[exc_cls_name]
         msg = get_arg_default(raw_args, 'msg', self.default_msg)
-        reactor.callFromThread(self.run, exc_cls, msg)
+        threaded_reactor = asserted_cast(IReactorFromThreads, reactor)
+        threaded_reactor.callFromThread(self.run, exc_cls, msg)
         return b'OK: no side-effects\n'
 
 
@@ -185,7 +188,8 @@ class DebugMessAroundResource(Resource):
         mess = get_arg_default(get_args(request), 'with', self.default_mess)
         assert mess in self.mess_map
         mess_func = self.mess_map[mess]
-        reactor.callFromThread(mess_func)
+        threaded_reactor = asserted_cast(IReactorFromThreads, reactor)
+        threaded_reactor.callFromThread(mess_func)
         return b'OK: database yanked, full-node will break\n'
 
 
--- a/hathor/p2p/sync_v1/agent.py
+++ b/hathor/p2p/sync_v1/agent.py
@@ -31,7 +31,8 @@ from hathor.p2p.sync_v1.downloader import Downloader
 from hathor.transaction import BaseTransaction
 from hathor.transaction.base_transaction import tx_or_block_from_bytes
 from hathor.transaction.storage.exceptions import TransactionDoesNotExist
-from hathor.util import Reactor, json_dumps, json_loads, verified_cast
+from hathor.util import Reactor, json_dumps, json_loads
+from hathor.utils.zope import asserted_cast
 
 settings = HathorSettings()
 logger = get_logger()
@@ -60,8 +61,8 @@ class SendDataPush:
         self.node_sync = node_sync
         self.protocol: 'HathorProtocol' = node_sync.protocol
         assert self.protocol.transport is not None
-        self.consumer = verified_cast(IConsumer, self.protocol.transport)
-
+        consumer = asserted_cast(IConsumer, self.protocol.transport)
+        self.consumer = consumer
         self.is_running: bool = False
         self.is_producing: bool = False
 
--- a/hathor/p2p/sync_v2/streamers.py
+++ b/hathor/p2p/sync_v2/streamers.py
@@ -21,7 +21,7 @@ from zope.interface import implementer
 
 from hathor.transaction import BaseTransaction, Block, Transaction
 from hathor.transaction.storage.traversal import BFSOrderWalk
-from hathor.util import verified_cast
+from hathor.utils.zope import asserted_cast
 
 if TYPE_CHECKING:
     from hathor.p2p.protocol import HathorProtocol
@@ -60,7 +60,8 @@ class _StreamingBase:
         self.node_sync = node_sync
         self.protocol: 'HathorProtocol' = node_sync.protocol
         assert self.protocol.transport is not None
-        self.consumer = verified_cast(IConsumer, self.protocol.transport)
+        consumer = asserted_cast(IConsumer, self.protocol.transport)
+        self.consumer = consumer
 
         self.counter = 0
         self.limit = limit
--- a/hathor/simulator/simulator.py
+++ b/hathor/simulator/simulator.py
@@ -25,7 +25,7 @@ from hathor.conf import HathorSettings
 from hathor.daa import TestMode, _set_test_mode
 from hathor.manager import HathorManager
 from hathor.p2p.peer_id import PeerId
-from hathor.simulator.clock import HeapClock
+from hathor.simulator.clock import HeapClock, MemoryReactorHeapClock
 from hathor.simulator.miner.geometric_miner import GeometricMiner
 from hathor.simulator.tx_generator import RandomTransactionGenerator
 from hathor.transaction.genesis import _get_genesis_transactions_unsafe
@@ -118,7 +118,7 @@ class Simulator:
         self.rng = Random(self.seed)
         self.settings = HathorSettings()
         self._network = 'testnet'
-        self._clock = HeapClock()
+        self._clock = MemoryReactorHeapClock()
         self._peers: OrderedDict[str, HathorManager] = OrderedDict()
         self._connections: list['FakeConnection'] = []
         self._started = False
@@ -178,6 +178,7 @@ class Simulator:
             .build()
 
         artifacts.manager.start()
+        self._clock.run()
         self.run_to_completion()
 
         # Don't use it anywhere else. It is unsafe to generate mnemonic words like this.
--- a/hathor/utils/zope.py
+++ b/hathor/utils/zope.py
@@ -1,0 +1,45 @@
+#  Copyright 2023 Hathor Labs
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from typing import Any, Optional, TypeVar, cast
+
+from zope.interface import Interface
+from zope.interface.exceptions import Invalid
+from zope.interface.verify import verifyObject
+
+T = TypeVar('T', bound=Interface)
+
+
+def verified_cast(interface_class: type[T], obj: Any) -> Optional[T]:
+    """
+    Receive a zope interface and an object, and return a cast to this interface if the object implements it.
+    Return None otherwise.
+    """
+    try:
+        if verifyObject(interface_class, obj):
+            return cast(T, obj)
+    except Invalid:
+        pass
+
+    return None
+
+
+def asserted_cast(interface_class: type[T], obj: Any) -> T:
+    """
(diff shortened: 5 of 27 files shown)


hathor/simulator/simulator.py
```diff
@@ -178,6 +178,7 @@ def create_artifacts(self, builder: Optional[Builder] = None) -> BuildArtifacts:
             .build()
 
         artifacts.manager.start()
+        self._clock.running = True
```
glevco, line 181: @msbrogli this will probably be a relevant line to change in https://github.com/HathorNetwork/hathor-core/pull/584

tests/unittest.py
```diff
@@ -170,6 +169,7 @@ def create_peer_from_builder(self, builder, start_manager=True):
 
         if start_manager:
             manager.start()
+            self.clock.running = True
```
glevco, line 172: @msbrogli this will probably be a relevant line to change in https://github.com/HathorNetwork/hathor-core/pull/584

hathor/simulator/simulator.py
glevco: Referenced in https://github.com/HathorNetwork/hathor-core/pull/584#issuecomment-1632904296

tests/unittest.py
glevco: Referenced in https://github.com/HathorNetwork/hathor-core/pull/584#issuecomment-1632904296

jansegre (dismissed):

jansegre (approved):

hathor/utils/zope.py
```diff
@@ -0,0 +1,39 @@
...
+    Return False otherwise.
+
+    Note: this return type loosely simulates what would be a Result type in other languages.
+    """
+    result: Union[T, Literal[False]] = False
+
+    try:
+        if verifyObject(interface_class, obj):
+            result = cast(T, obj)
+    except Invalid:
```
msbrogli, line 36: Why not just raise the exception?

tests/p2p/test_connections.py
```diff
@@ -23,6 +23,6 @@ def test_manager_connections(self):
         endpoint = 'tcp://127.0.0.1:8005'
         manager.connections.connect_to(endpoint, use_ssl=True)
 
-        self.assertNotIn(endpoint, manager.connections.iter_not_ready_endpoints())
+        self.assertIn(endpoint, manager.connections.iter_not_ready_endpoints())
```
msbrogli, line 26: Why have you inverted the logic?

tests/unittest.py
```diff
@@ -174,6 +173,7 @@ def create_peer_from_builder(self, builder, start_manager=True):
...
```
msbrogli, line 176: Why do you have to manually mark it as running?

hathor/simulator/simulator.py
msbrogli, line 181: Why do you have to manually mark it as running?

msbrogli (changes requested):

tests/others/test_metrics.py
```diff
@@ -74,7 +74,7 @@ def test_connections_manager_integration(self):
         self.assertEquals(manager.metrics.known_peers, 3)
         self.assertEquals(manager.metrics.connected_peers, 2)
         self.assertEquals(manager.metrics.handshaking_peers, 1)
-        self.assertEquals(manager.metrics.connecting_peers, 0)
+        self.assertEquals(manager.metrics.connecting_peers, 1)
```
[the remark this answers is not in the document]
glevco: Because before we were using a `Clock` as a reactor for tests, and `Clock` does not implement the `IReactorTCP` interface, which it should since this interface's methods are actually used by us. For that, when trying to make the connection, the reactor would raise when exception when the non-existent method was called.

By changing our reactor tests to the `MemoryReactorClock`, which [does implement](https://github.com/twisted/twisted/blob/39ee213fca02e08bc1aadad61607d7e7e123a776/src/twisted/internet/testing.py#L407-L409) `IReactorTCP`, the connection we create correctly gets to the "connecting" state, and that's why the test fixture is changed (it never get's to "connected", though).

tests/p2p/test_connections.py
glevco: For the same reason described in https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1286423313

tests/unittest.py
glevco: Because before we were using a `Clock` as a reactor for tests, and `Clock` does not implement the `IReactorCore` interface, which it should since this interface's methods are actually used by us. More specifically, `IReactorCore` defines the `run()` method. It also defines the `running` attribute that is being used by our current PubSub implementation (it ignores the attribute when testing since `Clock` doesn't implement it).

By changing our reactor tests to the `MemoryReactorClock`, which does implement `IReactorCore`, the `run()` method and `running` attribute are defined. The ideal then would be to simply call `self.clock.run()`, but `MemoryReactorClock`'s fake `run()` implementation [weirdly calls `stop()`](https://github.com/twisted/twisted/blob/39ee213fca02e08bc1aadad61607d7e7e123a776/src/twisted/internet/testing.py#L516) before returning, making it unusable to us, since we need the reactor in a `running=True` state for the tests. For that, instead of calling `run()`, we manually set the attribute value.

Another option would be to subclass `MemoryReactorClock` and override our own `run()` implementation, but since the only thing it would do would be to set `running=True`, I thought it seemed a bit overkill.

hathor/simulator/simulator.py
glevco: For the same reason described in https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1286429491

hathor/utils/zope.py
glevco: Because I don't want it to fail in the logic used by the new PubSub implementation, instead I'm just checking the cast.

I have implemented a new utils function that uses this one and raises if the cast fails in https://github.com/HathorNetwork/hathor-core/pull/694/commits/ddcde1028e1c8866b376a185a5ae74b4e80348f4.

hathor/debug_resources.py
```diff
@@ -61,7 +63,9 @@ def render_GET(self, request: Request) -> bytes:
         assert exc_cls_name in self.exc_class_map
         exc_cls = self.exc_class_map[exc_cls_name]
         msg = get_arg_default(raw_args, 'msg', self.default_msg)
-        reactor.callFromThread(self.run, exc_cls, msg)
+        threaded_reactor = verified_cast(IReactorFromThreads, reactor)
+        assert threaded_reactor is not False
```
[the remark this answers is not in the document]
glevco: Updated to use the new utils function described in https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1286432637

hathor/debug_resources.py
```diff
@@ -185,7 +189,9 @@ def render_GET(self, request: Request) -> bytes:
         mess = get_arg_default(get_args(request), 'with', self.default_mess)
         assert mess in self.mess_map
         mess_func = self.mess_map[mess]
-        reactor.callFromThread(mess_func)
+        threaded_reactor = verified_cast(IReactorFromThreads, reactor)
+        assert threaded_reactor is not False
```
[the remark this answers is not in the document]
glevco: Updated to use the new utils function described in https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1286432637

hathor/p2p/sync_v1/agent.py
```diff
@@ -60,8 +61,9 @@ def __init__(self, node_sync: 'NodeSyncTimestamp'):
         self.node_sync = node_sync
         self.protocol: 'HathorProtocol' = node_sync.protocol
         assert self.protocol.transport is not None
-        self.consumer = verified_cast(IConsumer, self.protocol.transport)
-
+        consumer = verified_cast(IConsumer, self.protocol.transport)
+        assert consumer is not False
```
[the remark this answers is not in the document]
glevco: Updated to use the new utils function described in https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1286432637

hathor/p2p/sync_v2/streamers.py
```diff
@@ -60,7 +60,9 @@ def __init__(self, node_sync: 'NodeBlockSync', *, limit: int = DEFAULT_STREAMING
         self.node_sync = node_sync
         self.protocol: 'HathorProtocol' = node_sync.protocol
         assert self.protocol.transport is not None
-        self.consumer = verified_cast(IConsumer, self.protocol.transport)
+        consumer = verified_cast(IConsumer, self.protocol.transport)
+        assert consumer is not False
```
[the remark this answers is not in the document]
glevco: Updated to use the new utils function described in https://github.com/HathorNetwork/hathor-core/pull/694#discussion_r1286432637

hathor/utils/zope.py
msbrogli: Got it. Thanks for the clarification. I still think that returning `None` and then asserting it's not `None` has the same effect then just not catching the exception. Anyway, the implementation seems good to me now. Thanks!

hathor/p2p/sync_v2/streamers.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/p2p/sync_v1/agent.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/debug_resources.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/debug_resources.py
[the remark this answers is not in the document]
msbrogli: Thanks!

tests/unittest.py
msbrogli: I agree it's very weird to have the `run()` method calling `stop()`. But, with the current implementation, we have to remember this fact in the future. So I think we should subclass the clock and override the `run()` implementation (and leave a note about this weird behavior from the original class).

tests/others/test_metrics.py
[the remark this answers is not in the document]
msbrogli: Got it. Why wasn't the test failing when the non-existent method was called? Anyway, I agree with your change and you can mark it as resolved.

tests/p2p/test_connections.py
msbrogli: Got it. Thanks!

msbrogli (dismissed):

tests/others/test_metrics.py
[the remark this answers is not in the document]
glevco: Because there's an error callback on the deferred that tries to make the connection: https://github.com/HathorNetwork/hathor-core/blob/master/hathor/p2p/manager.py#L570-L574

This callback simply emits a warning log and a PubSub event, and removes the connection from `self.connecting_peers`. That's why it was `0` in the old code. In the new code this error callback is not called, so the connection stays in `self.connecting_peers`, keeping it equal to `1`.

tests/unittest.py
glevco: Fair enough, done in https://github.com/HathorNetwork/hathor-core/pull/694/commits/aa53857b5974f4082741ced56e763fb75de54e3d

hathor/simulator/simulator.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/694/commits/aa53857b5974f4082741ced56e763fb75de54e3d

hathor/utils/zope.py
glevco: I wanted to return `None` instead of raising the exception because it simplifies this code in the PubSub:

```python
if not isInIOThread() and (threaded_reactor := verified_cast(IReactorFromThreads, self.reactor)):
    ...
```

By doing it like this, I don't need a `try-except` block in the PubSub.

tests/others/test_metrics.py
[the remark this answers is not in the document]
msbrogli: Then it was a bug in the test. Thanks!

hathor/utils/zope.py
msbrogli: Got it. Thanks again for the explanation.

msbrogli (approved):

jansegre (approved):

jansegre (approved):

msbrogli (approved):

PR #697, opened 2023-07-12
refactor: Prepare for sync-v2

msbrogli:
### Motivation

This PR contains refactors needed by sync v2. These changes DO NOT change any external behavior. They are mostly adjustments and fixes so tests can be easily adapted to run with sync v2.

### Acceptance Criteria

1. Add asserts on `FeatureService` to gracefully fail and prevent infinite loop.
2. Fix scope verification on `TransactionStorage.get_all_transactions()`.
3. Assert that a transaction is fully validated on `ConsensusAlgorithm.update()`.
4. Fix `RocksDBHeightIndex._add()` method. It was wrongfully identifying a reorg when a reorg wasn't happening.
5. Prepare `HathorManager` to initialize with partial and invalid transactions stored in the database.
6. Add missing `@abstractmethod` decorators on `SyncManager`.
7. Improve simulator `FakeConnection` with an auto-reconnect feature.
8. Add `FakeConnection.reconnect()` and `FakeConnection.is_both_synced()` methods.
9. Add simulator triggers (`StopWhenTrue`, `StopWhenSynced`, and `All`).
10. Add private attribute `TransactionRocksDBStorage. _rocksdb_storage: RocksDBStorage`.
11. Change `tests/p2p/test_sync_rate_limiter.py` to run for sync v1 only since it tests a rate limiter that exists only in sync v1.

### Checklist

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -51,6 +51,7 @@ from hathor.transaction import BaseTransaction, Block, MergeMinedBlock, Transact
 from hathor.transaction.exceptions import TxValidationError
 from hathor.transaction.storage import TransactionStorage
 from hathor.transaction.storage.exceptions import TransactionDoesNotExist
+from hathor.transaction.storage.tx_allow_scope import TxAllowScope
 from hathor.types import Address, VertexId
 from hathor.util import EnvironmentInfo, LogDuration, Random, Reactor, calculate_min_significant_weight, not_none
 from hathor.wallet import BaseWallet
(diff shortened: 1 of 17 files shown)


jansegre (approved):

jansegre (approved):

hathor/manager.py
```diff
@@ -264,6 +265,8 @@ def start(self) -> None:
 
         # Disable get transaction lock when initializing components
         self.tx_storage.disable_lock()
+        # Open scope for initialization.
+        self.tx_storage.set_allow_scope(TxAllowScope.VALID | TxAllowScope.PARTIAL | TxAllowScope.INVALID)
```
glevco, line 269: Should we use `set_allow_scope()` here? In its docstring, there's: `This method should not normally be used directly, use one of the allow_*_scope methods instead`

pedroferreira1 (approved):

hathor/manager.py
msbrogli: This is a special occasion because we need to access all transactions during initialization. So, we allow full access to the DAG and restore normality after the full node is initialized.

msbrogli: I'll merge it and change it later if you disagree.

PR #698, opened 2023-07-12
refactor: do not use built-in random and add proper linter check

jansegre:
### Motivation

Some tests are still using the built-in `random` module, which makes it so we can't reproduce a random case.

### Acceptance Criteria

- use `self.rng` (from `tests.TestCase`) insead of `random`
- make `self.rng` use the seed from `self.seed_config`, using a random seed when it is `None`
- introduce a custom linter script mechanism, porting the version check into it
- add a custom linter to make sure we don't use the built-in random again

### Checklist


tests/unittest.py
```diff
@@ -108,7 +110,9 @@ def setUp(self):
         self.clock.advance(time.time())
         self.log = logger.new()
         self.reset_peer_id_pool()
-        self.rng = Random()
+        self.seed = secrets.randbits(64) if self.seed_config is None else self.seed_config
+        self.log.debug('with', seed=self.seed)
```
msbrogli, line 114: Maybe `set seed`?

msbrogli (approved):

tests/unittest.py
jansegre: Done.

msbrogli (dismissed):

Makefile
```diff
@@ -78,15 +78,15 @@ isort-check:
 yamllint:
 	yamllint .
 
-.PHONY: check-version
-check-version:
-	bash ./extras/check_version.sh $(VERSION)
```
luislhl, line 83: I think we will need to adapt the `.github/workflows/docker.yml`, because we used this arg there.

Since you ported it to be an env var, we should set it as an env var there instead.

jansegre: Oh, right, I'll fix it.

jansegre: Done. I've fixed it and tested with `act` with a wrong and correct tag and both work as expected.

msbrogli (approved):

PR #701, opened 2023-07-12
chore(extras): script to generate list of PRs that master is ahead of rc

jansegre:
### Motivation

Getting the list of PRs to include in the release-candidate description was getting cumbersome and error prone.

### Acceptance Criteria

- new script in `extras` that will generate a markdown list of PRs to be used when creating a new release-candidate
- script requires `gh` to be installed and authenticated in order to use Github's API

### Checklist

### DIFF
--- a/extras/gen_release_candidate_changes.py
+++ b/extras/gen_release_candidate_changes.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python
+"""
+This script finds all PRs that have been merged into the `master` branch but not into the `release-candidate` branch in a given GitHub repository.
+
+Usage:
+
+    ./extras/gen_release_candidate_changes.py
+
+Example output:
+
+```
+- #701
+- #697
+- #686
+```
+"""
+
+import yaml
+import os
+import requests
+
+BASE_API_URL = 'https://api.github.com'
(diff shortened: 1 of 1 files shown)


extras/gen_release_candidate_changes.py
```diff
@@ -0,0 +1,64 @@
+#!/bin/env python
```
msbrogli, line 1: Add some docs on how to execute and give an example of output.

msbrogli (approved):

extras/gen_release_candidate_changes.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
```
luislhl, line 1: We need a way for people doing releases to know that this exists (in case someone else needs to do them)

I think we should guide them to use this in our release guide.
What do you think?

luislhl (approved):

extras/gen_release_candidate_changes.py
jansegre: That would be great.

extras/gen_release_candidate_changes.py
jansegre: Done.

PR #703, opened 2023-07-14
feat(feature-activation): implement bit signal in block templates

glevco:
### Motivation

### Acceptance Criteria

### Checklist


jansegre (approved):

hathor/simulator/miner/geometric_miner.py
```diff
@@ -75,7 +75,10 @@ def _generate_mining_block(self) -> 'Block':
         except IndexError:
             signal_bits = 0
 
-        return self._manager.generate_mining_block(signal_bits=signal_bits)
+        block = self._manager.generate_mining_block()
+        block.signal_bits = signal_bits
```
msbrogli, line 79: Shouldn't we use a `BitSignalingService` here as well?

hathor/transaction/base_transaction.py
```diff
@@ -1035,6 +1035,7 @@ def to_json(self, decode_script: bool = False, include_metadata: bool = False) -
         data['timestamp'] = self.timestamp
         data['version'] = int(self.version)
         data['weight'] = self.weight
+        data['signal_bits'] = self.signal_bits
```
msbrogli, line 1038: Should we serialize it as a string of zeros and ones?

msbrogli (approved):

hathor/simulator/miner/geometric_miner.py
glevco: The `BitSignalingService` generates a "fixed" set of signal bits that is the same for all generated blocks (assuming the feature states do not change). Conversely, in [some tests](https://github.com/HathorNetwork/hathor-core/blob/master/tests/feature_activation/test_feature_service.py#L35-L68) I define a different signal bits combination for each block, to test specific scenarios, using the `GeometricMiner`.

In other words, the signal bits are mocked manually here in the `GeometricMiner`, since the `BitSignalingService` doesn't allow such behavior.

hathor/transaction/base_transaction.py
glevco: I don't see a reason why, what are your thoughts?

msbrogli: I feel it's weird to get an integer number. But it's ok too.

hathor/simulator/miner/geometric_miner.py
msbrogli: Got it. Thanks for the clarification.

Issue #683, opened 2023-06-28
Test GET /status


alexruzenhack:
Design: https://github.com/HathorNetwork/internal-rfcs/pull/2

Issue #682, opened 2023-06-28
Update /status payload in API documentation


alexruzenhack:
Design: https://github.com/HathorNetwork/internal-rfcs/pull/2

Implemented by: https://github.com/HathorNetwork/hathor-core/pull/704

Issue #680, opened 2023-06-28
Add best_blockchain for each connected peer in the /status payload


alexruzenhack:
Design: https://github.com/HathorNetwork/internal-rfcs/pull/2

Implemented by: https://github.com/HathorNetwork/hathor-core/pull/704

PR #704, opened 2023-07-14
feat: add best_blockchain in /status payload

alexruzenhack:
### Motivation

Expose the best_blockchain through an interface.

### Acceptance Criteria

- Must return the` peer_best_blockchain` property in the connected_peer body when `/status` is requested
- Must return the `best_blockchain` property in the dag when `/status` is requested
- Make some adjustments in the test and fix some issues from the previews get_best_blockchain PR

### Checklist

Closes:
- https://github.com/HathorNetwork/hathor-core/issues/680
- https://github.com/HathorNetwork/hathor-core/issues/681
- https://github.com/HathorNetwork/hathor-core/issues/682
- https://github.com/HathorNetwork/hathor-core/issues/683


tests/test_utils/test_list.py
```diff
@@ -34,3 +34,11 @@ def test_single_or_none_more_than_one():
         single_or_none([1, 2, 3])
 
     assert exc_info.value.args[0] == 'expected one value at most'
+
+
+def test_decreasing_monotonicity(sequence):
```
msbrogli, line 39: Missing types and docstring.

tests/p2p/test_get_best_blockchain.py
```diff
@@ -257,6 +262,60 @@ def test_node_without_get_best_blockchain_capability(self):
...
+            raw_block_info = best_blockchain[0]
+            block_info = BlockInfo.from_raw(raw_block_info)
+            # assert the first element height is from the lastest block mined
+            self.assertEqual(block_info.height, 20)
+        except ValueError:
+            self.fail('Block info not valid')
+
+        # assert decreasing order for a sequence of heights
+        height_sequence = [b[1] for b in best_blockchain]
+        self.assertTrue(test_list.test_decreasing_monotonicity(height_sequence))
```
msbrogli, line 317: I'm not sure whether we should generalize this `test_decreasing_monotonicity()` method. It seems we can easily iterate and generate a more friendly error message.

msbrogli (changes requested):

tests/p2p/test_get_best_blockchain.py
alexruzenhack: I removed the generalization, but brought the check to the test module as util function.

tests/test_utils/test_list.py
alexruzenhack: Done.

hathor/p2p/resources/openapi/status.py
alexruzenhack, line 1: I've created this file just to make the setup.cfg ignore the "lint" erro `E501` (max-line-length) for this directory.

alexruzenhack, line 1: I've created this file just to make the setup.cfg ignore the "lint" erro `E501` (max-line-length) for this directory `hathor/p2p/resources/openapi/`.

jansegre: I think this change is too specific, all the other resources could potentially "need" the same thing and we'd have to keep adding more paths to `setup.cfg`. A better alternative IMO would be to just extract one of the dicts nested too deep into its own variable, like:

```python
example_value = {
    'server':  {
        # ...
    },
    # ...
}
openapi = {
   '/status': {
        # ...
        'get': {
            # ...
            'responses': {
                '200': {
                    'description': 'Success',
                    'content': {
                        'application/json': {
                            'examples': {
                                'success': {
                                    'summary': 'Server and peers data',
                                    'value': example_value,  # <--- here
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

jansegre (changes requested):

hathor/p2p/resources/openapi/status.py
alexruzenhack: Well, the ignore is a little wilder, it applies for the `hathor/p2p/resources/openapi/` path as a whole. My ideia is to move all the OpenApi objects to this folder in another PR. I believe this segregation also makes sense because it is a resource that is not code, and the lint rules are for code. What do you think?

jansegre: Still, he have many `resource` dirs. If we only do this for this resource I'm not sure if we will remember to do for the others or if we will have two different "standards". I don't think it's bad for it to be together with the code. IMO if we want to leave it separate we shouldn't even leave it in a `.py` file, it'd be better to just use a json file and not have to worry about adding an exception to the `.py` linter.

alexruzenhack: Ok. Now I see. I will remove this configuration and use your suggestion.

jansegre (approved):

msbrogli (approved):

msbrogli (changes requested):

hathor/p2p/utils.py
```diff
@@ -200,3 +202,28 @@ def format_address(addr: IAddress) -> str:
...
+        raise ValueError(f"id (hash) must be valid. We got {id}.")
+    if not isinstance(height, int):
+        raise ValueError(f"height must be an integer. We got {height}.")
+    if height < 0:
+        raise ValueError(f"height must be greater than or equal to 0. We got {height}.")
+
+    return HeightInfo(height, bytes.fromhex(id))
+
+
+def to_serializable_best_blockchain(best_blockchain: list[HeightInfo]) -> list[tuple[int, str]]:
```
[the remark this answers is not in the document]
alexruzenhack: Well, it would not change much, right? Because the result of this function will pass though a json_dumps anyway in the ready state, and also will pass through a formatter in the resource response. I believe we can keep this function and the HeightInfo as it is. Besides this two points, the ready state and status resource, HeightInfo do not need to be converted to json.

hathor/p2p/resources/status.py
```diff
@@ -89,6 +94,8 @@ def render_GET(self, request):
             best_block_tips.append({'hash': tx.hash_hex, 'height': meta.height})
 
         best_block = self.manager.tx_storage.get_best_block()
+        best_blockchain = to_serializable_best_blockchain(
```
msbrogli, line 97: Being nitpick here, as discussed, just fix the formatting.

msbrogli (dismissed):

jansegre (approved):

hathor/p2p/utils.py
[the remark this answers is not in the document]
msbrogli: Ok.

PR #708, opened 2023-07-18
feat(feature-activation): fix criteria validation

glevco:
### Motivation

Due to the way validations were performed on the `Criteria` class, some tests were flaky and behaving in unexpected ways. This PRs changes the way Criteria attributes are validated to a more robust solution.

### Acceptance Criteria

### Checklist

### DIFF
--- a/hathor/feature_activation/model/criteria.py
+++ b/hathor/feature_activation/model/criteria.py
@@ -12,7 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import TYPE_CHECKING, Any, ClassVar, Optional
+from typing import TYPE_CHECKING, Any, Optional
 
 from pydantic import Field, NonNegativeInt, validator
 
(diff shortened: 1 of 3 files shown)


jansegre (dismissed):

hathor/feature_activation/model/criteria.py
```diff
@@ -58,29 +60,45 @@ class Criteria(BaseModel, validate_all=True):
     lock_in_on_timeout: bool = False
     version: str = Field(..., regex=version.BUILD_VERSION_REGEX)
 
+    def to_validated(self, evaluation_interval: int, max_signal_bits: int) -> 'ValidatedCriteria':
+        """Create a validated version of self, including attribute validations that have external dependencies."""
+        attributes: dict[str, Any] = self.dict() | dict(
```
msbrogli, line 65: I don't understand why we are using a dict here. Shouldn't we use a typed `NamedTuple` so mypy can check types?

msbrogli (changes requested):

msbrogli (approved):

jansegre (approved):

PR #732, opened 2023-07-26
refactor(sync-v2): improve deferreds handling

glevco:
### Motivation

Simplify deferred-handling code by removing repetition and improving typing robustness.

### Acceptance Criteria

- Remove usage of string keys for "singular" deferreds (`tips`, `best-block`, and `peer-block-hashes`), by using a specific, typed attribute for each of them, instead of storing them in a generic dictionary.
- By doing that, the deferreds dictionary is now only used for transactions, so it has been renamed and typed accordingly. The key type has also been changed to `VertexId`, so the `origin` is removed from it.
- Change `None` handling in `get_tx()` from `if/raise` to `assert`, as the typing guarantees `tx` will never be `None`.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/manager.py
+++ b/hathor/p2p/sync_v2/manager.py
@@ -98,7 +98,10 @@ class NodeBlockSync(SyncAgent):
         self.peer_height = 0
 
         # Latest deferred waiting for a reply.
-        self.deferred_by_key: dict[str, Deferred] = {}
+        self._deferred_txs: dict[VertexId, Deferred[BaseTransaction]] = {}
+        self._deferred_tips: Optional[Deferred[list[bytes]]] = None
+        self._deferred_best_block: Optional[Deferred[dict[str, Any]]] = None
+        self._deferred_peer_block_hashes: Optional[Deferred[list[tuple[int, bytes]]]] = None
 
         # When syncing blocks we start streaming with all peers
         # so the moment I get some repeated blocks, I stop the download
@@ -372,14 +375,12 @@ class NodeBlockSync(SyncAgent):
     def get_tips(self) -> Deferred[list[bytes]]:
         """ Async method to request the remote peer's tips.
         """
-        key = 'tips'
-        deferred = self.deferred_by_key.get(key, None)
-        if deferred is None:
-            deferred = self.deferred_by_key[key] = Deferred()
+        if self._deferred_tips is None:
+            self._deferred_tips = Deferred()
             self.send_get_tips()
         else:
             assert self._receiving_tips is not None
-        return deferred
+        return self._deferred_tips
 
     def send_get_tips(self) -> None:
         """ Send a GET-TIPS message.
@@ -388,7 +389,7 @@ class NodeBlockSync(SyncAgent):
         self.send_message(ProtocolMessages.GET_TIPS)
         self._receiving_tips = []
 
-    def handle_get_tips(self, payload: str) -> None:
+    def handle_get_tips(self, _payload: str) -> None:
         """ Handle a GET-TIPS message.
         """
         assert self.tx_storage.indexes is not None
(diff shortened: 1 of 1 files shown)


hathor/p2p/sync_v2/manager.py
```diff
@@ -98,7 +98,10 @@ def __init__(self, protocol: 'HathorProtocol', reactor: Optional[Reactor] = None
         self.peer_height = 0
 
         # Latest deferred waiting for a reply.
-        self.deferred_by_key: dict[str, Deferred] = {}
+        self._deferred_txs: dict[str, Deferred[BaseTransaction]] = {}
```
jansegre, line 101: We could as well use `VertexId` here for the key.

hathor/p2p/sync_v2/manager.py
```diff
@@ -388,7 +389,7 @@ def send_get_tips(self) -> None:
         self.send_message(ProtocolMessages.GET_TIPS)
         self._receiving_tips = []
 
-    def handle_get_tips(self, payload: str) -> None:
+    def handle_get_tips(self, _payload: str) -> None:
```
jansegre, line 392: What's the benefit of prefixing the argument with `_`, the only reason I could see is making it so we don't want it to be used as a keyword argument, but for that we can make it positional only with `handle_get_tips(self, payload: str, /)` which is available since Python 3.8 I think.

This comment applies to all the renames of `payload` to `_payload`.

jansegre (approved):

hathor/p2p/sync_v2/manager.py
glevco: This is just convention since this argument is not used in this method, so that intention is made clear by using `_payload`. It could even be just `_`, but in this case I prefer the named unused parameter, also for clarity

hathor/p2p/sync_v2/manager.py
glevco: Indeed, but the current implementation uses `key = f'{origin}:{tx_id.hex()}'`, where `origin` is always equal to `mempool`:

https://github.com/HathorNetwork/hathor-core/blob/master/hathor/p2p/sync_v2/manager.py#L1003

It looks like we could actually remove this origin argument altogether, right? It would also simplify some of the downstream code.

jansegre: I'd say we can use the dict only for deferreds with `mempool` origin, rather than removing the origin argument from the sync command (if that's what you meant). And if we need to support a different argument (or drop it altogether) it's independent from this dict.

hathor/p2p/sync_v2/manager.py
jansegre: Oooh, got it, if it's unused but needs to exist it makes sense, but can't we just remove them instead?

hathor/p2p/sync_v2/manager.py
glevco: Good idea, I think it's better that I do it on a separate PR, what do you think?

hathor/p2p/sync_v2/manager.py
jansegre: Oh, because of the signature of the dict that stores the handlers we might need the argument. I'm OK using `_payload` then.

glevco: Exactly, we can't because it's used in the commands dict:

```python
def get_cmd_dict(self) -> dict[ProtocolMessages, Callable[[str], None]]:
    ...
```

So the callable needs the string argument. The other option would be to remove the argument from the method and then use a lambda in the cmd_dict, `lambda _: self.handle_get_tips()`, so it conforms to the callable type

hathor/p2p/sync_v2/manager.py
jansegre: Sure, but I think it's probably like a 4 lines change that is really in scope with this PR.

glevco: True, I thought I would have to remove the `origin` from other places. But that can be done separately. I updated the key type to `VertexId` in https://github.com/HathorNetwork/hathor-core/pull/732/commits/a54f623fb8c531e23ce939007ef5d4a7a9cdd558

jansegre: Great, that's exactly what I meant. Removing `origin` is something else entirely, we'd have to revise if it isn't needed for the implementations of the rest of the sync-v2 protocol that will come in the future.

jansegre (approved):

hathor/p2p/sync_v2/manager.py
msbrogli: I've never seen this convention but I'm ok with it as well.

msbrogli (approved):

PR #737, opened 2023-07-28
refactor(sync-v2): abstract request-response commands

glevco:
### Motivation

### Acceptance Criteria

### Checklist


hathor/p2p/sync_v2/request_response_command.py
msbrogli, line 1: Maybe we should add a request/response capability to the p2p protocol (instead of just to sync-v2). It might be useful for other networking tasks. I feel that it's worth preparing a design for it but a simple implementation would be to have a sequence number of the messages so the response can refer to the request sequence number. Such approach would allow each side to have more than one request and to receive responses out-of-order which seems to be a desirable feature.

msbrogli (changes requested):

PR #739, opened 2023-07-31
refactor: extracts SysctlRunner from SysctlProtocol

alexruzenhack:
### Motivation

Segregates the line parsing from the protocol.

### Acceptance Criteria

- Extracts a SysctlRunner from SysctlProtocol
- Instantiates the SysctlRunner in the RunNode
- Passes the runner to the SysctlFactory

### Checklist

### DIFF
--- a/hathor/sysctl/protocol.py
+++ b/hathor/sysctl/protocol.py
@@ -13,85 +13,69 @@
 # limitations under the License.
 
 import inspect
-import json
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from typing import Callable, Optional
 
 from pydantic import ValidationError
 from twisted.protocols.basic import LineReceiver
 
-from hathor.sysctl.exception import SysctlEntryNotFound, SysctlException, SysctlReadOnlyEntry, SysctlWriteOnlyEntry
-
-if TYPE_CHECKING:
-    from hathor.sysctl.sysctl import Sysctl
+from hathor.sysctl.exception import (
+    SysctlEntryNotFound,
+    SysctlException,
+    SysctlReadOnlyEntry,
+    SysctlRunnerException,
+    SysctlWriteOnlyEntry,
+)
+from hathor.sysctl.runner import SysctlRunner
 
 
 class SysctlProtocol(LineReceiver):
     delimiter = b'\n'
 
-    def __init__(self, root: 'Sysctl') -> None:
-        self.root = root
+    def __init__(self, runner: SysctlRunner) -> None:
+        self.runner = runner
 
     def lineReceived(self, raw: bytes) -> None:
         try:
             line = raw.decode('utf-8').strip()
         except UnicodeDecodeError:
             self.sendError('command is not utf-8 valid')
+
         if line.startswith('!help'):
             _, _, path = line.partition(' ')
             self.help(path)
             return
-        elif line == '!backup':
+        elif line.startswith('!backup'):
             self.backup()
             return
-        head, separator, tail = line.partition('=')
-        head = head.strip()
-        tail = tail.strip()
-        if separator == '=':
-            self.set(head, tail)
-        else:
-            self.get(head)
+
+        try:
+            feedback = self.runner.run(line)
+            if feedback:
+                self.sendLine(feedback)
+        except SysctlEntryNotFound:
+            path, _, _ = self.runner.get_line_parts(line)
+            self.sendError(f'{path} not found')
+        except SysctlReadOnlyEntry:
+            path, _, _ = self.runner.get_line_parts(line)
+            self.sendError(f'cannot write to {path}')
+        except SysctlWriteOnlyEntry:
+            path, _, _ = self.runner.get_line_parts(line)
+            self.sendError(f'cannot read from {path}')
+        except SysctlException as e:
+            self.sendError(str(e))
+        except ValidationError as e:
+            self.sendError(str(e))
+        except SysctlRunnerException as e:
+            self.sendError(str(e))
 
     def sendError(self, msg: str) -> None:
         """Send an error message to the client. Used when a command fails."""
         self.sendLine(f'[error] {msg}'.encode('utf-8'))
 
-    def set(self, path: str, value_str: str) -> None:
-        """Run a `set` command in sysctl."""
-        try:
-            value = self._deserialize(value_str)
-        except json.JSONDecodeError:
-            self.sendError('value: wrong format')
-            return
-
-        try:
-            self.root.set(path, value)
-        except SysctlEntryNotFound:
-            self.sendError(f'{path} not found')
-        except SysctlReadOnlyEntry:
-            self.sendError(f'cannot write to {path}')
-        except SysctlException as e:
-            self.sendError(str(e))
-        except ValidationError as e:
-            self.sendError(str(e))
-        except TypeError as e:
-            self.sendError(str(e))
-
-    def get(self, path: str) -> None:
-        """Run a `get` command in sysctl."""
-        try:
-            value = self.root.get(path)
-        except SysctlEntryNotFound:
-            self.sendError(f'{path} not found')
-        except SysctlWriteOnlyEntry:
-            self.sendError(f'cannot read from {path}')
-        else:
-            output = self._serialize(value)
-            self.sendLine(output.encode('utf-8'))
-
     def backup(self) -> None:
         """Run a `backup` command, sending all parameters to the client."""
-        for key, value in self.root.get_all():
-            output = f'{key}={self._serialize(value)}'
+        for key, value in self.runner.root.get_all():
+            output = f'{key}={self.runner.serialize(value)}'
             self.sendLine(output.encode('utf-8'))
 
     def help(self, path: str) -> None:
@@ -100,7 +84,7 @@ class SysctlProtocol(LineReceiver):
             self._send_all_commands()
             return
         try:
-            cmd = self.root.get_command(path)
+            cmd = self.runner.root.get_command(path)
         except SysctlEntryNotFound:
             self.sendError(f'{path} not found')
             return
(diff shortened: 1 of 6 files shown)


tests/sysctl/test_sysctl.py
```diff
@@ -175,15 +177,15 @@ def test_proto_get_tuple(self) -> None:
 
     def test_proto_get_unknown(self) -> None:
         self.proto.lineReceived(b'net.unknown')
-        self.assertEqual(b'[error] net.unknown not found\n', self.tr.value())
+        self.assertEqual(b'[error] path net.unknown not found\n', self.tr.value())
```
msbrogli, line 180: I don't think we should change the error message in this PR.

tests/sysctl/test_sysctl.py
```diff
@@ -175,15 +177,15 @@ def test_proto_get_tuple(self) -> None:
...
 
     def test_proto_get_readonly(self) -> None:
         self.proto.lineReceived(b'net.readonly')
         self.assertEqual(b'0.25\n', self.tr.value())
 
     def test_proto_get_writeonly(self) -> None:
         self.proto.lineReceived(b'core.writeonly')
-        self.assertEqual(b'[error] cannot read from core.writeonly\n', self.tr.value())
+        self.assertEqual(b'[error] cannot read from the path core.writeonly\n', self.tr.value())
```
msbrogli, line 188: Same here.

tests/sysctl/test_sysctl.py
```diff
@@ -203,11 +205,11 @@ def test_proto_set_str(self) -> None:
 
     def test_proto_set_readonly(self) -> None:
         self.proto.lineReceived(b'net.readonly=0.50')
-        self.assertEqual(b'[error] cannot write to net.readonly\n', self.tr.value())
+        self.assertEqual(b'[error] cannot write to the path net.readonly\n', self.tr.value())
```
msbrogli, line 208: Same here.

tests/sysctl/test_sysctl.py
```diff
@@ -203,11 +205,11 @@ def test_proto_set_str(self) -> None:
...
 
     def test_proto_set_unknown(self) -> None:
         self.proto.lineReceived(b'net.unknown=0.50')
-        self.assertEqual(b'[error] net.unknown not found\n', self.tr.value())
+        self.assertEqual(b'[error] path net.unknown not found\n', self.tr.value())
```
msbrogli, line 212: Same here.

hathor/sysctl/protocol.py
```diff
@@ -12,138 +12,30 @@
...
-        self.root = root
+    def __init__(self, runner: SysctlRunner) -> None:
+        self.runner = runner
 
     def lineReceived(self, raw: bytes) -> None:
         try:
             line = raw.decode('utf-8').strip()
         except UnicodeDecodeError:
             self.sendError('command is not utf-8 valid')
-        if line.startswith('!help'):
```
msbrogli, line 39: Why don't we keep custom commands `!help` and `!backup` here? I'm not sure that we should move them to the runner.

What do you think @jansegre ?

hathor/sysctl/runner.py
```diff
@@ -0,0 +1,145 @@
...
+
+        head, separator, tail = line.partition('=')
+        head = head.strip()
+        tail = tail.strip()
+        if separator == '=':
+            return self._set(head, tail)
+        else:
+            return self._get(head)
+
+    def _set(self, path: str, value_str: str) -> bytes:
```
msbrogli, line 58: I don't like the fact that we are capturing all specific exceptions and turning them into `SysctlRunnerException` exceptions.

msbrogli (changes requested):

hathor/sysctl/runner.py
alexruzenhack: Hmm... yeah we loose the original stack trace, however we gain the opportunity to deliver a better error message.

tests/sysctl/test_sysctl.py
alexruzenhack: Ok

tests/sysctl/test_sysctl.py
alexruzenhack: Ok

tests/sysctl/test_sysctl.py
alexruzenhack: Ok

tests/sysctl/test_sysctl.py
alexruzenhack: 97d561b46c3a3b53ec1fed8e2925439843597518

tests/sysctl/test_sysctl.py
alexruzenhack: Done: 97d561b46c3a3b53ec1fed8e2925439843597518

hathor/sysctl/runner.py
msbrogli: I think that we just need better error messages for end users. Runners are used internally and they can fail raising a custom exception that can be easily captured by the caller. In this case, the protocol would capture these exceptions and generate friendly error messages.

I feel it's not a good design to convert all specific exceptions into a single type `SysctlRunnerException` in an internal class that might be used in different parts of the code (e.g., on the initialization of the node).

alexruzenhack: Ok. Just to confirm, in this case I just need to drop the custom exception, right?

alexruzenhack: I did this commit to remove the custom exceptions: 65ef8dbfb23fdaa5e6cd6248afe86b91c4169cf9

alexruzenhack: I made this commit to treat the exceptions in the protocol: 09bde1c1744e81f3790f73f60e29d28579ff44cc

hathor/sysctl/protocol.py
jansegre: `!help` and `!backup` in particular seem to be useful only for an interactive CLI, I don't think there's too much benefit in having them available directly in the runner, so I agree with @msbrogli that maybe it makes more sense to leave them here.

jansegre (approved):

hathor/sysctl/protocol.py
alexruzenhack: Done: 487964a80dbc2feaac3b0aa40f4d1a21c6821208

jansegre (dismissed):

tests/sysctl/test_sysctl.py
```diff
@@ -175,7 +177,7 @@ def test_proto_get_tuple(self) -> None:
 
     def test_proto_get_unknown(self) -> None:
         self.proto.lineReceived(b'net.unknown')
-        self.assertEqual(b'[error] net.unknown not found\n', self.tr.value())
+        self.assertEqual(b'[error] unknown not found\n', self.tr.value())
```
msbrogli, line 180: Why aren't we presenting the whole path in the error message?

tests/sysctl/test_sysctl.py
```diff
@@ -207,7 +209,7 @@ def test_proto_set_readonly(self) -> None:
 
     def test_proto_set_unknown(self) -> None:
         self.proto.lineReceived(b'net.unknown=0.50')
-        self.assertEqual(b'[error] net.unknown not found\n', self.tr.value())
+        self.assertEqual(b'[error] unknown not found\n', self.tr.value())
```
msbrogli, line 212: Why aren't we presenting the whole path in the error message?

msbrogli (approved):

hathor/sysctl/protocol.py
msbrogli: Thanks!

hathor/sysctl/runner.py
msbrogli: Great! Thanks!

tests/sysctl/test_sysctl.py
alexruzenhack: Because now the error message comes directly from `ConnectionsManagerSysctl` and `SysctlProtocol` has no clue about the path because it is the of `SysctlRunner` now. Same bellow.

alexruzenhack: I will proceed, but if you find this problematic I can open another PR.

hathor/sysctl/protocol.py
```diff
@@ -13,85 +13,66 @@
...
-            self.root.set(path, value)
-        except SysctlEntryNotFound:
-            self.sendError(f'{path} not found')
-        except SysctlReadOnlyEntry:
-            self.sendError(f'cannot write to {path}')
+            feedback = self.runner.run(line)
+            if feedback:
+                self.sendLine(feedback)
+        except SysctlEntryNotFound as e:
+            self.sendError(f'{str(e)} not found')
```
msbrogli, line 56: Why not use `line` here? Or ask the path to the runner?

msbrogli (changes requested):

tests/sysctl/test_sysctl.py
msbrogli: I guess we should fix the error message. Showing just the last part of the path is misleading. We can keep this conversation closed since I just started a new one.

hathor/sysctl/protocol.py
alexruzenhack: We can't ask the path to the runner because the runner is a singleton and don't hold any line state. We can use the `line` here, however it will come with the assigned value to the path as well. Maybe the following exceptions should be in the runner side. There we can put the path in the exception message.

```python
        except SysctlEntryNotFound as e:
            self.sendError(f'{str(e)} not found')
        except SysctlReadOnlyEntry as e:
            self.sendError(f'cannot write to {str(e)}')
        except SysctlWriteOnlyEntry as e:
            self.sendError(f'cannot read from {str(e)}')
```

msbrogli: But the runner knows how to parse the line and provide the key. So we can move it to separate method and use it here.

alexruzenhack: Oh! Ok, I will see this possibility.

alexruzenhack: Done: dc44366f1b24bfc91eeae13e63ee36f0cc29e0c0

hathor/sysctl/runner.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.root.set(path, value)
+        return b''
+
+    def _get(self, path: str) -> bytes:
+        """Run a `get` command in sysctl."""
+        value = self.root.get(path)
+        return self.serialize(value).encode('utf-8')
+
+    def get_line_parts(self, line: str) -> tuple[str, ...]:
```
msbrogli, line 59: I guess the returning type is `tuple[str, str, str]`.

msbrogli (approved):

hathor/sysctl/protocol.py
msbrogli: Thanks!

hathor/sysctl/runner.py
alexruzenhack: Yes, I changed it to `tuple[str, ...]`

jansegre (approved):

PR #742, opened 2023-08-02
feat(feature-activation): update feature endpoint visibility to public

glevco:
### Motivation

Currently, trying to access Feature Activation endpoints on our public nodes returns 403 Forbidden, like on https://node1.mainnet.hathor.network/v1a/feature. This PR updates the endpoint visibility so it can be used by the Hathor Explorer UIs.

### Acceptance Criteria

- Change `GET /feature` visibility to public and add rate limits (the same used by `GET /transaction`)

### Checklist


glevco:
Moved to https://github.com/HathorNetwork/hathor-core/pull/743

PR #743, opened 2023-08-02
chore(feature-activation): update feature endpoint visibility to public

glevco:
### Motivation

Currently, trying to access Feature Activation endpoints on our public nodes returns 403 Forbidden, like on https://node1.mainnet.hathor.network/v1a/feature. This PR updates the endpoint visibility so it can be used by the Hathor Explorer UIs.

### Acceptance Criteria

- Change `GET /feature` visibility to public and add rate limits (the same used by `GET /transaction`)

### Checklist

### DIFF
--- a/hathor/feature_activation/resources/feature.py
+++ b/hathor/feature_activation/resources/feature.py
@@ -171,7 +171,7 @@ FeatureResource.openapi = {
             ],
             'per-ip': [
                 {
-                    'rate': '3r/s',
+                    'rate': '1r/s',
                     'burst': 10,
                     'delay': 3
                 }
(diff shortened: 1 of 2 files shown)


hathor/feature_activation/resources/feature.py
```diff
@@ -160,7 +160,23 @@ class GetFeaturesResponse(Response):
...
+            'global': [
+                {
+                    'rate': '50r/s',
+                    'burst': 100,
+                    'delay': 50
+                }
+            ],
+            'per-ip': [
+                {
+                    'rate': '3r/s',
```
msbrogli, line 174: That much? Maybe 1r/s?

jansegre (approved):

hathor/feature_activation/resources/feature.py
glevco: Sure, changed in https://github.com/HathorNetwork/hathor-core/pull/743/commits/0d44bb5217e53a7728897aa36533a4f58dafc7f7

PR #744, opened 2023-08-02
fix(event-queue): improve memory usage during the load phase

glevco:
### Motivation

The previous code was consuming too much memory when loading events from an existing database.

### Acceptance Criteria

- Create new `batch_iterator()` utils function that converts an iterator of items into an iterator of batches.
- Change `EventManager.handle_load_phase_vertices()` so it saves events to RocksDB in write batches of 10.000 events instead of a single batch with all events.

### Checklist

### DIFF
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Callable, Iterable, Iterator, Optional
+from typing import Callable, Iterator, Optional
 
 from structlog import get_logger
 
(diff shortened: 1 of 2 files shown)


hathor/event/event_manager.py
```diff
@@ -258,20 +261,28 @@ def handle_load_phase_vertices(
...
-                event = self._handle_event_creation(
-                    event_type=EventType.NEW_VERTEX_ACCEPTED,
-                    event_args=EventArguments(tx=vertex)
-                )
+        self.log.info('Finished creating events from existing database.')
 
-                yield event
-                self._event_ws_factory.broadcast_event(event)
-                self._last_event = event
+    def _create_event_iterator(
```
msbrogli, line 273: Missing docstring.

msbrogli (approved):

jansegre (approved):

msbrogli (approved):

PR #768, opened 2023-09-02
refactor: move get settings [part 1]

glevco:
### Motivation

This PR is the first on of a series of PRs focused on improving settings handling in the project, aiming to remove "global settings" that often cause importing order problems and limit varying settings for tests.

Those improvements will be made in two phases:

1. Renaming the `HathorSettings()` function to `get_settings()` and removing all invocations from module-level, moving them to inside functions/methods. This will be done incrementally in multiple PRs, and this PR is part 1 of that.
2. Changing all usages of settings to use injection instead of direct access.

Eventually the objective is that `get_settings()` is only called once in the whole system, and then injected as arguments on all objects/functions that need it.

In this PR, changes are very simple, code is mostly just renamed and the function call is moved. No logic should be changed. The only exceptions are changes to conform to linters and such.

### Acceptance Criteria

- Create new `get_settings()` function that is just a forward of `HathorSettings()`. This is just so we can do an "incremental rename" without having to touch all code that uses `HathorSettings()` at once. When it's done, `HathorSettings()` will be removed.
- Remove some usages of `HathorSettings()` from the module level and move it to inside functions/methods
  - For functions, the function just calls the new `get_settings()`
  - For classes we set `self._settings = get_settings()` in the class' `__init__()`, and update all methods to use that new attribute

### Checklist

### DIFF
--- a/hathor/conf/get_settings.py
+++ b/hathor/conf/get_settings.py
@@ -33,6 +33,10 @@ class _SettingsMetadata(NamedTuple):
 _settings_singleton: Optional[_SettingsMetadata] = None
 
 
+def get_settings() -> Settings:
+    return HathorSettings()
+
+
 def HathorSettings() -> Settings:
     """
     Returns the configuration named tuple.
--- a/tests/unittest.py
+++ b/tests/unittest.py
@@ -11,6 +11,7 @@ from twisted.trial import unittest
 
 from hathor.builder import BuildArtifacts, Builder
 from hathor.conf import HathorSettings
+from hathor.conf.get_settings import get_settings
 from hathor.daa import TestMode, _set_test_mode
 from hathor.p2p.peer_id import PeerId
 from hathor.p2p.sync_version import SyncVersion
@@ -113,6 +114,7 @@ class TestCase(unittest.TestCase):
         self.log.debug('set seed', seed=self.seed)
         self.rng = Random(self.seed)
         self._pending_cleanups = []
+        self._settings = get_settings()
 
     def tearDown(self):
         self.clean_tmpdirs()
--- a/hathor/p2p/sync_v1/agent.py
+++ b/hathor/p2p/sync_v1/agent.py
@@ -24,7 +24,7 @@ from twisted.internet.defer import Deferred, inlineCallbacks
 from twisted.internet.interfaces import IConsumer, IDelayedCall, IPushProducer
 from zope.interface import implementer
 
-from hathor.conf import HathorSettings
+from hathor.conf.get_settings import get_settings
 from hathor.p2p.messages import GetNextPayload, GetTipsPayload, NextPayload, ProtocolMessages, TipsPayload
 from hathor.p2p.sync_agent import SyncAgent
 from hathor.p2p.sync_v1.downloader import Downloader
@@ -34,7 +34,6 @@ from hathor.transaction.storage.exceptions import TransactionDoesNotExist
 from hathor.util import Reactor, json_dumps, json_loads
 from hathor.utils.zope import asserted_cast
 
-settings = HathorSettings()
 logger = get_logger()
 
 if TYPE_CHECKING:
@@ -191,6 +190,7 @@ class NodeSyncTimestamp(SyncAgent):
         :param reactor: Reactor to schedule later calls. (default=twisted.internet.reactor)
         :type reactor: Reactor
         """
+        self._settings = get_settings()
         self.protocol = protocol
         self.manager = protocol.node
         self.downloader = downloader
@@ -228,7 +228,7 @@ class NodeSyncTimestamp(SyncAgent):
 
         # Maximum difference between our latest timestamp and synced timestamp to consider
         # that the peer is synced (in seconds).
-        self.sync_threshold: int = settings.P2P_SYNC_THRESHOLD
+        self.sync_threshold: int = self._settings.P2P_SYNC_THRESHOLD
 
         # Indicate whether the sync manager has been started.
         self._started: bool = False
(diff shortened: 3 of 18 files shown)


jansegre (dismissed):

hathor/pycoin/htr.py
```diff
@@ -14,14 +14,19 @@
...
+    return create_bitcoinish_network(
+        symbol='HTR', network_name='Hathor', subnet_name='mainnet',
+        wif_prefix_hex='80',
+        address_prefix_hex=settings.P2PKH_VERSION_BYTE.hex(),
+        pay_to_script_prefix_hex=settings.MULTISIG_VERSION_BYTE.hex(),
+        bip32_prv_prefix_hex='0488ade4', bip32_pub_prefix_hex='0488B21E',
+    )
+
+
+network = get_network()
```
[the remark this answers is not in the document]
glevco: You're right, thanks! But I couldn't find anywhere that is using this, do you know where it's used?

The closest I found was the `_register_pycoin_networks()` function [here](https://github.com/HathorNetwork/hathor-core/blob/master/hathor/wallet/hd_wallet.py#L37-L50), but all it does is add the module to the `PYCOIN_NETWORK_PATHS` env var, which is not read anywhere.

hathor/conf/get_settings.py
```diff
@@ -33,6 +33,10 @@ class _SettingsMetadata(NamedTuple):
 _settings_singleton: Optional[_SettingsMetadata] = None
 
 
+def get_settings() -> Settings:
```
msbrogli, line 36: I know it will cause changes in all files but I feel that `get_global_settings()` is more meaningful than just `get_settings()`.

tests/unittest.py
```diff
@@ -102,6 +103,10 @@ class TestCase(unittest.TestCase):
     use_memory_storage: bool = USE_MEMORY_STORAGE
     seed_config: Optional[int] = None
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._settings = get_settings()
```
msbrogli, line 108: Why not do it in the `setUp()` method?

hathor/pycoin/htr.py
```diff
@@ -14,14 +14,19 @@
...
-settings = HathorSettings()
 
-network = create_bitcoinish_network(
-    symbol='HTR', network_name='Hathor', subnet_name='mainnet',
-    wif_prefix_hex='80',
-    address_prefix_hex=settings.P2PKH_VERSION_BYTE.hex(),
-    pay_to_script_prefix_hex=settings.MULTISIG_VERSION_BYTE.hex(),
-    bip32_prv_prefix_hex='0488ade4', bip32_pub_prefix_hex='0488B21E',
-)
+def get_network():
```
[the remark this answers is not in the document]
msbrogli: I agree with @jansegre . I also think that we should rename it to `__get_network()` and add a note in the docstring that it is not supposed to be called directly.

hathor/pycoin/htr.py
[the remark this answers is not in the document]
msbrogli: I think that this change does not actually solve the issue of instantiating the settings in import time. So we should leave this file out of this PR.

msbrogli (approved):

hathor/conf/get_settings.py
msbrogli, line 36: Should we cache and always return the same instance?

glevco: Ok, but may I do it in a separate PR? It'll be simpler to do it after parts 1, 2, and 3 are merged, so I don't have to update all of them separately (they are all close to getting merged). Then, it will be a simple rename action in the IDE, and a simple PR just for that.

tests/unittest.py
glevco: Could be, but in both cases the same settings instance will be used for all tests. Doing it in `setUp()` would just override the same instance with itself. After this whole migration is complete and we begin to use injected settings, tests will be changed so each test can use its own custom settings instance, so we'll remove this.

hathor/conf/get_settings.py
glevco: We do, the current `HathorSettings()` function caches the same instance in the `_settings_singleton` attribute that you can see above. After the renaming process is complete, `HathorSettings()` will be removed and `get_settings()` will absorb its implementation.

hathor/pycoin/htr.py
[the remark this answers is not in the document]
glevco: > So we should leave this file out of this PR.

Ok, done in https://github.com/HathorNetwork/hathor-core/pull/768/commits/def3c1c0e64c29590080f0839b563c00cc27be6a

hathor/pycoin/htr.py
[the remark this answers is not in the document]
glevco: I removed this file from this PR as suggested in https://github.com/HathorNetwork/hathor-core/pull/768#discussion_r1320893737

hathor/pycoin/htr.py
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/conf/get_settings.py
msbrogli: Ok. Thanks!

msbrogli: Got it. Thanks!

tests/unittest.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/768/commits/d6e19ea3591bc301f5ee5be5e251326c0d1d08de

msbrogli: Thanks!

msbrogli (approved):

PR #774, opened 2023-09-20
fix(feature-activation): add signal bits to minimally valid block

glevco:
### Motivation

The `BlockTemplate.generate_minimally_valid_block()` method is not setting signal bits as it should.

### Acceptance Criteria

- Add signal bits to minimally valid block template

### Checklist

### DIFF
--- a/hathor/mining/block_template.py
+++ b/hathor/mining/block_template.py
@@ -36,7 +36,7 @@ class BlockTemplate(NamedTuple):
     score: float  # metadata
     signal_bits: int  # signal bits for blocks generated from this template
 
-    def generate_minimaly_valid_block(self) -> BaseTransaction:
+    def generate_minimally_valid_block(self) -> BaseTransaction:
         """ Generates a block, without any extra information that is valid for this template. No random choices."""
         from hathor.transaction import TxOutput, TxVersion
         return TxVersion(min(self.versions)).get_cls()(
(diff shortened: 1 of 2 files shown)


hathor/mining/block_template.py
```diff
@@ -36,14 +36,15 @@ class BlockTemplate(NamedTuple):
...
-    def generate_minimaly_valid_block(self) -> BaseTransaction:
+    def generate_minimally_valid_block(self) -> BaseTransaction:
         """ Generates a block, without any extra information that is valid for this template. No random choices."""
         from hathor.transaction import TxOutput, TxVersion
         return TxVersion(min(self.versions)).get_cls()(
             timestamp=self.timestamp_min,
             parents=self.parents[:] + sorted(self.parents_any)[:(3 - len(self.parents))],
             outputs=[TxOutput(self.reward, b'')],
             weight=self.weight,
+            signal_bits=self.signal_bits,
```
msbrogli, line 47: Shouldn't we add a test that would fail without this change?

msbrogli (approved):

msbrogli (approved):

hathor/mining/block_template.py
msbrogli: Thanks!

jansegre (approved):

PR #775, opened 2023-09-20
fix: array decoding in sysctl-runner

alexruzenhack:
### Motivation

Fix SysctlRunner array decode.

### Acceptance Criteria

- Test decode and array with more than 1 element with success. See `test_command_option_sysctl_init_file` in `tests/cli/test_sysctl_init.py`.

### Checklist

### DIFF
--- a/hathor/sysctl/runner.py
+++ b/hathor/sysctl/runner.py
@@ -21,11 +21,21 @@ from hathor.sysctl.exception import SysctlRunnerException
 if TYPE_CHECKING:
     from hathor.sysctl.sysctl import Sysctl
 
-# - It starts with an opening square bracket [.
-# - It ends with a closing square bracket ].
-# - The elements are separated by commas and can be followed by optional whitespace.
-# - There can be zero or more elements in the array (an empty array is allowed).
-array_pattern = r'^\s*\[\s*(?:[^\[\],]+(?:\s*,\s*[^\[\],]+)*)?\s*\]\s*$'
+# Top level:
+# - Allow numbers, integers and float
+# - Allow string in double quote
+# - Allow mix of numbers with strings
+# - Allow space before, after and between elements
+# - Don't allow empty element like [1,,2]; empty element between 1 and 2
+# - Don't allow comma (,) after last element
+# - Don't allow double quote (") inside string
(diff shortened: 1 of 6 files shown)


hathor/sysctl/runner.py
```diff
@@ -13,13 +13,20 @@
...
 from hathor.sysctl.exception import SysctlRunnerException
 
 if TYPE_CHECKING:
     from hathor.sysctl.sysctl import Sysctl
 
+# - It starts with an opening square bracket [.
+# - It ends with a closing square bracket ].
+# - The elements are separated by commas and can be followed by optional whitespace.
+# - There can be zero or more elements in the array (an empty array is allowed).
+array_pattern = r'^\s*\[\s*(?:[^\[\],]+(?:\s*,\s*[^\[\],]+)*)?\s*\]\s*$'
```
jansegre, line 28: The regex pattern is hard to understand for me. I understand the "positive" cases, but what are the "negative" cases: what are we trying to prevent from being parsed as a list/array by using this regex?

hathor/sysctl/runner.py
```diff
@@ -76,6 +83,9 @@ def deserialize(self, value_str: str) -> Any:
         if len(value_str) == 0:
             return ()
 
+        if re.match(array_pattern, value_str):
+            return list(json.loads(value_str))
```
jansegre, line 87: Won't this produce a list inside a list? I don't understand how this is working.

hathor/sysctl/runner.py
alexruzenhack: I've update it and tried to make the regex clearer.

hathor/sysctl/runner.py
alexruzenhack: No. The `list(...)` is consolidating to list type, apparently. Take a look:
image

msbrogli: I guess this solution would not work if the method returns a `tuple[int, list[str]]`, for instance. But I haven't really tested against this case yet.

alexruzenhack:
Replaced by https://github.com/HathorNetwork/hathor-core/pull/820

PR #781, opened 2023-09-22
feat(p2p): Add a maximum number of connections per IP address

msbrogli:
### Motivation

Enforcing connection limits per IP Address enhances sybil attack resistance.

### Acceptance Criteria

1. Limit the number of connections per IP address.

### Checklist


hathor/p2p/protocol.py
```diff
@@ -92,11 +92,12 @@ class WarningFlags(str, Enum):
     capabilities: set[str]  # capabilities received from the peer in HelloState
 
     def __init__(self, network: str, my_peer: PeerId, p2p_manager: 'ConnectionsManager',
-                 *, use_ssl: bool, inbound: bool) -> None:
+                 *, use_ssl: bool, inbound: bool, remote_address: 'IAddress') -> None:
```
jansegre, line 95: Is it possible to use `remote_address: IAddress` instead of using a string annotation?

hathor/p2p/protocol.py
```diff
@@ -181,8 +182,11 @@ def is_state(self, state_enum: PeerState) -> bool:
 
     def get_short_remote(self) -> str:
         """Get remote for logging."""
-        assert self.transport is not None
-        return format_address(self.transport.getPeer())
+        return format_address(self.remote_address)
```
jansegre, line 185: If we want to make sure these have the same values we could check that `self.remote_address` and `self.transport.getPeer()` have the same value right when the transport is attributed to the `protocol` instance. Apparently this will happen _after_ `Factory.buildProtocol` is called, which is called by a `Connector`, which calls `protocol.makeConnection` on the instance returned by the factory. In our case this method is being inherited from `LineReceiver` into `HathorLineReceiver`, we could do something like:

```python
class HathorLineReceiver(LineReceiver, HathorProtocol):
    # ...
    def makeConnection(self, transport):
        assert self.remote_address == transport.getPeer()  # injected verification
        super().makeConnection(transport)
```

or we can also add it to `HathorLineReceiver.connectionMade` which is called when the transport has already been set, in which case we'd just use `self.transport` for the comparison.

hathor/p2p/protocol.py
msbrogli: Done!

hathor/p2p/protocol.py
msbrogli: Great! I did on the `HathorProtocol.on_connect()`. Would you review it again, please?

jansegre (dismissed):

glevco (dismissed):

jansegre (dismissed):

PR #782, opened 2023-09-22
feat(p2p): relay all recently-seen peers for neighbors

msbrogli:
### Motivation

The full node currently relays only connections in the ready state. So, if a connection drops for just a few seconds, we might not relay this peer information to our neighbors.

The current behavior of the full node involves requesting peer information from a neighboring node exclusively during the initial connection establishment phase. Consequently, if the full node attempts to establish a connection with a peer and that peer rejects the connection, there is a risk that the full node will subsequently remove this peer from its peer storage. As a result, there would be no further attempts made to connect to this particular peer in the future.

### Acceptance Criteria

1. Clean up aged peers in `ConnectionsManager.peer_storage`.
2. Add `PeerId.last_seen: float`.
3. Update `protocol.peer.last_seen` when the connection is ready and when new messages are received.
4. Add `known_peers.last_seen` to `/v1a/status` API.
5. Remove `time.time()` from `hathor.p2p.resources.StatusResource`.
6. Add a looping call to call `send_get_peers()` every 5 minutes.
7. Relay only peer information that contains at least one entrypoint.

### Checklist

### DIFF
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -157,6 +157,9 @@ class ConnectionsManager:
         # List of known peers.
         self.peer_storage = PeerStorage()  # dict[string (peer.id), PeerId]
 
+        # Maximum unseen time before removing a peer (seconds).
+        self.max_peer_unseen_dt: float = 30 * 60   # 30-minutes
+
         # A timer to try to reconnect to the disconnect known peers.
         self.lc_reconnect = LoopingCall(self.reconnect_to_all)
         self.lc_reconnect.clock = self.reactor
@@ -394,11 +397,15 @@ class ConnectionsManager:
             protocol.enable_sync()
 
         # Notify other peers about this new peer connection.
+        self.relay_peer_to_ready_connections(protocol.peer)
+
+    def relay_peer_to_ready_connections(self, peer: PeerId) -> None:
+        """Relay peer to all ready connections."""
         for conn in self.iter_ready_connections():
-            if conn != protocol:
-                assert conn.state is not None
-                assert isinstance(conn.state, ReadyState)
-                conn.state.send_peers([protocol])
+            if conn.peer == peer:
+                continue
+            assert isinstance(conn.state, ReadyState)
+            conn.state.send_peers([peer])
 
     def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
         """Called when a peer disconnect."""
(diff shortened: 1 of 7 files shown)


hathor/p2p/manager.py
msbrogli, line 1: @jansegre @glevco We might have to think more about item 8 since it might be exploited to waste bandwidth. I think we should remove item 8 and discuss it in a separate PR.

msbrogli: Done!

jansegre (approved):

hathor/p2p/manager.py
```diff
@@ -436,15 +440,39 @@ def on_receive_peer(self, peer: PeerId, origin: Optional[ReadyState] = None) ->
         """
         if peer.id == self.my_peer.id:
             return
+        if peer.id not in self.received_peer_storage:
+            for conn in self.iter_ready_connections():
+                if conn.peer != peer:
+                    assert isinstance(conn.state, ReadyState)
+                    conn.state.send_peers([peer])
```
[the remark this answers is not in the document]
msbrogli: Yes, we should. Let me do this.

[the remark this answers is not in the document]
msbrogli: I removed this line. But I did another minor refactor in a separate commit. Can you check it, please?

glevco (approved):

hathor/p2p/manager.py
```diff
@@ -342,6 +345,7 @@ def on_peer_ready(self, protocol: HathorProtocol) -> None:
         assert protocol.peer is not None
         protocol.peer = self.peer_storage.add_or_merge(protocol.peer)
         assert protocol.peer.id is not None
+        protocol.peer.last_seen = self.reactor.seconds()
```
jansegre, line 348: Since `last_seen` is updated on `recv_message` is there any effect updating here too? If there is I think it would be good to leave a comment.

hathor/p2p/manager.py
```diff
@@ -440,12 +447,28 @@ def on_receive_peer(self, peer: PeerId, origin: Optional[ReadyState] = None) ->
         peer = self.received_peer_storage.add_or_merge(peer)
         self.connect_to_if_not_connected(peer, 0)
 
+    def peers_cleanup(self) -> None:
```
jansegre, line 450: This function is only called in `reconnect_to_all`, is there no periodic clean up, will peers never expire in the current implementation?

jansegre (dismissed):

hathor/p2p/manager.py
```diff
@@ -375,11 +379,14 @@ def on_peer_ready(self, protocol: HathorProtocol) -> None:
             protocol.enable_sync()
 
         # Notify other peers about this new peer connection.
+        self.relay_peer_to_ready_connections(protocol.my_peer)
+
+    def relay_peer_to_ready_connections(self, peer: PeerId):
```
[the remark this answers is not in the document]
msbrogli: Nice catch. Thanks! Fixed!

hathor/p2p/manager.py
msbrogli: You're right. No need. Thanks!

hathor/p2p/manager.py
msbrogli: I didn't understand. There's a looping call for `reconnect_to_all()` that calls it every 5 seconds.

hathor/p2p/manager.py
msbrogli: Done!

hathor/p2p/manager.py
jansegre: Oh, I missed that. I didn't see that `reconnect_to_all` is called periodically, and I assumed it is only called to close and reopen connections, and not to only try to open new connections like it actually is.

jansegre (approved):

PR #645, opened 2023-06-01
feat(feature-activation): implement caching mechanism

glevco:
### Acceptance Criteria

- Implement `TransactionMetadata.feature_states` attribute
- Implement `Block.get_feature_state()` and `Block.update_feature_state()` methods
- Implement feature state caching in `FeatureService` using block metadata

Resolves https://github.com/HathorNetwork/hathor-core/issues/630

PR #785, opened 2023-09-25
feat(feature-activation): implement must signal

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/800

### Motivation

Implement the mandatory signaling mechanism required by the Feature Activation process.

### Acceptance Criteria

- Improvements in `Builder`:
  - Add `FeatureService` to `BuildArtifacts`
  - Change implementation of some `_get_or_create_*` methods so they can all be used without arguments.
  - Change private `_get_or_create_settings()` and `_get_or_create_feature_service()` methods to public so they can be used externally (specifically, so the `Simulator` can properly customize the builder).
- Implement `FeatureService.check_must_signal()`.
- Fix problem in caching mechanism in `FeatureService.get_state()` caused by must signal verification.
- Remove the original `Block.get_feature_activation_bit_counts()`, and change `calculate_feature_activation_bit_counts()` to `get_feature_activation_bit_counts()`.
- Change `Block.update_feature_state()` to `set_feature_state()`.
- Remove `BaseTransaction._update_feature_activation_bit_counts_metadata()` as it's not necessary anymore.
- Implement `BlockVerifier.verify_mandatory_signaling()`.

### Checklist

### DIFF
--- a/hathor/feature_activation/feature_service.py
+++ b/hathor/feature_activation/feature_service.py
@@ -12,28 +12,75 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, TypeAlias
+
 from hathor.feature_activation.feature import Feature
 from hathor.feature_activation.model.feature_description import FeatureDescription
 from hathor.feature_activation.model.feature_state import FeatureState
 from hathor.feature_activation.settings import Settings as FeatureSettings
-from hathor.transaction import Block
-from hathor.transaction.storage import TransactionStorage
+
+if TYPE_CHECKING:
+    from hathor.transaction import Block
+    from hathor.transaction.storage import TransactionStorage
+
+
+@dataclass(frozen=True, slots=True)
+class BlockIsSignaling:
+    """Represent that a block is correctly signaling support for all currently mandatory features."""
+    pass
+
+
+@dataclass(frozen=True, slots=True)
+class BlockIsMissingSignal:
+    """Represent that a block is not signaling support for at least one currently mandatory feature."""
+    feature: Feature
+
+
+BlockSignalingState: TypeAlias = BlockIsSignaling | BlockIsMissingSignal
 
 
 class FeatureService:
     __slots__ = ('_feature_settings', '_tx_storage')
 
-    def __init__(self, *, feature_settings: FeatureSettings, tx_storage: TransactionStorage) -> None:
+    def __init__(self, *, feature_settings: FeatureSettings, tx_storage: 'TransactionStorage') -> None:
         self._feature_settings = feature_settings
         self._tx_storage = tx_storage
 
-    def is_feature_active(self, *, block: Block, feature: Feature) -> bool:
+    def is_feature_active(self, *, block: 'Block', feature: Feature) -> bool:
         """Returns whether a Feature is active at a certain block."""
         state = self.get_state(block=block, feature=feature)
 
         return state == FeatureState.ACTIVE
 
-    def get_state(self, *, block: Block, feature: Feature) -> FeatureState:
+    def is_signaling_mandatory_features(self, block: 'Block') -> BlockSignalingState:
+        """
+        Return whether a block is signaling features that are mandatory, that is, any feature currently in the
+        MUST_SIGNAL phase.
+        """
+        bit_counts = block.get_feature_activation_bit_counts()
+        height = block.get_height()
+        offset_to_boundary = height % self._feature_settings.evaluation_interval
+        remaining_blocks = self._feature_settings.evaluation_interval - offset_to_boundary - 1
+        descriptions = self.get_bits_description(block=block)
+
+        must_signal_features = (
+            feature for feature, description in descriptions.items()
+            if description.state is FeatureState.MUST_SIGNAL
+        )
+
+        for feature in must_signal_features:
+            criteria = self._feature_settings.features[feature]
+            threshold = criteria.get_threshold(self._feature_settings)
+            count = bit_counts[criteria.bit]
+            missing_signals = threshold - count
+
+            if missing_signals > remaining_blocks:
+                return BlockIsMissingSignal(feature=feature)
+
+        return BlockIsSignaling()
+
+    def get_state(self, *, block: 'Block', feature: Feature) -> FeatureState:
         """Returns the state of a feature at a certain block. Uses block metadata to cache states."""
 
         # per definition, the genesis block is in the DEFINED state for all features
@@ -54,6 +101,9 @@ class FeatureService:
         previous_boundary_block = self._get_ancestor_at_height(block=block, height=previous_boundary_height)
         previous_boundary_state = self.get_state(block=previous_boundary_block, feature=feature)
 
+        # We cache _and save_ the state of the previous boundary block that we just got.
+        previous_boundary_block.set_feature_state(feature=feature, state=previous_boundary_state, save=True)
+
         if offset_to_boundary != 0:
             return previous_boundary_state
 
@@ -63,14 +113,16 @@ class FeatureService:
             previous_state=previous_boundary_state
         )
 
-        block.update_feature_state(feature=feature, state=new_state)
+        # We cache the just calculated state of the current block _without saving it_, as it may still be unverified,
+        # so we cannot persist its metadata. That's why we cache and save the previous boundary block above.
+        block.set_feature_state(feature=feature, state=new_state)
 
         return new_state
 
     def _calculate_new_state(
         self,
         *,
-        boundary_block: Block,
+        boundary_block: 'Block',
         feature: Feature,
         previous_state: FeatureState
     ) -> FeatureState:
@@ -136,7 +188,7 @@ class FeatureService:
 
         raise ValueError(f'Unknown previous state: {previous_state}')
 
-    def get_bits_description(self, *, block: Block) -> dict[Feature, FeatureDescription]:
+    def get_bits_description(self, *, block: 'Block') -> dict[Feature, FeatureDescription]:
         """Returns the criteria definition and feature state for all features at a certain block."""
         return {
             feature: FeatureDescription(
@@ -146,7 +198,7 @@ class FeatureService:
             for feature, criteria in self._feature_settings.features.items()
         }
 
-    def _get_ancestor_at_height(self, *, block: Block, height: int) -> Block:
+    def _get_ancestor_at_height(self, *, block: 'Block', height: int) -> 'Block':
         """
         Given a block, returns its ancestor at a specific height.
         Uses the height index if the block is in the best blockchain, or search iteratively otherwise.
@@ -158,13 +210,14 @@ class FeatureService:
         metadata = block.get_metadata()
 
         if not metadata.voided_by and (ancestor := self._tx_storage.get_transaction_by_height(height)):
+            from hathor.transaction import Block
             assert isinstance(ancestor, Block)
             return ancestor
 
         return _get_ancestor_iteratively(block=block, ancestor_height=height)
 
 
-def _get_ancestor_iteratively(*, block: Block, ancestor_height: int) -> Block:
+def _get_ancestor_iteratively(*, block: 'Block', ancestor_height: int) -> 'Block':
     """Given a block, returns its ancestor at a specific height by iterating over its ancestors. This is slow."""
     # TODO: there are further optimizations to be done here, the latest common block height could be persisted in
     #  metadata, so we could still use the height index if the requested height is before that height.
--- a/hathor/simulator/simulator.py
+++ b/hathor/simulator/simulator.py
@@ -22,7 +22,9 @@ from structlog import get_logger
 
 from hathor.builder import BuildArtifacts, Builder
 from hathor.conf.get_settings import get_settings
+from hathor.conf.settings import HathorSettings
 from hathor.daa import TestMode, _set_test_mode
+from hathor.feature_activation.feature_service import FeatureService
 from hathor.manager import HathorManager
 from hathor.p2p.peer_id import PeerId
 from hathor.simulator.clock import HeapClock, MemoryReactorHeapClock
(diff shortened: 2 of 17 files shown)


jansegre (approved):

msbrogli (changes requested):

hathor/feature_activation/feature_service.py
```diff
@@ -12,28 +12,56 @@
...
+        offset_to_boundary = height % self._feature_settings.evaluation_interval
+        remaining_blocks = self._feature_settings.evaluation_interval - offset_to_boundary - 1
+        descriptions = self.get_bits_description(block=block)
+
+        must_signal_features = (
+            feature for feature, description in descriptions.items()
+            if description.state is FeatureState.MUST_SIGNAL
+        )
+
+        for feature in must_signal_features:
```
msbrogli, line 53: I don't get it. Shouldn't we just check if the block has the correct bit signaling for features in the MUST_SIGNAL phase? Why is there a threshold and remaining blocks in the algorithm?

hathor/feature_activation/feature_service.py
```diff
@@ -54,6 +82,9 @@ def get_state(self, *, block: Block, feature: Feature) -> FeatureState:
         previous_boundary_block = self._get_ancestor_at_height(block=block, height=previous_boundary_height)
         previous_boundary_state = self.get_state(block=previous_boundary_block, feature=feature)
 
+        # We cache _and save_ the state of the previous boundary block that we just got.
+        previous_boundary_block.set_feature_state(feature=feature, state=previous_boundary_state, save=True)
```
msbrogli, line 86: Why are we changing the metadata in a `get_state()` method? I feel that this method is not supposed to change anything.

hathor/transaction/base_transaction.py
```diff
@@ -672,20 +672,13 @@ def get_metadata(self, *, force_reload: bool = False, use_storage: bool = True)
             #        happens include generating new mining blocks and some tests
             height = self.calculate_height() if self.storage else None
             score = self.weight if self.is_genesis else 0
-            kwargs: dict[str, Any] = {}
-
-            if self.is_block:
-                from hathor.transaction import Block
-                assert isinstance(self, Block)
-                kwargs['feature_activation_bit_counts'] = self.calculate_feature_activation_bit_counts()
```
msbrogli, line 680: Why you don't have to do it anymore? Isn't this metadata used anymore?

hathor/transaction/block.py
```diff
@@ -110,21 +110,27 @@ def calculate_min_height(self) -> int:
...
+
+        if metadata.feature_activation_bit_counts is not None:
+            return metadata.feature_activation_bit_counts
+
         previous_counts = self._get_previous_feature_activation_bit_counts()
         bit_list = self._get_feature_activation_bit_list()
 
         count_and_bit_pairs = zip_longest(previous_counts, bit_list, fillvalue=0)
         updated_counts = starmap(add, count_and_bit_pairs)
+        metadata.feature_activation_bit_counts = list(updated_counts)
```
msbrogli, line 131: Should you save it?

tests/feature_activation/test_feature_service.py
```diff
@@ -43,8 +43,8 @@ def _get_blocks_and_storage() -> tuple[list[Block], TransactionStorage]:
         0b0011,
         0b0001,
 
-        0b0000,  # 8: boundary block
-        0b0000,
+        0b0010,  # 8: boundary block
```
msbrogli, line 46: Are these bits activated because of MUST_SIGNAL?

tests/resources/transaction/test_mining.py
```diff
@@ -39,7 +39,7 @@ def test_get_block_template_with_address(self):
                 'height': 1,
                 'min_height': 0,
                 'first_block': None,
-                'feature_activation_bit_counts': [0, 0, 0, 0]
+                'feature_activation_bit_counts': None
```
msbrogli, line 42: Why has is changed to `None`?

msbrogli (changes requested):

hathor/feature_activation/feature_service.py
glevco: This follows the specification in the [RFC](https://github.com/HathorNetwork/rfcs/blob/master/projects/feature-activation/0001-feature-activation-for-blocks.md#invalidation-of-new-blocks-in-must_signal):

> During the `MUST_SIGNAL` state, the threshold must be enforced by invalidating blocks that don't signal support for the feature.
>
> That is, during one evaluation interval, if there are `EVALUATION_INTERVAL - threshold` non-signaling blocks, all new blocks are required to signal support and will be invalidated otherwise.

In other words, what it does is enforcing the same threshold that is used to activate the feature in the first place. For example, if 75% of blocks are required to activate a feature, then we'll tolerate up to 25% of non-signaling blocks during the `MUST_SIGNAL` phase. This is also analogous to Bitcoin's BIP 8.

The reason I see for this is, the `MUST_SIGNAL` phase serves to create a buffer period between a feature getting accepted through voting, and the feature becoming `LOCKED_IN`. Let's say a miner represents 10% of the hash rate, and it votes to _not_ support some feature during its voting phase, that is, it signals `0` for that feature. The feature will have 90% of blocks signaling support (from other miners), so it will eventually become activated. As soon as the voting phase ends, the mechanism shifts to `MUST_SIGNAL`, and all blocks that come from this miner will be rejected. So, we wouldn't be giving him any chance to update his full node and/or enable the signal during the `MUST_SIGNAL` phase.

By allowing for the same threshold used for voting, we give any non-signaling miners time to start signaling before their blocks are rejected. Otherwise, non-signaling miners would have to keep checking if they should enable the signal _before_ it becomes mandatory, which could create a weird game between signaling and non-signaling miners, trying to keep their wish until the last minute before the threshold is reached.

hathor/feature_activation/feature_service.py
glevco: This is the caching mechanism created to prevent repeating long-running calculations of feature states for blocks. Without caching, each time we want to cache for a block's state, we would have to calculate the state of all boundary blocks down to the genesis. This cache is stored in metadata.

Saving this in the `get_state()` method is actually not new,  it was firstly implemented in https://github.com/HathorNetwork/hathor-core/pull/645. Once a block is received and its state is calculated, it never changes. It's fixed forever, as it only depends on its ancestors. Therefore, this `get_state()` works like a lazy calculation: as soon as the state is queried for the first time, it's calculated and persisted in metadata. Then, any subsequent calls simply return the saved state.

The change here is that before, we were saving the state of the new, just queried block. This was the straightforward solution. However, now the `get_state()` method is used indirectly during the block verification process. That occurs before the block's `metadata.validation` is marked, so I cannot save the metadata at this point. Now, to workaround this, I delayed the caching. Instead of saving the current block's metadata in the storage, I only save it in memory. But I also save the previous boundary block's state in the storage (which is the line you commented).

hathor/transaction/base_transaction.py
glevco: It's still used, but I have reorganized it in a way to try to improve it. Before, there were two methods:

1. `Block.calculate_feature_activation_bit_counts()`
2. `BaseTransaction._update_feature_activation_bit_counts_metadata()`

The metadata initialization in `BaseTransaction.get_metadata()` called the first one and stored its value in metadata. Then, `BaseTransaction.update_initial_metadata()` called the second one, which also called the first one, calculating and storing the value in metadata, too. This was implemented in analogy to the height metadata, which is calculated and saved in the same places.

There was also a third method, `Block.get_feature_activation_bit_counts()`, that simply returned the value from metadata. Having all these methods was redundant and hard to maintain. Using a similar "lazy get" strategy, I simply removed all pre-calculation of this value, and left a single method, a new `Block.get_feature_activation_bit_counts()` that calculates this value only the first time it's called, stores it in metadata, and returns the stored value in all subsequent calls.

hathor/transaction/block.py
glevco: Maybe it would be ideal, but I can't do that here as this method is called in verification context, so I would have the same problem described in https://github.com/HathorNetwork/hathor-core/pull/785#discussion_r1377035035.

It is going to be indirectly saved though, as the metadata is saved by the manager after the verification process is complete.

tests/feature_activation/test_feature_service.py
glevco: Yes, those specific bits in those block heights were not being used by any other test in this file, so they could be freely changed. I simply configured the new `MUST_SIGNAL` tests to use them.

tests/resources/transaction/test_mining.py
glevco: Because of what was described in https://github.com/HathorNetwork/hathor-core/pull/785#discussion_r1377048608. Since this metadata value is not pre-calculated anymore, but instead lazily calculated (and then stored), its value is now initialized with `None`. This metadata is only internal and is irrelevant for the API being tested there.

hathor/transaction/exceptions.py
```diff
@@ -146,6 +146,10 @@ class CheckpointError(BlockError):
     """Block hash does not match checkpoint hash for its height"""
 
 
+class BlockMustSignal(BlockError):
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/785/commits/b05202359426d371c839a0eb6d64201877efa066

hathor/feature_activation/feature_service.py
```diff
@@ -12,28 +12,56 @@
...
 
-    def is_feature_active(self, *, block: Block, feature: Feature) -> bool:
+    def is_feature_active(self, *, block: 'Block', feature: Feature) -> bool:
         """Returns whether a Feature is active at a certain block."""
         state = self.get_state(block=block, feature=feature)
 
         return state == FeatureState.ACTIVE
 
-    def get_state(self, *, block: Block, feature: Feature) -> FeatureState:
+    def check_must_signal(self, block: 'Block') -> bool:
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/785/commits/b05202359426d371c839a0eb6d64201877efa066

hathor/transaction/block.py
glevco: I will test this to make sure it's saved, actually.

hathor/simulator/simulator.py
```diff
@@ -168,17 +168,25 @@ def create_artifacts(self, builder: Optional[Builder] = None) -> BuildArtifacts:
         wallet = HDWallet(gap_limit=2)
         wallet._manually_initialize()
 
+        builder = builder \
+            .set_reactor(self._clock) \
+            .set_rng(Random(self.rng.getrandbits(64))) \
+            .set_wallet(wallet)
+
+        feature_service = builder.get_or_create_feature_service()
```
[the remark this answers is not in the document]
glevco: Yes, the problem is that to create the verifiers, I need a `FeatureService`, and for that I need a `TransactionStorage`, that needs indexes... the dependency tree only grows, and the builder does exactly that, so I just reused its methods.

But what you said makes sense. I reverted the methods back to private and implemented an alternative approach in https://github.com/HathorNetwork/hathor-core/pull/785/commits/e8e9a08a921b6b7874d63571694027dc8443c424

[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/transaction/exceptions.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/feature_activation/feature_service.py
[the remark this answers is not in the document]
msbrogli: Thanks!

tests/feature_activation/test_feature_service.py
msbrogli: Got it. Thanks!

tests/resources/transaction/test_mining.py
msbrogli: Got it. Thanks!

hathor/feature_activation/feature_service.py
msbrogli: Got it. Thanks for the explanation.

hathor/feature_activation/feature_service.py
msbrogli: Got it. Thanks for the detailed explanation. It makes sense to store cache information. Should we add a note to this metadata saying that it's a cache and can be safely cleared up?

hathor/transaction/block.py
msbrogli: Got it. Thanks!

hathor/transaction/base_transaction.py
msbrogli: Got it. So you just initialize it as `None` during metadata creation, right?

hathor/verification/block_verifier.py
```diff
@@ -107,3 +117,13 @@ def verify_outputs(self, block: BaseTransaction) -> None:
...
+
+    def verify_mandatory_signaling(self, block: Block) -> None:
+        """Verify whether this block is missing mandatory signaling for any feature."""
+        if not self._settings.FEATURE_ACTIVATION.enable_usage:
+            return
+
+        assert self._feature_service is not None
+
+        if not self._feature_service.is_signaling_mandatory_features(block):
+            raise BlockMustSignalError()
```
msbrogli, line 129: Should we add information about which feature failed?

msbrogli (approved):

hathor/transaction/base_transaction.py
glevco: Exactly, this is done in `TransactionMetadata.__init__()`

hathor/feature_activation/feature_service.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/785/commits/c07032019b78c1ce3cbcaa067d1d65748c337353

hathor/verification/block_verifier.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/785/commits/c07032019b78c1ce3cbcaa067d1d65748c337353. Instead of using something like a `(success: bool, error: Optional[str])` tuple, I used the sum types pattern that we discussed the other day. Check out what you think.

jansegre (approved):

hathor/transaction/base_transaction.py
msbrogli: Thanks!

hathor/verification/block_verifier.py
msbrogli: Nice. Thanks!

msbrogli (approved):

PR #787, opened 2023-09-25
refactor(settings): change HathorManager to use injected settings

glevco:
### Motivation

Continuing with the settings-related refactors, this PR updates `HathorManager` so it starts using injected settings, instead of global. This refactor is necessary for the Feature Activation mandatory signaling implementation.

### Acceptance Criteria

- Add `settings` attribute to `HathorManager`, removing the module level settings
- Move `DEFAULT_CAPABILITIES` from module level to `HathorManager.get_default_capabilities()`
- Update `Builder` and `CliBuilder` accordingly

### Checklist

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -27,7 +27,7 @@ from twisted.python.threadpool import ThreadPool
 
 from hathor import daa
 from hathor.checkpoint import Checkpoint
-from hathor.conf import HathorSettings
+from hathor.conf.settings import HathorSettings
 from hathor.consensus import ConsensusAlgorithm
 from hathor.event.event_manager import EventManager
 from hathor.exception import (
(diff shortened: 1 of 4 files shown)


hathor/manager.py
```diff
@@ -130,6 +123,7 @@ def __init__(self,
                 'Either enable it, or use the reset-event-queue CLI command to remove all event-related data'
             )
 
+        self.settings = settings
```
jansegre, line 126: Other classes use `self._settings` should we do this here too?

hathor/manager.py
```diff
@@ -233,6 +227,13 @@ def __init__(self,
         self.lc_check_sync_state.clock = self.reactor
         self.lc_check_sync_state_interval = self.CHECK_SYNC_STATE_INTERVAL
 
+    def get_default_capabilities(self) -> list[str]:
```
msbrogli, line 230: Missing docstring.

jansegre (approved):

msbrogli (dismissed):

hathor/manager.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/787/commits/914dd775149b55db26cf5e9d0e8e29d3ef118681

hathor/manager.py
glevco: Makes sense, done in https://github.com/HathorNetwork/hathor-core/pull/787/commits/748afab81c05dbd564a00c5274bacaec0526cc5e

hathor/manager.py
msbrogli: Thanks!

msbrogli (approved):

PR #794, opened 2023-09-28
refactor(settings): refactor genesis module

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/787

### Motivation

This PR not only removes any usage of global settings from the `genesis` module, but it also substitutes all global variables defined in this module by functions. It also includes some simplifications in code using that module.

### Acceptance Criteria

- Remove `BLOCK_GENESIS, TX_GENESIS1, TX_GENESIS2, GENESIS, GENESIS_HASHES, GENESIS_HASH` global variables and implement, respectively, `get_genesis_block(), get_genesis_tx1(), get_genesis_tx2(), get_all_genesis(), get_all_genesis_hashes(), get_genesis_hash()`. As these are not shared instances anymore, it's safe to manipulate the returned instance in tests, for example.
- Simplify `get_genesis_hash()` and `is_genesis()` implementations.
- Remove `_get_genesis_transactions_unsafe()` function as after all refactors it was only used in one place.
- Change `CliBuilder` to use the existing `get_genesis_short_hash()` function.
- Removed `height_index.BLOCK_GENESIS_ENTRY` global variable and refactored it into `HeightIndex.get_genesis_block_entry()`.
- Update all code using removed global variables.

### Checklist


hathor/transaction/genesis.py
jansegre, line 1: I'm not sure if this has a significant impact, but there's one subtle behavioral change in this refactor, previously using `hathor.transaction.genesis.BLOCK_GENESIS` (and others) would always give you the exact same instance, now when using `hathor.transaction.genesis.get_genesis_block()` instead, you'll get a new instance for every call.

While the impact of having multiple _should_ probably be low, I'm not sure how it interacts with edge cases of the cache system (which tries to *prevent* at runtime from having multiple instances of the same vertex). So, in my opinion, it would be safest to preserve the singleton behavior, possibly using `@functools.cache` is enough.

Also, this should be testable (although it was never tested since `hathor.transaction.genesis.BLOCK_GENESIS` is a direct access to a Block instance) using `is` for different calls.

jansegre (approved):

hathor/transaction/genesis.py
glevco: This was actually an intended change (maybe I should have made it clear in the acceptance criteria). I prefer to return different instances, since we have methods that modify the instance but are not too explicit about it — for example, `get_metadata()` can actually set the internal `_metadata` attribute. This can generate hard to debug problems between different users of this instance.

Specially considering tests, I had used `BLOCK_GENESIS` in a very simple Feature Activation test. Then, when I implemented the mandatory signaling, that uses metadata, a lot of completely unrelated tests began to fail because the instance metadata was being changed. I took a long time to debug this.

Good point about the cache system, though. Do you think it's unfeasible to go though its code and determine if this change affects it? I would prefer to do that, but if we think it would be too complicated, I'm fine going back to the single instance.

glevco: P.S.: Even if we decide to investigate the cache and keep the multiple-instances behavior, I could revert this PR to use the singleton instance and implement the change to multiple in a separate PR, to keep this refactor pure

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
-    weight=settings.MIN_TX_WEIGHT,
-)
-
-GENESIS = [BLOCK_GENESIS, TX_GENESIS1, TX_GENESIS2]
-
-GENESIS_HASHES = [settings.GENESIS_BLOCK_HASH, settings.GENESIS_TX1_HASH, settings.GENESIS_TX2_HASH]
-
-
-def _get_genesis_hash() -> bytes:
+def get_genesis_block(settings: HathorSettings) -> Block:
```
msbrogli, line 24: I feel that we should assert `block.hash == GENESIS_BLOCK_HASH`.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
+        nonce=settings.GENESIS_BLOCK_NONCE,
+        timestamp=settings.GENESIS_TIMESTAMP,
+        weight=settings.MIN_BLOCK_WEIGHT,
+        outputs=[
+            TxOutput(settings.GENESIS_TOKENS, settings.GENESIS_OUTPUT_SCRIPT),
+        ],
+    )
+
+
+def get_genesis_tx1(settings: HathorSettings) -> Transaction:
```
msbrogli, line 37: I feel that we should assert `tx.hash == GENESIS_TX1_HASH`.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
+    """Return the genesis tx1."""
+    return Transaction(
+        hash=settings.GENESIS_TX1_HASH,
+        nonce=settings.GENESIS_TX1_NONCE,
+        timestamp=settings.GENESIS_TIMESTAMP + 1,
+        weight=settings.MIN_TX_WEIGHT,
+    )
+
+
+def get_genesis_tx2(settings: HathorSettings) -> Transaction:
```
msbrogli, line 47: I feel that we should assert `tx.hash == GENESIS_TX2_HASH`.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
-GENESIS = [BLOCK_GENESIS, TX_GENESIS1, TX_GENESIS2]
-
-GENESIS_HASHES = [settings.GENESIS_BLOCK_HASH, settings.GENESIS_TX1_HASH, settings.GENESIS_TX2_HASH]
-
-
-def _get_genesis_hash() -> bytes:
+def get_genesis_block(settings: HathorSettings) -> Block:
+    """Return the genesis block."""
+    return Block(
+        hash=settings.GENESIS_BLOCK_HASH,
```
msbrogli, line 27: We should calculate the hash instead.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
+        outputs=[
+            TxOutput(settings.GENESIS_TOKENS, settings.GENESIS_OUTPUT_SCRIPT),
+        ],
+    )
+
+
+def get_genesis_tx1(settings: HathorSettings) -> Transaction:
+    """Return the genesis tx1."""
+    return Transaction(
+        hash=settings.GENESIS_TX1_HASH,
```
msbrogli, line 40: We should calculate the hash instead.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
+        nonce=settings.GENESIS_TX1_NONCE,
+        timestamp=settings.GENESIS_TIMESTAMP + 1,
+        weight=settings.MIN_TX_WEIGHT,
+    )
+
+
+def get_genesis_tx2(settings: HathorSettings) -> Transaction:
+    """Return the genesis tx2."""
+    return Transaction(
+        hash=settings.GENESIS_TX2_HASH,
```
msbrogli, line 50: We should calculate the hash instead.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
+def get_all_genesis_hashes(settings: HathorSettings) -> list[bytes]:
+    """Return all genesis hashes."""
+    return [
+        settings.GENESIS_BLOCK_HASH,
+        settings.GENESIS_TX1_HASH,
+        settings.GENESIS_TX2_HASH
+    ]
+
+
+def get_genesis_hash(settings: HathorSettings) -> bytes:
```
msbrogli, line 75: I think this method name is misleading.

hathor/transaction/genesis.py
```diff
@@ -12,71 +12,75 @@
...
+    """Return the genesis tx2."""
+    return Transaction(
+        hash=settings.GENESIS_TX2_HASH,
+        nonce=settings.GENESIS_TX2_NONCE,
+        timestamp=settings.GENESIS_TIMESTAMP + 2,
+        weight=settings.MIN_TX_WEIGHT,
+    )
+
+
+def get_all_genesis(settings: HathorSettings) -> list[BaseTransaction]:
```
msbrogli, line 57: Add to the docstring that all instances will be freshly created.

msbrogli (changes requested):

hathor/transaction/genesis.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

hathor/transaction/genesis.py
glevco: Removed in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377

glevco:
@msbrogli @jansegre as we discussed, in https://github.com/HathorNetwork/hathor-core/pull/794/commits/a86c1756e42ce554a47e0048bbc19dec4614c377 I made the following changes:

- In `HathorSettings`:
  - Rename `GENESIS_TIMESTAMP` to `GENESIS_BLOCK_TIMESTAMP`.
  - Add `GENESIS_TX1_TIMESTAMP` and `GENESIS_TX2_TIMESTAMP` calculated properties. This simplifies some use cases of the removed `genesis` module functions.
- In `genesis` module:
  - Rename `get_genesis_hash()` to `get_representation_for_all_genesis()`.
  - Remove `get_genesis_block()`, `get_genesis_tx1()`, `get_genesis_tx2()`, and `get_all_genesis()`.
- In `TransactionStorage`:
  - Remove `_get_genesis_from_settings()`.
  - Add `_construct_genesis_block()`, `_construct_genesis_tx1()`, and `_construct_genesis_tx2()`.
- In `test_tx_storage` I had to remove part of a test, since it doesn't make sense anymore.
- In `tests/utils` remove `get_min_timestamp()`.

hathor/transaction/genesis.py
msbrogli: Thanks!

hathor/transaction/genesis.py
msbrogli: Thanks!

msbrogli (approved):

hathor/transaction/genesis.py
msbrogli: Thanks!

PR #796, opened 2023-09-29
refactor(verification): create vertex verifiers [part 1/5]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/794

### Motivation

This is the first PR in a 5-part series of PRs that completely move verification-related code out of the vertex model classes. This series is a continuation of the already merged https://github.com/HathorNetwork/hathor-core/pull/779.

In this PR, the module scaffolding introduced in the previous PR is changed to a classes structure that mimics the existing model classes structure. This allows easier incremental migration of code, as the inheritance tree is kept the same. In the future, if we untangle the inherited calls in verification methods, we may convert this back to modules, which would be the ideal solution in my opinion.

### Acceptance Criteria

- Move `block_verification`, `transaction_verification`, and `token_creation_tx_verification` modules to each respective new `Verifier` class. No logic should be changed, only moving module functions to class methods.
- Also create a `VertexVerifier` superclass and a `Verifier` class for `MergeMinedBlock`, completing verifiers for all 4 vertex types, and setting up the same inheritance tree as the vertex class themselves.
- Update `VerificationService` to receive the new `VertexVerifiers` named tuple, for injecting verifiers (this will be useful for the simulator). Also, update methods to use each respective verifier.
- Update `Builder` and `CliBuilder` accordingly.

### Checklist


jansegre (dismissed):

hathor/verification/verification_service.py
```diff
@@ -12,15 +12,40 @@
...
+
+
+class VertexVerifiers(NamedTuple):
+    block: BlockVerifier
+    merge_mined_block: MergeMinedBlockVerifier
+    tx: TransactionVerifier
+    token_creation_tx: TokenCreationTransactionVerifier
+
+    @classmethod
+    def create(cls, *, settings: HathorSettings) -> 'VertexVerifiers':
```
msbrogli, line 35: Maybe rename to `create_defaults()`.

msbrogli, line 35: Missing docstring.

msbrogli (changes requested):

msbrogli (approved):

hathor/verification/verification_service.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/796/commits/f35ffcea3b95d6a9ce173a6e5ad18f851aeee033

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/796/commits/f35ffcea3b95d6a9ce173a6e5ad18f851aeee033

msbrogli: Thanks!

msbrogli: Thanks!

msbrogli (approved):

PR #797, opened 2023-09-29
refactor(verification): move verification methods signatures [part 2/5]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/796

### Motivation

This is the second PR in a 5-part series of PRs that completely move verification-related code out of the vertex model classes.

The objective in this PR is to copy all verification methods _signatures_ from the vertex model classes to each respective `Verifier` class. The new methods simply call the original methods, effectively providing a new interface for existing implementations.

By doing this, we can update all callers of any verification method to use the new `Verifier` methods, while not touching the original model classes. This is the scaffolding necessary to actually move method implementations in the following PRs.

### Acceptance Criteria

- Copy all existing verification methods to each respective `Verifier` class. The implementation is simply a forward to the original method implemented by each vertex model class.
- Implement `VerificationService.verify_without_storage()`.
- Update every usage of any verification method to use the `Verifier` method, instead of the original model method.

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

PR #798, opened 2023-09-29
refactor(verification): move block-only verification methods [part 3/5]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/797

### Motivation

This is the third PR in a 5-part series of PRs that completely move verification-related code out of the vertex model classes.

This PR moves the verification method implementations of the `Block` inheritance branch.

### Acceptance Criteria

- Remove all `Block` and `MergeMinedBlock` verification methods from the original model classes, moving the implementation to each respective `Verifier` class. The only exception is the `verify_outputs()` method, which is part of the inheritance tree up to `BaseTransaction`, and will be moved in a separate PR.

### Checklist

### DIFF
--- a/hathor/verification/block_verifier.py
+++ b/hathor/verification/block_verifier.py
@@ -12,8 +12,16 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from hathor import daa
 from hathor.profiler import get_cpu_profiler
 from hathor.transaction import BaseTransaction, Block
+from hathor.transaction.exceptions import (
+    BlockWithInputs,
+    InvalidBlockReward,
+    RewardLocked,
+    TransactionDataError,
+    WeightError,
+)
 from hathor.verification.vertex_verifier import VertexVerifier
 
 cpu = get_cpu_profiler()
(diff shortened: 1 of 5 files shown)


jansegre (approved):

hathor/verification/block_verifier.py
```diff
@@ -52,25 +60,45 @@ def verify(self, block: Block) -> None:
...
+        meta = block.get_metadata()
+        assert meta.height is not None
+        assert meta.min_height is not None
+        if meta.height < meta.min_height:
+            raise RewardLocked(f'Block needs {meta.min_height} height but has {meta.height}')
 
     def verify_weight(self, block: Block) -> None:
         """Validate minimum block difficulty."""
-        block.verify_weight()
+        block_weight = daa.calculate_block_difficulty(block)
```
msbrogli, line 79: Rename to `min_block_weight`? I think having both `block.weight` and `block_weight` is error prone.

msbrogli (dismissed):

hathor/verification/block_verifier.py
msbrogli: Maybe add it as a separate commit in this PR? Or else I think we should get this new PR merged before moving forward with the other parts.

glevco: Sure, done in https://github.com/HathorNetwork/hathor-core/pull/798/commits/61cc04be1b99a23a5d2c899b89fe512fa1a7228c

msbrogli: Thanks!

msbrogli (approved):

jansegre (approved):

PR #805, opened 2023-10-06
feat[healthcheck]: make the p2p healthcheck use the GET-BEST-BLOCKCHAIN capability if enabled

luislhl:
### Motivation

Improving our current healthcheck so we detect more situations where we may be out of sync in relation to the network.

### Acceptance Criteria

We should include a new check in our healthcheck that will use the GET-BEST-BLOCKCHAIN capability to compare our current best blockchain with the other connected peers.

The previous logic was able to do something similar, but only for peers with which we are syncing now.

Including all other connected peers in this check will expand our capacity to detect out-of-sync situations and warn about this in the healthcheck.

### Checklist

### DIFF
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -23,6 +23,7 @@ from twisted.protocols.tls import TLSMemoryBIOFactory, TLSMemoryBIOProtocol
 from twisted.python.failure import Failure
 
 from hathor.conf import HathorSettings
+from hathor.indexes.height_index import HeightInfo
 from hathor.p2p.netfilter.factory import NetfilterFactory
 from hathor.p2p.peer_discovery import PeerDiscovery
 from hathor.p2p.peer_id import PeerId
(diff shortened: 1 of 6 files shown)


hathor/p2p/manager.py
```diff
@@ -302,10 +303,32 @@ def has_synced_peer(self) -> bool:
...
+        so use this method with caution. We only consider the one with the
+        highest height regardless of whether the blockchains are the same.
+        """
+        connections = list(self.iter_ready_connections())
+        best_height_info = None
+
+        for conn in connections:
+            assert conn.state is not None
+            assert isinstance(conn.state, ReadyState)
+            peer_best_blockchain = list(sorted(conn.state.peer_best_blockchain, key=lambda x: x.height, reverse=True))
```
alexruzenhack, line 323: Usually the `peer_best_blockchain` will come already sorted in descending order. If this sorting is already optimized, then the time complexity would be `O(n)`, assuming `n` as the number of elements in the `peer_best_blockchain`. Given `n` always equal `10`, it yields `O(1)` for this instruction. But everything depends on the `sorted` implementation.

alexruzenhack (approved):

jansegre (approved):
Since the scope of use of this is limited to `/v1a/p2p/readiness` (and `/v1a/health` in a subsequent PR), I'm OK merging this even if the change isn't ideal.

But there is a fundamental issue using self-reported unverified height from other peers. We might want to open an issue to improve this in the future. I'm not sure what the solution would be, maybe only considering trusted peers in this check, but for now we should be aware that a peer could simply report incorrect information and this could immediately affect the p2p healthcheck of all peers.

luislhl:
> Since the scope of use of this is limited to `/v1a/p2p/readiness` (and `/v1a/health` in a subsequent PR), I'm OK merging this even if the change isn't ideal.
> 
> But there is a fundamental issue using self-reported unverified height from other peers. We might want to open an issue to improve this in the future. I'm not sure what the solution would be, maybe only considering trusted peers in this check, but for now we should be aware that a peer could simply report incorrect information and this could immediately affect the p2p healthcheck of all peers.

Makes sense. In this case I prefer to close this PR, it's not essential for the health check, it was just a nice-to-have. But with this flaw I think it becomes unworthy.

PR #806, opened 2023-10-06
chore: improve push_tx log

glevco:
### Motivation

The `/push_tx` API currently generates alerts when it fails. It should simply log the reason, but not generate an alert. For more info, read https://github.com/HathorNetwork/on-call-incidents/issues/97.

### Acceptance Criteria

- Substitute `exc_info` by an human-readable message from the warning logged by `/push_tx`. By removing the "Traceback" from the log, an alert is not fired anymore.

### Checklist

### DIFF
--- a/hathor/transaction/resources/push_tx.py
+++ b/hathor/transaction/resources/push_tx.py
@@ -101,9 +101,9 @@ class PushTxResource(Resource):
             self.manager.push_tx(tx, allow_non_standard_script=self.allow_non_standard_script,
                                  max_output_script_size=self.max_output_script_size)
         except (InvalidNewTransaction, TxValidationError) as e:
-            self.log.warn('push tx rejected', exc_info=True)
             success = False
             message = str(e)
+            self.log.warn('push tx rejected', reason=repr(e))
         data = {'success': success, 'message': message}
         if success:
             data['tx'] = tx.to_json()


luislhl (approved):

hathor/transaction/resources/push_tx.py
```diff
@@ -101,9 +101,9 @@ def handle_push_tx(self, params: dict[str, Any], client_addr: str) -> dict[str,
             self.manager.push_tx(tx, allow_non_standard_script=self.allow_non_standard_script,
                                  max_output_script_size=self.max_output_script_size)
         except (InvalidNewTransaction, TxValidationError) as e:
-            self.log.warn('push tx rejected', exc_info=True)
             success = False
             message = str(e)
+            self.log.warn('push tx rejected', reason=message)
```
msbrogli, line 106: I agree that removing the stack trace is good but I think we should improve the error message so we can debug when a transaction fails to be pushed. Maybe add the exception type to the message?

msbrogli (changes requested):

hathor/transaction/resources/push_tx.py
glevco: I changed it in https://github.com/HathorNetwork/hathor-core/pull/806/commits/bd854c17ae6c8a38ebe99039d4cc1f6a5a27cd69 to use `repr(e)` instead of `str(e)`, which includes the exception type.

msbrogli: Thanks!

msbrogli (dismissed):

luislhl (approved):

luislhl (approved):

msbrogli (approved):

PR #807, opened 2023-10-06
feat[healthcheck]: new healthcheck response format

luislhl:
### Motivation

We are building a new healthcheck standard

### Acceptance Criteria

- Create a `/health` endpoint that will check the health of the p2p sync and return our new health check format

### TODO

- [ ] Generate the new Nginx image

### Checklist

### DIFF
--- a/hathor/healthcheck/resources/healthcheck.py
+++ b/hathor/healthcheck/resources/healthcheck.py
@@ -1,5 +1,5 @@
 import hathor
-from hathor.api_util import Resource
+from hathor.api_util import Resource, get_arg_default, get_args
 from hathor.cli.openapi_files.register import register_resource
 from hathor.healthcheck.models import ComponentHealthCheck, ComponentType, HealthCheckStatus, ServiceHealthCheck
 from hathor.manager import HathorManager
@@ -29,8 +29,15 @@ class HealthcheckResource(Resource):
         """ GET request /health/
             Returns the health status of the fullnode
 
+            The 'strict_status_code' argument can be used to return 200 even if the fullnode is unhealthy.
+            This can be useful when integrating with tools that could prefer to pass the response code only
+            in case the response is 200.
+
             :rtype: string (json)
         """
+        raw_args = get_args(request)
+        strict_status_code = get_arg_default(raw_args, 'strict_status_code', '0') == '1'
+
         components_health_checks = [
             build_sync_health_status(self.manager)
         ]
@@ -40,7 +47,10 @@ class HealthcheckResource(Resource):
             checks={c.component_name: [c] for c in components_health_checks},
         )
 
-        request.setResponseCode(health_check.get_http_status_code())
+        if not strict_status_code:
+            status_code = health_check.get_http_status_code()
+            request.setResponseCode(status_code)
+
         return json_dumpb(health_check.to_json())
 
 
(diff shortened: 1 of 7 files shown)


hathor/healthcheck/resources/healthcheck.py
```diff
@@ -0,0 +1,164 @@
...
+                                        'status': 'pass',
+                                        'description': 'Hathor-core v0.56.0',
+                                        'components': [
+                                            {
+                                                'componentName': 'sync',
+                                                'componentType': 'internal',
+                                                'status': 'pass',
+                                                'output': 'Healthy'
+                                            }
+                                        ]
```
alexruzenhack, line 100: Shouldn't it be?
```suggestion
                                        'checks': {
                                            'sync': [{
                                                'componentName': 'sync',
                                                'componentType': 'internal',
                                                'status': 'pass',
                                                'output': 'Healthy'
                                            }]
                                        }
```

luislhl: You're totally right, I'll fix this.

alexruzenhack (dismissed):

hathor/healthcheck/resources/healthcheck.py
```diff
@@ -0,0 +1,172 @@
...
+        },
+        'get': {
+            'tags': ['healthcheck'],
+            'operationId': 'get',
+            'summary': 'Health status of the fullnode',
+            'description': '''
+Returns 200 if the fullnode should be considered healthy.
+
+Returns 503 otherwise. The response will contain the components that were considered for the healthcheck
+and the reason why they were unhealthy.
```
jansegre, line 75: Is there any particular motivation for using 503? My intuition would be that the direct response from the fullnode would always be 200 if it was able to compute the request and the content of the request would have the health status. I'd expect 503 to be returned by a proxy that couldn't reach the fullnode, in case we use nginx (which we typically do) and the request timeouts, it will respond with a 503, however if the fullnode also responds with 503 normally we would only be able to differentiate by looking at the content (json response from fullnode vs html response from nginx).

If this is by design then I think the motivation should be commented more clearly. Otherwise it seems better to just use 200 and let the caller look at the content. As in, 200 always means the response came from the fullnode and has a predictable json format.

luislhl: Most infrastructural tooling that work with health checks expect them to work based on HTTP codes, so returning any code > 399 facilitates integration with them.

Even the uptime monitoring tools that we evaluated to use do a better job with HTTP codes than reading the response code. Most of them have at most a basic support to parsing the response code that usually just searches for the presence of a string on them, if at all.

AWS Load Balancers themselves are another instance of this. Their health check works solely based on HTTP codes.

So I think we should stick to this to have an easier integration with any such tool. But I could change the code from 503 to any other that is > 399.

What do you think?

jansegre: It looks like 503 would be useful in practice then. If in practice, the content of the response isn't important when the status is 503, I'm fine using it as is and just commenting that we use 503 because it integrates better with external tools.

Otherwise, if the content ends up being used in practice it could be useful to have some way of signaling (possibly with a GET parameter like `?strict_status_code=1`) that we want a code 200 with even with an error response (or the other way around depending on what we choose as a default), that way the tools that do use the content can more easily expect 200 for knowing they can parse the response content.

luislhl: I think it's interesting to allow choosing the behavior. I'll implement that.

luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/807/commits/456a63bd3125b248fece5ea53259774633b60349

hathor/healthcheck/resources/healthcheck.py
```diff
@@ -0,0 +1,200 @@
...
+        if strict_status_code:
+            request.setResponseCode(200)
+        else:
+            status_code = health_check.get_http_status_code()
+            request.setResponseCode(status_code)
+
+        return json_dumpb(health_check.to_json())
+
+
+# TODO: Fix below
```
luislhl, line 59: I'll remove this, it's an old TODO

luislhl:
```suggestion
```

jansegre (dismissed):

jansegre (approved):

alexruzenhack (approved):

jansegre (approved):

alexruzenhack (approved):

PR #809, opened 2023-10-12
feat(unrecoverable-error): implement halting of full node execution

glevco:
### Motivation

Currently, when an exception happens during a consensus update, it marks the tx as voided with a custom marker and continues to operate. This operation may be faulty, though, as the database is likely in an undesired state. For example, if such exception happens when a block is received, no following block will be accepted and the full node will not be able to sync anymore. If the full node is manually stopped and restarted, it starts up but continues to be unable to sync.

This PR's goal is to, instead, completely halt and exit the full node in those cases, forcing manual intervention. When some exception happens during a consensus update, the full node process exits with a non-zero exit code. This also guarantees that the database is marked as corrupted, so the full node cannot be restarted normally, only by a full verification or new database.

This PR is only _more_ restrictive than what's currently implemented, and will be necessary for the Feature Activation for Transactions.

### Acceptance Criteria

- Create new `ExecutionManager` with `crash_and_exit()` method.
- Create new related methods in `EventManager` and `TransactionStorage` for dealing with a full node crash.
- Change handling of exceptions during consensus so it halts and exit the full node, in addition to voiding the tx.

### Checklist

### DIFF
--- a/hathor/consensus/consensus.py
+++ b/hathor/consensus/consensus.py
@@ -18,6 +18,7 @@ from hathor.conf.get_settings import get_global_settings
 from hathor.consensus.block_consensus import BlockConsensusAlgorithmFactory
 from hathor.consensus.context import ConsensusAlgorithmContext
 from hathor.consensus.transaction_consensus import TransactionConsensusAlgorithmFactory
+from hathor.execution_manager import ExecutionManager
 from hathor.profiler import get_cpu_profiler
 from hathor.pubsub import HathorEvents, PubSubManager
 from hathor.transaction import BaseTransaction
@@ -55,13 +56,20 @@ class ConsensusAlgorithm:
     b0 will not be propagated to the voided_by of b1, b2, and b3.
     """
 
-    def __init__(self, soft_voided_tx_ids: set[bytes], pubsub: PubSubManager) -> None:
+    def __init__(
+        self,
+        soft_voided_tx_ids: set[bytes],
+        pubsub: PubSubManager,
+        *,
+        execution_manager: ExecutionManager
+    ) -> None:
         self._settings = get_global_settings()
         self.log = logger.new()
         self._pubsub = pubsub
         self.soft_voided_tx_ids = frozenset(soft_voided_tx_ids)
         self.block_algorithm_factory = BlockConsensusAlgorithmFactory()
         self.transaction_algorithm_factory = TransactionConsensusAlgorithmFactory()
+        self._execution_manager = execution_manager
 
     def create_context(self) -> ConsensusAlgorithmContext:
         """Handy method to create a context that can be used to access block and transaction algorithms."""
@@ -75,11 +83,11 @@ class ConsensusAlgorithm:
         assert meta.validation.is_valid()
         try:
             self._unsafe_update(base)
-        except Exception:
+        except BaseException:
             meta.add_voided_by(self._settings.CONSENSUS_FAIL_ID)
             assert base.storage is not None
             base.storage.save_transaction(base, only_metadata=True)
-            raise
+            self._execution_manager.crash_and_exit(reason=f'Consensus update failed for tx {base.hash_hex}')
 
     def _unsafe_update(self, base: BaseTransaction) -> None:
         """Run a consensus update with its own context, indexes will be updated accordingly."""
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -22,6 +22,7 @@ from hathor.event.model.event_type import EventType
 from hathor.event.model.node_state import NodeState
 from hathor.event.storage import EventStorage
 from hathor.event.websocket import EventWebsocketFactory
+from hathor.execution_manager import ExecutionManager
 from hathor.pubsub import EventArguments, HathorEvents, PubSubManager
 from hathor.reactor import ReactorProtocol as Reactor
 from hathor.transaction import BaseTransaction
@@ -70,6 +71,7 @@ class EventManager:
         event_storage: EventStorage,
         pubsub: PubSubManager,
         reactor: Reactor,
+        execution_manager: ExecutionManager,
         event_ws_factory: Optional[EventWebsocketFactory] = None,
     ) -> None:
         self.log = logger.new()
@@ -78,6 +80,7 @@ class EventManager:
         self._event_storage = event_storage
         self._event_ws_factory = event_ws_factory
         self._pubsub = pubsub
+        self._execution_manager = execution_manager
 
     def start(self, peer_id: str) -> None:
         """Starts the EventManager."""
@@ -85,6 +88,7 @@ class EventManager:
         assert self._event_ws_factory is not None, 'Cannot start, EventWebsocketFactory is not set'
         assert self.get_event_queue_state() is True, 'Cannot start, event queue feature is disabled'
 
+        self._execution_manager.register_on_crash_callback(self.on_full_node_crash)
         self._previous_node_state = self._event_storage.get_node_state()
 
         if self._should_reload_events():
@@ -133,7 +137,7 @@ class EventManager:
         for event in _SUBSCRIBE_EVENTS:
             self._pubsub.subscribe(event, self._handle_hathor_event)
 
-    def load_started(self):
+    def load_started(self) -> None:
         if not self._is_running:
             return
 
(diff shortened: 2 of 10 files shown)


hathor/consensus/consensus.py
```diff
@@ -75,11 +76,13 @@ def update(self, base: BaseTransaction) -> None:
         assert meta.validation.is_valid()
         try:
             self._unsafe_update(base)
-        except Exception:
-            meta.add_voided_by(self._settings.CONSENSUS_FAIL_ID)
-            assert base.storage is not None
-            base.storage.save_transaction(base, only_metadata=True)
-            raise
+        except BaseException:
```
msbrogli, line 79: I think we should keep handling `Exception`. Even if it is a bug in the code, the transaction should be marked as fail.

hathor/consensus/consensus.py
```diff
@@ -75,11 +76,13 @@ def update(self, base: BaseTransaction) -> None:
         assert meta.validation.is_valid()
         try:
             self._unsafe_update(base)
-        except Exception:
-            meta.add_voided_by(self._settings.CONSENSUS_FAIL_ID)
```
msbrogli, line 79: I feel we should keep adding this marker. It's important for debugging.

hathor/consensus/consensus.py
```diff
@@ -75,11 +76,13 @@ def update(self, base: BaseTransaction) -> None:
...
-            assert base.storage is not None
-            base.storage.save_transaction(base, only_metadata=True)
-            raise
+        except BaseException:
+            self.log.critical(
```
msbrogli, line 80: Should we mark the database as dirt and prevent the full node from restarting? I feel that this situation might happen in other cases (e.g., database writing failure), so we might generalize this to a `fail_and_exit()` function.

hathor/consensus/consensus.py
```diff
@@ -75,11 +76,13 @@ def update(self, base: BaseTransaction) -> None:
...
+                'Consensus update failed, causing the full node to halt execution. Manual intervention is required.',
+                tx=base.hash_hex,
+                exc_info=True
+            )
+            os._exit(-1)
```
msbrogli, line 85: Should we emit an event to let integration know about this error?

msbrogli (changes requested):

hathor/consensus/consensus.py
glevco: Sure, I brought back that transaction marking

hathor/consensus/consensus.py
glevco: Sure, I brought back that transaction marking

hathor/consensus/consensus.py
glevco: Done!

glevco:
@msbrogli the code changed a lot, so I just rebased it. Could you please re-review all files?

jansegre (commented):
We could start using this now when a storage write fails. Do you think it'd be worth to include this in this PR?

hathor/event/event_manager.py
```diff
@@ -153,6 +153,15 @@ def load_finished(self):
         )
         self._event_storage.save_node_state(NodeState.SYNC)
 
+    def full_node_crashed(self) -> None:
```
msbrogli, line 156: Maybe rename to `on_full_node_crash()`.

hathor/execution_manager.py
```diff
@@ -0,0 +1,51 @@
...
+
+from structlog import get_logger
+
+from hathor.event import EventManager
+from hathor.transaction.storage import TransactionStorage
+
+logger = get_logger()
+
+
+class ExecutionManager:
```
msbrogli, line 26: Maybe add a `register_on_crash(self, callback, *, priority)` that will keep record of all calls to be executed when the full node crashes. You might use the `priority` to order the calls.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -978,6 +981,14 @@ def is_running_manager(self) -> bool:
         """
         return self.get_value(self._manager_running_attribute) == '1'
 
+    def full_node_crashed(self) -> None:
```
msbrogli, line 984: Maybe rename to `on_full_node_crash()`.

hathor/execution_manager.py
```diff
@@ -0,0 +1,51 @@
...
+        self._tx_storage.full_node_crashed()
+        self._event_manager.full_node_crashed()
+        self._log.critical(
+            'Critical failure occurred, causing the full node to halt execution. Manual intervention is required.',
+            reason=reason,
+            exc_info=True
+        )
+        # We use os._exit() instead of sys.exit() or any other approaches because this is the only one Twisted
+        # doesn't catch.
+        os._exit(-1)
```
msbrogli, line 51: I prefer to crash using `reactor.crash()` or `sys.exit()` which would allow flushes and other activities to happen.

msbrogli (changes requested):

hathor/execution_manager.py
msbrogli: Notice that we have to handle all exceptions raised by these callbacks so the `exit()` method is called.

glevco:
> We could start using this now when a storage write fails. Do you think it'd be worth to include this in this PR?

@jansegre I think we can do it in a separate PR

hathor/event/event_manager.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/809/commits/2938c0855a3efcceece808b8116061e25489dce9

hathor/execution_manager.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/809/commits/2938c0855a3efcceece808b8116061e25489dce9

hathor/transaction/storage/transaction_storage.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/809/commits/2938c0855a3efcceece808b8116061e25489dce9

hathor/execution_manager.py
glevco: Initially my idea was to purposefully halt execution _immediately_, preventing any other actions such as pending `Deferred`s getting executed. They pollute the logs with unrelated errors, but I guess it's okay. I've changed it in https://github.com/HathorNetwork/hathor-core/pull/809/commits/6978d71f9a18363e78d1d7d4724b512efd84dc73

jansegre (dismissed):

hathor/execution_manager.py
msbrogli: I've seen things like:

```python
try:
    sys.exit(-1)
except:
    os._exit(-1)
```

What do you think?

glevco: I'm not sure this works, since `sys.exit(-1)` simply raises a `SystemExit(-1)` exception. This would prevent the exception from bubbling up, and would call `os._exit(-1)` every time.

glevco: Unless that's the intended behavior, which I guess could make sense. The `sys.exit()` would perform its finalization operations, and then we would force the exit with the `os._exit()`, always. Is that your idea?

hathor/consensus/consensus.py
msbrogli: Thanks!

hathor/execution_manager.py
msbrogli: Great! Thanks!

hathor/consensus/consensus.py
msbrogli: Great! Thanks!

hathor/consensus/consensus.py
msbrogli: Great! Thanks!

hathor/event/event_manager.py
msbrogli: Thanks!

hathor/transaction/storage/transaction_storage.py

hathor/execution_manager.py
msbrogli: I saw the proposed changelog in another project, but I agree with your remarks on how it would actually work. I guess the current implementation is fine and it should be simple to change this behavior in the future if needed. So we need to move forward the way it is now. Thanks!

msbrogli (dismissed):

msbrogli (approved):

jansegre (approved):

PR #810, opened 2023-10-13
fix(sync-v2): Stop streaming if current block becomes voided

msbrogli:
### Motivation

Fix one of the conditions that makes the `tests/simulation/test_simulator.py::SyncV2RandomSimulatorTestCase::test_many_miners_since_beginning` flaky.

If we do not stop the streaming, the later call to `get_next_block_best_chain()` will raise an assertion error since the next block will be voided.

### Acceptance Criteria

1. Stop streaming blocks if the current block gets voided.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/streamers.py
+++ b/hathor/p2p/sync_v2/streamers.py
@@ -143,6 +143,12 @@ class BlockchainStreaming(_StreamingBase):
         assert cur is not None
         assert cur.hash is not None
 
+        meta = cur.get_metadata()
+        if meta.voided_by:
+            self.stop()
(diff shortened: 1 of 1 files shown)


jansegre (commented):
Does this really improve it? The test matrix failed in all but one job (at least on the first run). Also I think we've discussed this before, we previously had this behavior (stopping a stream when it reaches a voided block) but we changed it, and I don't remember why.

glevco:
@jansegre maybe it helped with one test, but others failed in the run. One way to check is this adding, for example, `@flaky(min_passes=20, max_runs=20)` to force the supposedly fixed test to run 20 times and pass all of them. We can run this locally just to check if all runs are passing (but keep it out of the PR).

jansegre (approved):

PR #811, opened 2023-10-17
refactor(scripts): modularize script files

glevco:
### Motivation

Refactor code so the single `scripts.py` file is broken into multiple modules, improving organization and maintainability. No behavior is changed by this PR, code is only moved between files. The only active changes are in imports to fix circular dependencies after the refactor.

### Acceptance Criteria

1. Move HathorScript to its own file 
1. Move BaseScript to its own file 
1. Move P2PKH to its own file 
1. Move MultiSig to its own file 
1. Move NanoContractMatchValues to its own file 
1. Move Opcode and functions to their own file
1. Remaining functions from the `scripts` file moved to either `construct` or `execute` modules
1. Fix imports and linter issues 

### Checklist


hathor/transaction/script/scripts.py
msbrogli, line 1: It seems this file still mixes things from "script execution" to "script utils to parse and build scripts".

msbrogli (changes requested):

hathor/transaction/script/scripts.py
glevco: Done! In two separate commits: https://github.com/HathorNetwork/hathor-core/pull/811/commits/6919f49bfaac10a147c0a1d5bd8e1df05628af84 and https://github.com/HathorNetwork/hathor-core/pull/811/commits/15050e3aaaffeea2957bb1651e3dea08ff21adfa

msbrogli: Thanks!

hathor/transaction/script/opcode.py
```diff
@@ -0,0 +1,640 @@
...
+    EqualVerifyFailed,
+    InvalidScriptError,
+    InvalidStackData,
+    MissingStackItems,
+    OracleChecksigFailed,
+    ScriptError,
+    TimeLocked,
+    VerifyFailed,
+)
+from hathor.transaction.script.construct import (
```
msbrogli, line 41: I feel all these methods should be in the `hathor.transaction.script.execution` module. Even though they are also used by parsers and constructors, they were developed to be used during script execution.

hathor/transaction/script/__init__.py
msbrogli, line 1: Why not import common methods and classes here so we don't have to change so many imports all over the project?

msbrogli (changes requested):

hathor/transaction/script/opcode.py
glevco: Done!

hathor/transaction/script/__init__.py
glevco: Good idea, done! I ended up rebasing and pushing a single commit, as now the number of changed files was dramatically reduced.

msbrogli (approved):

hathor/transaction/script/opcode.py
msbrogli: Thanks!

hathor/transaction/script/__init__.py
msbrogli: Well done! Thanks!

msbrogli (approved):

jansegre (approved):

PR #814, opened 2023-10-17
test: Use INFO as default log level to reduce amount of logs

msbrogli:
### Motivation

Current log size is too big when a test fails. It's so big that it does not provide any help to understand what went wrong.

### Acceptance Criteria

1. Print seed using INFO log level.
2. Change default log level for INFO for all tests.

### Checklist

### DIFF
--- a/hathor/cli/util.py
+++ b/hathor/cli/util.py
@@ -157,6 +157,9 @@ def setup_logging(
     else:
         handlers = ['pretty']
 
+    # Flag to enable debug level for both sync-v1 and sync-v2.
+    debug_sync = False and debug
+
     # See: https://docs.python.org/3/library/logging.config.html#configuration-dictionary-schema
     logging.config.dictConfig({
             'version': 1,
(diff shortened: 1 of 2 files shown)


glevco (approved):

hathor/cli/util.py
```diff
@@ -157,6 +157,9 @@ def setup_logging(
     else:
         handlers = ['pretty']
 
+    # Flag to enable debug level for both sync-v1 and sync-v2.
+    debug_sync = False and debug
```
glevco, line 161: Why keep this if it always evaluates to `False`?

jansegre (approved):

hathor/cli/util.py
msbrogli: I thought it might be useful when someone is debugging in the future. Just change this flag and both sync-v1 and sync-v2 will log on DEBUG level.

glevco: Ah, ok!

PR #815, opened 2023-10-17
test(simulation): Fix flaky test test_many_miners_since_beginning

msbrogli:
### Background

The following tests were too flaky:

- `tests/simulation/test_simulator.py::SyncV1RandomSimulatorTestCase::test_many_miners_since_beginning`
- `tests/simulation/test_simulator.py::SyncV2RandomSimulatorTestCase::test_many_miners_since_beginning`
- `tests/simulation/test_simulator.py::SyncBridgeRandomSimulatorTestCase::test_many_miners_since_beginning`

This PR does not fix the `HeightIndex` behavior. It just fixes the test itself.

Before this fix, one of these tests were failing every 3 executions. After this fix, I could locally run each of the three tests more than 50 times with only one failure coming from some other edge case. For clarity, 1 fail out of 150+ executions.

The only fail I got happened with seed `4231149425068099915`. Here is the log: `FAILED tests/simulation/test_simulator.py::SyncV2RandomSimulatorTestCase::test_many_miners_since_beginning - twisted.trial.unittest.FailTest: Items in the first set but not the second:`.

#### Root cause analysis

The `self.assertTrue(self.simulator.run(3600, trigger=AllTriggers(stop_triggers)))` was failing. In other words, the `run(...)` method was returning `False`.

Why?

Because the `AllTriggers(stop_triggers))` was not stopping the execution.

Why?

Because the `StopWhenSynced` trigger was using the `tx_storage.get_n_height_tips()` which was returning different block hashes for the nodes.

Why?

Because there's an edge case where all blockchains had exactly the same size and the same score (i.e., a tie between all blockchains). In this case, the `HeightIndex` does not update at all and keep returning the previous best blockchain, which is different for each full node. As `get_n_height_tips()` method uses the `HeightIndex`, the block hashes were different.

### Acceptance Criteria

1. Add some debugging tool to `hathor.simulator.FakeConnection.is_both_synced()`.
2. Remove `assertTrue(...)` in `self.assertTrue(self.simulator.run(3600, trigger=AllTriggers(stop_triggers)))`.

### Checklist

### DIFF
--- a/tests/simulation/test_simulator.py
+++ b/tests/simulation/test_simulator.py
@@ -87,7 +87,9 @@ class BaseRandomSimulatorTestCase(SimulatorTestCase):
         for miner in miners:
             miner.stop()
 
-        self.assertTrue(self.simulator.run(3600, trigger=AllTriggers(stop_triggers)))
+        # TODO Add self.assertTrue(...) when the trigger is fixed.
+        #      For further information, see https://github.com/HathorNetwork/hathor-core/pull/815.
+        self.simulator.run(3600, trigger=AllTriggers(stop_triggers))
 
         for node in nodes[1:]:
             self.assertTipsEqual(nodes[0], node)
--- a/hathor/simulator/fake_connection.py
+++ b/hathor/simulator/fake_connection.py
@@ -71,13 +71,16 @@ class FakeConnection:
         self._proto1.disable_idle_timeout()
         self._proto2.disable_idle_timeout()
 
-    def is_both_synced(self) -> bool:
+    def is_both_synced(self, *, errmsgs: Optional[list[str]] = None) -> bool:
         """Short-hand check that can be used to make "step loops" without having to guess the number of iterations."""
+        if errmsgs is None:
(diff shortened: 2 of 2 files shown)


glevco (approved):

tests/simulation/test_simulator.py
```diff
@@ -87,7 +87,7 @@ def test_many_miners_since_beginning(self):
         for miner in miners:
             miner.stop()
 
-        self.assertTrue(self.simulator.run(3600, trigger=AllTriggers(stop_triggers)))
+        self.simulator.run(3600, trigger=AllTriggers(stop_triggers))
```
jansegre, line 90: Should we add a TODO to use `self.assertTrue` again in the future when the trigger's behavior is fixed?

jansegre (approved):

tests/simulation/test_simulator.py
msbrogli: Done!

jansegre (approved):

PR #826, opened 2023-10-24
chore: refactor python healthcheck lib

luislhl:
### Motivation

Refactor the [former healthcheck logic](https://github.com/HathorNetwork/hathor-core/pull/807) to use the new healthcheck lib we've built in https://github.com/HathorNetwork/python-healthcheck-lib

### Acceptance Criteria

- The /health endpoint should keep working like described in https://github.com/HathorNetwork/hathor-core/pull/807

### TODO

- [x] The lib is being installed directly from the Git repo to make sure we can test it. But we should install it from PyPI when it's available there.

### Checklist


pyproject.toml
```diff
@@ -79,6 +79,7 @@ hathorlib = "0.3.0"
 pydantic = "~1.10.13"
 pyyaml = "^6.0.1"
 typing-extensions = "~4.8.0"
+python-healthchecklib = {git = "https://github.com/hathornetwork/python-healthcheck-lib"}
```
glevco, line 82: Should we pin a specific commit/release?

glevco (dismissed):

pyproject.toml
luislhl: Done in https://github.com/HathorNetwork/hathor-core/pull/826/commits/fb612398edbb9ece11b4ad190ae2512688306cad

glevco (dismissed):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.6.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
```
jansegre, line 1: It seems this was generated with an outdated version of Poetry. I don't think there's any issue in this case, but sometimes there is. I think we probably don't have a clear guideline for adding/updating dependencies but for when we have we should keep in mind that Poetry should be updated before making any changes.

jansegre (dismissed):

jansegre (approved):

glevco (approved):

jansegre (approved):

glevco (approved):

PR #828, opened 2023-10-26
fix(feature-activation): add missing metadata migration

glevco:
### Motivation

During the latest release, we noticed that full node initialization could fail depending on the snapshot state (only on testnet). This was caused by a missing migration on the new Feature Activation metadata attribute.

### Acceptance Criteria

- Rename `BitSignalingService.start()` to `generate_initial_logs()` and add docstring. Also move call to after migrations are applied in `HathorManager`.
- Implement `add_feature_activation_bit_counts_metadata` migration.

### Checklist

### DIFF
--- a/hathor/feature_activation/bit_signaling_service.py
+++ b/hathor/feature_activation/bit_signaling_service.py
@@ -54,6 +54,10 @@ class BitSignalingService:
         self._validate_support_intersection()
 
     def start(self) -> None:
+        """
+        Log information related to bit signaling. Must be called after the storage is ready and migrations have
+        been applied.
+        """
         best_block = self._tx_storage.get_best_block()
 
         self._warn_non_signaling_features(best_block)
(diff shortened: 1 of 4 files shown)


hathor/feature_activation/bit_signaling_service.py
```diff
@@ -53,7 +53,11 @@ def __init__(
 
         self._validate_support_intersection()
 
-    def start(self) -> None:
+    def generate_initial_logs(self) -> None:
```
msbrogli, line 56: Even though I understand the expected pre-conditions to run this method, I think that it should be named `start()`.

msbrogli (changes requested):

hathor/feature_activation/bit_signaling_service.py
glevco: Ok, changed it back in https://github.com/HathorNetwork/hathor-core/pull/828/commits/366e6ae39b96da97bd98bc11fc20f47eef39de4b

msbrogli: Thanks!

msbrogli (approved):

jansegre (approved):

PR #829, opened 2023-10-26
refactor(sync-v2): Refactor sync_v2 agent to hold (block_height, block_id) information in an internal namedtuple

msbrogli:
### Motivation

Sync-v2 agent code uses the pair `(block_height, block_hash)` multiple times. So this PR refactors it to use a namedtuple `BlockInfo` (instead of handling with both information separately).

### Acceptance Criteria

1. Add `hathor.p2p.sync_v2.agent.BlockInfo`.
2. Add `hathor.p2p.sync_v2.agent.NodeBlockSync.get_my_best_block()` method.
3. Refactor `find_best_common_block()` and `run_sync_blocks()` methds to use `BlockInfo`.
4. Rename variables `(synced, not_synced)` to `(lo, hi)` in the n-ary search.
5. There are two behavior changes in this PR: (i) change of keys in the `get_status()`, and (ii) using `lo=0` instead of `lo=self.synced_block.height`.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/agent.py
+++ b/hathor/p2p/sync_v2/agent.py
@@ -18,7 +18,7 @@ import math
 import struct
 from collections import OrderedDict
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Generator, Optional, cast
+from typing import TYPE_CHECKING, Any, Callable, Generator, NamedTuple, Optional, cast
 
 from structlog import get_logger
 from twisted.internet.defer import Deferred, inlineCallbacks
@@ -42,6 +42,20 @@ if TYPE_CHECKING:
 logger = get_logger()
 
 MAX_GET_TRANSACTIONS_BFS_LEN: int = 8
+
+
+class _HeightInfo(NamedTuple):
+    height: int
+    id: VertexId
+
+    def __repr__(self):
+        return f'_HeightInfo({self.height}, {self.id.hex()})'
+
+    def to_json(self) -> dict[str, Any]:
+        return {
+            'height': self.height,
+            'id': self.id.hex(),
+        }
 
 
 class PeerState(Enum):
@@ -92,16 +106,16 @@ class NodeBlockSync(SyncAgent):
         self.receiving_stream = False
 
         # highest block where we are synced
-        self.synced_height = 0
+        self.synced_block: Optional[_HeightInfo] = None
 
         # highest block peer has
-        self.peer_height = 0
+        self.peer_best_block: Optional[_HeightInfo] = None
 
         # Latest deferred waiting for a reply.
         self._deferred_txs: dict[VertexId, Deferred[BaseTransaction]] = {}
         self._deferred_tips: Optional[Deferred[list[bytes]]] = None
-        self._deferred_best_block: Optional[Deferred[dict[str, Any]]] = None
-        self._deferred_peer_block_hashes: Optional[Deferred[list[tuple[int, bytes]]]] = None
+        self._deferred_best_block: Optional[Deferred[_HeightInfo]] = None
+        self._deferred_peer_block_hashes: Optional[Deferred[list[_HeightInfo]]] = None
 
         # When syncing blocks we start streaming with all peers
         # so the moment I get some repeated blocks, I stop the download
@@ -151,8 +165,8 @@ class NodeBlockSync(SyncAgent):
         """
         res = {
             'is_enabled': self.is_sync_enabled(),
-            'peer_height': self.peer_height,
-            'synced_height': self.synced_height,
+            'peer_best_block': self.peer_best_block.to_json() if self.peer_best_block else None,
+            'synced_block': self.synced_block.to_json() if self.synced_block else None,
             'synced': self._synced,
             'state': self.state.value,
         }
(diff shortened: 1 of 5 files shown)


hathor/p2p/sync_v2/agent.py
```diff
@@ -44,6 +44,20 @@
 MAX_GET_TRANSACTIONS_BFS_LEN: int = 8
 
 
+class BlockInfo(NamedTuple):
```
msbrogli, line 47: Maybe rename to `HeightInfo`?

glevco: We have a `HeightInfo` class in `hathor.indexes` that is exactly the same. Maybe move that one to some common module and use it in both places? I feel this tuple is so simple that it makes sense for it to be more general.

hathor/p2p/sync_v2/agent.py
```diff
@@ -332,37 +346,43 @@ def run_sync_transactions(self) -> None:
                       end_block_height=block_height)
         self.send_get_transactions_bfs(needed_txs, block.hash)
 
+    def get_my_best_block(self) -> BlockInfo:
+        """Return my best block info."""
+        bestblock = self.tx_storage.get_best_block()
+        assert bestblock.hash is not None
+        meta = bestblock.get_metadata()
+        # assert not meta.voided_by
```
glevco, line 354: Remove or uncomment this?

hathor/p2p/sync_v2/agent.py
```diff
@@ -813,21 +833,23 @@ def handle_get_best_block(self, payload: str) -> None:
         """
         best_block = self.tx_storage.get_best_block()
         meta = best_block.get_metadata()
+        assert meta.validation.is_fully_connected()
+        # assert not meta.voided_by
```
glevco, line 837: Remove or uncomment this?

glevco (dismissed):

hathor/p2p/sync_v2/agent.py
```diff
@@ -151,8 +165,8 @@ def get_status(self) -> dict[str, Any]:
         """
         res = {
             'is_enabled': self.is_sync_enabled(),
-            'peer_height': self.peer_height,
-            'synced_height': self.synced_height,
+            'peer_best_block': self.peer_best_block.to_json() if self.peer_best_block else None,
+            'synced_block': self.synced_block.to_json() if self.synced_block else None,
```
jansegre, line 169: Although the change is good, we should probably create an issue to update the monitoring tools, which I believe already implemented support for the current API. Maybe even keeping the current fields alongside the new ones before completely removing them.

jansegre (dismissed):

hathor/p2p/sync_v2/agent.py
msbrogli: We can easily keep the old format. I think this format is weird because the height doesn't tell the whole story. I mean, the fact that two peers have the height does not mean they are synced.

hathor/p2p/sync_v2/agent.py
msbrogli: I'll leave a comment. I think this assert should be there but the current implementation does not work with it.

hathor/p2p/sync_v2/agent.py
msbrogli: Same here. I'll leave a comment. I think this assert should be there but the current implementation does not work with it.

hathor/p2p/sync_v2/agent.py
jansegre: Yes, but it gives a good idea on the sync progress, from an external point of view, if this API only returns the block's hash, and we don't have that block, we'd need to query the node again on another API to learn that block's height.

hathor/p2p/sync_v2/agent.py
msbrogli: Even though both tuples have the same name `HeightInfo`, they are both used internally and might be independently modified in the future. So I guess we should keep them separate. What do you think?

hathor/p2p/sync_v2/agent.py
jansegre: I think the assertion is true when there is no tie, otherwise it's possible that the block returned by `get_best_block` becomes voided when the block with a tie is added (because the previous block won't be removed from the height index).

hathor/p2p/sync_v2/agent.py
msbrogli: Just renamed it to `_HeightInfo`. Both the commit message and the PR title have been updated.

hathor/p2p/sync_v2/agent.py
msbrogli: I agree that we can remove it for now. Done! f880a62fac1873f7d8abb384f6209d1b919e6e46

hathor/p2p/sync_v2/agent.py
msbrogli: I agree that we can remove it for now. Done! f880a62fac1873f7d8abb384f6209d1b919e6e46

hathor/p2p/sync_v2/agent.py
msbrogli: The current change returns both the block id and the height.

msbrogli: Should I add the hashes on separate fields? That's another way to go here and keep compatibility.

hathor/p2p/sync_v2/agent.py
glevco: Ok, looks good to me

glevco (dismissed):

hathor/p2p/sync_v2/agent.py
jansegre: Oh, right it's a block-info, not a hash. I think it's good as is, the monitoring should be able to catch up to this change.

jansegre (dismissed):

glevco (approved):

Issue #825, opened 2023-10-24
Move `verify_unsigned_skip_pow()`


glevco:
Maybe we should open a KTLO to rename `Transaction.verify_unsigned_skip_pow()` to something else as it's not part of the verification process. I checked that it is used by an HTTP Endpoint so we might move the code to this specific endpoint.

_Originally posted by @msbrogli in https://github.com/HathorNetwork/hathor-core/pull/824#discussion_r1369336691_

PR #830, opened 2023-10-27
refactor(verification): organization and typing improvements [part 1/9]

glevco:
### Motivation

This PR is the first one in a series of 9 PRs that finishes the vertex verification refactors. The objective is to completely remove all inheritance from the verification process, removing implicit dependencies between verifiers, streamlining the verification process and improving clarity on the scope/origin of each verification method, all to improve code readability, maintainability, and patching (for the simulator).

This is done incrementally. In previous PRs, we moved the verification out of the model classes. Now, in this series, we'll first change inheritance to composition, and then completely remove the composition too, leaving each verifier with the least possible amount of dependencies.

Also, in this PR, we install the [typing_extensions](https://github.com/python/typing_extensions) dev dependency. This is a project from the official Python organization, and what it does is enable use of new type system features on older Python versions. In other words, it simply allows us to use features from the `typing` module that are not available in all Python versions we support.

Specifically, for this refactor we'll use [`assert_never()`](https://typing.readthedocs.io/en/latest/source/unreachable.html), a function to be used were we would normally use something like `assert False`, but that can be correctly type-checked by mypy. This improves robustness of `match` statements for example, that can be made exhaustive with this. We also use the [`@override`](https://peps.python.org/pep-0698/) annotation in next PRs, improving robustness and safety guarantees of inheritance. We may also find other useful typing features in the future that wouldn't be available otherwise.

### Acceptance Criteria

- Change `BaseTransaction` and all subclasses to use `TxVersion` as the `version` type, instead of `int`.
- Install the `typing_extensions` dev dependency.
- Move `verify_unsigned_skip_pow()` from `TransactionVerifier` to `create_tx.py`. This resolves https://github.com/HathorNetwork/hathor-core/issues/825.
- In `VerificationService`:
  - Change all `isinstance()` checks to `type() is`, which is more restrictive.
  - Change `raise NotImplementedError` usages in `match` statements to `assert_never()`, which can be checked by the static type checker.
  - Remove `validate_vertex_error()`, inlining it `HathorManager`.
- Set `self.verifiers` variable in `test_verification`, updating all usages. This will be useful in the next PRs.
- No behavior should be changed by this PR.

### Checklist

### DIFF
--- a/hathor/verification/verification_service.py
+++ b/hathor/verification/verification_service.py
@@ -14,11 +14,12 @@
 
 from typing import NamedTuple
 
+from typing_extensions import assert_never
+
 from hathor.conf.settings import HathorSettings
 from hathor.daa import DifficultyAdjustmentAlgorithm
 from hathor.feature_activation.feature_service import FeatureService
 from hathor.transaction import BaseTransaction, Block, MergeMinedBlock, Transaction, TxVersion
-from hathor.transaction.exceptions import TxValidationError
 from hathor.transaction.token_creation_tx import TokenCreationTransaction
 from hathor.transaction.validation_state import ValidationState
 from hathor.verification.block_verifier import BlockVerifier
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -44,6 +44,7 @@ pytest-cov = "~4.0.0"
 flaky = "~3.7.0"
 pytest-xdist = "~3.2.0"
 yamllint = "~1.31.0"
+typing-extensions = "~4.8.0"
```
msbrogli, line 47: I'm not sure we should add this dependency only for `assert_nerver()`. I think we should keep doing `assert False` and we can change all these asserts after we deprecate python 3.10.

hathor/verification/verification_service.py
```diff
@@ -101,74 +102,59 @@ def verify_basic(self, vertex: BaseTransaction, *, skip_block_weight_verificatio
         Used by `self.validate_basic`. Should not modify the validation state."""
         match vertex.version:
             case TxVersion.REGULAR_BLOCK:
-                assert isinstance(vertex, Block)
+                assert type(vertex) is Block
```
msbrogli, line 105: Just add a comment saying that we cannot use `isinstance()` here because we're not interest in inheritance.

msbrogli (approved):

hathor/verification/verification_service.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/830/commits/6d844b94a9889a236b30f525f03754db04a7b087

msbrogli: Thanks!

pyproject.toml
msbrogli: You're right. Ok!

msbrogli (dismissed):

jansegre (approved):

msbrogli (approved):

pyproject.toml
jansegre: I think it can only be considered a dev-dependency if it's not needed in runtime. But since `from typing_extensions import assert_never` is used directly in a module runtime would fail if `typing_extensions` is not installed. The only thing preventing it from failing in runtime is that some other runtime dependency pulls this lib. So I think we should put it in the `[tool.poetry.dependencies]` section.

jansegre (approved):

pyproject.toml
glevco: Makes sense, I got it mixed up there. I'll change this. Thanks!

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/830/commits/99af497b6300f65304ed647aa26b2562cb1a16bf

jansegre (approved):

msbrogli (approved):

PR #831, opened 2023-10-27
refactor(verification): refactor verify_sum [part 2/9]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/830

### Motivation

Currently, `TokenCreationTransactionVerifier.verify_sum()` has duplicate code from `TransactionVerifier.verify_sum()`. This PR changes it so it calls the super method instead.

### Acceptance Criteria

- Implement `TransactionVerifier.get_complete_token_info()` and update its `verify_sum()` to use it.
- Implement `TokenCreationTransaction.get_token_info_from_inputs()`, overriding the `Transaction` implementation.
- Update `TokenCreationTransaction.verify_sum()` to use the new `get_complete_token_info()`.
- No behavior should be changed by this PR.

### Checklist

### DIFF
--- a/hathor/verification/token_creation_transaction_verifier.py
+++ b/hathor/verification/token_creation_transaction_verifier.py
@@ -12,10 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from hathor.transaction import Transaction
 from hathor.transaction.exceptions import InvalidToken, TransactionDataError
 from hathor.transaction.token_creation_tx import TokenCreationTransaction
-from hathor.transaction.transaction import TokenInfo, Transaction
 from hathor.transaction.util import clean_token_string
+from hathor.util import not_none
 from hathor.verification.transaction_verifier import TransactionVerifier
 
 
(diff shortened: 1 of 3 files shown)


hathor/verification/token_creation_transaction_verifier.py
```diff
@@ -39,20 +40,14 @@ def verify_sum(self, tx: Transaction) -> None:
...
+        token_dict = self.get_complete_token_info(tx)
 
         # make sure tokens are being minted
-        token_info = token_dict[tx.hash]
+        token_info = token_dict[not_none(tx.hash)]
         if token_info.amount <= 0:
             raise InvalidToken('Token creation transaction must mint new tokens')
 
-        self.verify_authorities_and_deposit(token_dict)
+        super().verify_sum(tx)
```
msbrogli, line 50: I guess we are calculating `token_info` twice if we change this call to `super().verify_sum(tx)`, right? In this case, it would also double the number of hits in the storage to load the inputs.

msbrogli (changes requested):

hathor/verification/token_creation_transaction_verifier.py
glevco: That's true, but it would most likely hit the cache. Do you think it could be a problem?

If yes, an alternative is changing the code so every dependency is fetched before the verification process starts, we've talked about this change. For example, we would fetch all inputs and parents, and provide them in a dict to the verifier methods, so no method would need to call the storage. The calculated `token_info` could be in this dependencies object, too. This change will also be useful for the future project of running verifications on a separate process, as it won't have to access the storage in parallel.

If considering the cache this is not problem, I would rather do this after the current refactors. If it is, I can do it now in a separate PR. What do you think?

msbrogli: I guess we should do it now in a separate PR because vertex validation is the bottleneck and I feel we shouldn't calculate the same thing more than once.

glevco: Ok, makes sense! I'll move this PR back to WIP.

glevco: This has been implemented in 2 PRs, that depend on this one. Here they are: https://github.com/HathorNetwork/hathor-core/pull/858 and https://github.com/HathorNetwork/hathor-core/pull/859. Let's review all 3 of them before merging this one, so they call all be merged together.

glevco: Considering what was discussed in a meeting, disregard the comment above. This has been fixed in a new PR: https://github.com/HathorNetwork/hathor-core/pull/868

msbrogli: Ok. Thanks!

msbrogli (approved):

PR #840, opened 2023-10-27
chore: update dependencies

glevco:
### Motivation

Update dependencies to improve static analysis and developer experience.

### Acceptance Criteria

- Update all dev-dependencies.
- Update Pydantic.
- Fix some new linter issues.

### Checklist

### DIFF
--- a/hathor/utils/named_tuple.py
+++ b/hathor/utils/named_tuple.py
@@ -46,8 +46,8 @@ def validated_named_tuple_from_dict(
 
     # This intermediate step shouldn't be necessary, but for some reason pydantic.create_model_from_namedtuple
     # doesn't support default attribute values, so we do this to add them
-    all_attributes = named_tuple_type(**attributes_dict)
+    all_attributes = named_tuple_type(**attributes_dict)  # type: ignore[call-overload]
     validated_attributes = model(**all_attributes._asdict())
     validated_attributes_dict = {k: v for k, v in validated_attributes}
 
-    return named_tuple_type(**validated_attributes_dict)
+    return named_tuple_type(**validated_attributes_dict)  # type: ignore[call-overload]
--- a/hathor/event/model/base_event.py
+++ b/hathor/event/model/base_event.py
@@ -66,7 +66,7 @@ class BaseEvent(BaseModel, use_enum_values=True):
         event_type = EventType(values['type'])
         expected_data_type = event_type.data_type()
 
-        if type(v) != expected_data_type:
+        if type(v) is not expected_data_type:
             raise ValueError('event data type does not match event type')
 
         return v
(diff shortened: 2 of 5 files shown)


hathor/utils/named_tuple.py
```diff
@@ -46,8 +46,8 @@ def validated_named_tuple_from_dict(
 
     # This intermediate step shouldn't be necessary, but for some reason pydantic.create_model_from_namedtuple
     # doesn't support default attribute values, so we do this to add them
-    all_attributes = named_tuple_type(**attributes_dict)
+    all_attributes = named_tuple_type.__call__(**attributes_dict)
```
msbrogli, line 49: Why do we need to use the `__call__` method? The previous code does not pass linters anymore?

msbrogli (approved):

hathor/utils/named_tuple.py
glevco: Yes. Here's the new error with the original code:

```python
hathor/utils/named_tuple.py:49: error: No overload variant of "NamedTuple" matches argument type "dict[str, Any]"  [call-overload]
        all_attributes = named_tuple_type(**attributes_dict)
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
hathor/utils/named_tuple.py:49: note: Possible overload variants:
hathor/utils/named_tuple.py:49: note:     def __init__(self, str, Iterable[tuple[str, Any]], /) -> T
hathor/utils/named_tuple.py:49: note:     def __init__(self, str, None = ..., /, **kwargs: Any) -> T
```

The `named_tuple_type` is a `type[T]` where `T = TypeVar('T', bound=NamedTuple)`. That's setup in a way so that I can receive a type as an argument (such as `HathorSettings`) that is a `NamedTuple`. I then need to perform operations using the type itself, and finally return a `T`, which is an instance of this type (a concrete `HathorSettings` object).

The problem is that mypy thinks `named_tuple_type` is a call to the `NamedTuple` _function_, instead of a call to initialize the object itself.

I couldn't find a better way around this, so I added the `__call__()` which worked. But now that you commented on it, I've reconsidered and decided to keep the original code, adding a mypy `ignore[call-overload]` on those lines (https://github.com/HathorNetwork/hathor-core/pull/840/commits/ce0a57751e45a9dbd68daa48a6ac2a129661fe5c). This function is already a bit complicated as it is, I think it's better to avoid introducing a new call to a non-standard method and being explicit about ignoring some specific mypy error.

msbrogli: I agree with your solution. Thanks for the explanation.

msbrogli (approved):

jansegre (approved):

PR #842, opened 2023-10-27
fix(sync-v2): Fix n-ary search to handle reorgs during its execution

msbrogli:
### Motivation

The n-ary search executed by sync-v2 was not taking into consideration that a reorg might occur during the search as it runs multiple async queries on the connected peer.

### Acceptance Criteria

1. Detect when a reorg occurs on the peer and stop the search. Notice that a new search will be executed on the next sync loop.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/agent.py
+++ b/hathor/p2p/sync_v2/agent.py
@@ -364,13 +364,17 @@ class NodeBlockSync(SyncAgent):
         # Get my best block.
         my_best_block = self.get_my_best_block()
 
-        # Find peer's best block
+        # Get peer's best block
         self.peer_best_block = yield self.get_peer_best_block()
         assert self.peer_best_block is not None
 
-        # find best common block
+        # Find best common block
         self.synced_block = yield self.find_best_common_block(my_best_block, self.peer_best_block)
-        assert self.synced_block is not None
+        if self.synced_block is None:
+            # Find best common block failed. Try again soon.
+            # This might happen if a reorg occurs during the search.
+            return
+
         self.log.debug('run_sync_blocks',
                        my_best_block=my_best_block,
                        peer_best_block=self.peer_best_block,
@@ -516,7 +520,7 @@ class NodeBlockSync(SyncAgent):
     @inlineCallbacks
     def find_best_common_block(self,
                                my_best_block: _HeightInfo,
-                               peer_best_block: _HeightInfo) -> Generator[Any, Any, _HeightInfo]:
+                               peer_best_block: _HeightInfo) -> Generator[Any, Any, Optional[_HeightInfo]]:
         """ Search for the highest block/height where we're synced.
         """
         self.log.debug('find_best_common_block', peer_best_block=peer_best_block, my_best_block=my_best_block)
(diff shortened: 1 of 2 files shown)


hathor/p2p/sync_v2/agent.py
```diff
@@ -545,36 +549,42 @@ def find_best_common_block(self,
         # Run an n-ary search in the interval [lo, hi).
         # `lo` is always a height where we are synced.
         # `hi` is always a height where sync state is unknown.
-        hi = min(peer_best_block.height, my_best_block.height)
-        lo = 0
-
-        lo_block_hash = self._settings.GENESIS_BLOCK_HASH
+        hi = min(peer_best_block, my_best_block)
```
glevco, line 552: **issue:** Using `min` on a `NamedTuple` is error prone.

I suggest using the `key` argument to select the block height, just like it's done in the `sort` below. Otherwise, this hinges on a specific ordering in the named tuple.

hathor/p2p/sync_v2/agent.py
```diff
@@ -545,36 +549,42 @@ def find_best_common_block(self,
...
+            block_info_list.sort(key=lambda x: x.height, reverse=True)
+
+            # As we are supposed to be always synced at `lo`, we expect to receive a response
+            # with at least one item equals to lo. If it does not happen, we stop the search
+            # and return None. This might be caused when a reorg occurs during the search.
+            if not block_info_list:
+                return None
+            lo_block_info = block_info_list[-1]
+            if lo_block_info != lo:
+                return None
```
glevco, line 571: **question**: Why -1?

glevco, line 571: **suggestion:** Do not use `None` to convey meaning.

I think those are clear and specific cases that can happen, and they have a known reason. We should only use `None` for the absence of value, but here it's representing information. We also loose information on which case happened, as they both return `None` (it isn't clear to me if the second case is also caused by a reorg).

In my opinion it would be better for future debugging and maintainability to create specific types that represent the information that those failures happened. They could be as simple as an empty NamedTuple/dataclass (these are known as zero-sized types). Then the return type would be something like `_HeightInfo | ReorgFailure | SomeOtherError`. That value can be easily matched against later, if needed.

Also, maybe add some logs?

hathor/p2p/sync_v2/agent.py
glevco, line 1: Do we have tests that target this now known reorg case?

glevco (changes requested):

hathor/p2p/sync_v2/agent.py
msbrogli: Because `lo` has the lowest height and `block_info_list` is sorted by decreasing height. So, if it belongs to the list, it will be at the last position.

jansegre: I really like typed errors on the return values. I'd prefer to introduce them using some ergonomic lib though, to make use of match. To me it's enough if we log the "reason" even if it still returns `None`.

jansegre (approved):

hathor/p2p/sync_v2/agent.py
msbrogli: Fixed! Thanks!

hathor/p2p/sync_v2/agent.py
msbrogli: No, we don't. We would have to come up with a way to pause the simulator in a specific moment.

hathor/p2p/sync_v2/agent.py
msbrogli: I added log messages for these cases. Another possibility is to raise custom exceptions instead of returning None. What do you think?

jansegre: Because the use of this method is very local, I think an exception would just make it more annoying to use. But also it's not too bad.

jansegre (dismissed):

hathor/p2p/sync_v2/agent.py
glevco: Thanks!

hathor/p2p/sync_v2/agent.py
glevco: Ok, thanks!

hathor/p2p/sync_v2/agent.py
glevco: Couldn't we just handcraft protocol messages, pretending it's a peer, or something like that? Not sure how doable it is, though. But maybe something that tests the unit, not necessarily a complete simulation

hathor/p2p/sync_v2/agent.py
glevco: > I really like typed errors on the return values. I'd prefer to introduce them using some ergonomic lib though, to make use of match.

@jansegre In this case it's not necessarily an error, but a possible state that is considered invalid. We don't really need a lib for some simple use cases, for example:

```python
@dataclass
class NoDataState:
    pass

@dataclass
class SomeIntState:
    a: int

@dataclass
class SomeStrState:
    b: str

MyState: TypeAlias = NoDataState | SomeIntState | SomeStrState

def handle_state(state: MyState) -> str:
    match state:
        case NoDataState(): return "no data in this state."
        case SomeIntState(some_int): return f"state with number: {some_int}"
        case SomeStrState(some_str): return some_str
        case _: assert_never(state)
```

I think we could use something analogous in this and other cases. What do you think?

Specifically for handling errors, maybe we could use the [Result](https://github.com/rustedpy/result) lib, that we've talked about before, now that we have `match` available. I also don't think exceptions would be very ergonomic.

Anyway, @msbrogli I'm fine leaving it like this, the logging is enough for now. But I suggest we keep this pattern in mind

msbrogli: I did the coding to use typed errors with @glevco and here are the results:

1. It seems too much for this specific case.
2. We had a weird `mypy` error when attributing a variable inside the match.
3. The code seems to get more complex than it currently is.

hathor/p2p/sync_v2/agent.py
msbrogli: I just added a test as we discussed. It does not simulate the reorg itself but it tests if the `find_common_best_block()` returns `None` if specific conditions arise. Can you review it again, please?

hathor/p2p/sync_v2/agent.py
jansegre: I'm OK with the current implementation.

jansegre (approved):

glevco (approved):

hathor/p2p/sync_v2/agent.py
glevco: Great, thanks!

PR #845, opened 2023-10-31
feat(sync-v2): Wait for sync internal methods to finish before initiating next syncing cycle

msbrogli:
### Motivation

Previously, the `_run_sync()` method checked for ongoing mempool syncs or open streamings to prevent concurrent cycle runs. Despite the presence of the `_is_running: bool` flag, it was prematurely marked as `false` due to mempool syncs and streamings returning at the start, not the end, of their execution. This led to complexity and potential errors in `_run_sync()`. This PR streamlines this process for better clarity and reliability.

### Acceptance Criteria

1. Create the deferred `_deferred_blockchain_streaming` when we request the peer to open a block streaming.
2. Resolve the `_deferred_blockchain_streaming` when the streaming ends.
3. Create the deferred `_deferred_transactions_streaming` when we request the peer to open a transactions streaming.
4. Resolve the `_deferred_transactions_streaming` when the streaming ends.
5. Change `run_sync_transactions()` to wait for its streaming to end.
6. Change `run_sync_blocks()`to wait for its streaming to end.
7. Change `mempool_manager.run()` to return a deferred that is resolved when the execution is complete.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/agent.py
+++ b/hathor/p2p/sync_v2/agent.py
@@ -117,6 +117,10 @@ class NodeBlockSync(SyncAgent):
         self._deferred_best_block: Optional[Deferred[_HeightInfo]] = None
         self._deferred_peer_block_hashes: Optional[Deferred[list[_HeightInfo]]] = None
 
+        # Deferreds used when we are receiving a streaming of vertices.
+        self._deferred_blockchain_streaming: Optional[Deferred[None]] = None
+        self._deferred_transactions_streaming: Optional[Deferred[None]] = None
+
         # When syncing blocks we start streaming with all peers
         # so the moment I get some repeated blocks, I stop the download
         # because it's probably a streaming that I've just received
@@ -289,16 +293,8 @@ class NodeBlockSync(SyncAgent):
     def _run_sync(self) -> Generator[Any, Any, None]:
         """ Actual implementation of the sync step logic in run_sync.
         """
-        if self.receiving_stream:
-            # If we're receiving a stream, wait for it to finish before running sync.
-            # If we're sending a stream, do the sync to update the peer's synced block
-            self.log.debug('receiving stream, try again later')
-            return
-
-        if self.mempool_manager.is_running():
-            # It's running a mempool sync, so we wait until it finishes
-            self.log.debug('running mempool sync, try again later')
-            return
+        assert not self.receiving_stream
+        assert not self.mempool_manager.is_running()
 
         assert self.protocol.connections is not None
         assert self.tx_storage.indexes is not None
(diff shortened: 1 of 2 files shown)


hathor/p2p/sync_v2/agent.py
```diff
@@ -308,16 +304,17 @@ def _run_sync(self) -> Generator[Any, Any, None]:
             self.log.debug('needed tx exist, sync transactions')
             self.update_synced(False)
             # TODO: find out whether we can sync transactions from this peer to speed things up
-            self.run_sync_transactions()
+            yield self.run_sync_transactions()
```
glevco, line 307: **question:** don't we have to add `@inlineCallbacks` to this method?

hathor/p2p/sync_v2/mempool.py
```diff
@@ -52,21 +54,28 @@ def is_running(self) -> bool:
         """Whether the sync-mempool is currently running."""
         return self._is_running
 
-    def run(self) -> None:
+    def run(self) -> Deferred[None]:
         """Starts _run in, won't start again if already running."""
         if self.is_running():
             self.log.warn('already started')
-            return
+            return succeed(None)
```
glevco, line 61: **question:** shouldn't we return `self.deferred` here?

I got that the deferred is only resolved when the execution is complete, but here it's still running.

msbrogli: I agree that returning the current deferred is better than creating a new one that is resolved. I just made the change. Thanks!

hathor/p2p/sync_v2/agent.py
msbrogli: It already does, doesn't it?

glevco: 🤔 I probably misread it, nevermind. Thanks!

glevco (dismissed):

glevco (dismissed):

hathor/p2p/sync_v2/mempool.py
```diff
@@ -39,6 +39,8 @@ def __init__(self, sync_agent: 'NodeBlockSync'):
         self.tx_storage = self.manager.tx_storage
         self.reactor = self.sync_agent.reactor
 
+        self.deferred: Optional[Deferred[None]] = None
```
jansegre, line 42: Should it be `self._deferred` since it's not meant to be used outside of the class?

hathor/p2p/sync_v2/mempool.py
```diff
@@ -52,21 +54,29 @@ def is_running(self) -> bool:
         """Whether the sync-mempool is currently running."""
         return self._is_running
 
-    def run(self) -> None:
+    def run(self) -> Deferred[None]:
         """Starts _run in, won't start again if already running."""
```
jansegre, line 58: The doctstring should be updated with the purpose of the deferred. From what I understood it's purely so we know that mempool sync finished successfully, but no further information is given, and the deferred will never complete unless `self.missing_tips` is zeroed. Is that correct?

jansegre (dismissed):

hathor/p2p/sync_v2/mempool.py
msbrogli: Indeed. However, it's important to note that by the time we reach this stage, our node will have activated real-time relay. Consequently, even during mempool synchronization, the other peer will continue to relay new blocks and transactions, ensuring we remain updated.

hathor/p2p/sync_v2/mempool.py
msbrogli: Done!

jansegre (approved):

glevco (approved):

PR #846, opened 2023-11-03
chore(custom_checks): Skip binary files when checking for deprecated typing

msbrogli:
### Motivation

If a file is open in vim, this skip wrongfully fails because the `.swp` file concatenate lines using a custom separator so the whole file is considered as in one line.

### Acceptance Criteria

1. Use grep `-I` option to skip binary files.

### Checklist


jansegre (approved):
Makes sense. I set up vim to store temp files in a dedicated dir, so I never experienced this.

glevco (approved):

PR #847, opened 2023-11-03
feat(simulator): Add StopWhenSendLineMatch trigger

msbrogli:
### Motivation

This trigger gives control to stop the simulator when a given message is sent from one node to another.

### Acceptance Criteria

1. Add `StopWhenSendLineMatch` trigger which monitors a connection and stops the simulator when a message matches a regex pattern.

### Checklist

### DIFF
--- a/hathor/simulator/trigger.py
+++ b/hathor/simulator/trigger.py
@@ -16,6 +16,9 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Callable
 
 if TYPE_CHECKING:
+    from re import Match, Pattern
+
+    from hathor.p2p.protocol import HathorLineReceiver
     from hathor.simulator.fake_connection import FakeConnection
     from hathor.simulator.miner import AbstractMiner
     from hathor.simulator.tx_generator import RandomTransactionGenerator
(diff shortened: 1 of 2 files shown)


hathor/simulator/trigger.py
```diff
@@ -107,3 +110,32 @@ def __init__(self, sub_triggers: list[Trigger]) -> None:
...
+
+
+class StopWhenSendLineMatch(Trigger):
+    """Stop the simulation when the node sends a line that matches a designated regex pattern.
+    """
+
+    def __init__(self, protocol: 'HathorLineReceiver', regex: 'Pattern') -> None:
+        # patches protocol.sendLine
+        self.original_send_line = protocol.sendLine
+        protocol.sendLine = self._send_line_wrapper  # type: ignore[method-assign]
```
msbrogli, line 122: mypy fails here:

```
mypy -p hathor -p tests
hathor/simulator/trigger.py: note: In member "__init__" of class "StopWhenSendLineMatch":
hathor/simulator/trigger.py:122: error: Cannot assign to a method  [method-assign]
            protocol.sendLine = self._send_line_wrapper
            ^~~~~~~~~~~~~~~~~
```

Any idea on how to do it differently?

jansegre (approved):

hathor/simulator/trigger.py
glevco: Maybe this patching could be done through subclassing instead? Especially since we're moving away from simulator patches. Maybe you could use the existing builder scaffolding to do something like that, similar to how we did for the vertex verifiers. You would create a `SimulatorHathorLineReceiver` that would integrate with this trigger, instead of patching it manually.

If you deem that the current structure wouldn't support this, I'm ok with leaving the `ignore` comment.

jansegre: An alternative is using `setattr` (see https://github.com/python/mypy/issues/2427#issuecomment-480263443). I'm OK either way.

msbrogli: I used `setattr()`. Thanks for your help!

jansegre (approved):

glevco (approved):

PR #848, opened 2023-11-03
feat(sync-v2): Add both BlockchainStreamingClient and TransactionStreamingClient to manage streamings from the client side

msbrogli:
### Motivation

Even though streaming servers were implemented on separated classes, the clients were intertwined within the sync agent.

### Acceptance Criteria

1. Add `BlockchainStreamingClient` to handle blockchain streamings.
2. Add `TransactionStreamingClient` to handle transaction streamings.
3. Rename `until_first_block` to `last_block_hash` in GET-TRANSACTION-BFS message.
4. Add `first_block_hash` parameter to GET-TRANSACTION-BFS message.
5. Besides the changes in the GET-TRANSACTION-BFS, the changes in this PR can be considered a refactor and no other behavior should be changed.
6. Add pydantic models to GET-NEXT-BLOCKS, BEST-BLOCK and GET-TRANSACTIONS-BFS payloads.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/agent.py
+++ b/hathor/p2p/sync_v2/agent.py
@@ -27,17 +27,26 @@ from twisted.internet.task import LoopingCall
 from hathor.conf.get_settings import get_settings
 from hathor.p2p.messages import ProtocolMessages
 from hathor.p2p.sync_agent import SyncAgent
+from hathor.p2p.sync_v2.blockchain_streaming_client import BlockchainStreamingClient, StreamingError
 from hathor.p2p.sync_v2.mempool import SyncMempoolManager
-from hathor.p2p.sync_v2.streamers import DEFAULT_STREAMING_LIMIT, BlockchainStreaming, StreamEnd, TransactionsStreaming
+from hathor.p2p.sync_v2.payloads import BestBlockPayload, GetNextBlocksPayload, GetTransactionsBFSPayload
+from hathor.p2p.sync_v2.streamers import (
+    DEFAULT_STREAMING_LIMIT,
+    BlockchainStreamingServer,
+    StreamEnd,
+    TransactionsStreamingServer,
+)
+from hathor.p2p.sync_v2.transaction_streaming_client import TransactionStreamingClient
 from hathor.transaction import BaseTransaction, Block, Transaction
 from hathor.transaction.base_transaction import tx_or_block_from_bytes
 from hathor.transaction.exceptions import HathorError
 from hathor.transaction.storage.exceptions import TransactionDoesNotExist
 from hathor.types import VertexId
-from hathor.util import Reactor, collect_n
+from hathor.util import Reactor, collect_n, not_none
 
 if TYPE_CHECKING:
     from hathor.p2p.protocol import HathorProtocol
+    from hathor.transaction.storage import TransactionStorage
 
 logger = get_logger()
 
@@ -82,7 +91,7 @@ class NodeBlockSync(SyncAgent):
         self._settings = get_settings()
         self.protocol = protocol
         self.manager = protocol.node
-        self.tx_storage = protocol.node.tx_storage
+        self.tx_storage: 'TransactionStorage' = protocol.node.tx_storage
         self.state = PeerState.UNKNOWN
 
         self.DEFAULT_STREAMING_LIMIT = DEFAULT_STREAMING_LIMIT
@@ -96,11 +105,6 @@ class NodeBlockSync(SyncAgent):
 
         # Create logger with context
         self.log = logger.new(peer=self.protocol.get_short_peer_id())
-
-        # Extra
-        self._blk_size = 0
-        self._blk_end_hash = self._settings.GENESIS_BLOCK_HASH
-        self._blk_max_quantity = 0
 
         # indicates whether we're receiving a stream from the peer
         self.receiving_stream = False
@@ -117,18 +121,13 @@ class NodeBlockSync(SyncAgent):
         self._deferred_best_block: Optional[Deferred[_HeightInfo]] = None
         self._deferred_peer_block_hashes: Optional[Deferred[list[_HeightInfo]]] = None
 
-        # Deferreds used when we are receiving a streaming of vertices.
-        self._deferred_blockchain_streaming: Optional[Deferred[None]] = None
-        self._deferred_transactions_streaming: Optional[Deferred[None]] = None
+        # Clients to handle streaming messages.
+        self._blk_streaming_client: Optional[BlockchainStreamingClient] = None
+        self._tx_streaming_client: Optional[TransactionStreamingClient] = None
 
-        # When syncing blocks we start streaming with all peers
-        # so the moment I get some repeated blocks, I stop the download
-        # because it's probably a streaming that I've just received
-        self.max_repeated_blocks = 10
-
-        # Streaming objects
-        self.blockchain_streaming: Optional[BlockchainStreaming] = None
-        self.transactions_streaming: Optional[TransactionsStreaming] = None
+        # Streaming server objects
+        self._blk_streaming_server: Optional[BlockchainStreamingServer] = None
+        self._tx_streaming_server: Optional[TransactionsStreamingServer] = None
 
         # Whether the peers are synced, i.e. we have the same best block.
         # Notice that this flag ignores the mempool.
@@ -136,11 +135,6 @@ class NodeBlockSync(SyncAgent):
 
         # Indicate whether the sync manager has been started.
         self._started: bool = False
-
-        # Saves the last received block from the block streaming # this is useful to be used when running the sync of
-        # transactions in the case when I am downloading a side chain. Starts at the genesis, which is common to all
-        # peers on the network
-        self._last_received_block: Optional[Block] = None
 
         # Saves if I am in the middle of a mempool sync
         # we don't execute any sync while in the middle of it
@@ -158,9 +152,6 @@ class NodeBlockSync(SyncAgent):
 
         # Whether we propagate transactions or not
         self._is_relaying = False
-
-        # This stores the final height that we expect the last "get blocks" stream to end on
-        self._blk_end_height: Optional[int] = None
 
         # Whether to sync with this peer
         self._is_enabled: bool = False
@@ -328,7 +319,7 @@ class NodeBlockSync(SyncAgent):
                                   MAX_GET_TRANSACTIONS_BFS_LEN)
 
         # Start with the last received block and find the best block full validated in its chain
-        block = self._last_received_block
+        block = self._blk_streaming_client._last_received_block if self._blk_streaming_client else None
         if block is None:
             block = cast(Block, self.tx_storage.get_genesis(self._settings.GENESIS_BLOCK_HASH))
         else:
@@ -341,7 +332,10 @@ class NodeBlockSync(SyncAgent):
 
         self.log.info('run sync transactions', start=[i.hex() for i in needed_txs], end_block_hash=block.hash.hex(),
                       end_block_height=block_height)
-        yield self.start_transactions_streaming(needed_txs, block.hash)
+        try:
+            yield self.start_transactions_streaming(needed_txs, block.hash, block.hash)
+        except StreamingError:
+            self.receiving_stream = False
 
     def get_my_best_block(self) -> _HeightInfo:
         """Return my best block info."""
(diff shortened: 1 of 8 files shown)


hathor/p2p/sync_v2/agent.py
```diff
@@ -407,10 +392,12 @@ def run_sync_blocks(self) -> Generator[Any, Any, bool]:
...
         # Sync from common block
-        yield self.start_blockchain_streaming(self.synced_block.id,
-                                              self.synced_block.height,
-                                              self.peer_best_block.id,
-                                              self.peer_best_block.height)
+        try:
+            yield self.start_blockchain_streaming(self.synced_block,
+                                                  self.peer_best_block)
+        except StreamingError:
+            self.send_stop_block_streaming()
```
msbrogli, line 399: Should we do anything else here? Maybe set state to ERROR? Or just close connection?

hathor/p2p/sync_v2/agent.py
```diff
@@ -341,7 +326,7 @@ def run_sync_transactions(self) -> Generator[Any, Any, None]:
 
         self.log.info('run sync transactions', start=[i.hex() for i in needed_txs], end_block_hash=block.hash.hex(),
                       end_block_height=block_height)
-        yield self.start_transactions_streaming(needed_txs, block.hash)
+        yield self.start_transactions_streaming(needed_txs, block.hash, block.hash)
```
msbrogli, line 329: What should we do if the deferred fails? Maybe set state to ERROR? Or just close connection?

hathor/p2p/sync_v2/payloads.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class PayloadBaseModel(BaseModel):
+
+    @classmethod
+    def convert_hex_to_bytes(cls, value: str | bytes) -> bytes:
+        if isinstance(value, str):
+            return bytes.fromhex(value)
+        elif isinstance(value, bytes):
+            return value
```
jansegre, line 27: This behavior is a bit surprising to me. Looking only at the signature I would have assumed the that it supports encoded/unencoded characters, as in if bytes is given it would still try to hex decode, but assume ascii (or utf-8, shouldn't matter for hex chars).

Is there any practical application for not doing anything if bytes are given? Otherwise I'd rather have it more straightforward and only accept str.

hathor/p2p/sync_v2/payloads.py
```diff
@@ -0,0 +1,66 @@
...
+
+from pydantic import validator
+
+from hathor.utils.pydantic import BaseModel
+
+
+class PayloadBaseModel(BaseModel):
+
+    @classmethod
+    def convert_hex_to_bytes(cls, value: str | bytes) -> bytes:
```
jansegre, line 23: Missing docstring.

hathor/p2p/sync_v2/payloads.py
```diff
@@ -0,0 +1,66 @@
...
+        raise ValueError('invalid type')
+
+    class Config:
+        json_encoders = {
+            bytes: lambda x: x.hex()
+        }
+
+
+class GetNextBlocksPayload(PayloadBaseModel):
+    start_hash: bytes
```
jansegre, line 37: Should we use VertexId for all of these `bytes` fields?

jansegre (changes requested):

hathor/p2p/sync_v2/blockchain_streaming_client.py
```diff
@@ -0,0 +1,140 @@
...
+
+        self._last_received_block: Optional[Block] = None
+
+        self._partial_blocks: list[Block] = []
+
+    def run(self) -> Deferred[StreamEnd]:
+        return self._deferred
+
+    def fails(self, reason: 'StreamingError') -> None:
+        self.sync_agent.send_stop_block_streaming()
```
jansegre, line 74: Should this method be controlled by the sync agent? As in letting the agent register it as an errback for the deferred of the streaming client?

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -0,0 +1,118 @@
...
+        self._tx_received: int = 0
+        self._tx_repeated: int = 0
+
+        self._tx_max_quantity = DEFAULT_STREAMING_LIMIT
+
+    def run(self) -> Deferred[StreamEnd]:
+        return self._deferred
+
+    def fails(self, reason: 'StreamingError') -> None:
+        self.sync_agent.send_stop_block_streaming()
```
jansegre, line 68: Should this method be controlled by the sync agent? As in letting the agent register it as an errback for the deferred of the streaming client?

Shouldn't this be `send_stop_transaction_streaming()`

jansegre (approved):

hathor/p2p/sync_v2/agent.py
msbrogli: I improved it in PR #850 .

hathor/p2p/sync_v2/agent.py
msbrogli: I agree. This code was completely removed in PR #850. So I think we should discuss the final implementation there. Is it ok for you?

hathor/p2p/sync_v2/agent.py
```diff
@@ -82,7 +86,7 @@ def __init__(self, protocol: 'HathorProtocol', reactor: Optional[Reactor] = None
         self._settings = get_settings()
         self.protocol = protocol
         self.manager = protocol.node
-        self.tx_storage = protocol.node.tx_storage
+        self.tx_storage: 'TransactionStorage' = protocol.node.tx_storage
```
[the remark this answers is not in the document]
msbrogli: mypy wasn't able to infer the type and was complaining about it. As we do not use anything specific of subtypes, I just used the base storage class here.

hathor/p2p/sync_v2/payloads.py
msbrogli: Yes, there's a practical application. The current implementation is very flexible and supports all the following object creation. Notice that the validator is always executed, no matter how we create the object.

```
> GetNextBlocksPayload(start_hash=b1.hash, end_hash=b2.hash)
> GetNextBlocksPayload(start_hash=b1.hash_hex, end_hash=b2.hash_hex)
> GetNextBlocksPayload.parse_raw(json.dumps({'start_hash': b1.hash_hex, 'end_hash': b2.hash_hex}))
```

hathor/p2p/sync_v2/blockchain_streaming_client.py
```diff
@@ -0,0 +1,140 @@
...
+
+        assert blk.hash is not None
+        is_duplicated = False
+        if self.partial_vertex_exists(blk.hash):
+            # We reached a block we already have. Skip it.
+            self._blk_repeated += 1
+            is_duplicated = True
+            if self._blk_repeated > self.max_repeated_blocks:
+                self.log.debug('too many repeated block received', total_repeated=self._blk_repeated)
+                self.fails(TooManyRepeatedVerticesError())
```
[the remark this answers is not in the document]
msbrogli: The previous implementation was not returning, so I decided to keep the same behavior. It was changed in PR #850, though.

hathor/p2p/sync_v2/blockchain_streaming_client.py
msbrogli: Actually, this method was being called twice. Fixed! https://github.com/HathorNetwork/hathor-core/pull/848/commits/c044befd04c533c13894a8d39258287fa5549e4d

hathor/p2p/sync_v2/payloads.py
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/848/commits/c044befd04c533c13894a8d39258287fa5549e4d

hathor/p2p/sync_v2/payloads.py
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/848/commits/c044befd04c533c13894a8d39258287fa5549e4d

hathor/p2p/sync_v2/blockchain_streaming_client.py
```diff
@@ -0,0 +1,140 @@
...
+        if self._blk_max_quantity < 0:
+            self._blk_max_quantity = -self._blk_max_quantity
+            self._reverse = True
+
+        self._last_received_block: Optional[Block] = None
+
+        self._partial_blocks: list[Block] = []
+
+    def run(self) -> Deferred[StreamEnd]:
+        return self._deferred
```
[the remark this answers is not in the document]
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/848/commits/bbdc1955af7cb57ba29375c2489c8219b24abbde

hathor/p2p/sync_v2/agent.py
msbrogli: I added a try/except block and just marked streaming as false. Check it out here: https://github.com/HathorNetwork/hathor-core/pull/848/commits/c044befd04c533c13894a8d39258287fa5549e4d

hathor/p2p/sync_v2/payloads.py
jansegre: If we already do this in other cases I think it's fine to keep the pattern, otherwise this flexibility feels error prone to me. Specially if we're not making use of the flexibility.

msbrogli: We use it. To send information, I use `GetNextBlocksPayload(start_hash=b1.hash, end_hash=b2.hash)` to generate a payload and, at the receiver, I use `GetNextBlocksPayload.parse_raw(payload)` to parse the payload.

jansegre: Got it. Seems a bit weird to me that `GetNextBlocksPayload(start_hash=b1.hash, end_hash=b2.hash)` would call effectively deserialization methods, I'm not too familiar with pydantic.

jansegre (dismissed):

hathor/p2p/sync_v2/payloads.py
```diff
@@ -0,0 +1,68 @@
...
+    first_block_hash: VertexId
+    last_block_hash: VertexId
+
+    @validator('first_block_hash', 'last_block_hash', pre=True)
+    def validate_bytes_fields(cls, value: str | VertexId) -> VertexId:
+        return cls.convert_hex_to_bytes(value)
+
+    @validator('start_from', pre=True)
+    def validate_start_from(cls, values: list[str | VertexId]) -> list[VertexId]:
+        return [cls.convert_hex_to_bytes(x) for x in values]
```
glevco, line 68: You can use `each_item=True` in the decorator so you don't have to do the loop. Also, shouldn't the `start_from` type be `list[VertexId]`?

glevco (dismissed):

hathor/p2p/sync_v2/payloads.py
glevco, line 1: Docstrings for the payloads?

hathor/p2p/sync_v2/exception.py
```diff
@@ -0,0 +1,32 @@
...
+class TooManyRepeatedVerticesError(StreamingError):
+    pass
+
+
+class BlockNotConnectedToPreviousBlock(StreamingError):
+    pass
+
+
+class InvalidVertexError(StreamingError):
+    pass
```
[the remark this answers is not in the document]
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/848/commits/8cb2fd15779d219c5cefeed42a372c46198e24e9

hathor/p2p/sync_v2/payloads.py
msbrogli: Thanks for the tip. Done! https://github.com/HathorNetwork/hathor-core/pull/848/commits/8cb2fd15779d219c5cefeed42a372c46198e24e9

hathor/p2p/sync_v2/payloads.py
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/848/commits/6dd0b5931d39609da58cd06f368ce980d62fd502

jansegre (approved):

hathor/p2p/sync_v2/exception.py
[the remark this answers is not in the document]
glevco: Thanks!

hathor/p2p/sync_v2/payloads.py
glevco: Thanks!

hathor/p2p/sync_v2/payloads.py
glevco: Thanks!

glevco (approved):

PR #850, opened 2023-11-03
feat(sync-v2): Implement new protocol for synchronizing transactions

msbrogli:
### Motivation

This PR addresses two significant issues in the current implementation:

1. In Sync-v2, a global dependencies index was utilized across all connections.
2. Sync-v2 agents insisted on the fulfillment of all dependencies before progressing with the synchronization.

These combined challenges led to potential deadlocks. Specifically, if one peer introduced a dependency that was unavailable from others, it could cause all agents to endlessly attempt to satisfy this elusive dependency from their peers—a futile endeavor.

To counter this, we've established a new transaction synchronization protocol from blocks. With this enhanced approach, every agent manages its own unique dependency list, allowing for independent synchronization with each peer. Furthermore, it allows the synchronization loop to function even with outstanding dependencies.

This PR changes the behavior of the `GET-TRANSACTIONS-BFS` message. The new parameters are `start_from`, `first_block_hash`, and `end_block_hash`. The response is a sequence of `TRANSACTION` messages, each containing one transaction.

#### New protocol

The new protocol syncs blocks first and keeps track of `partial_blocks` which are the blocks that cannot be fully validated because of a missing dependency. Then, the `partial_blocks` list is passed to the transaction streamer. The streamer will then send all transactions confirmed by the first block followed by all transactions confirmed by the second block and so on. To determine the precise order of transactions within each block, the streamer implements a Breadth-First Search (BFS) algorithm, tracing back from the block in question.

To illustrate the ordering, consider `partial_blocks` enumerated as `[b0, b1, b2, ..., bn]`, where `tx_i_j` represents the i-th transaction within the j-th block (`b_j`). The transaction streamer would then relay transactions in the ensuing sequence: `[tx_0_0, tx_1_0, tx_2_0, ... tx_N0_0, tx_0_1, tx_1_1, tx_2_1, ..., tx_N1_1, ..., tx_0_n, tx_1_n, tx_2_n, ..., tx_Nn_n]`. Here, `N0, N1, ..., Nn` correspond to the total number of transactions validated by blocks `b0, b1, ..., bn` respectively.

### Acceptance Criteria

1. Remove `DepsIndex`, `MemoryDepsIndex`, `RocksDBDepsIndex` because this index is not used anymore.
2. Remove `HathorManager._sync_v2_resume_validations()` and `HathorManager.sync_v2_step_validations()` because they are not used anymore.
3. Add `STOP_TRANSACTIONS_STREAMING` message to allow clients to request servers to stop streaming.
4. Remove `sync_v2.NodeBlockSync.run_sync_transactions()` because it is not used anymore.
5. Stop using partially validated vertex. All vertices added to the DAG are immediately fully validated.
6. Modify `TransactionsStreamingServer` to send transactions according to the new protocol.
7. Change fixed seed in `tests/consensus/test_soft_voided.py` because this test was failing.

### Checklist


hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -87,34 +92,91 @@ def handle_transaction(self, tx: BaseTransaction) -> None:
...
         if self._deferred.called:
             return
         self._deferred.callback(response_code)
+
+    def _execute_and_prepare_next(self) -> None:
+        self._buffer.sort(key=lambda vertex_id: self._db[vertex_id].timestamp)
+        for vertex_id in self._buffer:
+            tx = self._db[vertex_id]
+            try:
+                self.manager.on_new_tx(tx, propagate_to_peers=False, fails_silently=False)
```
msbrogli, line 154: It cannot verify reward lock.

hathor/p2p/sync_v2/streamers.py
```diff
@@ -191,31 +194,42 @@ class TransactionsStreamingServer(_StreamingServerBase):
     """
 
     def __init__(self,
-                 node_sync: 'NodeBlockSync',
+                 sync_agent: 'NodeBlockSync',
                  start_from: list[BaseTransaction],
```
msbrogli, line 198: `start_from` is not implemented yet

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -87,34 +92,91 @@ def handle_transaction(self, tx: BaseTransaction) -> None:
...
+                continue
+            yield dep
 
     def handle_transactions_end(self, response_code: StreamEnd) -> None:
         """This method is called by the sync agent when a TRANSACTIONS-END message is received."""
         if self._deferred.called:
             return
         self._deferred.callback(response_code)
+
+    def _execute_and_prepare_next(self) -> None:
```
msbrogli, line 149: Move it to the sync agent.

msbrogli: Done!

hathor/p2p/sync_v2/streamers.py
```diff
@@ -191,31 +194,42 @@ class TransactionsStreaming(_StreamingBase):
...
+        self.current_block: Optional[Block] = self.first_block
 
-        self.bfs = BFSOrderWalk(self.storage, is_dag_verifications=True, is_dag_funds=True, is_left_to_right=False)
-        self.iter = self.bfs.run(start_from, skip_root=False)
+        # TODO Validated start_from and use it when it's not empty.
+
+        self.bfs = BFSOrderWalk(self.tx_storage, is_dag_verifications=True, is_dag_funds=True, is_left_to_right=False)
+        self.iter = self.get_next_tx()
+
+    def get_next_tx(self) -> Iterator[BaseTransaction]:
```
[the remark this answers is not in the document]
msbrogli: Renamed to `get_iter()`. What do you think?

hathor/p2p/sync_v2/blockchain_streaming_client.py
```diff
@@ -93,6 +93,15 @@ def handle_blocks(self, blk: Block) -> None:
             self.fails(TooManyVerticesReceivedError())
             return
 
+        # Run basic verification.
+        # if not blk.is_genesis:
+        #     try:
+        #         self.manager.verification_service.validate_basic(blk)
+        #     except TxValidationError as e:
+        #         self.fails(InvalidVertexError(repr(e)))
+        #         return
```
jansegre, line 102: This is a TODO, right? Is it expected to be in the final version of this PR?

hathor/p2p/sync_v2/streamers.py
jansegre: It is now, right?

jansegre (dismissed):

hathor/p2p/sync_v2/streamers.py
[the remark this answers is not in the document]
jansegre: I think it's good, maybe it's not obvious what it does but it's definitely not misleading.

hathor/p2p/sync_v2/mempool.py
```diff
@@ -63,6 +63,8 @@ def run(self) -> Deferred[None]:
         self._is_running = True
         self.reactor.callLater(0, self._run)
 
+        # TODO Implement a stop() and call it after N minutes.
```
msbrogli, line 66: TODO

jansegre (dismissed):

hathor/p2p/sync_v2/streamers.py
msbrogli: Yes, it is.

hathor/p2p/sync_v2/streamers.py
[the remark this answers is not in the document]
msbrogli: Ok. I'm open for suggestions.

hathor/p2p/sync_v2/blockchain_streaming_client.py
msbrogli: I can't do it right now because I need to refactor the verification services. The issue here is that the chain of blocks is in memory (not in tx storage) so the current verifier will fail with a `TransactionDoesNotExist` error.

glevco: The refactor I'll do as we agreed [here](https://github.com/HathorNetwork/hathor-core/pull/831#discussion_r1381939630) will potentially solve this, as a list of dependencies will be passed to the verification service, and the storage won't be accessed during verification anymore. Then you can construct this dependency object from memory instead of the storage. I'll work on it today.

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -88,33 +94,85 @@ def handle_transaction(self, tx: BaseTransaction) -> None:
...
-            if is_duplicated:
-                self.log.debug('tx early terminate?', tx_id=tx.hash.hex())
-            else:
-                self.log.debug('tx received', tx_id=tx.hash.hex())
-            self.sync_agent.on_new_tx(tx, propagate_to_peers=False, quiet=True, reject_locked_reward=True)
-        except HathorError:
-            self.fails(InvalidVertexError())
+        self.log.debug('tx received', tx_id=tx.hash.hex())
+
+        # Run basic verification.
```
glevco, line 100: Add a TODO here, too?

hathor/p2p/sync_v2/streamers.py
```diff
@@ -180,44 +184,67 @@ def send_next(self) -> None:
...
-        self.bfs = BFSOrderWalk(self.storage, is_dag_verifications=True, is_dag_funds=True, is_left_to_right=False)
-        self.iter = self.bfs.run(start_from, skip_root=False)
+        super().__init__(sync_agent, limit=limit)
+
+        self.first_block: Block = cast(Block, self.tx_storage.get_transaction(first_block_hash))
+        self.last_block: Block = cast(Block, self.tx_storage.get_transaction(last_block_hash))
+        self.start_from = start_from
+
+        assert isinstance(self.first_block, Block)
+        assert isinstance(self.last_block, Block)
```
glevco, line 217: If you assert before, then you don't need the cast above

hathor/p2p/sync_v2/streamers.py
```diff
@@ -180,44 +184,67 @@ def send_next(self) -> None:
...
+
+        # TODO Validated start_from and use it when it's not empty.
```
glevco, line 219: Just marking this TODO for bookkeeping

hathor/p2p/sync_v2/blockchain_streaming_client.py
msbrogli: I guess I'll just leave a TODO note and get back to it later. Is it for you? I'll add this change as a separate task in the project spreadsheet.

hathor/p2p/sync_v2/mempool.py
msbrogli: I'm planning to do it in a separate PR. I feel that this PR is already too big and this change is unrelated to the changes in this PR.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: I guess we should check `tx_storage.allow_partially_validated_context()` instead of `allow_only_valid_context()`. I found the following cases:

```
$ grep -RIi allow_partially_validated_context hathor
hathor/transaction/storage/transaction_storage.py:    def allow_partially_validated_context(self) -> AbstractContextManager[None]:
hathor/p2p/sync_v2/blockchain_streaming_client.py:        with self.tx_storage.allow_partially_validated_context():
hathor/p2p/sync_v2/agent.py:        with self.tx_storage.allow_partially_validated_context():
```

I can easily remove all these `allow_partially_validated_context()` calls from this PR.

hathor/p2p/sync_v2/exception.py
```diff
@@ -35,3 +35,7 @@ class BlockNotConnectedToPreviousBlock(StreamingError):
 class InvalidVertexError(StreamingError):
     """Raised when the received vertex fails validation."""
     pass
+
+
+class UnexpectedVertex(StreamingError):
```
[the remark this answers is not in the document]
msbrogli: My bad. I'll do it.

hathor/p2p/sync_v2/streamers.py
msbrogli: I need to improve this validation. I'm not sure whether I should assert it here and actually validate it on the caller or just validate it here. The point is that these hashes are sent by the other peer so we need to validate them somewhere.

hathor/p2p/sync_v2/streamers.py
msbrogli: The "use it" is done already. I just need to validate it properly.

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: We can test if it fully syncs. As far as we know there isn't a practical case that would break it. But _not_ skipping the reward lock here could cause sync issues if a breaking case appears.

hathor/p2p/sync_v2/streamers.py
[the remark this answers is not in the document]
jansegre: Maybe `iter_next_txs`, to me it reads as "iterate over the next transactions". I've been using `iter_` prefix for methods that return an iterator. But either way I don't think `get_iter` is necessarily bad.

hathor/p2p/sync_v2/streamers.py
jansegre: I see. The cast will only trick mypy, but the assert is the actual runtime validation. If it helps you could use `if not self.first_block.is_block: # close connection or something`.

hathor/p2p/sync_v2/blockchain_streaming_client.py
glevco: Ok for me.

hathor/manager.py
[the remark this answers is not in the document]
msbrogli: We might be able to completely remove it but I guess we should finish the implementation before doing it. We still have to implement the sync of checkpoints which is supposed to considerably speed up the sync.

hathor/p2p/sync_v2/streamers.py
msbrogli: I want `self.first_block` and `self.last_block` to have type `Block`. But then the `self.tx_storage.get_transaction()` returns a `BaseTransaction` which causes mypy to fail. So I used a cast followed by an assert to make sure we have the right type (since cast does not actually check it).


I aim for `self.first_block` and `self.last_block` to specifically be typed as `Block`. However, the method `self.tx_storage.get_transaction()` returns a BaseTransaction, which causes mypy to fail. So, I initially used a cast to hint to mypy that the returned object should be treated as a `Block`. But since cast alone doesn't guarantee the object's type at runtime, I also added asserts to make sure that the objects are indeed instances of `Block`.

hathor/p2p/sync_v2/streamers.py
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/850/commits/93c2fb9b04d747d00a1524e8a2c59016c6a728b3

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/850/commits/93c2fb9b04d747d00a1524e8a2c59016c6a728b3

hathor/p2p/sync_v2/exception.py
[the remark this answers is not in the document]
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/850/commits/93c2fb9b04d747d00a1524e8a2c59016c6a728b3

hathor/p2p/sync_v2/blockchain_streaming_client.py
msbrogli: Done! https://github.com/HathorNetwork/hathor-core/pull/850/commits/93c2fb9b04d747d00a1524e8a2c59016c6a728b3

hathor/p2p/sync_v2/streamers.py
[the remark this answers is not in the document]
msbrogli: Got it. I'd left as `get_iter` if that's ok for you.

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: I agree. But I think we can do it in a separate PR because this one got too big already.

hathor/p2p/sync_v2/streamers.py
msbrogli: I guess this verification should be made at the agent level. This assert just checks if the parameters were properly placed.

hathor/p2p/sync_v2/exception.py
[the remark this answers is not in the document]
glevco: Thanks!

hathor/manager.py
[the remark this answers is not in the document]
glevco: Ok!

hathor/p2p/sync_v2/transaction_streaming_client.py
glevco: Thanks!

hathor/p2p/sync_v2/streamers.py
glevco: Is this being done in this PR?

glevco (approved):

hathor/p2p/sync_v2/streamers.py
msbrogli: It was partially implemented. It's complete now. https://github.com/HathorNetwork/hathor-core/pull/850/commits/35f8222bf3917636e72762f9375b3db9bc06052f

msbrogli: I missed one validation. Here it is: https://github.com/HathorNetwork/hathor-core/pull/850/commits/00253b4b8482d698f028b2870348195763c368b9

glevco: Ok!

glevco (approved):

jansegre (dismissed):

jansegre (approved):

glevco (dismissed):

jansegre (approved):

glevco (approved):

hathor/p2p/sync_v2/blockchain_streaming_client.py
glevco: @msbrogli you'll be able to use in-memory transactions for verification using the refactor introduced by these 3 PRs:

- https://github.com/HathorNetwork/hathor-core/pull/831
- https://github.com/HathorNetwork/hathor-core/pull/858
- https://github.com/HathorNetwork/hathor-core/pull/859

PR #851, opened 2023-11-05
refactor(scripts): implement ScriptContext

glevco:
### Motivation

Create the `ScriptContext` scaffolding that may be used by future opcodes.

### Acceptance Criteria

- Create new, currently empty, `ScriptContext` class.
- Change function opcodes execution to use a new  `execute_op_code()`, that takes a `ScriptContext` as an argument.
- Remove unused arguments from all opcode functions.

### Checklist

### DIFF
--- a/hathor/transaction/scripts/execute.py
+++ b/hathor/transaction/scripts/execute.py
@@ -16,7 +16,7 @@ import struct
 from typing import NamedTuple, Optional, Union
 
 from hathor.transaction import BaseTransaction, Transaction, TxInput
-from hathor.transaction.exceptions import DataIndexError, FinalStackInvalid, InvalidScriptError, OutOfData, ScriptError
+from hathor.transaction.exceptions import DataIndexError, FinalStackInvalid, InvalidScriptError, OutOfData
 
 
 class ScriptExtras(NamedTuple):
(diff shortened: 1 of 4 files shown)


hathor/transaction/scripts/opcode.py
```diff
@@ -157,7 +157,7 @@ def op_pushdata1(position: int, full_data: bytes, stack: Stack) -> int:
     return new_pos
 
 
-def op_dup(stack: Stack, log: list[str], extras: ScriptExtras) -> None:
```
msbrogli, line 160: Why have you removed the log in some opcodes?

hathor/transaction/scripts/execute.py
```diff
@@ -54,21 +54,19 @@ def execute_eval(data: bytes, log: list[str], extras: ScriptExtras) -> None:
...
             continue
-        # this is an opcode manipulating the stack
-        fn = MAP_OPCODE_TO_FN.get(opcode, None)
-        if fn is None:
-            # throw error
-            raise ScriptError('unknown opcode')
 
-        fn(stack, log, extras)
+        # this is an opcode manipulating the stack
+        execute_op_code(Opcode(opcode), stack, log, extras, context)
```
msbrogli, line 69: Maybe log which opcode is being executed?

msbrogli (changes requested):

hathor/transaction/scripts/execute.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/851/commits/b8ac57b0f42361e825abdbdcf58fec6e079d70a9 (inside the `execute_op_code` function)

hathor/transaction/scripts/opcode.py
glevco: Because they were not being used. Most opcodes don't use all 3 parameters (`stack`, `log`, and `extras`), and since I'm going to add a fourth one (`context`), I made this change so each opcode can have a different signature, eliminating unused parameters.

msbrogli: Should we create a context with all of them inside it? I mean, `Context` would carry the stack, log, extras, and whatever other stuff we might need in the future. So we don't have to change the method signatures anymore.

hathor/transaction/scripts/execute.py
msbrogli: Thanks!

hathor/transaction/scripts/opcode.py
glevco: Sure, done in https://github.com/HathorNetwork/hathor-core/pull/851/commits/f35564bb1f1696c6fdb87d1066646d8a60b0a3c1

msbrogli: Great! Thanks!

msbrogli (approved):

jansegre (approved):
Approved, but it's not a feature, it's a refactor, right?

PR #859, opened 2023-11-09
refactor(verification): externalize verification dependencies [part 1/2]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/896

### Motivation

This PR introduces the concept of verification dependencies. Before, to perform verifications, it was necessary to retrieve data from the vertex itself and also from multiple other sources, for example from the `TransactionStorage`.

Now, instead of doing this dependency retrieval one by one, interspersed in the whole verification process, a single point retrieves all necessary dependencies and everything is verified from the vertex data plus those pre-calculated dependencies. This prevents unnecessary duplicate calculations, and will be necessary for the Multiprocess Verification project.

There are also some other side-benefits:

- This is one step forward in completely removing the `TransactionStorage` from inside vertices.
- Sync-v2 may verify vertices using dependencies from its memory instead of the `TransactionStorage` (as discussed [here](https://github.com/HathorNetwork/hathor-core/pull/850#discussion_r1385422268)).
- Improved typing and ergonomics of `VerificationService` methods.

Note: in this PR, not all ad-hoc dependencies have been moved yet. This will be completed in the next PR.

### Acceptance Criteria

- Create new verification dependency models that are pre-calculated before verification.
- Update verifiers to use dependency models instead of directly accessing storage.
- Add new auxiliary methods to `SimpleMemoryStorage`.

### Checklist

### DIFF
--- a/hathor/verification/verification_dependencies.py
+++ b/hathor/verification/verification_dependencies.py
@@ -1,0 +1,88 @@
+#  Copyright 2023 Hathor Labs
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from dataclasses import dataclass
+
+from typing_extensions import Self
+
+from hathor.daa import DifficultyAdjustmentAlgorithm
+from hathor.feature_activation.feature import Feature
+from hathor.feature_activation.feature_service import BlockSignalingState, FeatureService
+from hathor.feature_activation.model.feature_description import FeatureInfo
+from hathor.transaction import Block
+from hathor.transaction.storage.simple_memory_storage import SimpleMemoryStorage
+from hathor.transaction.transaction import Transaction
+
+
+@dataclass(frozen=True, slots=True)
+class VertexDependencies:
+    """A dataclass of dependencies necessary for vertex verification."""
+    storage: SimpleMemoryStorage
+
+
+@dataclass(frozen=True, slots=True)
+class BasicBlockDependencies(VertexDependencies):
+    """A dataclass of dependencies necessary for basic block verification."""
+
+    @classmethod
+    def create(cls, block: Block, daa: DifficultyAdjustmentAlgorithm, *, skip_weight_verification: bool) -> Self:
+        """Create a basic block dependencies instance."""
+        assert block.storage is not None
+        simple_storage = SimpleMemoryStorage()
+        daa_deps = [] if skip_weight_verification else daa.get_block_dependencies(block)
+        deps = block.parents + daa_deps
+
+        simple_storage.add_vertices_from_storage(block.storage, deps)
+
+        return cls(simple_storage)
+
+
+@dataclass(frozen=True, slots=True)
+class BlockDependencies(VertexDependencies):
+    """A dataclass of dependencies necessary for block verification."""
+    signaling_state: BlockSignalingState
+    feature_info: dict[Feature, FeatureInfo]
+
+    @classmethod
+    def create(cls, block: Block, feature_service: FeatureService) -> Self:
+        """Create a block dependencies instance."""
+        assert block.storage is not None
+        signaling_state = feature_service.is_signaling_mandatory_features(block)
+        feature_info = feature_service.get_feature_info(block=block)
+        simple_storage = SimpleMemoryStorage()
+
+        simple_storage.add_vertices_from_storage(block.storage, block.parents)
+        simple_storage.add_vertex(block)  # we add the block itself so its metadata can be used as a dependency.
+
+        return cls(
+            storage=simple_storage,
+            signaling_state=signaling_state,
+            feature_info=feature_info,
+        )
+
+
+class TransactionDependencies(VertexDependencies):
+    """A dataclass of dependencies necessary for transaction verification."""
+
+    @classmethod
+    def create(cls, tx: Transaction) -> Self:
+        """Create a transaction dependencies instance."""
+        assert tx.storage is not None
+        simple_storage = SimpleMemoryStorage()
+        spent_txs = [tx_input.tx_id for tx_input in tx.inputs]
+        deps = tx.parents + spent_txs
+
+        simple_storage.add_vertices_from_storage(tx.storage, deps)
+
+        return cls(storage=simple_storage)
--- a/hathor/verification/verification_service.py
+++ b/hathor/verification/verification_service.py
@@ -14,25 +14,36 @@
 
 from typing_extensions import assert_never
 
-from hathor.feature_activation.feature import Feature
-from hathor.feature_activation.feature_service import BlockSignalingState, FeatureService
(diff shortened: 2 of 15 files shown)


hathor/verification/verification_model.py
```diff
@@ -0,0 +1,125 @@
...
+from hathor.transaction.exceptions import IncorrectParents, InexistentInput
+from hathor.transaction.storage.exceptions import TransactionDoesNotExist
+from hathor.transaction.token_creation_tx import TokenCreationTransaction
+from hathor.transaction.transaction import RewardLockedInfo, TokenInfo, Transaction
+from hathor.types import TokenUid
+from hathor.util import not_none
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class VertexDependencies:
```
msbrogli, line 31: I think we can simplify it a lot by just implementing a simple memory database.

```python
class SimpleMemoryDB:
    def __init__(self):
        self._db: Dict[VertexId, BaseTransaction] = {}

    def add_vertex(self, vertex: BaseTransaction) -> None:
        self._db[vertex.hash] = vertex

    def get_vertex(self, _id: VertexId) -> BaseTransaction:
        return self._db[_id]
```

It might have support for metadata too if needed.

msbrogli (changes requested):

hathor/verification/verification_model.py
glevco: This PR was completely changed, so I recommend re-reviewing it. Indeed the `SimpleMemoryDB` was a good solution for tx parents and spent txs, and I implemented it. But I needed to keep the verification model abstractions for some other dependencies.

hathor/transaction/vertex.py
```diff
@@ -0,0 +1,52 @@
...
+    """Vertex wrapper for TokenCreationTransaction."""
+
+
+"""
+A Vertex algebraic sum type, or tagged union, that unifies all vertex types.
+
+This is useful when dealing with vertices in functional code, for example when using `match` statements, so we don't
+have problems with inheritance in `case` branches.
+"""
+Vertex: TypeAlias = BlockType | MergeMinedBlockType | TransactionType | TokenCreationTransactionType
```
msbrogli, line 52: Why are we doing this? I think it is confusing to have two different types for each vertex type.

hathor/verification/verification_dependencies.py
```diff
@@ -0,0 +1,68 @@
...
+@dataclass(frozen=True, slots=True)
+class BlockDependencies(VertexDependencies):
+    """A dataclass of dependencies necessary for block verification."""
+    signaling_state: BlockSignalingState
+
+    @classmethod
+    def create(cls, block: Block, daa: DifficultyAdjustmentAlgorithm, feature_service: FeatureService) -> Self:
+        """Create a block dependencies instance."""
+        assert block.storage is not None
+        signaling_state = feature_service.is_signaling_mandatory_features(block)
```
msbrogli, line 41: Why is it part of the dependencies?

hathor/verification/verification_service.py
```diff
@@ -80,35 +90,43 @@ def validate_full(
         vertex.set_validation(validation)
         return True
 
-    def verify_basic(self, vertex: BaseTransaction, *, skip_block_weight_verification: bool = False) -> None:
+    def verify_basic(self, base_tx: BaseTransaction, *, skip_block_weight_verification: bool = False) -> None:
```
msbrogli, line 93: I felt like we are doing a refactor and add a feature in this PR.

msbrogli (changes requested):

hathor/transaction/vertex.py
glevco: Indeed this is not strictly necessary. It's an improvement, however, in my opinion. This uses the [tagged unions](https://en.wikipedia.org/wiki/Tagged_union) concept of type theory, and I believe we could benefit from using this in future code. This is available in other forms in many languages (like sealed interfaces in Kotlin, or Enums in Rust), and is the same concept behind `Result` and `Option` types (that I also believe we should start using).

They're simply a low-cost wrapper for the existing `BaseTransaction` subclasses, with the advantage that they represent a single type. If we have a `Block`, for example, we have no way to know if it is a `Block` or a `MergeMinedBlock`, unless we perform a check. With this new `BlockType`, we're sure this is strictly a `Block`, and not any subclass (well, restricted via semantics, but we could restrict this via code too). This is useful for verification, as we have different behavior for each subclass, and any other code that perform this kind of polymorphism.

I do concede it may be a bit confusing at this time, however I think the types are not exactly redundant as they provide different semantics. And I also think if we end up using them in other code, we may find we want to evolve the current models in the future, and maybe this would become the main form of abstraction—by using a data class for each vertex type, after we remove all dependencies from the existing classes (verification, storage, etc). So I like the trade-off.

I could certainly remove it now, if you and @jansegre disagree. But at worst I see this as an experiment that we could remove down the road, if we don't like it.

Specifically for the use case in this PR, it removed boilerplate, as it allowed to change this:

```python
# We assert with type() instead of isinstance() because each subclass has a specific branch.
match vertex.version:
    case TxVersion.REGULAR_BLOCK:
        assert type(vertex) is Block
        self._verify_without_storage_block(vertex)
    case TxVersion.MERGE_MINED_BLOCK:
        assert type(vertex) is MergeMinedBlock
        self._verify_without_storage_merge_mined_block(vertex)
    case TxVersion.REGULAR_TRANSACTION:
        assert type(vertex) is Transaction
        self._verify_without_storage_tx(vertex)
    case TxVersion.TOKEN_CREATION_TRANSACTION:
        assert type(vertex) is TokenCreationTransaction
        self._verify_without_storage_token_creation_tx(vertex)
    case _:
        assert_never(vertex.version)
```

To this:

```python
vertex = base_tx.as_vertex()
match vertex:
    case BlockType(block):
        self._verify_without_storage_block(block)
    case MergeMinedBlockType(block):
        self._verify_without_storage_merge_mined_block(block)
    case TransactionType(tx):
        self._verify_without_storage_tx(tx)
    case TokenCreationTransactionType(tx):
        self._verify_without_storage_token_creation_tx(tx)
    case _:
        assert_never(vertex)
```

In several places.

hathor/verification/verification_dependencies.py
glevco: Because since the Feature Activation process is dynamic, it is not straightforward to determine which blocks are necessary to calculate this state in advance (differently from the DAA, for example, where it's easy to determine the number of necessary parent blocks).

In order to determine the necessary blocks, we would have to run the algorithm anyway, so there's almost no difference in cost between determining all block dependencies in advance, or determining the state itself in advance.

I suggest we implement it like this for now, and then after the verification is moved to another process, we measure the cost of doing this calculation in main process, to decide if it's worth the refactor (we would need to change Feature Activation code in order to get just the required blocks).

hathor/verification/verification_service.py
glevco: Which feature?

hathor/transaction/vertex.py
glevco: I've removed this from this PR.

hathor/verification/verification_dependencies.py
glevco: As discussed in our call, we'll keep this for now. After completing the Multiprocess Verification, we'll measure its performance and determine if it should be improved or not.

hathor/verification/verification_service.py
glevco: I've removed this from this PR.

jansegre (approved):

glevco:
@jansegre @msbrogli requested re-review after rebase with conflicts.

glevco:
Moved to PRs https://github.com/HathorNetwork/hathor-core/pull/1022 and https://github.com/HathorNetwork/hathor-core/pull/1023.

PR #867, opened 2023-11-10
refactor(p2p): have two internal tiers for sync: available and enabled

jansegre:
### Motivation

This PR refactors internal code in order to make it easier to add sysctl controls for enabling and disabling specific sync versions.

### Acceptance Criteria

- Refactor `ConnectionsManager` so that instead of having `enable_sync_v1` and `enable_sync_v2` it has "available sync versions" and "enabled sync versions". A sync version being available means that it can be enabled or disabled at runtime. This distinction is useful because there are init-time requirements for making specific sync versions available and it is practical to deal with these separately from whether a specific sync version is enabled or not.
- Make both sync-v1 and sync-v2 available by default, there's no CLI option to make them not available, CLI options only control whether they will be enabled, which can subsequently be controlled in runtime, but this will need a future PR that adds the sysctl controls.

### Checklist

### DIFF
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -86,6 +86,7 @@ class ConnectionsManager:
     handshaking_peers: set[HathorProtocol]
     whitelist_only: bool
     _sync_factories: dict[SyncVersion, SyncAgentFactory]
+    _enabled_sync_versions: set[SyncVersion]
 
     rate_limiter: RateLimiter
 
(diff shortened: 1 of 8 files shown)


hathor/p2p/manager.py
```diff
@@ -184,23 +177,57 @@ def __init__(self,
...
+            self.log.info('tried to enable a sync verison that was already enabled, nothing to do')
+            return
+        self._enabled_sync_versions.add(sync_version)
+
+    def disable_sync_version(self, sync_version: SyncVersion) -> None:
+        """Disable using the given sync version, closes connections using this sync version."""
+        if sync_version not in self._enabled_sync_versions:
+            self.log.info('tried to disable a sync verison that was already disabled, nothing to do')
+            return
+        # TODO: close connections using the given sync version
```
msbrogli, line 224: I don't think we should do it. Maybe create a separate method that closes all connections using a specific version.

msbrogli (dismissed):

glevco (dismissed):

hathor/p2p/manager.py
jansegre: I agree, I removed this comment and amended the docstring.

msbrogli (approved):

hathor/p2p/manager.py

glevco (approved):

PR #874, opened 2023-11-11
fix(storage): Fix wrong update to the best_block_tips_cache

msbrogli:
### Motivation

The `tests/simulation/test_simulator.py::SyncV2RandomSimulatorTestCase::test_many_miners_since_beginning` was flaky because of this issue.

When the consensus algorithm updates the `best_block_tips_cache`, it does not include the current block being evaluated. Notice that the current block is not part of the `heads`.

### Acceptance Criteria

1. Include `block` to the list of best block tips used to update the `best_block_tips_cache`.
2. In the `assertTipsEqualSyncV2()`, we shouldn't assert `b1 == b2` because the height index keeps its previous entry when two or more blocks have the same score. So we can safely just check if `b1 in s1` and `b2 in s2`.
3. Notice that, when `len(s1) == len(s2) == 1`, checking if `b1 in s1` and `b2 in s2` is exactly the same as checking if `b1 == b2`.

### Checklist

### DIFF
--- a/hathor/consensus/block_consensus.py
+++ b/hathor/consensus/block_consensus.py
@@ -216,7 +216,9 @@ class BlockConsensusAlgorithm:
                         if common_block not in heads:
                             self.context.mark_as_reorg(common_block)
                 else:
-                    storage.update_best_block_tips_cache([not_none(blk.hash) for blk in heads])
+                    best_block_tips = [not_none(blk.hash) for blk in heads]
+                    best_block_tips.append(not_none(block.hash))
+                    storage.update_best_block_tips_cache(best_block_tips)
                     if not meta.voided_by:
                         self.context.mark_as_reorg(common_block)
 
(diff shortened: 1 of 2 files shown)


hathor/consensus/block_consensus.py
```diff
@@ -216,7 +216,9 @@ def update_voided_info(self, block: Block) -> None:
                         if common_block not in heads:
                             self.context.mark_as_reorg(common_block)
                 else:
-                    storage.update_best_block_tips_cache([not_none(blk.hash) for blk in heads])
+                    best_block_tips = [not_none(blk.hash) for blk in heads]
+                    best_block_tips.append(not_none(block.hash))
+                    storage.update_best_block_tips_cache(best_block_tips)
```
msbrogli, line 221: Should we just set to `None` and let the next call cache it?

jansegre: Either way is fine for me. I wish we can get rid of this cache in the future. Will we even need it if we don't have best block ties, because even for sync-v1 the block height index would have a ready `O(1)` (or at least however fast `best_blocks[-1]` is) response?

jansegre (approved):

glevco (approved):

PR #875, opened 2023-11-14
fix(sync-v2): adjust init assertion

jansegre:
### Motivation

During the QA of sync-v2 some errors were caught during a sync from scratch ([see here](https://github.com/HathorNetwork/internal-issues/issues/199#issuecomment-1808646707)). In particular this one:

```
2023-11-12 05:10:57 [warning  ] [hathor.p2p.protocol] send error                     msg=internal error peer_id=7181d23 remote=127.0.0.1:9001
2023-11-12 05:10:57 [error    ] [hathor.p2p.sync_v2.agent] unhandled exception            peer=7181d23
Traceback (most recent call last):
  File "/Users/jan/Projects/hathor-core/hathor/p2p/sync_v2/agent.py", line 296, in run_sync
    yield self._run_sync()
  File "/Users/jan/Library/Caches/pypoetry/virtualenvs/hathor-JIQwCn9w-py3.11/lib/python3.11/site-packages/twisted/internet/defer.py", line 1693, in _inlineCallbacks
    result = context.run(
             ^^^^^^^^^^^^
  File "/Users/jan/Library/Caches/pypoetry/virtualenvs/hathor-JIQwCn9w-py3.11/lib/python3.11/site-packages/twisted/python/failure.py", line 518, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jan/Projects/hathor-core/hathor/p2p/sync_v2/agent.py", line 311, in _run_sync
    is_block_synced = yield self.run_sync_blocks()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jan/Library/Caches/pypoetry/virtualenvs/hathor-JIQwCn9w-py3.11/lib/python3.11/site-packages/twisted/internet/defer.py", line 1697, in _inlineCallbacks
    result = context.run(gen.send, result)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jan/Projects/hathor-core/hathor/p2p/sync_v2/agent.py", line 397, in run_sync_blocks
    reason = yield self.start_transactions_streaming(partial_blocks)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jan/Projects/hathor-core/hathor/p2p/sync_v2/agent.py", line 834, in start_transactions_streaming
    self._tx_streaming_client = TransactionStreamingClient(self,
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jan/Projects/hathor-core/hathor/p2p/sync_v2/transaction_streaming_client.py", line 87, in __init__
    assert self._waiting_for
AssertionError
2023-11-12 05:10:57 [info     ] [hathor.p2p.protocol] disconnected                   peer_id=7181d23 reason=Connection was closed cleanly. remote=127.0.0.1:9001
```

This indicates that the assertion inside `TransactionStreamingClient.__init__` is not correct.

### Acceptance Criteria

- Adjust `TransactionStreamingClient.__init__` assertion to account for `_existing_deps` and not only `_waiting_for`.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/transaction_streaming_client.py
+++ b/hathor/p2p/sync_v2/transaction_streaming_client.py
@@ -84,7 +84,6 @@ class TransactionStreamingClient:
         self._existing_deps: set[VertexId] = set()
 
         self._prepare_block(self.partial_blocks[0])
-        assert self._waiting_for
 
     def wait(self) -> Deferred[StreamEnd]:
         """Return the deferred."""


hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -84,7 +84,10 @@ def __init__(self,
         self._existing_deps: set[VertexId] = set()
 
         self._prepare_block(self.partial_blocks[0])
-        assert self._waiting_for
+
+        # XXX: it could be that we already have all dependencies so _waiting_for ends up empty, but then it would have
+        #      to be added to _existing_deps, so it must be impossible for the union of both to be empty
+        assert len(self._waiting_for) + len(self._existing_deps) > 0
```
msbrogli, line 90: I think this might fix but shouldn't we try to process all blocks before starting downloading them? I mean, this fix should work but we will download transactions that we already have.

msbrogli (changes requested):

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: What do you think of moving this assert to `_prepare_block` and possibly even using a builder and call `_prepare_block` on `build()` instead of on this classes `__init__` (I'd prefer if `__init__` does minimal work and avoid doing stuff like db access).

jansegre: I pushed how it would look like in a new commit, if you think it doesn't make sense I'll revert it.

jansegre: It's actually not correct to have that in `_prepare_block` it broke several tests.

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -37,12 +38,31 @@
 logger = get_logger()
 
 
+@dataclass
+class TransactionStreamingClientBuilder:
```
msbrogli, line 42: The builder seems to be a too complex solution for the issue. I feel that it should be handled by the agent instead because it must align the request to open the stream with the client configuration. And, as a race condition might always occur between the request and the actual stream, maybe we can just remove the assert for now.

msbrogli (changes requested):

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: The builder is not intended to be a solution for the issue itself. It is used as a better pattern for doing work during initialization (which basically is just calling a private method) and avoid doing it during `__init__`. It would be even an even better fit if the work was async.

But I understand your point about making the agent more aware of this. I'll leave it to just removing the assert for now then. There would be a lot of repeated logic if both the agent and the streamer client implement the logic of advancing over the list of blocks and preparing which transactions need to be downloaded if the agent has to do that before instantiating the streamer client.

Maybe an improved solution would move the implementation of that logic outside the streaming client and into a common structure that the agent uses and would be passed to the streaming client.

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: This assert was removed.

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: Done. Assert removed.

msbrogli (approved):

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: I agree. We can refactor it between now and the official release of sync-v2. We might add a `start()` method that would do whatever is needed without the builder.

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: Thanks!

glevco (approved):

PR #877, opened 2023-11-20
fix(sync-v2): stop _process_transaction on error

jansegre:
### Motivation

During the last QA when a node was syncing from scratch, we caught a case where more than one unexpected transaction was sent on a streamer, the first one correctly failed the streamer client, but the second one lead to an error because the deferred had already been called.

### Acceptance Criteria

- Add test to reproduce the case;
- Make it so `TransactionStreamingClient.fails` will not call `self._deferred.errback` if the deferred has already been called.

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/transaction_streaming_client.py
+++ b/hathor/p2p/sync_v2/transaction_streaming_client.py
@@ -99,6 +99,9 @@ class TransactionStreamingClient:
 
     def fails(self, reason: 'StreamingError') -> None:
         """Fail the execution by resolving the deferred with an error."""
+        if self._deferred.called:
+            self.log.warn('already failed before', new_reason=repr(reason))
+            return
         self._deferred.errback(reason)
 
     def handle_transaction(self, tx: BaseTransaction) -> None:
@@ -125,6 +128,9 @@ class TransactionStreamingClient:
     @inlineCallbacks
     def process_queue(self) -> Generator[Any, Any, None]:
         """Process next transaction in the queue."""
+        if self._deferred.called:
+            return
+
         if self._is_processing:
             return
 
--- a/tests/p2p/test_sync_v2.py
+++ b/tests/p2p/test_sync_v2.py
@@ -1,12 +1,22 @@
+import base64
+import re
+
 import pytest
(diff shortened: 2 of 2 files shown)


hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -100,6 +100,9 @@ def resume(self) -> Deferred[StreamEnd]:
 
     def fails(self, reason: 'StreamingError') -> None:
         """Fail the execution by resolving the deferred with an error."""
+        if self._deferred.called:
+            self.log.warn('already failed before', new_reason=str(reason))
```
msbrogli, line 104: Maybe `repr(reason)`? I feel it's worthy logging the name of the exception.

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli, line 1: The multiple calls to `self.fails()` was caused by the scheduling of `process_queue()`. Should we stop this processing queue if an error occurs? It seems enough to just check for `self._deferred.active` in the `process_queue()` method.

msbrogli (changes requested):

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: Done.

hathor/p2p/sync_v2/transaction_streaming_client.py
jansegre: I added a check for `self._deferred.called` on `process_queue`, I checked that it also fixed the test by itself (without the change in `fails()`).

msbrogli (approved):

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: Thanks!

msbrogli (approved):

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: Thanks!

glevco (approved):

Issue #881, opened 2023-11-28
QA issue during release-candidate v0.58.0-rc1


jansegre:
During the QA of the release-candidate v0.58.0-rc1, the following issue has been identified:

- Excessive memory usage during sync from scratch using sync-v1, causing the host to hang.

Typical memory usage on our production hosts hover just bellow 8GB, hosts are deployed with 16GB. During the testing of syncing from scratch, specifically only when using sync-v1 (when using sync-v2 it completes successfully). It's most likely to have happened because now the mempool-tips index is enabled even if sync-v2 is not enabled, this was made to facilitate enabling sync-v2 later in runtime.

Other than that the nodes ran really well, and CPU usage on sync-v2 nodes is noticeably lower.

jansegre:
Reviewers : @msbrogli @luislhl

PR #882, opened 2023-11-28
fix(cli): do not include sync-v2 indexes when not explicitly enabled

jansegre:
### Motivation

This PR addresses the issue caught on the QA of RC-1, detailed on #881 

### Acceptance Criteria

- Do not add sync-v2 factory when sync-v2 is not explicitly enabled, in turn making the sync-v2 index (the mempool-tips-index) to not be always turned on, the downside being that sync-v2 cannot be enabled through sysctl if a node starts with sync-v1 only.

### Checklist

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -353,12 +353,17 @@ class Builder:
             whitelist_only=False,
             rng=self._rng,
         )
+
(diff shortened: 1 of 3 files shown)


jansegre:
Not sure if any test will break, will let the CI run.

glevco (approved):

PR #887, opened 2023-12-06
chore(custom-checks): add check to prevent circular imports

glevco:
### Motivation

Sometimes, when something is imported from the `hathor` module in an entrypoint file (from CLI commands), a circular import could happen, causing the full node node startup to fail. This has happened a couple times in the past ([example](https://github.com/HathorNetwork/hathor-core/pull/762)).

This problem would not be caught by our linters and neither tests, so it would only appear after the merge to `master`, when the docker image is generated. This PR aims to prevent this by adding a custom lint check that prohibits importing from `hathor` in the module-level of CLI entrypoints. For specific cases, a `# no-custom-check` comment can be added to opt-out of this check. [Here's an example](https://github.com/HathorNetwork/hathor-core/actions/runs/7121270335/job/19390180920?pr=887) of a failing output.

This will also be useful in https://github.com/HathorNetwork/hathor-core/pull/889, as it prevents indirect calls that could automatically install the default Twisted Reactor before our code had a chance to customize it.

### Acceptance Criteria

- Add new custom check to prevent importing from the `hathor` module in the module-level of CLI related files.
- Update code to conform with the new check.

### Checklist

### DIFF
--- a/extras/custom_checks.sh
+++ b/extras/custom_checks.sh
@@ -89,12 +89,25 @@ function check_do_not_import_tests_in_hathor() {
 	return 0
 }
 
+function check_do_not_import_from_hathor_in_entrypoints() {
+    PATTERN='^import .*hathor.*\|^from .*hathor.* import'
+
+    if grep -R "$PATTERN" "hathor/cli" | grep -v 'from hathor.cli.run_node import RunNode' | grep -v '# skip-cli-import-custom-check'; then
+        echo 'do not import from `hathor` in the module-level of a CLI entrypoint.'
+        echo 'instead, import locally inside the function that uses the import.'
+        echo 'alternatively, comment `# skip-cli-import-custom-check` to exclude a line.'
+        return 1
+    fi
+    return 0
+}
+
 # List of functions to be executed
 checks=(
 	check_version_match
 	check_do_not_use_builtin_random_in_tests
 	check_deprecated_typing
 	check_do_not_import_tests_in_hathor
+	check_do_not_import_from_hathor_in_entrypoints
 )
 
 # Initialize a variable to track if any check fails
--- a/hathor/builder/cli_builder.py
+++ b/hathor/builder/cli_builder.py
@@ -21,7 +21,7 @@ from typing import Any, Optional
 
 from structlog import get_logger
 
-from hathor.cli.run_node import RunNodeArgs
+from hathor.cli.run_node_args import RunNodeArgs
 from hathor.consensus import ConsensusAlgorithm
 from hathor.daa import DifficultyAdjustmentAlgorithm
 from hathor.event import EventManager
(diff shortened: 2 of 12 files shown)


extras/custom_checks.sh
```diff
@@ -89,12 +89,25 @@ function check_do_not_import_tests_in_hathor() {
 	return 0
 }
 
+function check_do_not_import_from_hathor_in_entrypoints() {
+    PATTERN='^import .*hathor.*\|^from .*hathor.* import'
+
+    if grep -R "$PATTERN" "hathor/cli" | grep -v '# no-custom-check'; then
```
msbrogli, line 95: Maybe `# skip-cli-import-custom-check`?

extras/custom_checks.sh
```diff
@@ -89,12 +89,25 @@ function check_do_not_import_tests_in_hathor() {
 	return 0
 }
 
+function check_do_not_import_from_hathor_in_entrypoints() {
```
msbrogli, line 92: Why don't we allow `import hathor.cli`? It seems most of the exceptions would fit in this case.

extras/custom_checks.sh
msbrogli, line 1: I think we can merge this PR but I don't see much use for these changes. In fact, I feel that preventing importing is the wrong way to prevent indirect calls that could automatically install the default Twisted Reactor before our code had a chance to customize it.

msbrogli (changes requested):

extras/custom_checks.sh
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/887/commits/1ac0f096bb9e39d44dddc541015fc472420a88f5, although I added a more strict version of this, specifically allowing to import `RunNode`.

extras/custom_checks.sh
glevco: Do you have any other ideas on how we could prevent it? Anyway, this is most likely temporary. After we finish refactoring all global accesses of the settings and reactor instances, neither of those problems would happen (initializing settings multiple times, and initializing the reactor before customization).

msbrogli: I did it here: https://github.com/HathorNetwork/hathor-core/pull/889#discussion_r1419416771

msbrogli: I just checked that it's already implemented. Great!

msbrogli (approved):

extras/custom_checks.sh
msbrogli: Thanks!

extras/custom_checks.sh
msbrogli: Great! Thanks!

jansegre (approved):

PR #889, opened 2023-12-06
feat(reactor): implement option to use asyncio reactor

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/888

### Motivation

Considering the general improvements we plan to make in the full node related to concurrent and parallel execution, using async/await and multiprocessing, it may be useful to interop with Python's native `asyncio`. Currently we used Twisted's default reactor, but they provide a reactor that has compatibility with asyncio: [AsyncioSelectorReactor](https://docs.twisted.org/en/stable/api/twisted.internet.asyncioreactor.AsyncioSelectorReactor.html). In theory, it is a drop-in replacement for what we're currently using.

This PR adds an experimental CLI option to enable this custom reactor, so we can test its usage in some full nodes for a while. Then, we'll have data to assess whether it's worth it using it as the default, in the future.

### Acceptance Criteria

- Add `--x-asyncio-reactor` CLI option to enable using the asyncio reactor.
- Update `get_global_reactor()` to add support for installing the asyncio reactor.
- Add reactor type to initial logs.

### Checklist

### DIFF
--- a/hathor/builder/cli_builder.py
+++ b/hathor/builder/cli_builder.py
@@ -100,6 +100,7 @@ class CliBuilder:
             python=python,
             platform=platform.platform(),
             settings=settings_source,
+            reactor_type=type(reactor).__name__,
         )
 
         tx_storage: TransactionStorage
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -33,7 +33,8 @@ class RunNode:
         ('--enable-crash-api', lambda args: bool(args.enable_crash_api)),
         ('--x-sync-bridge', lambda args: bool(args.x_sync_bridge)),
         ('--x-sync-v2-only', lambda args: bool(args.x_sync_v2_only)),
-        ('--x-enable-event-queue', lambda args: bool(args.x_enable_event_queue))
+        ('--x-enable-event-queue', lambda args: bool(args.x_enable_event_queue)),
+        ('--x-asyncio-reactor', lambda args: bool(args.x_asyncio_reactor))
     ]
 
     @classmethod
@@ -117,6 +118,8 @@ class RunNode:
                             help=f'Signal support for a feature. One of {possible_features}')
         parser.add_argument('--signal-not-support', default=[], action='append', choices=possible_features,
                             help=f'Signal not support for a feature. One of {possible_features}')
+        parser.add_argument('--x-asyncio-reactor', action='store_true',
+                            help='Use asyncio reactor instead of Twisted\'s default.')
         return parser
 
     def prepare(self, *, register_resources: bool = True) -> None:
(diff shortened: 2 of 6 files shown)


hathor/builder/cli_builder.py
```diff
@@ -100,6 +100,7 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
             python=python,
             platform=platform.platform(),
             settings=settings_source,
+            reactor_type=type(reactor),
```
msbrogli, line 103: Maybe `repr()` works best since it can show the initialization parameters?

hathor/cli/run_node.py
```diff
@@ -117,6 +117,8 @@ def create_parser(cls) -> ArgumentParser:
                             help=f'Signal support for a feature. One of {possible_features}')
         parser.add_argument('--signal-not-support', default=[], action='append', choices=possible_features,
                             help=f'Signal not support for a feature. One of {possible_features}')
+        parser.add_argument('--x-asyncio-reactor', action='store_true',
```
msbrogli, line 120: Maybe add it to the list of unsafe parameters?

hathor/reactor/reactor.py
```diff
@@ -14,25 +14,55 @@
...
 from hathor.reactor.reactor_protocol import ReactorProtocol
 
+logger = get_logger()
+
 # Internal variable that should NOT be accessed directly.
 _reactor: ReactorProtocol | None = None
 
 
-def get_global_reactor() -> ReactorProtocol:
+def get_global_reactor(*, use_asyncio_reactor: bool | None = None) -> ReactorProtocol:
```
msbrogli, line 29: Maybe create an `initialize_global_reactor()` that must be called first.

msbrogli (changes requested):

hathor/cli/run_node.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/889/commits/af2bb36f82eef6f6c5fdd9d6e33ae99e355ae1be

hathor/builder/cli_builder.py
jansegre: I'm even OK even with just `reactor_type=type(reactor).__name__`.

hathor/reactor/reactor.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/889/commits/70f2ad036acfa65217d08613d98cd9c86ae46c00

hathor/cli/run_node.py
msbrogli: Thanks!

hathor/builder/cli_builder.py
msbrogli: Got it. I'm ok either way here. You choose. ;)

hathor/reactor/reactor.py
msbrogli: Thanks!

msbrogli (dismissed):

hathor/builder/cli_builder.py
glevco: I changed it to @jansegre's suggestion in https://github.com/HathorNetwork/hathor-core/pull/889/commits/15ba731c56ee99802a0021d0e6791c66441ed121

jansegre (dismissed):

hathor/builder/cli_builder.py
msbrogli: Thanks!

msbrogli (approved):

PR #894, opened 2023-12-14
feat(feature-activation): add feature settings validation

glevco:
### Motivation

This PR implements a validation that compares the Feature Activation settings from the configuration file, against the previous value that is stored in the database. This runs during full node initialization and is just to prevent human errors when editing those files, since one cannot remove or change existing features from there.

### Acceptance Criteria

- Create new `FeatureActivationStorage` that validates configured feature activation settings against the previous value from the database.
- Add new `reset-feature-settings` CLI tool for erasing saved feature settings.

### Checklist

### DIFF
--- a/hathor/feature_activation/feature.py
+++ b/hathor/feature_activation/feature.py
@@ -16,7 +16,7 @@ from enum import Enum, unique
 
 
 @unique
-class Feature(Enum):
+class Feature(str, Enum):
     """
     An enum containing all features that participate in the feature activation process, past or future, activated
     or not, for all networks. Features should NOT be removed from this enum, to preserve history. Their values
--- a/hathor/feature_activation/storage/feature_activation_storage.py
+++ b/hathor/feature_activation/storage/feature_activation_storage.py
@@ -1,0 +1,101 @@
+#  Copyright 2023 Hathor Labs
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from structlog import get_logger
+
+from hathor.conf.settings import HathorSettings
+from hathor.exception import InitializationError
+from hathor.feature_activation.feature import Feature
+from hathor.feature_activation.model.criteria import Criteria
+from hathor.feature_activation.settings import Settings as FeatureActivationSettings
+from hathor.storage import RocksDBStorage
+
+_CF_NAME_META = b'feature-activation-metadata'
+_KEY_SETTINGS = b'feature-activation-settings'
+
+logger = get_logger()
+
+
+class FeatureActivationStorage:
+    __slots__ = ('_log', '_settings', '_db', '_cf_meta')
+
+    def __init__(self, *, settings: HathorSettings, rocksdb_storage: RocksDBStorage) -> None:
+        self._log = logger.new()
+        self._settings = settings
+        self._db = rocksdb_storage.get_db()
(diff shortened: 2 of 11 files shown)


hathor/feature_activation/feature.py
```diff
@@ -16,7 +16,7 @@
 
 
 @unique
-class Feature(Enum):
+class Feature(str, Enum):
```
msbrogli, line 19: Why this change?

hathor/manager.py
```diff
@@ -98,6 +99,7 @@ def __init__(self,
                  event_manager: EventManager,
                  feature_service: FeatureService,
                  bit_signaling_service: BitSignalingService,
+                 feature_storage: FeatureActivationStorage | None,
```
msbrogli, line 102: Why optional? Shouldn't it be part of the `BitSignalingService`?

hathor/feature_activation/storage/feature_activation_storage.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class FeatureActivationStorage:
+    __slots__ = ('_log', '_settings', '_db', '_cf_meta')
+
+    def __init__(self, *, settings: HathorSettings, rocksdb_storage: RocksDBStorage) -> None:
+        self._log = logger.new()
+        self._settings = settings
+        self._db = rocksdb_storage.get_db()
+        self._cf_meta = rocksdb_storage.get_or_create_column_family(_CF_NAME_META)
```
msbrogli, line 36: Should we use a single column family for all settings related stuff?

msbrogli (changes requested):

hathor/feature_activation/feature.py
glevco: Because now we're serializing this to json (as it's stored in the database), so adding this make it work (otherwise it tries to serialize a raw enum and fails). It's equivalent to using `StrEnum` from Python 3.11.

hathor/manager.py
glevco: It's optional because when the full node is running with `--memory-storage`, this check doesn't make sense. As this is only here to prevent human error in the configuration file, it's only necessary when running the full node "for real", with a rocksdb storage.

Also, I left it separate from the `BitSignalingService` for the same reason, as this is just a pre-initialization check. Maybe the name `FeatureActivationStorage` is not so good, it could be something more specific like `FeatureActivationSettingsValidator`?

hathor/feature_activation/storage/feature_activation_storage.py
glevco: Makes sense, but what other related stuff do you mean? Like using it to store the whole `HathorSettings` instance?

hathor/feature_activation/feature.py
msbrogli: Got it. Thanks!

hathor/manager.py
msbrogli: Got it. But why would HathorManager have to understand the validation of the feature service? Shouldn't the FeatureService or the BitSignalingService do it themselves? Maybe they will have to have a `start()` method?

glevco: I moved it to the `BitSignalingService` on https://github.com/HathorNetwork/hathor-core/pull/894/commits/dfb7c66a17ae0452453ea1ffe4bbb9307a19676e

jansegre (approved):

hathor/manager.py

hathor/feature_activation/storage/feature_activation_storage.py
msbrogli: Now the whole HathorSettings but we do have some flags stored in the database here and there in the code.

msbrogli (dismissed):

hathor/feature_activation/storage/feature_activation_storage.py
msbrogli: I guess we don't have to do this refactor in this PR. So I'll approve this PR and we can discuss this refactor separately. Thanks!

msbrogli (approved):

jansegre (approved):

jansegre (dismissed):

msbrogli (approved):

jansegre (approved):

PR #895, opened 2023-12-14
refactor(daa): externalize block dependencies

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/909

### Motivation

Change the DAA so it does not access the storage directly. Instead, block verification can pass pre-fetched dependencies through a new `SimpleMemoryStorage`. This will be useful for Multiprocess Verification.

### Acceptance Criteria

- Create new `SimpleMemoryStorage` class that will be used for storing pre-fetched verification dependencies in memory.
- Change DAA methods `calculate_block_difficulty` and `calculate_next_weight` to accept a `SimpleMemoryStorage`, which is used to retrieve pre-fetched parent blocks.
- Update block's `verify_weight` to pre-fetch DAA dependencies.

### Checklist

### DIFF
--- a/hathor/transaction/storage/simple_memory_storage.py
+++ b/hathor/transaction/storage/simple_memory_storage.py
@@ -1,0 +1,99 @@
+#  Copyright 2023 Hathor Labs
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from hathor.transaction import Block, Transaction
+from hathor.transaction.base_transaction import BaseTransaction
+from hathor.transaction.storage import TransactionStorage
+from hathor.transaction.storage.exceptions import TransactionDoesNotExist
+from hathor.types import VertexId
+
+
+class SimpleMemoryStorage:
+    """
+    Instances of this class simply facilitate storing some data in memory, specifically for pre-fetched verification
+    dependencies.
+    """
+    __slots__ = ('_blocks', '_transactions',)
+
+    def __init__(self) -> None:
+        self._blocks: dict[VertexId, BaseTransaction] = {}
+        self._transactions: dict[VertexId, BaseTransaction] = {}
+
+    @property
+    def _vertices(self) -> dict[VertexId, BaseTransaction]:
+        """Blocks and Transactions together."""
+        return {**self._blocks, **self._transactions}
+
+    def get_block(self, block_id: VertexId) -> Block:
+        """Return a block from the storage, throw if it's not found."""
+        block = self._get_vertex(self._blocks, block_id)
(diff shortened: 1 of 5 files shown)


hathor/verification/block_verifier.py
```diff
@@ -51,7 +53,11 @@ def verify_height(self, block: Block) -> None:
 
     def verify_weight(self, block: Block) -> None:
         """Validate minimum block difficulty."""
-        min_block_weight = self._daa.calculate_block_difficulty(block)
+        memory_storage = SimpleMemoryStorage()
+        dependencies = self._daa.get_block_dependencies(block)
+        memory_storage.add_vertices_from_storage(not_none(block.storage), dependencies)
+
+        min_block_weight = self._daa.calculate_block_difficulty(block, memory_storage)
```
glevco, line 60: This is done "inline" here temporarily. In next PRs it will be moved to outside the verification.

hathor/transaction/storage/simple_memory_storage.py
```diff
@@ -0,0 +1,105 @@
...
+from hathor.transaction import Block, Transaction, TransactionMetadata
+from hathor.transaction.base_transaction import BaseTransaction, tx_or_block_from_bytes
+from hathor.transaction.storage import TransactionStorage
+from hathor.transaction.storage.exceptions import TransactionDoesNotExist
+from hathor.types import VertexId
+
+
+@dataclass(frozen=True, slots=True)
+class _SimpleMemoryRecord:
+    vertex_bytes: bytes
```
msbrogli, line 27: Why don't we store `BaseTransaction` and `TransactionMetadata`? I mean, I see no use to serialize/deserialize since they will be consuming memory anyway.

msbrogli (changes requested):

hathor/transaction/storage/simple_memory_storage.py
glevco: Because this is going to be sent to another process for the multiprocess verification. Python has to (de)serialize objects to do that mandatorily, and by default it uses pickling.

Therefore, storing `bytes` and `dict` directly has some advantages:

1. Since we perform manual (de)serialization ourselves, it's more performatic than pickling. For instance, `bytes` are sent directly to the other process, without pickling.
2. We want the verification to be pure, that is, it's going to receive all pre-calculated dependencies and it won't have any side effects. This means that every input should be passed by value (that is, the transaction and its metadata). One way of guaranteeing that would be to use the `clone()` methods. What they do is exactly (de)serializing using our (de)serialization.
3. Some types cannot be passed between processes, for example, weak refs. Both the `TransactionStorage` (which is inside transactions) and `TransactionMetadata` are complex objects that store weak refs. Therefore, in order to make them be passable between processes, we would have to clean that. For example, remove the weak ref before sending them to the other process, and then restoring it after verification is complete. This is unnecessary and more complicated. Is also a good thing to retrieve deserialized transaction instances from the `SimpleMemoryStorage`, as it guarantees it won't point to a `TransactionStorage`, so it's impossible to manipulate the storage from another process (which is desirable, in the current implementation).

In conclusion, storing raw types has advantages considering the `SimpleMemoryDatabase` is going to be passed around between processes.

glevco: Changed to use `BaseTransaction` in https://github.com/HathorNetwork/hathor-core/pull/895/commits/e774d522352b5055e70ccd533021e978c86428b8

msbrogli: Thanks!

msbrogli (approved):

PR #904, opened 2023-12-22
refactor(verification): externalize verification dependencies [part 2/2]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/859

### Motivation

Finish the refactor started in #859 (read its description for more info). After this PR, all verification dependencies are externalized and pre-calculated.

### Acceptance Criteria

- Add missing verification dependencies for transactions: token info and best block tips.
- Add new auxiliary methods to `SimpleMemoryStorage`.
- Update verifiers and `VerificationService` accordingly.
- Update `Transaction` methods related to token info to receive an optional `SimpleMemoryStorage`, which may be used to retrieve data.

### Checklist

### DIFF
--- a/hathor/transaction/transaction.py
+++ b/hathor/transaction/transaction.py
@@ -23,7 +23,7 @@ from hathor.profiler import get_cpu_profiler
 from hathor.reward_lock import iter_spent_rewards
 from hathor.transaction import BaseTransaction, TxInput, TxOutput, TxVersion
 from hathor.transaction.base_transaction import TX_HASH_SIZE
-from hathor.transaction.exceptions import InvalidToken
+from hathor.transaction.exceptions import InexistentInput, InvalidToken
 from hathor.transaction.util import VerboseCallback, unpack, unpack_len
 from hathor.types import TokenUid, VertexId
 from hathor.util import not_none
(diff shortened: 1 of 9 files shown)


hathor/transaction/transaction.py
```diff
@@ -349,10 +355,10 @@ def _update_token_info_from_outputs(self, *, token_dict: dict[TokenUid, TokenInf
                     sum_tokens = token_info.amount + tx_output.value
                     token_dict[token_uid] = TokenInfo(sum_tokens, token_info.can_mint, token_info.can_melt)
 
-    def iter_spent_rewards(self) -> Iterator[Block]:
+    def iter_spent_rewards(self, storage: Optional['SimpleMemoryStorage'] = None) -> Iterator[Block]:
```
msbrogli, line 358: It seems like we are juggling to make the new memory storage work alongside the current code. I think we can spawn an ephemeral verifier that carries a storage and use it to run the verification. It can be either set to `tx.storage` or to any other storage. What do you think?

msbrogli (changes requested):

hathor/transaction/transaction.py
```diff
@@ -305,7 +306,12 @@ def _get_token_info_from_inputs(self) -> dict[TokenUid, TokenInfo]:
 
         for tx_input in self.inputs:
             spent_tx = self.get_spent_tx(tx_input)
-            spent_output = spent_tx.outputs[tx_input.index]
+            try:
+                spent_output = spent_tx.outputs[tx_input.index]
+            except IndexError:
+                raise InexistentInput(
```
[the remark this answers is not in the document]
glevco: It's just that before, this method was only called during verification. Now, it's called before verification (when retrieving the verification dependencies). So, if the tx is invalid, it will raise this exception during that retrieval.

hathor/transaction/transaction.py
glevco: Indeed, this was not a good solution. I have solved this using a `StorageProtocol`, so both `TransacionStorage` and `SimpleMemoryStorage` have an unified interface that can be received by those methods. It was implemented in https://github.com/HathorNetwork/hathor-core/pull/922, and after it's merged to master I'll rebase this PR so this is updated.

jansegre (approved):

glevco:
Moved to PRs https://github.com/HathorNetwork/hathor-core/pull/1022 and https://github.com/HathorNetwork/hathor-core/pull/1023.

PR #910, opened 2024-01-05
feat(merged-mining): configure new max merkle path length on testnet

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/931

### Motivation

Create a new feature to change the maximum merkle path length of merged mining through Feature Activation, and configure it on testnet to start the process in roughly two weeks.

### Acceptance Criteria

- Create `INCREASE_MAX_MERKLE_PATH_LENGTH` feature and configure it on testnet.
- Change `BitcoinAuxPow.verify` to receive a `max_merkle_path_length` instead of using a constant.
- Change `MergeMinedBlockVerifier.verify_aux_pow` to use Feature Activation to determine the `max_merkle_path_length`.
- Update tests.

### Checklist


hathor/transaction/aux_pow.py
```diff
@@ -19,7 +19,8 @@
 logger = get_logger()
 
 
-MAX_MERKLE_PATH_LENGTH: int = 12
+OLD_MAX_MERKLE_PATH_LENGTH: int = 12
+NEW_MAX_MERKLE_PATH_LENGTH: int = 15
```
glevco, line 23: The reason I chose `15` is because I ran a query on our AWS logs for the last 12 weeks, and in that time we had:

- 1591 instances of `merkle_path too long: 13 > 12`
- 757 instances of `merkle_path too long` (from before the logging was improved)
- 14 instances of `merkle_path too long: 15 > 12`
- 1 instance of `merkle_path too long: 14 > 12`

This of course includes only miners connected to our own nodes. We can still change this number, but I will detail this reasoning further on the RFC.

jansegre, line 23: I think these parameters should be moved to settings.

jansegre: For the testnet, I think 15 (or mostly whatever increase), works, mostly because we'll be able to test. We should also prepare to be able to merge mine, at least against a "dummy chain" (we can use zero'd or random bytes to build dummy data). But for the mainnet, I think we should:

1. at least estimate what parameter would fit the chains most commonly merged-mined with (BTC/DGC/BSV/BCH/BTG), some of those have unlimited blocks, so the theoretical limit of those doesn't matter, but at least scanning the blocks with most transactions on those chains
2. and maybe most importantly, getting feedback from the main miners, since they are directly affected by it

The most pressing counterpoint to a high limit is the block size, which increases with `merkle_path_len * 32 bytes`, the `merkle_path_len` doesn't increase proportionally with the max allowed limit though, the max allowed limit only expands the distribution of allowed values. I'm not sure what's a good way to measure this, I think it's probably not harmful to err a little bit to a higher limit.

hathor/verification/merge_mined_block_verifier.py
```diff
@@ -12,14 +12,29 @@
...
         """ Verify auxiliary proof-of-work (for merged mining).
         """
         assert block.aux_pow is not None
-        block.aux_pow.verify(block.get_base_hash())
+
+        is_feature_active = self._feature_service.is_feature_active(
+            block=block,
+            feature=Feature.INCREASE_MAX_MERKLE_PATH_LENGTH
+        )
+        max_merkle_path_length = NEW_MAX_MERKLE_PATH_LENGTH if is_feature_active else OLD_MAX_MERKLE_PATH_LENGTH
```
jansegre, line 38: Already mentioned previously, but I'd prefer this value to be in settings, we then could have different values for the testnet/mainnet/privatenets, not that it is useful in practice, but it will be useful for starting the experiment on the testnet with a non-optimal value, and choosing an optimal value for the mainnet without interfering with the testnet.

jansegre (changes requested):

hathor/transaction/aux_pow.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/910/commits/f4b6182ad58640dc0a0f51cdeb967e178c298873

hathor/verification/merge_mined_block_verifier.py
glevco: Makes sense, changed in https://github.com/HathorNetwork/hathor-core/pull/910/commits/f4b6182ad58640dc0a0f51cdeb967e178c298873

hathor/transaction/aux_pow.py
glevco: Agreed. I'm keeping it as 15 then, for testnet. We can do that investigation to chose a better value for mainnet before configuring its activation process.

jansegre (dismissed):

glevco:
- [ ] TODO: I'll delay the start of the process so it can be done after changing the Evaluation Interval from 2 weeks to 1 week.

hathor/conf/testnet.yml
glevco, line 1: - [ ] TODO: I'll delay the start of the process so it can be done after changing the Evaluation Interval from 2 weeks to 1 week.

hathor/merged_mining/coordinator.py
```diff
@@ -624,7 +624,10 @@ def handle_submit(self, params: list[Any], msgid: Optional[str]) -> None:
 
         try:
             aux_pow = job.build_aux_pow(work)
-            aux_pow.verify(block_base_hash)
+            # This constant usually comes from full node settings, but it's hardcoded here as this runs as a
+            # standalone CLI tool.
+            old_max_merkle_path_length = 12
```
msbrogli, line 629: Add a note that it must be equal to `settings.NEW_MAX_MERKLE_PATH_LENGTH`.

msbrogli (approved):

hathor/merged_mining/coordinator.py
glevco: I'll separate the aux_pow verify method and not verify the merkle path length here

hathor/transaction/aux_pow.py
glevco: Changed to 20 in https://github.com/HathorNetwork/hathor-core/pull/910/commits/8aa74cba4e8edb7108bf64d45c7af7ba7adca292, as it's most likely the number we'll also use on mainnet.

hathor/conf/testnet.yml
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/910/commits/8aa74cba4e8edb7108bf64d45c7af7ba7adca292, @msbrogli @jansegre could you please re-review it?

hathor/merged_mining/coordinator.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/910/commits/8aa74cba4e8edb7108bf64d45c7af7ba7adca292

jansegre (dismissed):

hathor/transaction/aux_pow.py
msbrogli: Perfect! I guess you can mark this conversation as resolved.

hathor/conf/testnet.py
```diff
@@ -54,5 +56,16 @@
     ],
     FEATURE_ACTIVATION=FeatureActivationSettings(
         default_threshold=15_120,  # 15120 = 75% of evaluation_interval (20160)
+        features={
+            Feature.INCREASE_MAX_MERKLE_PATH_LENGTH: Criteria(
+                bit=3,
+                start_height=3_548_160,  # N (right now the best block is 3_510_700 on testnet)
```
msbrogli, line 62: Should we add the expected date and time in the comments?

hathor/feature_activation/feature.py
```diff
@@ -33,3 +33,5 @@ class Feature(Enum):
     NOP_FEATURE_4 = 'NOP_FEATURE_4'
     NOP_FEATURE_5 = 'NOP_FEATURE_5'
     NOP_FEATURE_6 = 'NOP_FEATURE_6'
+
+    INCREASE_MAX_MERKLE_PATH_LENGTH = 'INCREASE_MAX_MERKLE_PATH_LENGTH'
```
msbrogli, line 37: Should we prefix it with `TESTNET_`?

hathor/transaction/aux_pow.py
```diff
@@ -44,23 +41,26 @@ def calculate_hash(self, base_block_hash: bytes) -> bytes:
...
         return sha256d_hash(self.header_head + merkle_root + self.header_tail)
 
-    def verify(self, _base_block_hash: bytes) -> None:
+    def verify(self, _base_block_hash: bytes, max_merkle_path_length: int) -> None:
         """ Check for inconsistencies, raises instance of TxValidationError on error.
         """
+        self.verify_magic_number(_base_block_hash)
+        self.verify_merkle_path(_base_block_hash, max_merkle_path_length)
+
+    def verify_magic_number(self, _base_block_hash: bytes) -> None:
```
msbrogli, line 50: Missing docstring.

hathor/transaction/aux_pow.py
```diff
@@ -44,23 +41,26 @@ def calculate_hash(self, base_block_hash: bytes) -> bytes:
...
-            AuxPowUnexpectedMagicError,
-        )
+        from hathor.transaction.exceptions import AuxPowNoMagicError, AuxPowUnexpectedMagicError
         magic_index = self.coinbase_head.find(MAGIC_NUMBER)
         if magic_index == -1:
             raise AuxPowNoMagicError('cannot find MAGIC_NUMBER')
         if magic_index < len(self.coinbase_head) - len(MAGIC_NUMBER):
             raise AuxPowUnexpectedMagicError('unexpected MAGIC_NUMBER')
+
+    def verify_merkle_path(self, _base_block_hash: bytes, max_merkle_path_length: int) -> None:
```
msbrogli, line 59: Missing docstring.

msbrogli (approved):

hathor/merged_mining/coordinator.py
msbrogli: Great! Thanks!

hathor/transaction/aux_pow.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/910/commits/863db2aaa85a62ae1e6ea60dc5a5175868221da1

hathor/transaction/aux_pow.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/910/commits/863db2aaa85a62ae1e6ea60dc5a5175868221da1

hathor/conf/testnet.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/910/commits/863db2aaa85a62ae1e6ea60dc5a5175868221da1

hathor/feature_activation/feature.py
glevco: We could if you see any advantage, however it's designed to not be necessary. This is simply a "feature identifier", and defining it here doesn't enable anything. To actually configure the process, this identifier has to be referenced in the network's configuration file. In other words, this can be reused for both testnet and mainnet, even if they use completely separate processes.

hathor/conf/testnet.py
msbrogli: Great! Thanks!

hathor/feature_activation/feature.py
msbrogli: Got it. So it makes sense to not use it unless it's something exclusive for a network. Thanks!

hathor/transaction/aux_pow.py
msbrogli: Great! Thanks!

hathor/transaction/aux_pow.py
msbrogli: Great! Thanks!

msbrogli (approved):

jansegre (approved):

PR #911, opened 2024-01-08
feat(sighash): implement sighash bitmask [part 1/7]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/851

### Motivation

Implement the new SIGHASH structure. This PR implements everything necessary for creating P2PKH inputs/outputs using `OP_SIGHASH_BITMASK`. That is, creating atomic swap txs is already fully functional, but gated to prevent a fork (this will later be toggled by Feature Activation).

Adding support for `OP_SIGHASH_RANGE` and MultiSig is straightforward and will be done in separate PRs. Tests for new functionality will also be implemented in a separate PR.

### Acceptance Criteria

- New models `SighashAll`, `SighashBitmask`, `SighashRange`, and `SighashLimit`.
- Implement `evaluate_final_context()` and change `execute_eval()` to call it.
- Change `execute_eval()` and `script_eval()` to return the selected outputs from the context.
- Create new opcodes and implement `op_sighash_bitmask()` and `op_max_inputs_outputs()`.
- Update `op_checksig()` to use the `ScriptContext` for signature verification.
- Implement `is_opcode_valid()` and use it to gate new opcodes in `execute_op_code()`.
- Update `P2PKH.create_input_data()` to support `SighashBitmask` and `SighashLimit`.
- Update `ScriptContext` with methods for interacting with sighash.
- Implement `Transaction.get_custom_sighash_data()` and change `get_sighash_all()` to unify implementations.
- Update `TransactionVerifier.verify_inputs()` to check for unsigned outputs.
- Update existing tests.

### DIFF
--- a/hathor/transaction/exceptions.py
+++ b/hathor/transaction/exceptions.py
@@ -74,6 +74,10 @@ class ConflictingInputs(TxValidationError):
     """Inputs in the tx are spending the same output"""
 
 
+class OutputNotSelected(TxValidationError):
+    """At least one output is not selected for signing by some input."""
+
+
 class TooManyOutputs(TxValidationError):
     """More than 256 outputs"""
 
@@ -202,3 +206,25 @@ class VerifyFailed(ScriptError):
 
 class TimeLocked(ScriptError):
     """Transaction is invalid because it is time locked"""
+
+
+class InputNotSelectedError(ScriptError):
+    """Raised when an input does not select itself for signing in its script."""
+
+
+class MaxInputsExceededError(ScriptError):
+    """The transaction has more inputs than the maximum configured in the script."""
+
+
+class MaxOutputsExceededError(ScriptError):
+    """The transaction has more outputs than the maximum configured in the script."""
+
+
+class InputsOutputsLimitModelInvalid(ScriptError):
+    """
+    Raised when the inputs outputs limit model could not be constructed from the arguments provided in the script.
+    """
+
+
+class CustomSighashModelInvalid(ScriptError):
+    """Raised when the sighash model could not be constructed from the arguments provided in the script."""
--- a/hathor/transaction/scripts/execute.py
+++ b/hathor/transaction/scripts/execute.py
@@ -13,16 +13,32 @@
 #  limitations under the License.
 
 import struct
-from typing import NamedTuple, Optional, Union
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, NamedTuple, Optional, Union
 
+from hathor.conf.get_settings import get_global_settings
 from hathor.transaction import BaseTransaction, Transaction, TxInput
 from hathor.transaction.exceptions import DataIndexError, FinalStackInvalid, InvalidScriptError, OutOfData
 
+if TYPE_CHECKING:
+    from hathor.transaction.scripts.script_context import ScriptContext
 
-class ScriptExtras(NamedTuple):
+
+@dataclass(slots=True, frozen=True, kw_only=True)
+class ScriptExtras:
+    """
+    A simple container for auxiliary data that may be used during execution of scripts.
+    """
     tx: Transaction
-    txin: TxInput
+    input_index: int
     spent_tx: BaseTransaction
+
+    @property
+    def txin(self) -> TxInput:
+        return self.tx.inputs[self.input_index]
+
+    def __post_init__(self) -> None:
+        assert self.txin.tx_id == self.spent_tx.hash
 
 
 # XXX: Because the Stack is a heterogeneous list of bytes and int, and some OPs only work for when the stack has some
@@ -39,7 +55,7 @@ class OpcodePosition(NamedTuple):
     position: int
 
 
-def execute_eval(data: bytes, log: list[str], extras: ScriptExtras) -> None:
+def execute_eval(data: bytes, log: list[str], extras: ScriptExtras) -> 'ScriptContext':
     """ Execute eval from data executing opcode methods
 
         :param data: data to be evaluated that contains data and opcodes
@@ -56,8 +72,9 @@ def execute_eval(data: bytes, log: list[str], extras: ScriptExtras) -> None:
     """
     from hathor.transaction.scripts.opcode import Opcode, execute_op_code
     from hathor.transaction.scripts.script_context import ScriptContext
+    settings = get_global_settings()
     stack: Stack = []
-    context = ScriptContext(stack=stack, logs=log, extras=extras)
+    context = ScriptContext(settings=settings, stack=stack, logs=log, extras=extras)
     data_len = len(data)
     pos = 0
     while pos < data_len:
@@ -69,6 +86,8 @@ def execute_eval(data: bytes, log: list[str], extras: ScriptExtras) -> None:
         execute_op_code(Opcode(opcode), context)
 
     evaluate_final_stack(stack, log)
+
+    return context
 
 
 def evaluate_final_stack(stack: Stack, log: list[str]) -> None:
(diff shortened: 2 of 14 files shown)


hathor/transaction/exceptions.py
glevco, line 1: @msbrogli as you had already reviewed this PR in https://github.com/HathorNetwork/sec-hathor-core/pull/125, I'm hijacking this thread here so we can keep track of the threads in the old PR. I'll keep this thread open until we resolve all of them.

hathor/transaction/transaction.py
```diff
@@ -217,29 +218,34 @@ def get_sighash_all(self) -> bytes:
         if self._sighash_cache:
             return self._sighash_cache
 
+        sighash = self._get_sighash(inputs=self.inputs, outputs=self.outputs)
+        self._sighash_cache = sighash
```
msbrogli, line 222: Maybe rename it to `_sighash_all_cache`?

hathor/verification/transaction_verifier.py
msbrogli, line 1: I couldn't find where we enforce that at least one sighash all must be used.

msbrogli (changes requested):

hathor/transaction/scripts/opcode.py
```diff
@@ -617,6 +628,59 @@ def op_integer(opcode: int, stack: Stack) -> None:
...
+    assert isinstance(inputs, bytes)
+    assert isinstance(outputs, bytes)
+
+    try:
+        sighash = SighashBitmask(
+            inputs=bytes_to_int(inputs),
+            outputs=bytes_to_int(outputs)
+        )
+    except Exception as e:
+        raise CustomSighashModelInvalid('Could not construct sighash bitmask.') from e
```
msbrogli, line 647: How can we differentiate a code bug from a bad input? I feel that we should capture only the exceptions raised by the `bytes_to_int()` and let all others blow up.

hathor/transaction/scripts/opcode.py
```diff
@@ -617,6 +628,59 @@ def op_integer(opcode: int, stack: Stack) -> None:
...
+    max_inputs = context.stack.pop()
+    assert isinstance(max_inputs, bytes)
+    assert isinstance(max_outputs, bytes)
+
+    try:
+        limit = InputsOutputsLimit(
+            max_inputs=bytes_to_int(max_inputs),
+            max_outputs=bytes_to_int(max_outputs)
+        )
+    except Exception as e:
```
msbrogli, line 672: Same here.

hathor/transaction/scripts/opcode.py
```diff
@@ -617,6 +628,59 @@ def op_integer(opcode: int, stack: Stack) -> None:
...
+    tx_inputs_len = len(context.extras.tx.inputs)
+    if tx_inputs_len > limit.max_inputs:
+        raise MaxInputsExceededError(f'Maximum number of inputs exceeded ({tx_inputs_len} > {limit.max_inputs}).')
+
+    tx_outputs_len = len(context.extras.tx.outputs)
+    if tx_outputs_len > limit.max_outputs:
+        raise MaxOutputsExceededError(f'Maximum number of outputs exceeded ({tx_outputs_len} > {limit.max_outputs}).')
+
+
 def execute_op_code(opcode: Opcode, context: ScriptContext) -> None:
```
msbrogli, line 684: Why not have a list of available upcodes in the context?

hathor/transaction/exceptions.py
msbrogli: All threads have been resolved there. Thanks for the reminder.

tests/tx/test_scripts.py
```diff
@@ -176,22 +176,22 @@ def test_pushdata1(self):
 
     def test_dup(self):
         with self.assertRaises(MissingStackItems):
-            op_dup(ScriptContext(stack=[], logs=[], extras=Mock()))
+            op_dup(ScriptContext(stack=[], logs=[], extras=Mock(), settings=Mock()))
```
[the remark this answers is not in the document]
glevco: Simply because this opcode doesn't use the settings, so it doesn't really matter what value is there. I think using a `Mock()` semantically represents that this is not going to be used.

hathor/verification/transaction_verifier.py
glevco: This is implemented in Part 7: https://github.com/HathorNetwork/hathor-core/pull/917

hathor/transaction/scripts/opcode.py
glevco: I think it would convey a semantic that available opcodes could be different depending on execution, and that opcode availability could somehow change between different executions.

Instead, opcode availability is defined by whether it has been activated or not (through Feature Activation), and that is defined in the code, not in the context during execution.

hathor/transaction/scripts/execute.py
```diff
@@ -13,16 +13,21 @@
...
 
+if TYPE_CHECKING:
+    from hathor.transaction.scripts.script_context import ScriptContext
+
 
 class ScriptExtras(NamedTuple):
     tx: Transaction
     txin: TxInput
     spent_tx: BaseTransaction
+    input_index: int
```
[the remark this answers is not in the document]
glevco: I have changed it to a validated dataclass in https://github.com/HathorNetwork/hathor-core/pull/911/commits/b6112976da77aa63fb70f2eb22b1616ea933eecd, so instead of a note, I assert that an invalid instance cannot be created.

hathor/transaction/transaction.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/911/commits/b6112976da77aa63fb70f2eb22b1616ea933eecd

hathor/transaction/scripts/opcode.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/911/commits/b6112976da77aa63fb70f2eb22b1616ea933eecd to catch only `pydantic.ValidationError`. The `bytes_to_int()` function does not raise any exceptions explicitly, so the only expected exception to be raised here is the model validation from pydantic.

However, I believe this is a shortcoming caused by the try-except model itself. There's no way to formally differentiate between exceptions caused by bugs or bad inputs, because there's no formally typed list of expected exceptions. In my opinion we should improve this by using result monads throughout the code.

hathor/transaction/scripts/opcode.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/commit/b6112976da77aa63fb70f2eb22b1616ea933eecd, see https://github.com/HathorNetwork/hathor-core/pull/911#discussion_r1470406237

hathor/transaction/scripts/execute.py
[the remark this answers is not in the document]
msbrogli: Great! Thanks!

tests/tx/test_scripts.py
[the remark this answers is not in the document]
msbrogli: Got it. Ok. Thanks!

hathor/transaction/transaction.py
msbrogli: Thanks!

hathor/verification/transaction_verifier.py
msbrogli: Ok! Thanks for the clarification.

hathor/transaction/scripts/opcode.py
msbrogli: Makes sense. Thanks!

jansegre (dismissed):

hathor/transaction/scripts/execute.py
```diff
@@ -88,7 +107,7 @@ def evaluate_final_stack(stack: Stack, log: list[str]) -> None:
         raise FinalStackInvalid('\n'.join(log))
 
 
-def script_eval(tx: Transaction, txin: TxInput, spent_tx: BaseTransaction) -> None:
+def script_eval(tx: Transaction, spent_tx: BaseTransaction, *, input_index: int) -> 'ScriptContext':
```
msbrogli, line 110: As you're already changing the method signature, you can put it all as kwargs only, right?

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/911/commits/2f8579f7c83882245a077f7d354f97457d51925e

hathor/verification/transaction_verifier.py
```diff
@@ -121,7 +125,9 @@ def verify_inputs(self, tx: Transaction, *, skip_script: bool = False) -> None:
                 ))
 
             if not skip_script:
-                self.verify_script(tx=tx, input_tx=input_tx, spent_tx=spent_tx)
+                script_context = self.verify_script(tx=tx, spent_tx=spent_tx, input_index=input_index)
+                selected_outputs = script_context.get_selected_outputs()
+                all_selected_outputs = all_selected_outputs.union(selected_outputs)
```
msbrogli, line 130: You can use `all_selected_outputs.update(selected_outputs)` to prevent creating a new set.

msbrogli (approved):

hathor/verification/transaction_verifier.py
glevco: Done

jansegre (approved):

Issue #125, opened 2020-07-30
Ops: Implement consensus check


msbrogli:
Design: https://github.com/HathorNetwork/hathor-core/issues/98

jansegre:
I think Luis has implemented something like this already. Can you confirm? @luislhl

PR #912, opened 2024-01-08
tests(sighash): add sighash bitmask tests [part 2/7]

glevco:
Depends on https://github.com/HathorNetwork/sec-hathor-core/pull/125

### Motivation

The previous PR introduced the structure for sighash types, including sighash bitmask for P2PKH scripts. It fixed existing tests, but did not introduce any new tests. This PR does this.

### Acceptance Criteria

- Implement new tests for features introduced in PR #125


tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        atomic_swap_tx = Transaction(
+            weight=1,
+            inputs=[tokens_input],
+            outputs=[htr_output],
+            parents=parents,
+            tokens=[token_uid],
+            storage=self.manager1.tx_storage,
+            timestamp=token_creation_tx.timestamp + 1
+        )
+        self.manager1.cpu_mining_service.resolve(atomic_swap_tx)
```
msbrogli, line 83: Why are you resolving the tx here if it's gonna be modified just after this point?

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        assert self.manager1.wallet
+        public_bytes1, signature1 = self.manager1.wallet.get_input_aux_data(data_to_sign1, self.private_key1)
+        tokens_input.data = P2PKH.create_input_data(
+            public_key_bytes=public_bytes1,
+            signature=signature1,
+            sighash=sighash_bitmask,
+        )
+
+        # At this point, the tx is partial. The inputs are valid, but they're mismatched with outputs
+        self.manager1.verification_service.verifiers.tx.verify_inputs(atomic_swap_tx)
```
msbrogli, line 97: Shouldn't it fail since there's no sighash all?

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        with pytest.raises(InputOutputMismatch):
+            self.manager1.verification_service.verify(atomic_swap_tx)
+
+        # Alice sends the tx bytes to Bob, represented here by cloning the tx
+        atomic_swap_tx_clone = cast(Transaction, atomic_swap_tx.clone())
+        self.manager1.cpu_mining_service.resolve(atomic_swap_tx_clone)
```
msbrogli, line 103: We just have to resolve the transaction before pushing it to the network. Any previous solution is a waste of resources.

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        atomic_swap_tx = Transaction(
+            weight=1,
+            inputs=[tokens_input],
+            outputs=[htr_output],
+            parents=parents,
+            tokens=[token_uid],
+            storage=self.manager1.tx_storage,
+            timestamp=token_creation_tx.timestamp + 1
+        )
+        self.manager1.cpu_mining_service.resolve(atomic_swap_tx)
```
msbrogli, line 154: Same here.

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        )
+
+        # At this point, the tx is partial. The inputs are valid, but they're mismatched with outputs
+        self.manager1.verification_service.verifiers.tx.verify_inputs(atomic_swap_tx)
+        with pytest.raises(InputOutputMismatch):
+            self.manager1.verification_service.verify(atomic_swap_tx)
+
+        # Alice sends the tx bytes to Bob, represented here by cloning the tx
+        atomic_swap_tx_clone = cast(Transaction, atomic_swap_tx.clone())
+        self.manager1.cpu_mining_service.resolve(atomic_swap_tx_clone)
```
msbrogli, line 176: Same here.

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        htr_input = TxInput(not_none(self.genesis_block.hash), 0, b'')
+        atomic_swap_tx_clone.inputs.append(htr_input)
+
+        # Bob adds an output to receive all test tokens
+        bob_output_script = P2PKH.create_output_script(self.address2)
+        tokens_output = TxOutput(token_creation_utxo.value, bob_output_script, 1)
+        atomic_swap_tx_clone.outputs.append(tokens_output)
+
+        # Bob adds two change outputs for his HTR, which violates the maximum tx outputs set by Alice
+        htr_output1 = TxOutput(int(genesis_utxo.value / 4), bob_output_script)
```
msbrogli, line 188: Nitpick: You can use integer division `genesis_utxo.value // 4`.

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        atomic_swap_tx_clone.outputs.append(htr_output2)
+
+        # Bob signs his input using sighash_all to complete the tx
+        data_to_sign2 = atomic_swap_tx_clone.get_sighash_all()
+        assert self.manager2.wallet
+        public_bytes2, signature2 = self.manager2.wallet.get_input_aux_data(data_to_sign2, self.genesis_private_key)
+        htr_input.data = P2PKH.create_input_data(public_bytes2, signature2)
+
+        # The atomic swap tx is not valid and cannot be propagated
+        with pytest.raises(InvalidInputData) as e:
```
msbrogli, line 200: Should we create a specific exception for this case? It can inherit from `InvalidInputData`.

tests/tx/scripts/test_sighash.py
```diff
@@ -0,0 +1,321 @@
...
+        # Alice signs her genesis input using the same sighash, so the genesis input is not selected in the bitmask.
+        public_bytes1, signature1 = self.manager1.wallet.get_input_aux_data(data_to_sign1, self.genesis_private_key)
+        genesis_input.data = P2PKH.create_input_data(
+            public_key_bytes=public_bytes1,
+            signature=signature1,
+            sighash=sighash_bitmask,
+        )
+
+        # The inputs are invalid, since one of them doesn't select itself.
+        with pytest.raises(InvalidInputData) as e:
```
msbrogli, line 253: Shouldn't we create a specific exception for this case?

tests/tx/test_scripts.py
```diff
@@ -972,3 +982,105 @@ def test_get_sigops_count(self):
...
         self.assertEqual(get_sigops_count(input_script.data), 1)
+
+    def test_op_sighash_bitmask(self) -> None:
+        with self.assertRaises(MissingStackItems):
+            op_sighash_bitmask(ScriptContext(stack=[], extras=Mock(), logs=[], settings=Mock()))
+
+        with self.assertRaises(MissingStackItems):
+            op_sighash_bitmask(ScriptContext(stack=[b''], extras=Mock(), logs=[], settings=Mock()))
+
+        with self.assertRaises(AssertionError):
```
msbrogli, line 993: Why would it be an assertion error? Is it because they are not bytes?

tests/tx/test_scripts.py
```diff
@@ -972,3 +982,105 @@ def test_get_sigops_count(self):
...
+        stack: list[bytes | int | str] = [bytes([0b0]), bytes([0xFF, 0xFF])]
+        context = Mock(spec_set=ScriptContext)
+        extras = Mock(spec_set=ScriptExtras)
+        context.stack = stack
+        context.extras = extras
+
+        with self.assertRaises(CustomSighashModelInvalid):
+            op_sighash_bitmask(context)
+
+        context.stack = [bytes([0b111]), bytes([0b101])]
```
msbrogli, line 1005: Maybe add tests with different endianness.

tests/tx/test_scripts.py
```diff
@@ -972,3 +982,105 @@ def test_get_sigops_count(self):
...
+        context.extras.tx.outputs = ['a', 'b']
+
+        op_max_inputs_outputs(context)
+
+        self.assertEqual(context.stack, [])
+
+    def test_execute_op_code(self) -> None:
+        with (
+            patch('hathor.transaction.scripts.opcode.is_opcode_valid', lambda _: False),
+            self.assertRaises(ScriptError)
```
msbrogli, line 1067: Should it be an `INVALID_OPCODE` or `UNKNOWN_OPCODE`?

tests/tx/test_scripts.py
```diff
@@ -972,3 +982,105 @@ def test_get_sigops_count(self):
...
+        ):
+            execute_op_code(opcode=Mock(), context=Mock())
+
+        with (
+            patch('hathor.transaction.scripts.opcode.is_opcode_valid', lambda _: True),
+            self.assertRaises(ScriptError)
+        ):
+            execute_op_code(opcode=Opcode.OP_0, context=Mock())
```
msbrogli, line 1075: What are we testing here?

tests/tx/test_scripts.py
```diff
@@ -972,3 +982,105 @@ def test_get_sigops_count(self):
...
+
+        with patch('hathor.transaction.scripts.opcode.op_dup') as op_mock:
+            execute_op_code(opcode=Opcode.OP_DUP, context=Mock())
```
msbrogli, line 1078: What are we testing here?

msbrogli (changes requested):

tests/tx/scripts/__init__.py
msbrogli, line 1: Maybe add tests to make sure that a sighash all will mark all inputs and outputs as selected at least once.

tests/tx/scripts/test_script_context.py
```diff
@@ -0,0 +1,95 @@
...
+        context.set_sighash(sighash)
+
+
+@pytest.mark.parametrize(
+    ['outputs_bitmask', 'selected_outputs'],
+    [
+        (0b00, set()),
+        (0b01, {0}),
+        (0b10, {1}),
+        (0b11, {0, 1}),
```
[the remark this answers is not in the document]
glevco: You mean testing `0b111`, for example? That would raise an exception, as the test tx only has 2 outputs. This error is checked in `test_sighash.py`.

tests/tx/scripts/test_sighash.py
glevco: Because the tx is going to be verified on line 99, so it needs to be resolved.

tests/tx/scripts/test_sighash.py
glevco: This requirement is only implemented in Part 7: https://github.com/HathorNetwork/hathor-core/pull/917

tests/tx/scripts/test_sighash.py
glevco: Same as above, the tx is verified on line 125, so it need to be resolved

tests/tx/scripts/test_sighash.py
glevco: the tx is verified on line 172, so it needs to be resolved

tests/tx/scripts/test_sighash.py
glevco: The tx is verified on line 201, so it needs to be resolved

tests/tx/scripts/test_sighash.py
glevco: Same as above, please see https://github.com/HathorNetwork/hathor-core/pull/912#discussion_r1470447371

tests/tx/test_scripts.py
glevco: Exactly!

tests/tx/test_scripts.py
glevco: What do you mean? Endianness has no meaning in this context as this string of bits is never interpreted as a number, it's simply a string of bits

tests/tx/scripts/__init__.py
glevco: This is implemented in Part 7: https://github.com/HathorNetwork/hathor-core/pull/917/files

tests/tx/scripts/test_script_context.py
```diff
@@ -0,0 +1,95 @@
...
+
+from hathor.transaction import Transaction, TxInput, TxOutput
+from hathor.transaction.exceptions import ScriptError
+from hathor.transaction.scripts.script_context import ScriptContext
+from hathor.transaction.scripts.sighash import SighashAll, SighashBitmask
+
+
+def test_defaults() -> None:
+    settings = Mock()
+    settings.MAX_NUM_OUTPUTS = 99
```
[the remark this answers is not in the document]
glevco: I think doing it like this asserts that the code is correctly retrieving information from settings, instead of having a hardcoded value that happens to be the same as the one we have in settings. I added a test for the current value in https://github.com/HathorNetwork/hathor-core/pull/912/commits/2bf75499d3e715b87646a1b5ec0f018d4655847c, though

tests/tx/scripts/test_sighash.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/912/commits/2bf75499d3e715b87646a1b5ec0f018d4655847c

tests/tx/test_scripts.py
glevco: That when `is_opcode_valid` returns `True`, execution must fail if it's not a "function opcode". I added a comment for it in https://github.com/HathorNetwork/hathor-core/pull/912/commits/2bf75499d3e715b87646a1b5ec0f018d4655847c

tests/tx/test_scripts.py
glevco: That the function associated with a valid opcode is correctly called. I added a comment for it in https://github.com/HathorNetwork/hathor-core/pull/912/commits/2bf75499d3e715b87646a1b5ec0f018d4655847c

tests/tx/scripts/test_p2pkh.py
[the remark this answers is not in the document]
glevco: The script regex is only made to match output scripts, isn't it? It wouldn't match any of those input data

tests/tx/scripts/test_sighash.py
msbrogli: But you are modifying the transaction on line 90, so the hash will become invalid again. Am I missing something?

tests/tx/scripts/test_sighash.py
msbrogli: My point is that this exception might be generated by another verification (instead of the one you are expecting to fail).

tests/tx/test_scripts.py
msbrogli: I meant that `0b1101` should be different from `0b1011`.

tests/tx/test_scripts.py
msbrogli: I got it. But shouldn't we capture the more specific exception? We have to be very careful in these changes because it's a critical part of the full node.

msbrogli (approved):

tests/tx/scripts/test_sighash.py
glevco: It passes the verification like this anyway. But I moved it closer to the verification line in https://github.com/HathorNetwork/hathor-core/pull/912/commits/9147127aa7f05fec78fefec2a4cd112833a589d2

tests/tx/scripts/test_sighash.py
glevco: Moved to be closer to the verification line in https://github.com/HathorNetwork/hathor-core/pull/912/commits/9147127aa7f05fec78fefec2a4cd112833a589d2

tests/tx/scripts/test_sighash.py
glevco: Moved to be closer to the verification line in https://github.com/HathorNetwork/hathor-core/pull/912/commits/9147127aa7f05fec78fefec2a4cd112833a589d2

tests/tx/scripts/test_sighash.py
glevco: Moved to be closer to the verification line in https://github.com/HathorNetwork/hathor-core/pull/912/commits/9147127aa7f05fec78fefec2a4cd112833a589d2

tests/tx/test_scripts.py
glevco: Ah ok, done in https://github.com/HathorNetwork/hathor-core/pull/912/commits/75790a3c8ae93a735588360e1985e9fbda4ade8f

tests/tx/test_scripts.py
glevco: Got it, I added assertions for the exception message in https://github.com/HathorNetwork/hathor-core/pull/912/commits/75790a3c8ae93a735588360e1985e9fbda4ade8f

PR #913, opened 2024-01-08
feat(sighash): implement sighash range  [part 3/7]

glevco:
Depends on https://github.com/HathorNetwork/sec-hathor-core/pull/126

### Motivation

Implement the new sighash range opcode.

### Acceptance Criteria

- Implement `SighashRange` model.
- Implement `op_sighash_range()` function.
- Add sighash range support for P2PKH input scripts.


hathor/transaction/scripts/opcode.py
```diff
@@ -655,6 +655,38 @@ def op_sighash_bitmask(context: ScriptContext) -> None:
...
+    assert isinstance(input_start, bytes)
+
+    try:
+        sighash = SighashRange(
+            input_start=bytes_to_int(input_start),
+            input_end=bytes_to_int(input_end),
+            output_start=bytes_to_int(output_start),
+            output_end=bytes_to_int(output_end),
+        )
+    except Exception as e:
```
msbrogli, line 679: Why are you capturing `Exception` instead of the specific exception types?

hathor/transaction/scripts/sighash.py
```diff
@@ -60,7 +60,37 @@ def _get_indexes(bitmask: int) -> list[int]:
         return [index for index in range(8) if (bitmask >> index) & 1]
 
 
-SighashType: TypeAlias = SighashAll | SighashBitmask
+class SighashRange(CustomSighash):
+    """A model representing the sighash range type config. Range ends are not inclusive."""
```
msbrogli, line 64: Shouldn't range ends be included? I feel it's more intuitive than not including them.

msbrogli (changes requested):

hathor/transaction/scripts/opcode.py
glevco: I changed it as explained in https://github.com/HathorNetwork/hathor-core/pull/911#discussion_r1470406237

hathor/transaction/scripts/sighash.py
glevco: I think it's intuitive like this when you include the whole range, which would be from 0 to 255, which is the max number of inputs/outputs. If we do it inclusive, the user would have to set 254 to include the whole range. I can change it if you want.

PR #914, opened 2024-01-08
feat(sighash): implement sighash for MultiSig [part 4/7]

glevco:
Depends on https://github.com/HathorNetwork/sec-hathor-core/pull/127

### Motivation

Add custom sighash support for MultiSig scripts.

### Acceptance Criteria

- Move code from `P2PKH` to `HathorScript`, generalizing it (`push_sighash()` and `push_inputs_outputs_limit()`).
- Add support for sighash and limit on `MultiSig.create_input_data()`.
- Create new utils function `generate_signature_for_data()` for MultiSig.
- Fix issue caused by unused context when `op_checkmultisig()` calls `op_checksig()`.


hathor/transaction/scripts/hathor_script.py
```diff
@@ -49,3 +58,30 @@ def pushData(self, data: Union[int, bytes]) -> None:
...
+            case SighashRange():
+                self.pushData(sighash.input_start)
+                self.pushData(sighash.input_end)
+                self.pushData(sighash.output_start)
+                self.pushData(sighash.output_end)
+                self.addOpcode(Opcode.OP_SIGHASH_RANGE)
+            case _:
+                assert_never(sighash)
+
+    def push_inputs_outputs_limit(self, limit: InputsOutputsLimit | None) -> None:
```
msbrogli, line 80: It seems weird to accept None and simply do nothing. I guess the caller should take care of skipping this call.

hathor/transaction/scripts/opcode.py
```diff
@@ -591,21 +591,25 @@ def op_checkmultisig(context: ScriptContext) -> None:
...
     # Signatures must be in order (same as the public keys in the multi sig wallet)
     pubkey_index = 0
+    old_stack = context.stack
     for signature in signatures:
         while pubkey_index < len(pubkeys):
             pubkey = pubkeys[pubkey_index]
             new_stack = [signature, pubkey]
-            op_checksig(ScriptContext(stack=new_stack, logs=context.logs, extras=context.extras, settings=settings))
+            context.stack = new_stack
+            op_checksig(context)
```
msbrogli, line 600: I guess that's the bug fix, right? I'd like more information about the bug, its origin, and the fix.

tests/tx/scripts/test_sighash_range.py
```diff
@@ -125,6 +127,101 @@ def test_sighash_range(self) -> None:
...
+        parents = self.manager1.get_new_tx_parents()
+
+        public_keys = [
+            bytes.fromhex('0250bf5890c9c6e9b4ab7f70375d31b827d45d0b7b4e3ba1918bcbe71b412c11d7'),
+            bytes.fromhex('02d83dd1e9e0ac7976704eedab43fe0b79309166a47d70ec3ce8bbb08b8414db46'),
+            bytes.fromhex('02358c539fa7474bf12f774749d0e1b5a9bc6e50920464818ebdb0043b143ae2ba'),
+        ]
+
+        private_keys = [
+            '3081de304906092a864886f70d01050d303c301b06092a864886f70d01050c300e04089abeae5e8a8f75d302020800301d060960864801650304012a0410abbde27221fd302280c13fca7887c85e048190c41403f39b1e9bbc5b6b7c3be4729c054fae9506dc0f8361adcff0ea393f0bb3ca9f992fc2eea83d532691bc9a570ed7fb9e939e6d1787881af40b19fb467f06595229e29b5a6268d831f0287530c7935d154deac61dd4ced988166f9c98054912935b607e2fb332e11c95b30ea4686eb0bda7dd57ed1eeb25b07cea9669dde5210528a00653159626a5baa61cdee7f4',  # noqa: E501
```
msbrogli, line 146: You can break this huge line in smaller lines.

msbrogli (changes requested):

hathor/transaction/scripts/hathor_script.py
glevco: Ok, done in https://github.com/HathorNetwork/hathor-core/pull/914/commits/bde5cc4116610fc2ca447833bdb72197c6588966

tests/tx/scripts/test_sighash_range.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/914/commits/bde5cc4116610fc2ca447833bdb72197c6588966

hathor/transaction/scripts/opcode.py
glevco: It's not a preexisting bug, it's just an issue caused by the new sighash features.

`op_checkmultisig` calls `op_checksig` for each signature-pubkey pair, trying to find matches. Before, it would call `op_checksig` with a new ad-hoc `ScriptContext`, with a new stack.

Now, since the sighash configuration is persisted in the `ScriptContext` by the new sighash opcodes, this information was missing from this ad-hoc context passed to `op_checksig`. So I simply changed the code to reuse the same context, only changing the stack (and later changing it back to the original one).

The behavior is exactly the same as before, except now the same context is reused, guaranteeing the sighash config is considered.

PR #915, opened 2024-01-08
feat(sighash): implement OP_MAX_SIGHASH_SUBSETS [part 5/7]

glevco:
Depends on https://github.com/HathorNetwork/sec-hathor-core/pull/128

### Motivation

Implement new `OP_MAX_SIGHASH_SUBSETS`.

### Acceptance Criteria

- Create new `OP_MAX_SIGHASH_SUBSETS` and implement `op_max_sighash_subsets()`.
- Implement ability to push a `OP_MAX_SIGHASH_SUBSETS` in `HathorScript`, adding support in P2PKH and MultiSig.
- Implement `TooManySighashSubsets` verification in `TransactionVerifier`.


hathor/transaction/scripts/opcode.py
```diff
@@ -82,7 +82,8 @@ class Opcode(IntEnum):
     OP_DATA_MATCH_VALUE = 0xD1
     OP_SIGHASH_BITMASK = 0xE0
     OP_SIGHASH_RANGE = 0xE1
-    OP_MAX_INPUTS_OUTPUTS = 0xE2
+    OP_MAX_SIGHASH_SUBSETS = 0xE2
+    OP_MAX_INPUTS_OUTPUTS = 0xE3
```
msbrogli, line 86: Just to double check: It is safe to change the OPCODE because these opcodes haven't been activated in any network yet. Is that correct?

tests/tx/test_tx.py
```diff
@@ -184,6 +185,37 @@ def test_output_not_selected(self) -> None:
 
         self.assertEqual(str(e.value), "Output at index 0 is not signed by any input.")
 
+    @patch('hathor.transaction.scripts.opcode.is_opcode_valid', lambda _: True)
+    def test_too_many_sighash_subsets(self) -> None:
```
msbrogli, line 189: It seems this test covers only the sighash all. What about other possible subsets? Or even subsets generated by different sighash strategies (e.g., bitmask vs range)?

msbrogli (changes requested):

hathor/transaction/scripts/opcode.py
glevco: Exactly.

tests/tx/test_tx.py
glevco: I added a new more complex test case in https://github.com/HathorNetwork/hathor-core/pull/915/commits/44f39cf901aaad08ac481c4f9ce3f06870910fd9

PR #922, opened 2024-01-12
refactor(storage): create StorageProtocol

glevco:
### Motivation

To externalize storage dependencies from the verification process, we need to inject a storage into some methods used in verification, so it can either receive a `TransactionStorage` (rocksdb), or a simple memory storage that will be implemented in next PRs. To unify both of them in a single interface, a Python protocol is introduced, `StorageProtocol`.

### Acceptance Criteria

- Create `StorageProtocol`, a protocol for unifying a subset of storage methods.
- Add methods to `TransactionStorage` so it conforms to the `StorageProtocol`.
- Update DAA so its calculation depends on a `StorageProtocol`, and update usages accordingly.
- Update Reward Lock `Transaction` methods so they depend on a `StorageProtocol`, and update usages accordingly.

### Checklist

### DIFF
--- a/hathor/transaction/transaction.py
+++ b/hathor/transaction/transaction.py
@@ -15,12 +15,13 @@
 import hashlib
 from itertools import chain
 from struct import pack
-from typing import TYPE_CHECKING, Any, Iterator, NamedTuple, Optional
+from typing import TYPE_CHECKING, Any, NamedTuple, Optional
 
 from hathor.checkpoint import Checkpoint
 from hathor.exception import InvalidNewTransaction
 from hathor.profiler import get_cpu_profiler
-from hathor.transaction import BaseTransaction, Block, TxInput, TxOutput, TxVersion
+from hathor.reward_lock import iter_spent_rewards
+from hathor.transaction import BaseTransaction, TxInput, TxOutput, TxVersion
 from hathor.transaction.base_transaction import TX_HASH_SIZE
 from hathor.transaction.exceptions import InvalidToken
 from hathor.transaction.util import VerboseCallback, unpack, unpack_len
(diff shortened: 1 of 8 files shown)


hathor/transaction/storage/storage_protocol.py
```diff
@@ -0,0 +1,41 @@
...
+#  limitations under the License.
+
+from abc import abstractmethod
+from typing import Protocol
+
+from hathor.transaction import BaseTransaction, Block
+from hathor.types import VertexId
+
+
+class StorageProtocol(Protocol):
```
jansegre, line 22: I think there could be some documentation here explaining at least a little bit of the motivation and some of the restrictions of using the interface. I'm not completely sure, but I would guess that objects returned from whatever class implements this protocol, might not necessarily have the `object.storage` pointer.

jansegre (approved):

hathor/transaction/storage/storage_protocol.py
```diff
@@ -0,0 +1,48 @@
...
```
msbrogli, line 22: `VertexStorageProtocol` or `VertexReducedStorageProtocol`

hathor/transaction/transaction.py
msbrogli, line 1: I think we should use this opportunity to move these methods to another class. So they would never be able to access the `self.storage` attribute from the storage by mistake (and any attempt would be raised by mypy).

msbrogli (changes requested):

hathor/transaction/storage/storage_protocol.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/922/commits/edcaf5693c5460ba64649640636c49d5afbf0aa4

hathor/transaction/transaction.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/922/commits/edcaf5693c5460ba64649640636c49d5afbf0aa4

jansegre (dismissed):

msbrogli (approved):

jansegre (approved):

msbrogli (approved):

hathor/transaction/storage/storage_protocol.py
msbrogli: Thanks!

hathor/transaction/transaction.py
msbrogli: Thanks!

PR #924, opened 2024-01-15
feat(cli): improve handling of SIGQUIT and SIGINT

jansegre:
### Motivation

Before this change we had these two not directly related but of similar nature:

- sending two SIGINT (Ctrl+C) in a row, could (depending on the timing) cause the second one to fail because the node is already existing
- sending a SIGQUIT causes the node to quit "abruptly" and not able to resume

### Acceptance Criteria

- Introduce `HathorManager.try_stop` which does not fail if it's called multiple times, and use it for `SIGINT` and `SIGQUIT`
- Add explicit handler for `SIGQUIT` and make sure to stop the reactor

### Checklist

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -231,12 +231,28 @@ class RunNode:
         if sigusr1 is not None:
             # USR1 is available in this OS.
             signal.signal(sigusr1, self.signal_usr1_handler)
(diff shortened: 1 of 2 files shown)


jansegre:
After some testing. It doesn't seem that these changes make any difference in practice, so I'm closing this.

PR #927, opened 2024-01-16
fix(feature-activation): add missing metadata migration

glevco:
### Motivation

The feature activation bit counts metadata would be empty causing a max recursion error due to a change caused by https://github.com/HathorNetwork/hathor-core/pull/785.

### Acceptance Criteria

- Add bit counts calculation in `update_initial_metadata()`.
- Add `add_feature_activation_bit_counts_metadata2` migration.
- Change `add_feature_activation_bit_counts_metadata` migration so it does nothing.

### Checklist

### DIFF
--- a/hathor/transaction/storage/migrations/add_feature_activation_bit_counts_metadata2.py
+++ b/hathor/transaction/storage/migrations/add_feature_activation_bit_counts_metadata2.py
@@ -17,6 +17,7 @@ from typing import TYPE_CHECKING
 from structlog import get_logger
 
 from hathor.transaction.storage.migrations import BaseMigration
+from hathor.util import progress
 
 if TYPE_CHECKING:
     from hathor.transaction.storage import TransactionStorage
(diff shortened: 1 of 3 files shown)


jansegre (approved):

hathor/transaction/storage/migrations/add_feature_activation_bit_counts_metadata2.py
```diff
@@ -0,0 +1,41 @@
...
+from hathor.transaction.storage.migrations import BaseMigration
+from hathor.util import progress
+
+if TYPE_CHECKING:
+    from hathor.transaction.storage import TransactionStorage
+
+logger = get_logger()
+
+
+class Migration(BaseMigration):
```
msbrogli, line 28: Should we add a description in the docstring? If I got it right, you skipped the previous migration because it could have already been run and you need to make sure it will run again. Is that correct?

msbrogli (approved):

hathor/transaction/storage/migrations/add_feature_activation_bit_counts_metadata2.py
glevco: > If I got it right, you skipped the previous migration because it could have already been run and you need to make sure it will run again. Is that correct?

Exactly. In fact, all of this is going to be removed after your PR: https://github.com/HathorNetwork/hathor-core/pull/572#discussion_r1459639106

msbrogli: Got it. Thanks!

PR #931, opened 2024-01-23
chore(feature-activation): phased testing adjustments

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/927

### Motivation

Clean up Phased Testing artifacts from testnet, so we can update its Evaluation Interval from 2 weeks to 1 week, making it the same as mainnet.

### Acceptance Criteria

- Remove NOP feature configs from testnet config files.
- Update the testnet Evaluation Interval from 40320 to 20160 blocks (1 week), so it's the same as mainnet.
- Update the testnet default threshold to 15120 blocks (75% of the Evaluation Interval).
- Fix NOP feature logging in `HathorManager` to support mainnet features.
- Add a migration to clean testnet Phased Testing metadata.

### Checklist

### DIFF
--- a/hathor/conf/testnet.py
+++ b/hathor/conf/testnet.py
@@ -14,8 +14,6 @@
 
 from hathor.checkpoint import Checkpoint as cp
 from hathor.conf.settings import HathorSettings
-from hathor.feature_activation.feature import Feature
-from hathor.feature_activation.model.criteria import Criteria
 from hathor.feature_activation.settings import Settings as FeatureActivationSettings
 
 SETTINGS = HathorSettings(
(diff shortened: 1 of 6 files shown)


jansegre (approved):

hathor/transaction/storage/migrations/remove_second_nop_features.py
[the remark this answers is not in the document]
glevco: That's a good idea! It would be good to have my PRs (this one and https://github.com/HathorNetwork/hathor-core/pull/927) merged before, then. So we can remove all migrations

hathor/manager.py
```diff
@@ -1077,7 +1077,7 @@ def _log_feature_states(self, vertex: BaseTransaction) -> None:
             features_states=state_by_feature
         )
 
-        features = [Feature.NOP_FEATURE_4, Feature.NOP_FEATURE_5, Feature.NOP_FEATURE_6]
+        features = [Feature.NOP_FEATURE_1, Feature.NOP_FEATURE_2]
```
msbrogli, line 1080: Why this change?

hathor/conf/testnet.yml
```diff
@@ -37,34 +37,4 @@ CHECKPOINTS:
   1_600_000: 00000000060adfdfd7d488d4d510b5779cf35a3c50df7bcff941fbb6957be4d2
 
 FEATURE_ACTIVATION:
-  evaluation_interval: 40_320
```
msbrogli, line 40: I guess it's missing the `evaluation_internal: 20160`, isn't it?

hathor/conf/testnet.py
```diff
@@ -55,36 +53,6 @@
...
-              bit=2,
-              start_height=3_386_880,  # N (right now the best block is 3_346_600 on testnet)
-              timeout_height=3_467_520,  # N + 2 * 40320 (4 weeks after the start)
-              minimum_activation_height=0,
-              lock_in_on_timeout=False,
-              version='0.57.0',
-              signal_support_by_default=False
-            )
-        }
+        default_threshold=15_120,  # 15120 = 75% of evaluation_interval (20160)
```
msbrogli, line 56: I guess it's missing the `evaluation_internal=20_160`, isn't it?

msbrogli (dismissed):

hathor/manager.py
glevco: This is simply limiting which features are logged when they're `ACTIVE`, for the phased testing. Before, it was selecting the features used by the testnet phased testing (4, 5, 6). Now that it's complete, I'm selecting only the ones that will be used by mainnet (1, 2). After the whole process is complete, all this code is going to be removed.

hathor/conf/testnet.yml
glevco: Nope, `20160` is the default value in the configuration class, so I'm just removing an override. For mainnet it's the same, this key is not being set. Now both mainnet and testnet don't need it.

hathor/conf/testnet.py
glevco: Nope, `20160` is the default value in the configuration class, so I'm just removing an override. For mainnet it's the same, this key is not being set. Now both mainnet and testnet don't need it.

hathor/manager.py
msbrogli: Got it. So it's a fix since the mainnet features were supposed to already be in the list, right?

hathor/conf/testnet.yml
msbrogli: Got it. Thanks!

hathor/conf/testnet.py
msbrogli: Got it. Thanks!

jansegre (approved):

hathor/manager.py
glevco: Yes, I didn't call it fix because the mainnet process hasn't started yet, but I guess it could be

msbrogli: Got it. Thanks!

msbrogli (approved):

PR #938, opened 2024-02-05
fix(indexes): efficient rocksdb mempool-tips initialization

jansegre:
### Motivation

When we made sync-v2 always available, even when not enabled, it implied having to initialize the mempool-tips index, but its RocksDB implementation had been disabled for having a very slow initialization but using a memory index has the downside of always initializing the index on every start.

The mempool is defined as the set of transactions that don't have a `first_block` and are not voided, the mempool-tips index only stores only the mempool transactions that are "tips" (in the sense that all the other transactions can be reached by traversing to the left), the practical definition is the set of all transactions that don't have a `first_block`, don't have a non-voided child and don't have a non-voided spent_output (that is, they aren't literal "tips" because they can have children or spent outputs, as long as those are voided).

Before this PR, the initialization was implemented as simply processing all transactions in topological order the same way they would be if a node received them from the network, with the disadvantage that the it would almost always cache miss due to how the update has to load transactions "to the right" of the iteration loop, making it very slow.

This PR simplifies the loading by looking at the `first_block`, `children`, `spent_outputs` and `voided_by` metadata fields to determine if a transaction is in the mempool and only adding it to the index if it is, instead of adding and removing transactions.

### Acceptance Criteria

- Optimize `MemoryTipsIndex.init_loop_step` by restricting the scope and making it only add transactions without a first_block, children or spent outputs.
- Use the RocksDB implementation when RocksDB indexes are being used.

### Checklist

### DIFF
--- a/hathor/indexes/mempool_tips_index.py
+++ b/hathor/indexes/mempool_tips_index.py
@@ -14,6 +14,7 @@
 
 from abc import abstractmethod
 from collections.abc import Collection
+from itertools import chain
 from typing import TYPE_CHECKING, Iterable, Iterator, Optional, cast
 
 import structlog
@@ -27,10 +28,25 @@ if TYPE_CHECKING:  # pragma: no cover
     from hathor.transaction.storage import TransactionStorage
 
 SCOPE = Scope(
-    include_blocks=True,
+    include_blocks=False,
     include_txs=True,
     include_voided=True,
+    topological_order=False,
 )
+
+
+def any_non_voided(tx_storage: 'TransactionStorage', hashes: Iterable[bytes]) -> bool:
+    """
+    If there's any vertex with hash in hashes that is not voided, this function returns True, otherwise False.
+
+    Notice that this means that an empty `hashes` also returns False.
+    """
+    for tx_hash in hashes:
+        tx = tx_storage.get_transaction(tx_hash)
+        tx_meta = tx.get_metadata()
+        if not tx_meta.voided_by:
+            return True
+    return False
 
 
 class MempoolTipsIndex(BaseIndex):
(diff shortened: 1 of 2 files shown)


hathor/indexes/rocksdb_mempool_tips_index.py
msbrogli, line 1: I don't understand how these changes will speed up the mempool-tips index initialization. Maybe it's worth updating the acceptance criteria with further information.

msbrogli (changes requested):

glevco (approved):

hathor/indexes/rocksdb_mempool_tips_index.py
jansegre: Done.

hathor/indexes/mempool_tips_index.py
```diff
@@ -27,22 +28,33 @@
     from hathor.transaction.storage import TransactionStorage
 
 SCOPE = Scope(
-    include_blocks=True,
+    include_blocks=False,
     include_txs=True,
     include_voided=True,
+    topological_order=False,
```
msbrogli, line 34: Is this configuration just for the sake of initialization? I mean, this index must receive transaction in topological order to work properly (except during its initialization loop which will be executed only one time).

msbrogli (approved):

hathor/indexes/mempool_tips_index.py
jansegre: The order is only used for initialization. Maybe it could be refactored to an `enum` called `InitializationOrder`, but I'd rather do that on a refactor PR.

msbrogli: Ok. Thanks for the clarification.

PR #941, opened 2024-02-06
feat(cli): Add feature flag and CLI parameters for Nano Contracts

msbrogli:
### Motivation

Add a feature flag and CLI parameters to run a localnet with NanoContracts enabled. These changes are temporary but they are useful to allow beta testers to experiment with Nano Contracts.

The activation of Nano Contracts for both mainnet and testnet will occur through the Feature Activation. The proper configuration will take place in a future PR.

### Acceptance Criteria

1. Add a protection to prevent enabling Nano Contracts on any network except `nano-testnet`.
2. Add the CLI parameters `--nano-testnet` that will run the full node on a localnet.
3. Set `nano-testnet` to use the same version bytes as the testnet but different genesis.
4. Add feature flag `ENABLE_NANO_CONTRACTS` which is disabled by default except on the `nano-testnet`.

### Checklist


hathor/cli/run_node.py
```diff
@@ -376,6 +377,8 @@ def __init__(self, *, argv=None):
             os.environ['HATHOR_CONFIG_YAML'] = self._args.config_yaml
         elif self._args.testnet:
             os.environ['HATHOR_CONFIG_YAML'] = TESTNET_SETTINGS_FILEPATH
+        elif args.nano_testnet:
```
pedroferreira1, line 380: Shouldn't it be `self._args.nano_testnet`?

msbrogli: Fixed! Thanks!

hathor/builder/cli_builder.py
```diff
@@ -103,6 +103,11 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
             reactor_type=type(reactor).__name__,
         )
 
+        if not self._args.nano_testnet:
+            # Add protection to prevent enabling Nano Contracts due to misconfigurations.
+            self.check_or_raise(not settings.ENABLE_NANO_CONTRACTS,
+                                'configuration error: NanoContracts can only be enabled on localnets for now')
```
pedroferreira1, line 109: If the user uses `--testnet` and `--nano-testnet`, this wouldn't raise an error, right?

jansegre (approved):
Looks good to me, just looks like some tests that check for the settings have to be updated. Also, `nano_testnet.yml` does not have a `.py` module counterpart, I think it's fine tho, I'll open a PR to remove those, since they have been deprecated for a while now.

hathor/builder/cli_builder.py
msbrogli: Fixed!

msbrogli:
> Looks good to me, just looks like some tests that check for the settings have to be updated. Also, `nano_testnet.yml` does not have a `.py` module counterpart, I think it's fine tho, I'll open a PR to remove those (done, #942), since they have been deprecated for a while now.

Fixed!

pedroferreira1 (approved):

jansegre (approved):

PR #942, opened 2024-02-08
chore(settings): deprecate and remove module based settings

jansegre:
### Motivation

Using `HATHOR_CONFIG_FILE` has been deprecated in favor of `HATHOR_CONFIG_YAML` for a while now. I feel like it's time to remove it.

### Acceptance Criteria

- Log a "critical" error when `HATHOR_CONFIG_FILE` is used and fail loading the settings (this will fail very early on when running any command);
- Clean up `hathor/conf/get_settings.py` now that it only has to load YAML settings;
- Remove `hathor/conf/mainnet.py`, `hathor/conf/testnet.py` and `hathor/conf/unittests.py`;
- Remove module settings tests from `tests/others/test_hathor_settings.py`;

### Checklist


hathor/conf/get_settings.py
```diff
@@ -41,18 +39,22 @@ def HathorSettings() -> Settings:
     """
     Returns the configuration named tuple.
 
-    Tries to get the configuration from a python module in the 'HATHOR_CONFIG_FILE' env var, which will be deprecated.
     If not found, tries to get it from a yaml filepath in the 'HATHOR_YAML_CONFIG', which will be the new standard.
```
pedroferreira1, line 42:
```suggestion
    Gets the configuration from a yaml filepath in the 'HATHOR_YAML_CONFIG'.
    Before it, tries to get from a python module, which was deprecated, and throws a RuntimeError in case it's found.
```

pedroferreira1 (approved):

PR #944, opened 2024-02-08
docs(README): update and organize readme and supplementary documentation of the repository

BrunoCampana:
### Motivation

- README was outdated.
- Overall documentation of this repository was outdated.
- Documentation of codebase isn't standardized.

### Acceptance Criteria

- Update README.
- Organize the documentation of repository.
- Update non-default docs to standardize them.

Standards that must be followed: (all from `engineering-guidelines` repository)
- [README standard](https://github.com/HathorNetwork/engineering-guidelines/blob/master/documentation/write-readme-doc.md)
- [README template](https://github.com/HathorNetwork/engineering-guidelines/blob/master/documentation/readme-template-sw-component-repo.md)
- [Non-default docs standard](https://github.com/HathorNetwork/engineering-guidelines/blob/master/metadocs/drafts/write-ad-hoc-doc.md)
- [Organization of documentation](https://github.com/HathorNetwork/engineering-guidelines/blob/master/metadocs/drafts/high-level-view.md)

### Checklist

- [X] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged


README.md
```diff
@@ -1,201 +1,37 @@
...
 
-At the moment, our mainnet is running on a whitelist basis. This means only authorized nodes will be able to connect. For testing purposes, you can connect to the testnet (using the `--testnet` parameter). If you want to connect to the mainnet, you have to [use a peer-id](#using-a-peer-id) and send this id to a team member. You can get in touch with us through [our channels](https://hathor.network/community/), preferrably Discord.
+**Hathor full node** is the computer system unit that is a peer in Hathor Network. Hathor full node is the primary building block that comprises Hathor Network.
 
-### Using Docker
+## Operation and usage
 
-The easiest way to run a full-node is to use our Docker image. If you don't have Docker installed, check out [this
-link](https://docs.docker.com/install/). So, just run:
+To know how to operate and use Hathor core, see [Hathor full node at Hathor docs — official technical documentation of Hathor](https://docs.hathor.network/pathways/components/full-node).
```
jansegre, line 13: The official technical documentation seems to be outdated:

- latest stable version is v0.58.0 and latest "unstable" version is v0.59.0-rc.1
- deprecated versions, we definitely don't support running too old versions, I can't recall exactly from what version but there certainly is a cutoff.
- not sure what is the best way to keep that document up to date, but maybe we should update that document before we start pointing to it

BrunoCampana: - The [Hathor full node](https://docs.hathor.network/pathways/components/full-node) page of the official technical documentation is already updated, indicating that the latest stable version is: v0.58.0.
- As for unstable versions, these should not be reported in the official documentation.
- My commitment will be to keep this page updated with the same version marked with the latest tag on GitHub. So whenever a new version is marked as stable on GitHub, I will update the page.
- Regarding deprecated versions: it is important to mark, if possible, which versions are deprecated. If you are not sure, I suggest we use as a cutoff point the most recent version that you know no node in the network is operating with.
- Regarding updating this page beforehand and pointing to it: note that the information we currently find in the README is much more outdated than what is in the official documentation.
- Because of this, I do not see it as a problem to update the README pointing there.
- Moreover, all the hathor core operation documentation is being updated: how to install, how to update, etc.

jansegre: Sounds good, I'm very positive about moving all of that away from the README. What I had in mind was something that could go either of these ways:

- Instead of having the version on the document we could just point to the Github releases page which always has the latest version
- Make the CI update the documentation (it would be simpler if the documentation was generated)
- Include "notify the documentation maintainer" to the release process; this is the closest to what you've proposed but I think it would be helpful to have this step documented on the release process

README.md
```diff
@@ -1,201 +1,35 @@
...
 
-The `--status 8080` will run our HTTP API on port 8080, while the `--stratum 8081` will run a stratum server on port
-8081. You can check your full-node status accessing `http://localhost:8080/v1a/status/`. Use `--help` for more
-parameters.
+If you observe an incorrect behavior while using Hathor core, we encourage you to [open an issue to report this failure](https://github.com/HathorNetwork/hathor-core/issues/new).
 
-For more information about our HTTP API, check out our [API Documentation](https://docs.hathor.network/).
+You can also [open an issue to request a new feature you wish to see](https://github.com/HathorNetwork/hathor-core/issues/new).
 
+Please do not open an issue to report a security breach or vulnerability. Instead, follow the guidelines described at [SECURITY](SECURITY.md) for safely reporting, fixing and disclosing security issues.
```
jansegre, line 21: These 3 lines seem all to be about contributing. And the third is exactly the same as the second line of "## Contributing".

What do you think of making the first line point to the Contributing section and removing the other 2 lines?

jansegre (dismissed):

README.md
BrunoCampana: Change made:
I created a new react component to our docs website.
Now, the latest version is not "hardcoded" as text anymore.
The react component fetches latest version at `https://github.com/HathorNetwork/<name_of_repo>/releases/latest` when client side renders the webpage.

Thus, latest version presented in the docs webpages of core, headless and others will always be the latest one.

README.md
BrunoCampana: Change made: 
I changed the sections "Support" and "Contributing" according to your suggestion.

jansegre (dismissed):

SECURITY.md
```diff
@@ -0,0 +1,3 @@
+# Security
+
+Hathor Labs has a bounty program to encourage white hat hackers to collaborate in identifying security breaches and vulnerabilities in Hathor core. To know more about this, see [https://immunefi.com/bounty/hathornetwork/](https://immunefi.com/bounty/hathornetwork/).
```
msbrogli, line 3: I think we should have a simple page about the Bounty Program in Hathor's website. So all links point there. This page will have a link to Immunefi or any other company handling our bug bounty programs.

hathor-core
msbrogli, line 1: Why are we creating this empty file?

msbrogli (changes requested):

hathor-core
BrunoCampana: OK: change made. Removed this file. It was probably some mistake.

BrunoCampana: OK: change made. This file was removed. It was probably by accident.

SECURITY.md
BrunoCampana: Proposal: keep as is.
Reasons:
- Currently the complete description of the bug bounty program can be found on Immunify.
- Every place where the subject "security/bug bounty" is mentioned, Immunify is mentioned. This already occurs in the following locations:
- Main website: https://hathor.network/developers/ (help us improve section)
- Documentation website: https://docs.hathor.network/references/besides-documentation#security
- And now, the proposal is to do the same in all READMEs in public repositories.

- It is good practice to have a "security" doc in each repository. To an outsider, it denotes a high level of quality in our codebase.

- In this context, we would have to create a new section, albeit small, perhaps repeating what is already in the `SECURITY.md` doc or adding content that is in Immunify.
- In both cases, there does not seem to be any benefit to this. On the contrary, it would make updating the content even more difficult.

- The only advantage seems to be being able to update a single link, when in the future the bug bounty program migrates from Immunify to another platform.
- However, the cost of changing links in markdowns and website is marginal.
- Furthermore, in any case the best solution would not be to add a new section to the website, but rather to make a note using the DNS of the hathor.network domain.

Now: if you understand that it is important to have more information about security than we currently have on the website, the solution would be to keep this PR as is, and add the content of `SECURITY.md` on the hathor.network website, just as we have on https ://docs.hathor.network/references/besides-documentation#security.

hathor-core
msbrogli: Thanks!

jansegre (approved):

msbrogli (approved):

PR #945, opened 2024-02-08
feat(sync-v2): Improve logging when unexpected TRANSACTIONS-END or BLOCKS-END are received

msbrogli:
### Motivation

### Acceptance Criteria

### Checklist


hathor/p2p/sync_v2/agent.py
```diff
@@ -721,7 +721,7 @@ def handle_blocks_end(self, payload: str) -> None:
         assert self.protocol.connections is not None
 
         if self.state is not PeerState.SYNCING_BLOCKS:
-            self.log.error('unexpected BLOCKS-END', state=self.state)
+            self.log.error('unexpected BLOCKS-END', state=self.state, response_code=response_code)
```
jansegre, line 724: Wouldn't it be better to use `response_code.name` instead? Since just `response_code` is considered an `int` and will end up as such in the json logs. We've already passed it through the enum so it is guaranteed to have a name.

hathor/p2p/sync_v2/agent.py
```diff
@@ -978,7 +978,7 @@ def handle_transactions_end(self, payload: str) -> None:
         assert self.protocol.connections is not None
 
         if self.state is not PeerState.SYNCING_TRANSACTIONS:
-            self.log.error('unexpected TRANSACTIONS-END', state=self.state)
+            self.log.error('unexpected TRANSACTIONS-END', state=self.state, response_code=response_code)
```
jansegre, line 981: Same here.

hathor/p2p/sync_v2/agent.py
msbrogli: Nice catch. Thanks!

hathor/p2p/sync_v2/agent.py
msbrogli: Nice catch. Thanks!

jansegre (approved):

PR #949, opened 2024-02-09
feat(cli): Use SIGUSR2 to run sysctl commands

msbrogli:
### Motivation

Sometimes the CPU is so high that sysctl become irresponsive. So this PR adds another method to send command to sysctl.

### Acceptance Criteria

1. Handle SIGUSR2 on cli `run_node`.
2. Create a named pipe to receive a list of sysctl commands, one per line.
3. Execute the sysctl commands and log their results.
4. The named pipe will be created at the data directory if it exists. Otherwise, a temporary folder will be created.
5. Add decorator `@signal_handler_safe` to mark setters as safe to be executed during a signal handling.
6. Add `require_signal_handler_safe: bool = False` to the `SysctlRunner.run()` method.
7. Rename `_get_getter()` and `_set_setter()` to `get_getter()` and `set_setter()` on `Sysctl`.
8. Rename `set()` to `unsafe_set()` on `Sysctl`. It also updates the docstring to explain that it is unsafe to call this method directly.
9. Mark `max_enabled_sync`, `enabled_sync_versions` and `kill_connection` as signal handler safe on `ConnectionsManagerSysctl`.

### Checklist

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -14,8 +14,10 @@
 
 import os
 import sys
+import tempfile
 from argparse import SUPPRESS, ArgumentParser, Namespace
-from typing import TYPE_CHECKING, Any, Callable, Optional
+from contextlib import contextmanager
+from typing import TYPE_CHECKING, Any, Callable, Iterator, Optional, TextIO
 
 from pydantic import ValidationError
 from structlog import get_logger
@@ -25,6 +27,25 @@ logger = get_logger()
 
 if TYPE_CHECKING:
     from hathor.cli.run_node_args import RunNodeArgs
+    from hathor.sysctl.runner import SysctlRunner
+
+
+@contextmanager
+def temp_fifo(filename: str, tempdir: str | None) -> Iterator[TextIO]:
+    """Context Manager for creating named pipes."""
+    mkfifo = getattr(os, 'mkfifo', None)
+    if mkfifo is None:
+        raise AttributeError('mkfifo is not available')
+
+    mkfifo(filename, mode=0o666)
+    fp = open(filename, 'r')
+    try:
+        yield fp
+    finally:
+        fp.close()
+        os.unlink(filename)
+        if tempdir is not None:
+            os.rmdir(tempdir)
 
 
 class RunNode:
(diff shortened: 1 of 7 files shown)


hathor/cli/run_node.py
```diff
@@ -232,13 +246,71 @@ def register_signal_handlers(self) -> None:
...
+        if os.path.exists(filename):
+            self.log.warn('[USR2] Pipe already exists. Delete it and try again.', pipe=filename)
+            return
+
+        runner = self.get_sysctl_runner()
+
+        with temp_fifo(filename, tempdir):
+            self.log.warn('[USR2] Waiting for commands...', pipe=filename)
+            fp = open(filename, 'r')
+            lines = fp.readlines()
```
msbrogli, line 289: I considered to keep the pipe open until the user decides to delete it. But it would require non-blocking reading so I skipped it for now. What do you think?

hathor/cli/run_node.py
```diff
@@ -27,6 +29,18 @@
     from hathor.cli.run_node_args import RunNodeArgs
 
 
+@contextmanager
+def temp_fifo(filename: str, tempdir: str | None) -> Iterator[str]:
+    """Context Manager for creating named pipes."""
+    os.mkfifo(filename, mode=0o666)
+    try:
+        yield filename
```
jansegre, line 37: It seems this result isn't being used. Should we move the opening of the file to here too? And use it like:

```python
with temp_fifo(path, dir) as fp:
    fp.readlines()
    # ...
```

hathor/cli/run_node.py
```diff
@@ -232,13 +246,71 @@ def register_signal_handlers(self) -> None:
...
+            # USR1 is available in this OS.
+            signal.signal(sigusr2, self.signal_usr2_handler)
 
     def signal_usr1_handler(self, sig: int, frame: Any) -> None:
         """Called when USR1 signal is received."""
         self.log.warn('USR1 received. Killing all connections...')
         if self.manager and self.manager.connections:
             self.manager.connections.disconnect_all_peers(force=True)
 
+    def signal_usr2_handler(self, sig: int, frame: Any) -> None:
```
jansegre, line 260: I think the implementation of this could be moved to a method in the runner, and make this method focus on not leaking any exceptions, with something like:

```python
    def signal_usr2_handler(self, sig: int, frame: Any) -> None:
        """Called when USR2 signal is received."""
        self.log.warn('USR2 received.')
        try:
            runner = self.get_sysctl_runner()
            runner.run_from_signal(self._args.data)
        except Exception:
            self.log.error('prevented exception from escaping the signal handler', exc_info=True)
```

hathor/cli/run_node.py
```diff
@@ -232,13 +246,71 @@ def register_signal_handlers(self) -> None:
...
+        """Called when USR1 signal is received."""
```
jansegre, line 261: USR1 -> USR2

jansegre (approved):

hathor/cli/run_node.py
msbrogli: Fixed. Thanks!

hathor/cli/run_node.py
msbrogli: Done!

hathor/cli/run_node.py
msbrogli: Nice suggestion. Done!

jansegre (approved):

hathor/cli/run_node.py
msbrogli: We decided to keep it this way for now.

PR #951, opened 2024-02-09
feat(sysctl): Add core module with profiler commands to sysctl

msbrogli:
### Motivation

We currently must use an HTTP API to start and stop a global profiler. Such an API might not be accessible due to networking issues so having the ability to do it through sysctl becomes relevant.

### Acceptance Criteria

1. Create commands `core.profiler.status`, `core.profiler.start`, and `core.profiler.stop` in sysctl.
2. Add `HathorManagerSysctl` which handles all these commands.
3. Add `HathorManager.is_profiler_running` to keep track whether the profiler is running or not.
4. Add `HathorManager.profiler_last_start_time` to be able to calculate how long the profiler has been running.

### Checklist

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -15,8 +15,9 @@
 import datetime
 import sys
 import time
+from cProfile import Profile
 from enum import Enum
-from typing import Any, Iterator, NamedTuple, Optional, Union
+from typing import Iterator, NamedTuple, Optional, Union
 
 from hathorlib.base_transaction import tx_or_block_from_bytes as lib_tx_or_block_from_bytes
 from structlog import get_logger
@@ -144,7 +145,11 @@ class HathorManager:
             add_system_event_trigger('after', 'shutdown', self.stop)
 
         self.state: Optional[HathorManager.NodeState] = None
-        self.profiler: Optional[Any] = None
+
+        # Profiler info
+        self.profiler: Optional[Profile] = None
+        self.is_profiler_running: bool = False
+        self.profiler_last_start_time: float = 0
 
         # Hostname, used to be accessed by other peers.
         self.hostname = hostname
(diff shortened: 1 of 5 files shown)


hathor/manager.py
```diff
@@ -144,7 +144,11 @@ def __init__(self,
             add_system_event_trigger('after', 'shutdown', self.stop)
 
         self.state: Optional[HathorManager.NodeState] = None
+
+        # Profiler info
         self.profiler: Optional[Any] = None
```
pedroferreira1, line 149: Don't we have the `Profile` type?

hathor/sysctl/core/manager.py
```diff
@@ -0,0 +1,53 @@
...
+            None,
+            self.set_profiler_start,
+        )
+        self.register(
+            'profiler.stop',
+            None,
+            self.set_profiler_stop,
+        )
+
+    def get_profiler_status(self) -> tuple[int, float]:
```
pedroferreira1, line 40: Missing docstring in all methods

pedroferreira1 (approved):

jansegre (approved):

hathor/manager.py
msbrogli: Fixed. Thanks!

hathor/sysctl/core/manager.py
msbrogli: Fixed. Thanks!

hathor/sysctl/core/manager.py
```diff
@@ -0,0 +1,63 @@
...
+            return (0, 0)
+        now = self.manager.reactor.seconds()
+        duration = now - self.manager.profiler_last_start_time
+        return (1, duration)
+
+    def set_profiler_start(self, reset: bool) -> None:
+        """Start the profiler. One can safely call start multiple times to reset it."""
+        self.manager.start_profiler(reset=reset)
+
+    def set_profiler_stop(self, save_to: str | None) -> None:
```
pedroferreira1, line 56: What's the difference between `str | None` and `Optional[str]`?

pedroferreira1 (approved):

jansegre (approved):

hathor/sysctl/core/manager.py
msbrogli: They are the same; just a different notation.

PR #955, opened 2024-02-15
feat(sysctl): Add sysctl tools for debugging

msbrogli:
### Motivation

Being able to stop the event loop for debugging is useful to get a better understanding of issues. This PR introduces two different sysctl commands: (i) pudb and (ii) ipython.

The pudb commands is a full-screen console debugger that allows us to run step-by-step and inspect the stack trace and its local variables. The event loop will be paused during its execution.

The ipython command will pause the event loop and open an ipython console so we can inspect the objects internal state and call methods.

In both cases, the event loop is resumed after the debugging is finished. But it's important to highlight that these tools might crash the full node so we should avoid using it on production.

These are experimental tools that will probably be improved in future PRs.

### Acceptance Criteria

1. Add command `core.pudb.set_trace.attach_tty` that runs a pudb console attached to a tty that already exists.
2. Add command `core.pudb.set_trace.create_tty` that runs a pudb console on a newly created tty. So user have to attach this tty in a terminal to get access to the console.
3. Add command `core.pudb.status` that returns whether pudb is running or not.
4. Add command `core.pudb.stop` that stops a pudb console and return the control to the event loop.
5. Add command `core.ipython.run.attach_tty` that runs an ipython console attached a tty that already exists.
6. All these new commands were decorated with `@signal_handler_safe`.
7. Decorate both `set_profiler_start` and `set_profiler_stop` with `@signal_handler_safe`.

### Checklist

### DIFF
--- a/hathor/sysctl/core/manager.py
+++ b/hathor/sysctl/core/manager.py
@@ -12,8 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
+import os
+import sys
+from typing import IO, Any, Optional
+
 from hathor.manager import HathorManager
-from hathor.sysctl.sysctl import Sysctl
+from hathor.sysctl.sysctl import Sysctl, signal_handler_safe
 
 
 class HathorManagerSysctl(Sysctl):
@@ -36,6 +41,31 @@ class HathorManagerSysctl(Sysctl):
             None,
             self.set_profiler_stop,
         )
+        self.register(
+            'pudb.set_trace.attach_tty',
+            None,
+            self.set_pudb_set_trace_attach_tty,
+        )
+        self.register(
+            'pudb.set_trace.create_tty',
+            None,
+            self.set_pudb_set_trace_create_tty,
+        )
+        self.register(
+            'pudb.status',
+            self.get_pudb_status,
+            None
+        )
+        self.register(
+            'pudb.stop',
+            None,
+            self.set_pudb_stop,
+        )
+        self.register(
+            'ipython.run.attach_tty',
+            None,
+            self.set_ipython_run,
+        )
 
     def get_profiler_status(self) -> tuple[int, float]:
         """Return (enabled, duration) as a profiler status.
@@ -49,10 +79,12 @@ class HathorManagerSysctl(Sysctl):
         duration = now - self.manager.profiler_last_start_time
         return (1, duration)
 
+    @signal_handler_safe
     def set_profiler_start(self, reset: bool) -> None:
         """Start the profiler. One can safely call start multiple times to reset it."""
         self.manager.start_profiler(reset=reset)
 
+    @signal_handler_safe
     def set_profiler_stop(self, save_to: str | None) -> None:
         """Stop the profiler and optionally dump the statistics to a file.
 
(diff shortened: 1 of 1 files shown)


jansegre (approved):

jansegre (dismissed):

hathor/sysctl/core/manager.py
```diff
@@ -61,3 +93,141 @@ def set_profiler_stop(self, save_to: str | None) -> None:
...
+        import termios
+
+        # Some of these methods are not available in some operating system (e.g. Windows).
+        # This is a way to avoid mypy errors.
+        openpty = getattr(os, 'openpty', None)
+        ioctl = getattr(fcntl, 'ioctl', None)
+        TIOCSWINSZ = getattr(termios, 'TIOCSWINSZ', None)
+
+        if openpty is None or ioctl is None or TIOCSWINSZ is None:
+            self.log.warn('Error opening pudb', openpty=openpty, ioctl=ioctl, TIOCSWINSZ=TIOCSWINSZ)
```
pedroferreira1, line 137: You don't need this when you use the method `set_pudb_set_trace_attach_tty`, is that correct?

Should we recommend the user to use the other way in the log message?

hathor/sysctl/core/manager.py
```diff
@@ -61,3 +93,141 @@ def set_profiler_stop(self, save_to: str | None) -> None:
...
+    @signal_handler_safe
+    def set_pudb_set_trace_create_tty(self, cols: int, rows: int) -> None:
+        """Stop execution and open pudb for debugging in a newly created tty.
+
+        ATTENTION: This command can be destructive and the full node might not work after running it.
+
+        You must provide the terminal size (cols, rows).
+
+        The newly created tty path will be printed in the logs. After you get it, you can
+        connect to it using `screen <ttyname>`.
```
pedroferreira1, line 124: The user does not set `ttyname`. How can he know this name?

hathor/sysctl/core/manager.py
```diff
@@ -61,3 +93,141 @@ def set_profiler_stop(self, save_to: str | None) -> None:
...
+        from pudb.debugger import Debugger
+
+        self.log.warn('main loop paused; pudb.set_trace running', tty=tty)
+
+        tty_file = io.TextIOWrapper(fp)
+        kwargs: dict[str, Any] = {
+            'stdin': tty_file,
+            'stdout': tty_file,
+        }
+        if term_size is not None:
```
pedroferreira1, line 161: If `term_size` is `None` what happens? Does it get a default value?

pedroferreira1 (dismissed):

hathor/sysctl/core/manager.py
msbrogli: Good idea. Done!

hathor/sysctl/core/manager.py
msbrogli: I improved the documentation. The name ttyname will be logged. So you have to check the logs and get the name there.

hathor/sysctl/core/manager.py
msbrogli: It tries to get the term size from the envvars and using ioctl. This is the default behavior from pudb.

pedroferreira1 (dismissed):

PR #957, opened 2024-02-21
fix: Fix slow tx processing on some edge cases

msbrogli:
### Motivation

We got this issue during the QA of hathor-core v0.59.0-rc1. This is triggered when a node does a sync from scratch using sync-v2 and, later on, switches to sync-v1. The sync-v1 works properly but the processing of some transactions get considerably slower.

Fixes https://github.com/HathorNetwork/internal-issues/issues/246.

### Acceptance Criteria

1. Stop marking transactions confirmed by the best blockchain as voided during conflict resolution. Notice that this edge case only occurs when the conflicting transaction is soft voided and is confirmed by the best blockchain.

### Checklist

### DIFF
--- a/hathor/consensus/transaction_consensus.py
+++ b/hathor/consensus/transaction_consensus.py
@@ -235,7 +235,12 @@ class TransactionConsensusAlgorithm:
             conflict_tx = cast(Transaction, tx.storage.get_transaction(h))
             conflict_tx_meta = conflict_tx.get_metadata()
             if conflict_tx_meta.voided_by:
-                self.mark_as_voided(conflict_tx)
+                if conflict_tx_meta.first_block is not None:
+                    # do nothing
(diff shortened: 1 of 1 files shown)


hathor/consensus/transaction_consensus.py
```diff
@@ -234,6 +234,8 @@ def update_voided_info(self, tx: Transaction) -> None:
         for h in meta.conflict_with or []:
             conflict_tx = cast(Transaction, tx.storage.get_transaction(h))
             conflict_tx_meta = conflict_tx.get_metadata()
+            if conflict_tx_meta.first_block is not None:
```
msbrogli, line 237: `assert bool(self.context.consensus.soft_voided_tx_ids & conflict_tx_meta.voided_by)`?

msbrogli: Done!

jansegre (approved):

jansegre (approved):

glevco (approved):

PR #958, opened 2024-02-22
feat(feature-activation): configure new NOP features on mainnet

glevco:
### Motivation

Since the miners did not update their nodes for the last test, we have to introduce new NOP features to test that bit signaling is working correctly.

The features are configured to start in approximately 1 week from now, and the voting period will last for 10 weeks, giving enough time for miners to udpate.

### Acceptance Criteria

- Configure NOP features for a new Feature Activation test on mainnet.

### Checklist

### DIFF
--- a/hathor/conf/mainnet.py
+++ b/hathor/conf/mainnet.py
@@ -203,7 +203,7 @@ SETTINGS = HathorSettings(
         features={
             Feature.NOP_FEATURE_1: Criteria(
                 bit=0,
-                start_height=4_213_440,  # N (right now the best block is 4_169_000 on mainnet)
+                start_height=4_213_440,  # N
                 timeout_height=4_253_760,  # N + 2 * 20160 (2 weeks after the start)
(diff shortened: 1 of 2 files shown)


hathor/conf/mainnet.py
```diff
@@ -212,13 +212,31 @@
...
             Feature.NOP_FEATURE_2: Criteria(
                 bit=1,
-                start_height=4_213_440,  # N (right now the best block is 4_169_000 on mainnet)
+                start_height=4_213_440,  # N
                 timeout_height=4_253_760,  # N + 2 * 20160 (2 weeks after the start)
                 minimum_activation_height=0,
                 lock_in_on_timeout=False,
                 version='0.59.0',
                 signal_support_by_default=False,
-            )
```
jansegre, line 221: I think we should be able to remove `NOP_FEATURE_1` and `NOP_FEATURE_2`, right?

msbrogli (approved):

jansegre (approved):

PR #959, opened 2024-02-22
chore(cli): make sync-bridge the default

jansegre:
### Motivation

The current default of only enabling `sync-v1` is proving to be taxing on CPU, while `sync-v2` has been running very well on tests and seems stable enough. Defaulting to bridge mode allows for the best connectivity (connecting to updated and outdated nodes) and will gradually and naturally improve along with nodes being updated.

### Acceptance Criteria

- Add new cli option `--x-sync-v1-only`
- Make the default be sync bridge (sync-v1 and sync-v2 enabled)
- Tweak some warnings accordingly
- Adapt tests to the new defaults and the new cli option

### Checklist

### DIFF
--- a/hathor/cli/run_node_args.py
+++ b/hathor/cli/run_node_args.py
@@ -66,6 +66,7 @@ class RunNodeArgs(BaseModel, extra=Extra.allow):
     x_sync_bridge: bool
     x_sync_v2_only: bool
     sync_bridge: bool
+    sync_v1_only: bool
     sync_v2_only: bool
     x_localhost_only: bool
     x_rocksdb_indexes: bool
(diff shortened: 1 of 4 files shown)


hathor/cli/run_node_args.py
```diff
@@ -64,6 +64,7 @@ class RunNodeArgs(BaseModel, extra=Extra.allow):
     enable_debug_api: bool
     enable_crash_api: bool
     x_sync_bridge: bool
+    x_sync_v1_only: bool
```
glevco, line 67: Why not `sync_v1_only`?

glevco (approved):

hathor/cli/run_node_args.py
jansegre: Yeah, I think that would actually fit better, I'll change to that.

glevco (approved):

msbrogli (approved):

PR #962, opened 2024-02-27
feat(feature-activation): implement bit signaling sysctl

glevco:
## Motivation

Add sysctl commands to control Feature Activation bit signaling without having to restart the full node (previously, this was only available through CLI options). The commands are the following:

#### `core.features.supported_features`

Return features that are explicitly enabled. Read-only.

#### `core.features.not_supported_features`

Return features that are explicitly disabled. Read-only.

####  `core.features.signaling_features`

Return which feature are currently in a signaling phase. Read-only.

#### `core.features.add_support`

Add support for a feature. Write-only.

#### `core.features.remove_support`

Remove support for a feature. Write-only.

### Example

```bash
$ nc -U /tmp/hathor.sock
core.features.signaling_features
["NOP_FEATURE_5"]
core.features.remove_support="NOP_FEATURE_5"
core.features.not_supported_features
["NOP_FEATURE_5"]
core.features.supported_features
[]
core.features.add_support="NOP_FEATURE_5"
core.features.not_support_features
[]
core.features.supported_features
["NOP_FEATURE_5"]
```

Then, the bits in the block template will be changed accordingly.

## Acceptance Criteria

- Add `core.features` sysctl command group with the following subcommands: `supported_features`, `not_supported_features`, `signaling_features`, `add_support`, and `remove_support`.

### Checklist

### DIFF
--- a/hathor/builder/sysctl_builder.py
+++ b/hathor/builder/sysctl_builder.py
@@ -13,7 +13,13 @@
 # limitations under the License.
 
 from hathor.builder import BuildArtifacts
-from hathor.sysctl import ConnectionsManagerSysctl, HathorManagerSysctl, Sysctl, WebsocketManagerSysctl
+from hathor.sysctl import (
+    ConnectionsManagerSysctl,
+    FeatureActivationSysctl,
+    HathorManagerSysctl,
+    Sysctl,
+    WebsocketManagerSysctl,
+)
 
 
 class SysctlBuilder:
@@ -25,7 +31,11 @@ class SysctlBuilder:
     def build(self) -> Sysctl:
         """Build the sysctl tree."""
         root = Sysctl()
-        root.put_child('core', HathorManagerSysctl(self.artifacts.manager))
+
+        core = HathorManagerSysctl(self.artifacts.manager)
+        core.put_child('features', FeatureActivationSysctl(self.artifacts.bit_signaling_service))
+
+        root.put_child('core', core)
         root.put_child('p2p', ConnectionsManagerSysctl(self.artifacts.p2p_manager))
 
         ws_factory = self.artifacts.manager.metrics.websocket_factory
--- a/hathor/sysctl/feature_activation/manager.py
+++ b/hathor/sysctl/feature_activation/manager.py
@@ -1,0 +1,72 @@
+#  Copyright 2024 Hathor Labs
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from hathor.feature_activation.bit_signaling_service import BitSignalingService
+from hathor.feature_activation.feature import Feature
+from hathor.sysctl.sysctl import Sysctl
+
+
+class FeatureActivationSysctl(Sysctl):
+    def __init__(self, bit_signaling_service: BitSignalingService) -> None:
+        super().__init__()
+        self._bit_signaling_service = bit_signaling_service
+
+        self.register(
+            path='supported_features',
+            getter=self.get_support_features,
+            setter=None,
+        )
+        self.register(
+            path='not_supported_features',
+            getter=self.get_not_support_features,
+            setter=None,
+        )
+        self.register(
+            path='signaling_features',
+            getter=self.get_signaling_features,
+            setter=None,
+        )
+        self.register(
+            path='add_support',
+            getter=None,
+            setter=self.add_feature_support,
+        )
+        self.register(
+            path='remove_support',
+            getter=None,
+            setter=self.remove_feature_support,
+        )
+
+    def get_support_features(self) -> list[str]:
+        """Get a list of feature names with enabled support."""
+        return [feature.value for feature in self._bit_signaling_service.get_support_features()]
+
+    def get_not_support_features(self) -> list[str]:
+        """Get a list of feature names with disabled support."""
+        return [feature.value for feature in self._bit_signaling_service.get_not_support_features()]
+
+    def add_feature_support(self, *features: str) -> None:
+        """Explicitly add support for a feature by enabling its signaling bit."""
+        for feature in features:
+            self._bit_signaling_service.add_feature_support(Feature[feature])
+
+    def remove_feature_support(self, *features: str) -> None:
+        """Explicitly remove support for a feature by disabling its signaling bit."""
+        for feature in features:
+            self._bit_signaling_service.remove_feature_support(Feature[feature])
+
+    def get_signaling_features(self) -> list[str]:
+        """Get a list of feature names that are currently in a signaling state."""
+        features = self._bit_signaling_service.get_best_block_signaling_features().keys()
+        return [feature.value for feature in features]
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -71,6 +71,7 @@ class BuildArtifacts(NamedTuple):
     consensus: ConsensusAlgorithm
     tx_storage: TransactionStorage
     feature_service: FeatureService
+    bit_signaling_service: BitSignalingService
     indexes: Optional[IndexesManager]
     wallet: Optional[BaseWallet]
     rocksdb_storage: Optional[RocksDBStorage]
@@ -247,6 +248,7 @@ class Builder:
             rocksdb_storage=self._rocksdb_storage,
             stratum_factory=stratum_factory,
             feature_service=feature_service,
+            bit_signaling_service=bit_signaling_service
         )
 
         return self.artifacts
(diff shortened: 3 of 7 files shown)


hathor/builder/sysctl_builder.py
```diff
@@ -27,6 +33,7 @@ def build(self) -> Sysctl:
         root = Sysctl()
         root.put_child('core', HathorManagerSysctl(self.artifacts.manager))
         root.put_child('p2p', ConnectionsManagerSysctl(self.artifacts.p2p_manager))
+        root.put_child('fa', FeatureActivationSysctl(self.artifacts.bit_signaling_service))
```
msbrogli, line 36: Maybe `core.features.`?

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from hathor.feature_activation.bit_signaling_service import BitSignalingService
+from hathor.feature_activation.feature import Feature
+from hathor.sysctl.sysctl import Sysctl
+
+
+class FeatureActivationSysctl(Sysctl):
```
msbrogli, line 20: Missing tests.

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, bit_signaling_service: BitSignalingService) -> None:
+        super().__init__()
+        self._bit_signaling_service = bit_signaling_service
+
+        self.register(
+            path='support_features',
```
msbrogli, line 26: `supported_features`?

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+            getter=self.get_support_features,
+            setter=None,
+        )
+        self.register(
+            path='not_support_features',
```
msbrogli, line 31: `not_supported_features`?

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+            setter=self.add_feature_support,
+        )
+        self.register(
+            path='remove_support',
+            getter=None,
+            setter=self.remove_feature_support,
+        )
+
+    def get_support_features(self) -> list[str]:
+        """Get a list of feature names with explicitly enabled support."""
```
msbrogli, line 52: What about the ones supported by default?

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+        return [feature.value for feature in self._bit_signaling_service.get_support_features()]
+
+    def get_not_support_features(self) -> list[str]:
+        """Get a list of feature names with explicitly disabled support."""
```
msbrogli, line 56: What about the ones not supported by default?

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+        return [feature.value for feature in self._bit_signaling_service.get_not_support_features()]
+
+    def add_feature_support(self, *features: str) -> None:
```
msbrogli, line 59: Why not `features: List[str]`? Maybe we should do one per call.

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+        """Add explicit support for a feature by enabling its signaling bit."""
+        for feature in features:
+            self._bit_signaling_service.add_feature_support(Feature[feature])
+
+    def remove_feature_support(self, *features: str) -> None:
```
msbrogli, line 64: Why not `features: List[str]`? Maybe we should do one per call.

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+        """Remove explicit support for a feature by disabling its signaling bit."""
+        for feature in features:
+            self._bit_signaling_service.remove_feature_support(Feature[feature])
+
+    def get_signaling_features(self) -> list[str]:
```
msbrogli, line 69: Why not receive a block hash or block height? Maybe we should support all these possibilities.

hathor/sysctl/feature_activation/manager.py
```diff
@@ -0,0 +1,72 @@
...
+            getter=self.get_not_support_features,
+            setter=None,
+        )
+        self.register(
+            path='signaling_features',
+            getter=self.get_signaling_features,
+            setter=None,
+        )
+        self.register(
+            path='add_support',
```
msbrogli, line 41: You can dynamically create a `support.[feature_name]` so one can assign 0/1 to disable/enable. What do you think?

msbrogli (changes requested):

hathor/sysctl/feature_activation/manager.py
glevco: Because the `SysctlRunner` calls it like this:

```python
if isinstance(value, tuple):
    setter(*value)
else:
    setter(value)
```

So we can simply call `core.features.add_support="NOP_FEATURE_1", "NOP_FEATURE_2"` in the sysctl instead of wrapping it in an array.

I think allowing the user to add/remove multiple features at once is a UX convenience, why not allow it?

hathor/sysctl/feature_activation/manager.py
glevco: This can be easily checked in the `mainnet.yml` config file. Do you think it's necessary?

hathor/sysctl/feature_activation/manager.py
glevco: See https://github.com/HathorNetwork/hathor-core/pull/962#discussion_r1509450686

hathor/sysctl/feature_activation/manager.py
glevco: See https://github.com/HathorNetwork/hathor-core/pull/962#discussion_r1509450113

hathor/sysctl/feature_activation/manager.py
glevco: I think it's not very useful, as it's not possible to calculate the signaling features of future blocks (for example, with a height higher then the current best height). For blocks from the past, this information doesn't really matter for the user. This command is only useful for the user to check which features he could enable/disable at the moment he's using it. Also, sysctl does not currently support getters with arguments, does it?

hathor/builder/sysctl_builder.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/962/commits/2fa908fd1cb31ad6f5b0c863d668509f5883b160

hathor/sysctl/feature_activation/manager.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/962/commits/2fa908fd1cb31ad6f5b0c863d668509f5883b160

hathor/sysctl/feature_activation/manager.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/962/commits/2fa908fd1cb31ad6f5b0c863d668509f5883b160

hathor/sysctl/feature_activation/manager.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/962/commits/2fa908fd1cb31ad6f5b0c863d668509f5883b160

hathor/builder/sysctl_builder.py
msbrogli: Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: Make sense. Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: Got it. Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: I was thinking about this internal method (not from its specific use for this PR). But we can safely do it later if needed.

hathor/sysctl/feature_activation/manager.py
glevco: Sure, changed in https://github.com/HathorNetwork/hathor-core/pull/962/commits/3ef4d621830b35b247f656d3457114db7b71ce17

hathor/sysctl/feature_activation/manager.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/962/commits/3ef4d621830b35b247f656d3457114db7b71ce17

jansegre (approved):

jansegre (approved):

hathor/sysctl/feature_activation/manager.py
msbrogli: Thanks!

hathor/sysctl/feature_activation/manager.py
msbrogli: Thanks!

msbrogli (approved):

PR #963, opened 2024-02-29
feat(feature-activation): automatically enable support in MUST_SIGNAL

glevco:
### Motivation

### Acceptance Criteria

### Checklist


hathor/feature_activation/bit_signaling_service.py
```diff
@@ -108,6 +109,13 @@ def remove_feature_support(self, feature: Feature) -> None:
         self._support_features.discard(feature)
         self._not_support_features.add(feature)
 
+    def on_must_signal(self, feature: Feature) -> None:
+        """
+        When the MUST_SIGNAL phase is reached, feature support is automatically enabled, unless explicitly disabled.
```
msbrogli, line 114: Shouldn't we enable it anyway?

msbrogli (changes requested):

hathor/feature_activation/bit_signaling_service.py
glevco: Maybe we could, but I feel a bit weird forcing this into the user. Maybe he has a reason to explicitly configure a "non-support". Here's @jansegre's initial suggestion for this improvement:

> For updated nodes, when a feature reaches MUST_SIGNAL, the node should signal support by default, unless explicitly disabled by the node operator. Maybe we could have some ternary config, like "always-signal", "whatever-works", and "never-signal"

I implemented it like he said, keeping it off if explicitly disabled. I ended up not implementing the ternary config to keep it simple.

msbrogli: I'm concerned enabling it, despite the node operator's explicit refusal, is necessary to prevent potential hard forks, as the operator's defeat in the voting mandates compliance.

jansegre: I agree. If they disagree to the point of not wanting to signal even on MUST_SIGNAL, then they could as well fork the code.

glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/963/commits/ed8023ecf701ade069195c55935c81957b3f299e

msbrogli: Thanks!

msbrogli (dismissed):

jansegre (dismissed):

jansegre (approved):

msbrogli (approved):

PR #969, opened 2024-03-13
refactor(mypy): add stricter rules to specific modules [part I/VI]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/967

### Motivation

We currently use a subset of static typing rules with mypy, for historic reasons. The aim should be to enable `strict=True` in the whole repo, or at least most of it. One way to do this incrementally is to configure a mypy override with stricter rules and set it for some specific modules.

This PR does this for some newer or important modules. We cannot use `strict=True` directly, because this rule is not available on per-module configs. Therefore, we get all strict rules from the mypy documentation and manually configure them.

### Acceptance Criteria

- Update `mypy` and `mypy-zope`.
- Configure stricter mypy for some specific modules and update those modules to conform.
- Some other minor typing/style changes.

### Checklist


msbrogli (dismissed):

jansegre (approved):
Looks good to me.

jansegre (approved):

msbrogli (approved):

PR #971, opened 2024-03-14
refactor(mypy): add stricter rules to some test modules [part II/VI]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/969

### Motivation

This PR continues to improve the strictness of mypy in test modules. For more info see the previous PR, https://github.com/HathorNetwork/hathor-core/pull/969.

### Acceptance Criteria

- Add stricter mypy rules to some test modules and update them to conform.

### Checklist

### DIFF
--- a/tests/consensus/test_soft_voided4.py
+++ b/tests/consensus/test_soft_voided4.py
@@ -1,7 +1,10 @@
+from typing import Iterator
+
 from hathor.graphviz import GraphvizVisualizer
-from hathor.simulator import FakeConnection, Simulator
+from hathor.simulator import FakeConnection, RandomTransactionGenerator, Simulator
 from hathor.simulator.trigger import StopAfterNTransactions
 from hathor.simulator.utils import gen_new_double_spending
+from hathor.types import VertexId
 from tests import unittest
 from tests.simulation.base import SimulatorTestCase
 from tests.utils import add_custom_tx
@@ -10,7 +13,11 @@ from tests.utils import add_custom_tx
 class BaseSoftVoidedTestCase(SimulatorTestCase):
     seed_config = 5988775361793628169
 
-    def _run_test(self, simulator, soft_voided_tx_ids):
+    def _run_test(
+        self,
+        simulator: Simulator,
+        soft_voided_tx_ids: list[VertexId]
+    ) -> Iterator[RandomTransactionGenerator]:
         manager1 = self.create_peer(soft_voided_tx_ids=set(soft_voided_tx_ids), simulator=simulator)
         manager1.allow_mining_without_peers()
 
@@ -54,6 +61,7 @@ class BaseSoftVoidedTestCase(SimulatorTestCase):
 
         gen_tx2.stop()
 
+        assert isinstance(soft_voided_tx_ids, list)
         self.assertEqual(2, len(soft_voided_tx_ids))
         txA_hash = soft_voided_tx_ids[0]
         txB_hash = soft_voided_tx_ids[1]
@@ -125,12 +133,12 @@ class BaseSoftVoidedTestCase(SimulatorTestCase):
         metaD = txD.get_metadata()
         self.assertEqual(metaD.voided_by, {tx_base.hash})
 
-    def _get_txA_hash(self):
+    def _get_txA_hash(self) -> VertexId:
         simulator = Simulator(seed=self.simulator.seed)
         simulator.start()
 
         try:
-            it = self._run_test(simulator, set())
+            it = self._run_test(simulator, [])
             gen_tx = next(it)
             txA_hash = gen_tx.latest_transactions[0]
         finally:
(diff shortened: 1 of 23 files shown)


tests/consensus/test_soft_voided4.py
```diff
@@ -10,7 +13,11 @@
 class BaseSoftVoidedTestCase(SimulatorTestCase):
     seed_config = 5988775361793628169
 
-    def _run_test(self, simulator, soft_voided_tx_ids):
+    def _run_test(
+        self,
+        simulator: Simulator,
+        soft_voided_tx_ids: set[VertexId] | list[VertexId]
```
msbrogli, line 19: I guess `soft_voided_tx_ids` must be a list. Maybe change it here and fix any wrong callers.

tests/event/test_base_event.py
```diff
@@ -70,8 +70,8 @@ def test_create_base_event(event_id, group_id):
     assert event.dict() == expected
 
 
-@pytest.mark.parametrize('event_id', [None, -1, -1000])
-def test_create_base_event_fail_id(event_id):
+@pytest.mark.parametrize('event_id', [-1, -1000])
```
msbrogli, line 73: Did you remove the `None` test case?

tests/pubsub/test_pubsub.py
```diff
@@ -1,10 +1,10 @@
-from hathor.pubsub import HathorEvents, PubSubManager
+from hathor.pubsub import EventArguments, HathorEvents, PubSubManager
 from tests.unittest import TestCase
 
 
 class PubSubTestCase(TestCase):
-    def test_duplicate_subscribe(self):
-        def noop():
+    def test_duplicate_subscribe(self) -> None:
+        def noop(event: HathorEvents, args: EventArguments) -> None:
```
msbrogli, line 7: Have you added arguments in the `noop()` function?

msbrogli (approved):

tests/event/test_base_event.py
glevco: Yes, `BaseEvent.id` is not `Optional` so there's no reason to test `None` here, this test was not necessary. Now that it's typed, mypy complained about using an optional value in a non-optional property.

tests/pubsub/test_pubsub.py
glevco: Yes, the callable used in `PubSub.subscribe` is fully typed and expects those arguments. This was not enforced before because this test was not typed.

tests/consensus/test_soft_voided4.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/971/commits/eea7f836df7ecdbaedb30ae29fd4ca6ede81c3fc

msbrogli: Thanks!

tests/event/test_base_event.py
msbrogli: Got it. Thanks!

tests/pubsub/test_pubsub.py
msbrogli: Got it. Thanks!

msbrogli (dismissed):

jansegre (dismissed):

tests/crypto/test_util.py
```diff
@@ -13,22 +15,24 @@
 
 
 class CryptoUtilTestCase(unittest.TestCase):
-    def setUp(self):
+    def setUp(self) -> None:
         super().setUp()
-        self.private_key = ec.generate_private_key(ec.SECP256K1(), default_backend())
+        self.private_key = cast(
+            EllipticCurvePrivateKeyWithSerialization, ec.generate_private_key(ec.SECP256K1(), default_backend())
+        )
```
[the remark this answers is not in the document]
glevco: It does, but the function returns an `EllipticCurvePrivateKey`, while we expect an `EllipticCurvePrivateKeyWithSerialization`.

In the source, this looks like it's actually just an alias: https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/asymmetric/ec.py#L125

But for some reason mypy complained about this, so I had to the cast. Maybe we could remove all mentions of `EllipticCurvePrivateKeyWithSerialization` from our code and simply use `EllipticCurvePrivateKey` everywhere.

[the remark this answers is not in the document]
jansegre: Yeah, I think it'd be better to refactor `EllipticCurvePrivateKeyWithSerialization` uses to `EllipticCurvePrivateKey` if that fixes any mypy complaints.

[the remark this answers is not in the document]
glevco: I was wrong about this, they do have different methods. Which is weird considering the definition in the `ec.py` file linked above, but they're different in the `ec.pyi` file, and confirmed in the [docs](https://cryptography.io/en/3.3.2/hazmat/primitives/asymmetric/ec.html#cryptography.hazmat.primitives.asymmetric.ec.EllipticCurvePrivateKeyWithSerialization).

I changed it to use `assert isinstance` instead of a cast in https://github.com/HathorNetwork/hathor-core/pull/971/commits/2c62ee453e903da1d510e8b4c0858493dbc03631, which is also done in other parts of the code.

msbrogli (approved):

jansegre (approved):

PR #973, opened 2024-03-14
refactor(mypy): add stricter rules to p2p tests [part IV/VI]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/972

### Motivation

This PR continues to improve the strictness of mypy in test modules. For more info see the first PR in this series, https://github.com/HathorNetwork/hathor-core/pull/969.

### Acceptance Criteria

- Add stricter mypy rules to `tests.p2p` module and update it to conform.

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

PR #974, opened 2024-03-14
refactor(mypy): add stricter rules to unittest and utils [part V/VI]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/973

### Motivation

This PR continues to improve the strictness of mypy in test modules. For more info see the first PR in this series, https://github.com/HathorNetwork/hathor-core/pull/969.

### Acceptance Criteria

- Add stricter mypy rules to `tests.unittest` and `tests.utils` modules and update them to conform.

### Checklist

### DIFF
--- a/tests/unittest.py
+++ b/tests/unittest.py
@@ -3,7 +3,7 @@ import secrets
 import shutil
 import tempfile
 import time
-from typing import Iterator, Optional
+from typing import Callable, Collection, Iterable, Iterator, Optional
 from unittest import main as ut_main
 
 from structlog import get_logger
@@ -16,13 +16,17 @@ from hathor.conf.get_settings import get_global_settings
 from hathor.daa import DifficultyAdjustmentAlgorithm, TestMode
 from hathor.event import EventManager
 from hathor.event.storage import EventStorage
+from hathor.manager import HathorManager
 from hathor.p2p.peer_id import PeerId
+from hathor.p2p.sync_v1.agent import NodeSyncTimestamp
+from hathor.p2p.sync_v2.agent import NodeBlockSync
 from hathor.p2p.sync_version import SyncVersion
 from hathor.pubsub import PubSubManager
 from hathor.reactor import ReactorProtocol as Reactor, get_global_reactor
 from hathor.simulator.clock import MemoryReactorHeapClock
-from hathor.transaction import BaseTransaction
+from hathor.transaction import BaseTransaction, Block, Transaction
 from hathor.transaction.storage.transaction_storage import TransactionStorage
+from hathor.types import VertexId
 from hathor.util import Random, not_none
 from hathor.wallet import BaseWallet, HDWallet, Wallet
 from tests.test_memory_reactor_clock import TestMemoryReactorClock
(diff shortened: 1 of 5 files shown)


tests/unittest.py
```diff
@@ -33,9 +37,8 @@
 USE_MEMORY_STORAGE = os.environ.get('HATHOR_TEST_MEMORY_STORAGE', 'false').lower() == 'true'
 
 
-def shorten_hash(container):
-    container_type = type(container)
-    return container_type(h[-2:].hex() for h in container)
+def shorten_hash(container: Collection[bytes]) -> Iterable[str]:
+    return map(lambda hash_bytes: hash_bytes[-2:].hex(), container)
```
msbrogli, line 41: I guess this new code does something different then the original code.

tests/unittest.py
```diff
@@ -169,15 +172,13 @@ def get_builder(self, network: str) -> TestBuilder:
...
 
-    def create_peer_from_builder(self, builder, start_manager=True):
+    def create_peer_from_builder(self, builder: Builder, start_manager: bool = True) -> HathorManager:
         artifacts = builder.build()
         manager = artifacts.manager
 
         if artifacts.rocksdb_storage:
             self._pending_cleanups.append(artifacts.rocksdb_storage.close)
 
-        manager.avg_time_between_blocks = 0.0001
```
msbrogli, line 179: Why are you changing this?

msbrogli (dismissed):

tests/unittest.py
glevco: After this test was typed, mypy complained that `HathorManager` does not have an `avg_time_between_blocks` property. Therefore, this was useless (probably an artifact from some older code) and removing it doesn't break the test.

tests/unittest.py
glevco: It is implemented differently, but the result should be the same. I changed it because it was hard to correctly type it with the previous implementation, and this function is only ever used in debug logs.

jansegre: Can we rename this to `short_hashes`? The signature makes it more clear that it takes a list/container instead of a single bytes instance, which is what I thought it did from the function name.

jansegre (dismissed):

tests/unittest.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/974/commits/6ff78af167152e8f0737a5f46f6096ccc0a6c419

tests/unittest.py
glevco: Readded as a commented line in https://github.com/HathorNetwork/hathor-core/pull/974/commits/6ff78af167152e8f0737a5f46f6096ccc0a6c419, as discussed in https://github.com/HathorNetwork/hathor-core/pull/973#discussion_r1534692226

msbrogli: Thanks!

tests/unittest.py
jansegre: I agree there's a slight difference in meaning, the original would return a `set` when the input is a `set`. Typing wise the original behavior could be described with generics:

```python
C = TypeVar('C', bound=Collection)
def short_hashes(container: C[bytes]) -> C[str]
```

But I'm OK either way.

jansegre (approved):

tests/unittest.py
glevco: @jansegre I don't think Python supports type arguments on type vars, at least not like this. I wish this was Rust 😂

jansegre: I see, apparently it's a really complex feature from mypy's POV: https://github.com/python/typing/issues/548

msbrogli: Ok. Thanks!

msbrogli (approved):

PR #977, opened 2024-03-14
fix(indexes): Fix wrong order in MemoryTxGroupIndex

msbrogli:
### Motivation

`MemoryTxGroupIndex._get_sorted_from_key()` is returning transactions in the wrong order. The correct order is sorted by (timestamp, hash) but it is actually only sorting by hash.

### Acceptance Criteria

1. Replace `set[bytes]` to `set[tuple[int, bytes]]` so we can simply sort before returning the results.
2. Fix `_get_sorted_from_key()` sorting the tuple but returning only the hashes.

### Checklist

### DIFF
--- a/hathor/indexes/memory_tx_group_index.py
+++ b/hathor/indexes/memory_tx_group_index.py
@@ -31,7 +31,7 @@ class MemoryTxGroupIndex(TxGroupIndex[KT]):
     """Memory implementation of the TxGroupIndex. This class is abstract and cannot be used directly.
     """
 
-    index: defaultdict[KT, set[bytes]]
+    index: defaultdict[KT, set[tuple[int, bytes]]]
 
     def __init__(self) -> None:
         self.force_clear()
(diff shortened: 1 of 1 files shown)


hathor/indexes/memory_tx_group_index.py
```diff
@@ -57,13 +57,14 @@ def remove_tx(self, tx: BaseTransaction) -> None:
...
+            self.index[key].discard((tx.timestamp, tx.hash))
 
     def _get_from_key(self, key: KT) -> Iterable[bytes]:
-        yield from self.index[key]
+        for _, h in self.index[key]:
+            yield h
 
     def _get_sorted_from_key(self, key: KT) -> Iterable[bytes]:
-        return sorted(self.index[key])
+        return [h for t, h in sorted(self.index[key])]
```
pedroferreira1, line 67:
```suggestion
        return [h for _, h in sorted(self.index[key])]
```

pedroferreira1 (approved):

hathor/indexes/memory_tx_group_index.py
glevco: This reminds me that I wanted to find some linter that is able to alert about unused variables. Sometimes they're the source of bugs and we have a lot of code that we could clean up with this. Our current setup doesn't catch it.

glevco (approved):

PR #978, opened 2024-03-15
refactor(indexes): Optimize RocksDBAddressIndex to handle pagination in O(log n)

pedroferreira1:
### Motivation

We had some users with wallets that had many transactions in single addresses, then the API request `address_history` got really slow.

More details [here](https://github.com/HathorNetwork/internal-issues/issues/256)

### Acceptance Criteria

- Add parameter in the `get_sorted_from_key` method of tx group index to find the first tx to return the iterator from.
- Refactor address history API to remove O(n) operation and use rocksdb tree to find tx with hash and improve performance.

### Checklist

### DIFF
--- a/hathor/indexes/address_index.py
+++ b/hathor/indexes/address_index.py
@@ -92,8 +92,11 @@ class AddressIndex(TxGroupIndex[str]):
         raise NotImplementedError
 
     @abstractmethod
-    def get_sorted_from_address(self, address: str, tx: Optional[BaseTransaction] = None) -> Iterable[bytes]:
+    def get_sorted_from_address(self, address: str, tx_start: Optional[BaseTransaction] = None) -> Iterable[bytes]:
         """ Get a sorted list of transaction hashes of an address
+
+        The parameter tx_start means the transaction from which the iterator will start, and it is used for pagination.
+        When it's None it means that the iterator will start from the beginning.
         """
         raise NotImplementedError
 
--- a/hathor/indexes/address_index.py
+++ b/hathor/indexes/address_index.py
@@ -95,8 +95,8 @@ class AddressIndex(TxGroupIndex[str]):
     def get_sorted_from_address(self, address: str, tx_start: Optional[BaseTransaction] = None) -> Iterable[bytes]:
         """ Get a sorted list of transaction hashes of an address
 
-        The parameter tx_start means the transaction from which the iterator will start, and it is used for pagination.
-        When it's None it means that the iterator will start from the beginning.
+        `tx_start` serves as a pagination marker, indicating the starting position for the iteration.
+        When tx_start is None, the iteration begins from the initial element.
         """
         raise NotImplementedError
 
--- a/hathor/indexes/address_index.py
+++ b/hathor/indexes/address_index.py
@@ -92,7 +92,7 @@ class AddressIndex(TxGroupIndex[str]):
         raise NotImplementedError
 
     @abstractmethod
-    def get_sorted_from_address(self, address: str) -> list[bytes]:
+    def get_sorted_from_address(self, address: str, tx: Optional[BaseTransaction] = None) -> Iterable[bytes]:
         """ Get a sorted list of transaction hashes of an address
         """
         raise NotImplementedError
--- a/hathor/indexes/rocksdb_tx_group_index.py
+++ b/hathor/indexes/rocksdb_tx_group_index.py
@@ -108,9 +108,15 @@ class RocksDBTxGroupIndex(TxGroupIndex[KT], RocksDBIndexUtils):
             self._db.delete((self._cf, self._to_rocksdb_key(key, tx)))
 
     def _get_from_key(self, key: KT) -> Iterable[bytes]:
+        return self._util_get_from_key(key)
+
+    def _get_sorted_from_key(self, key: KT, tx: Optional[BaseTransaction] = None) -> Iterable[bytes]:
+        return self._util_get_from_key(key, tx)
+
+    def _util_get_from_key(self, key: KT, tx: Optional[BaseTransaction] = None) -> Iterable[bytes]:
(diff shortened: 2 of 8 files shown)


hathor/indexes/address_index.py
```diff
@@ -92,7 +92,7 @@ def get_from_address(self, address: str) -> list[bytes]:
         raise NotImplementedError
 
     @abstractmethod
-    def get_sorted_from_address(self, address: str) -> list[bytes]:
+    def get_sorted_from_address(self, address: str, tx: Optional[BaseTransaction] = None) -> Iterable[bytes]:
```
glevco, line 95: Maybe add info about what is this new argument in the docstring? Or maybe a more expressive name than `tx`

hathor/indexes/tx_group_index.py
```diff
@@ -49,7 +49,7 @@ def _get_from_key(self, key: KT) -> Iterable[bytes]:
         raise NotImplementedError
 
     @abstractmethod
-    def _get_sorted_from_key(self, key: KT) -> Iterable[bytes]:
+    def _get_sorted_from_key(self, key: KT, tx: Optional[BaseTransaction] = None) -> Iterable[bytes]:
```
glevco, line 52: Maybe add info about what is this new argument in the docstring? Or maybe a more expressive name than `tx`

glevco (approved):

hathor/indexes/address_index.py
msbrogli: Maybe `after` or `after_tx`?

hathor/indexes/rocksdb_tx_group_index.py
```diff
@@ -108,9 +108,15 @@ def remove_tx(self, tx: BaseTransaction) -> None:
             self._db.delete((self._cf, self._to_rocksdb_key(key, tx)))
 
     def _get_from_key(self, key: KT) -> Iterable[bytes]:
```
msbrogli, line 110: Why not just add a `tx` here?

hathor/wallet/resources/thin_wallet/address_history.py
```diff
@@ -181,31 +176,29 @@ def get_address_history(self, addresses: list[str], ref_hash: Optional[str]) ->
...
+                except TransactionDoesNotExist:
                     return json_dumpb({
                         'success': False,
-                        'message': 'Hash {} is not a transaction from the address {}.'.format(ref_hash, address)
+                        'message': 'Hash {} is not a transaction hash.'.format(ref_hash)
                     })
 
-            # Slice the hashes array from the start_index
-            to_iterate = hashes[start_index:]
+            # In rocksdb we can get an iterator directly from the hash
```
msbrogli, line 189:
```suggestion
            # The address index returns an iterable that starts at `tx`.
```

msbrogli (approved):

hathor/indexes/rocksdb_tx_group_index.py
pedroferreira1: This method implements an abstract method of the `TxGroupIndex`.  Adding a `tx` parameter here means that we should add this to the abstract method and all other classes that extend this `TxGroupIndex` class and this is particular of the rocksdb implementation.

Apart from that, the `get_from_key` method doesn't need any `tx` because it doesn't do any kind of sorting, see the docstring:

> """Get all transactions that have a given key."""

That's why I decided to create a util method instead of changing this one.

hathor/indexes/address_index.py
pedroferreira1: I changed to `tx_start` the parameter name and improved the docstring. I decided not to use `after`/`after_tx` because we don't return the iterator after it, we return including it.

What do you think? https://github.com/HathorNetwork/hathor-core/pull/978/commits/57558a502daa480f03a942e50f8c5861fc676ff1

hathor/indexes/tx_group_index.py
pedroferreira1: Improved the argument name and docstring. What do you think? https://github.com/HathorNetwork/hathor-core/pull/978/commits/57558a502daa480f03a942e50f8c5861fc676ff1

hathor/indexes/address_index.py
glevco: Works for me! Thanks

glevco (dismissed):

hathor/indexes/address_index.py
```diff
@@ -92,8 +92,11 @@ def get_from_address(self, address: str) -> list[bytes]:
         raise NotImplementedError
 
     @abstractmethod
-    def get_sorted_from_address(self, address: str) -> list[bytes]:
+    def get_sorted_from_address(self, address: str, tx_start: Optional[BaseTransaction] = None) -> Iterable[bytes]:
         """ Get a sorted list of transaction hashes of an address
+
+        The parameter tx_start means the transaction from which the iterator will start, and it is used for pagination.
+        When it's None it means that the iterator will start from the beginning.
```
msbrogli, line 99:
```suggestion
        `tx_start` serves as a pagination marker, indicating the starting position for the iteration. When tx_start is None, the iteration begins from the initial element.
```

hathor/indexes/address_index.py
msbrogli, line 1: I think we should rename the commit title and the PR title to "refactor(indexes): Optimize AddressIndex to handle pagination in O(log n)".

msbrogli (approved):

hathor/indexes/rocksdb_tx_group_index.py
msbrogli: Got it. Thanks for the explanation.

hathor/indexes/address_index.py
msbrogli: Thanks!

msbrogli (approved):

glevco (approved):

PR #981, opened 2024-03-21
feat(p2p): add ability to update peer_id.json with SIGUSR1

glevco:
### Motivation

Implement ability to update the full node public IP while the sync is running, as requested in the issue: resolves https://github.com/HathorNetwork/internal-issues/issues/261.

Here are the new ways of interacting with entrypoints and the hostname:

#### Sysctl `hostname` command

Get and manually set the full node's hostname. Entrypoints will automatically reflect this change, and connections will be reset.

#### Sysctl `refresh_auto_hostname` command

Automatically set the hostname to an auto discovered one. Entrypoints will automatically reflect this change, and connections will be reset.

#### Sysctl `reload_entrypoints_and_connections` command

Reload the `peer_id.json` file, replacing the entrypoints list with the new one. Then, connections will be reset.

#### `SIGUSR1` command

This is just a shortcut to the sysctl `reload_entrypoints_and_connections` command.

### Acceptance Criteria

- Remove `CliBuilder.create_peer_id()` and add `PeerId.create_from_json_path()`.
- Refactor `ConnectionsManager.listen()` so it uses the listen deferred and persists the listened addresses.
- Implement `ConnectionsManager.update_hostname_entrypoints()` and `reset_all_connections()`.
- Move `peer_id.json` file path to the `PeerId` class, with the `reload_entrypoints_from_source_file()` method.
- Add sysctl commands for getting/setting `hostname`, refreshing the auto hostname, and resetting all connections.
- Add `peer_id.json` reload in `SIGUSR1`.

### Checklist

### DIFF
--- a/hathor/builder/cli_builder.py
+++ b/hathor/builder/cli_builder.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import getpass
-import json
 import os
 import platform
 import sys
@@ -96,8 +95,7 @@ class CliBuilder:
         self.log = logger.new()
         self.reactor = reactor
 
-        peer_id = self.create_peer_id()
-
+        peer_id = PeerId.create_from_json_path(self._args.peer) if self._args.peer else PeerId()
         python = f'{platform.python_version()}-{platform.python_implementation()}'
 
         self.log.info(
@@ -367,7 +365,7 @@ class CliBuilder:
             self.log.warn('--memory-indexes is implied for memory storage or JSON storage')
 
         for description in self._args.listen:
-            p2p_manager.add_listen_address(description)
+            p2p_manager.add_listen_address_description(description)
 
         if self._args.peer_id_blacklist:
             self.log.info('with peer id blacklist', blacklist=self._args.peer_id_blacklist)
@@ -396,14 +394,6 @@ class CliBuilder:
                 sys.exit(-1)
             print('Hostname discovered and set to {}'.format(hostname))
         return hostname
-
-    def create_peer_id(self) -> PeerId:
-        if not self._args.peer:
-            peer_id = PeerId()
-        else:
-            data = json.load(open(self._args.peer, 'r'))
-            peer_id = PeerId.create_from_json(data)
-        return peer_id
 
     def create_wallet(self) -> BaseWallet:
         if self._args.wallet == 'hd':
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -89,31 +89,33 @@ class HathorManager:
     # This is the interval to be used by the task to check if the node is synced
     CHECK_SYNC_STATE_INTERVAL = 30  # seconds
 
-    def __init__(self,
-                 reactor: Reactor,
-                 *,
-                 settings: HathorSettings,
-                 pubsub: PubSubManager,
-                 consensus_algorithm: ConsensusAlgorithm,
-                 daa: DifficultyAdjustmentAlgorithm,
-                 peer_id: PeerId,
-                 tx_storage: TransactionStorage,
-                 p2p_manager: ConnectionsManager,
-                 event_manager: EventManager,
-                 feature_service: FeatureService,
-                 bit_signaling_service: BitSignalingService,
-                 verification_service: VerificationService,
-                 cpu_mining_service: CpuMiningService,
-                 network: str,
-                 execution_manager: ExecutionManager,
-                 hostname: Optional[str] = None,
-                 wallet: Optional[BaseWallet] = None,
-                 capabilities: Optional[list[str]] = None,
-                 checkpoints: Optional[list[Checkpoint]] = None,
-                 rng: Optional[Random] = None,
-                 environment_info: Optional[EnvironmentInfo] = None,
-                 full_verification: bool = False,
-                 enable_event_queue: bool = False):
+    def __init__(
+        self,
+        reactor: Reactor,
+        *,
+        settings: HathorSettings,
+        pubsub: PubSubManager,
+        consensus_algorithm: ConsensusAlgorithm,
+        daa: DifficultyAdjustmentAlgorithm,
+        peer_id: PeerId,
+        tx_storage: TransactionStorage,
+        p2p_manager: ConnectionsManager,
+        event_manager: EventManager,
+        feature_service: FeatureService,
+        bit_signaling_service: BitSignalingService,
+        verification_service: VerificationService,
+        cpu_mining_service: CpuMiningService,
+        network: str,
+        execution_manager: ExecutionManager,
+        hostname: Optional[str] = None,
+        wallet: Optional[BaseWallet] = None,
+        capabilities: Optional[list[str]] = None,
+        checkpoints: Optional[list[Checkpoint]] = None,
+        rng: Optional[Random] = None,
+        environment_info: Optional[EnvironmentInfo] = None,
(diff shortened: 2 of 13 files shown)


hathor/manager.py
```diff
@@ -1173,6 +1177,15 @@ def get_cmd_path(self) -> Optional[str]:
         """Return the cmd path. If no cmd path is set, returns None."""
         return self._cmd_path
 
+    def update_peer_id(self) -> None:
+        """Update this manager's PeerId from the json file. Only entrypoints can be changed."""
+        if not self._peer_id_json_path:
+            return
+
+        new_peer_id = PeerId.create_from_json_path(self._peer_id_json_path)
+        assert new_peer_id.id == self.my_peer.id, 'The Peer ID cannot be changed.'
```
jansegre, line 1186: Since this comes from user data that the node loads, I think it should not be an assert. It could be an error log that says the new `peer.json` file was ignored because the peer_id doesn't match, or something like that. But without raising any exceptions.

glevco: Agreed, changed in https://github.com/HathorNetwork/hathor-core/pull/981/commits/cccfe886117aea2037a5c831fc23d8f1eeb14ab7

jansegre (approved):

hathor/manager.py
```diff
@@ -135,6 +138,7 @@ def __init__(self,
         self._settings = settings
         self.daa = daa
         self._cmd_path: Optional[str] = None
+        self._peer_id_json_path = peer_id_json_path
```
msbrogli, line 141: Why not use sysctl and get a file path as argument? On the other hand, we could have other commands such as `refresh_ip_address()` or `refresh_public_ip_address()`.

hathor/manager.py
msbrogli, line 1: Note that we currently detect the public ip address in `builder/cli_builder.py:L403: hostname = discover_hostname()`. So we could just mark whether it was manually given by the user or auto detected. Then we can create one more sysctl commands to update it. For example: `p2p.refresh_detected_hostname` or `p2p.update_peer_id.from_path`. What do you think?

msbrogli (changes requested):

hathor/manager.py
jansegre: The downside is that the `peer.json` will not reflect the current state and that a restart would revert it to the original state.

hathor/manager.py
jansegre: I think a sysctl command to reload the `peer.json` file would be good, but it would also need to refresh all the connections with SIGUSR1, so why not make SIGUSR1 also reload the `peer.json` file?

hathor/manager.py
msbrogli: But I don't think the entrypoints are commonly stored there. I'd say, in most cases, it is autodetected by the auto hostname features in the full node. So it wouldn't revert to the original state.

hathor/manager.py
msbrogli: I guess we should make SIGUSR1 refresh auto detect and reload peer.json. But we will probably deprecate this signal soon, right?

msbrogli (changes requested):

hathor/p2p/manager.py
```diff
@@ -661,20 +665,43 @@ def listen(self, description: str, use_ssl: Optional[bool] = None) -> IStreamSer
...
+        deferred.addCallback(self._on_listen_success, description)
+
+    def _on_listen_success(self, listening_port: IListeningPort, description: str) -> None:
+        """Callback to be called when listening to an endpoint succeeds."""
+        self.log.info('success listening on', endpoint=description)
+        address = listening_port.getHost()
+
+        if not isinstance(address, (IPv4Address, IPv6Address)):
+            self.log.error(f'unhandled address type for endpoint "{description}": {str(type(address))}')
+            return
```
glevco, line 679: Before, we were using `getattr()` instead of correctly handling the types. Should we handle anything other than Twisted's `IPv4Address` and `IPv6Address`?

glevco:
@msbrogli @jansegre This PR changed considerably, could you please re-review it?

hathor/builder/cli_builder.py
```diff
@@ -322,6 +320,7 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
             verification_service=verification_service,
             cpu_mining_service=cpu_mining_service,
             execution_manager=execution_manager,
+            peer_id_json_path=self._args.peer,
```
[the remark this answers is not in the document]
glevco: Because the created `PeerId` is used by other classes that receive it during building, so it must be ready before `HathorManager`. However, I changed this so the file path is persisted in the `PeerId` itself, simplifying things and allowing me to remove this argument from the manager.

hathor/manager.py
glevco: This was changed in https://github.com/HathorNetwork/hathor-core/pull/981/commits/83e2acafa848c1d610b95d23c3065dcf182f9bac

hathor/manager.py
glevco: This was changed in https://github.com/HathorNetwork/hathor-core/pull/981/commits/83e2acafa848c1d610b95d23c3065dcf182f9bac

jansegre (dismissed):

msbrogli (approved):

hathor/manager.py

hathor/manager.py
msbrogli: Thanks!

hathor/builder/cli_builder.py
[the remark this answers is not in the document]
msbrogli: Got it. Thanks!

hathor/p2p/peer_id.py
```diff
@@ -408,3 +420,20 @@ def validate_certificate(self, protocol: 'HathorProtocol') -> bool:
             return False
 
         return True
+
+    def reload_entrypoints_from_source_file(self) -> None:
+        """Update this PeerId's entrypoints from the json file."""
+        if not self.source_file:
+            return
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/981/commits/647f42365381121919e2ed32baa3e26843e72fbe

hathor/manager.py
```diff
@@ -1173,6 +1175,13 @@ def get_cmd_path(self) -> Optional[str]:
         """Return the cmd path. If no cmd path is set, returns None."""
         return self._cmd_path
 
+    def set_hostname(self, new_hostname: str) -> None:
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/981/commits/647f42365381121919e2ed32baa3e26843e72fbe

jansegre (dismissed):

hathor/p2p/manager.py
jansegre: I think `IPv4Address` and `IPv6Address` should be the only 2 we support (even though IPv6 support is not complete).

hathor/p2p/peer_id.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/manager.py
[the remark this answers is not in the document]

msbrogli (dismissed):

jansegre (approved):

msbrogli (approved):

PR #982, opened 2024-03-21
feat(cli): add argument to make sync-v1 unavailable

jansegre:
### Motivation

In preparation to phase-out sync-v1, we would benefit from having a method for starting the node without sync-v1 indexes. This PR does not go into the details of which index is used, it only makes it possible to start with a node where sync-v1 is "not available" as opposed to just "disabled"; "not available" means it cannot be enabled in runtime.

### Acceptance Criteria

- Refactor the builder/cli to make it clear to understand how each sync protocol is enabled or made available
- Add `--x-remove-sync-v1` CLI option to make sync-v1 unavailable from the start

### Checklist

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from enum import Enum
+from enum import Enum, IntEnum
 from typing import Any, Callable, NamedTuple, Optional, TypeAlias
 
 from structlog import get_logger
@@ -53,6 +53,34 @@ from hathor.vertex_handler import VertexHandler
 from hathor.wallet import BaseWallet, Wallet
 
 logger = get_logger()
+
+
+class SyncSupportLevel(IntEnum):
+    UNAVAILABLE = 0  # not possible to enable at runtime
+    DISABLED = 1  # available but disabled by default, possible to enable at runtime
+    ENABLED = 2  # available and enabled by default, possible to disable at runtime
+
+    @classmethod
+    def add_factories(cls,
+                      p2p_manager: ConnectionsManager,
+                      sync_v1_support: 'SyncSupportLevel',
+                      sync_v2_support: 'SyncSupportLevel',
+                      ) -> None:
+        """Adds the sync factory to the manager according to the support level."""
+        from hathor.p2p.sync_v1.factory import SyncV11Factory
+        from hathor.p2p.sync_v2.factory import SyncV2Factory
+        from hathor.p2p.sync_version import SyncVersion
+
+        # sync-v1 support:
+        if sync_v1_support > cls.UNAVAILABLE:
+            p2p_manager.add_sync_factory(SyncVersion.V1_1, SyncV11Factory(p2p_manager))
+        if sync_v1_support is cls.ENABLED:
+            p2p_manager.enable_sync_version(SyncVersion.V1_1)
+        # sync-v2 support:
+        if sync_v2_support > cls.UNAVAILABLE:
+            p2p_manager.add_sync_factory(SyncVersion.V2, SyncV2Factory(p2p_manager))
+        if sync_v2_support is cls.ENABLED:
+            p2p_manager.enable_sync_version(SyncVersion.V2)
 
 
 class StorageType(Enum):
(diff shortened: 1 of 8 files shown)


hathor/builder/builder.py
```diff
@@ -145,8 +151,8 @@ def __init__(self) -> None:
         self._enable_tokens_index: bool = False
         self._enable_utxo_index: bool = False
 
-        self._enable_sync_v1: bool = True
-        self._enable_sync_v2: bool = False
+        self._sync_v1_support: SyncSupportLevel = SyncSupportLevel.DISABLED
+        self._sync_v2_support: SyncSupportLevel = SyncSupportLevel.ENABLED
```
glevco, line 155: Are we purposefully changing the default value for each of them? Just checking if this was a mistake since it was not in the acceptance criteria.

hathor/builder/cli_builder.py
```diff
@@ -293,11 +300,15 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
...
+        if sync_v1_support > SyncSupportLevel.UNAVAILABLE:
+            p2p_manager.add_sync_factory(SyncVersion.V1_1, SyncV11Factory(p2p_manager))
+        if sync_v1_support is SyncSupportLevel.ENABLED:
             p2p_manager.enable_sync_version(SyncVersion.V1_1)
-        if enable_sync_v2:
+        # sync-v2 support:
+        if sync_v2_support > SyncSupportLevel.UNAVAILABLE:
+            p2p_manager.add_sync_factory(SyncVersion.V2, SyncV2Factory(p2p_manager))
+        if sync_v2_support is SyncSupportLevel.ENABLED:
             p2p_manager.enable_sync_version(SyncVersion.V2)
```
glevco, line 312: This code is duplicate from the builder, should we create a method somewhere?

hathor/cli/run_node.py
```diff
@@ -131,6 +131,8 @@ def create_parser(cls) -> ArgumentParser:
                                help='Enable running both sync protocols.')
         sync_args.add_argument('--sync-v1-only', action='store_true', help='Disable support for running sync-v2.')
         sync_args.add_argument('--sync-v2-only', action='store_true', help='Disable support for running sync-v1.')
+        sync_args.add_argument('--x-remove-sync-v1', action='store_true', help='Make sync-v1 unavailable, thus '
+                               'possible to be enable in runtime.')
```
glevco, line 135: Typo, "impossible"?

msbrogli (dismissed):

hathor/builder/builder.py
jansegre: I'll add to the acceptance criteria, but this is an internal default. Maybe it should have both protocols as unavailable by default. Both builders have a default mechanism that overrides this, tests also have a mechanism that always sets the sync version through the builder and the cli also has a mechanism that sets a version through the builder, so in the end the value here doesn't have any effect on any code.

hathor/cli/run_node.py
jansegre: Fixed.

hathor/builder/cli_builder.py
jansegre: Moved to a classmethod `SyncSupportLevel.add_factories`.

hathor/builder/builder.py
jansegre: I set both to `SyncSupportLevel.UNAVAILABLE` which is more consistent with all the other properties having an "empty" value as default.

hathor/builder/cli_builder.py
```diff
@@ -49,9 +49,10 @@
 
 
 class SyncChoice(Enum):
-    V1_ONLY = auto()
-    V2_ONLY = auto()
-    BRIDGE = auto()
+    V1_DEFAULT = auto()  # v2 can be enabled in runtime
```
msbrogli, line 52: Maybe `v1 enabled, v2 disabled but can be enabled in runtime`?

msbrogli (approved):

hathor/builder/builder.py
msbrogli: I like it. Thanks, Jan!

hathor/builder/cli_builder.py
jansegre: Done.

glevco (dismissed):

hathor/builder/builder.py
glevco: Thanks!

hathor/builder/cli_builder.py
glevco: Thanks!

hathor/cli/run_node.py
glevco: Thanks!

msbrogli (approved):

PR #983, opened 2024-03-25
tests(verification): add missing assertions

glevco:
### Motivation

I noticed some verification tests were incorrect by a missing assertion.

### Acceptance Criteria

- Fix wrong Mock used in assertions from `verify_pow_wrapped` to `verify_aux_pow_wrapped`.

### Checklist


jansegre (approved):
So, they weren't failing because `verify_pow_wrapped` is also called, but we want to make sure it calls `verify_aux_pow_wrapped`? So in theory we could leave both asserts then?

glevco:
@jansegre yes, we want to assert that both of them are called. `verify_pow_wrapped` is already asserted a few lines above each changed line

PR #985, opened 2024-04-01
fix(events): add missing signal_bits to event schema

glevco:
### Motivation

The `signal_bits` field was missing from Reliable Integration events. This is not a problem because no Reliable Integration client uses this field, but for completeness, this PR adds it.

### Acceptance Criteria

- Add `signal_bits` to `TxData`.

### Checklist


jansegre (approved):

hathor/event/model/event_data.py
```diff
@@ -101,6 +101,7 @@ class TxData(BaseEventData, extra=Extra.ignore):
     hash: str
     nonce: Optional[int] = None
     timestamp: int
+    signal_bits: int
```
msbrogli, line 104: Should we add the list of features related to these bits?

msbrogli (approved):

hathor/event/model/event_data.py
glevco: I'm not sure if we should bloat the event with that data, since it's calculated and not part of the block itself. This information is available though an API, and an use case could integrate it to aggregate this information.

For example, https://node1.mainnet.hathor.network/v1a/feature?block=000000000000000003ec51ac33c5f16196818e44b89606fa6f3536d86f15e7cd returns this:

```json
{
  "signal_bits": [
    {
      "bit": 2,
      "signal": 0,
      "feature": "NOP_FEATURE_3",
      "feature_state": "STARTED"
    },
    {
      "bit": 3,
      "signal": 0,
      "feature": "NOP_FEATURE_4",
      "feature_state": "STARTED"
    }
  ]
}
```

msbrogli (approved):

PR #987, opened 2024-04-02
refactor: create VertexHandler [part 1/2]

glevco:
### Motivation

Move all `HathorManager.on_new_tx()` related code to its own new class, `VertexHandler`, improving organization and modularization, and as a prerequisite for the upcoming changes related to multiprocess verification.

### Acceptance Criteria

- Move `HathorManager.on_new_tx()` implementation to new `VertexHandler`, along with other methods used only by it.
  - Everything should be moved as-is, no behavior should change. The only exception is that the `conn` and `skip_block_weight_verification` arguments were removed, since they were not used.
- Keep `HathorManager.on_new_tx()` with a call to `VertexHandler.on_new_tx()`, compatibility with current code.
- Update `Builder` and `CliBuilder` accordingly.
- Fix `TestBuilder._get_reactor()` (it was not persisting the newly created reactor instance).

### Checklist

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -49,6 +49,7 @@ from hathor.transaction.storage import (
 from hathor.util import Random, get_environment_info, not_none
 from hathor.verification.verification_service import VerificationService
 from hathor.verification.vertex_verifiers import VertexVerifiers
+from hathor.vertex_handler import VertexHandler
 from hathor.wallet import BaseWallet, Wallet
 
 logger = get_logger()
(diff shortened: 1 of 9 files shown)


hathor/builder/builder.py
```diff
@@ -565,16 +593,14 @@ def force_memory_index(self) -> 'Builder':
...
-            assert self._wallet_unlock is None
             return self._wallet
 
         if self._wallet_directory is None:
             return None
-        wallet = Wallet(directory=self._wallet_directory)
+        self._wallet = Wallet(directory=self._wallet_directory)
         if self._wallet_unlock is not None:
-            wallet.unlock(self._wallet_unlock)
-        return wallet
```
glevco, line 577: The created wallet was not being persisted here, so I changed it to `self._wallet`. For that, I had to remove the assertions on the top. Adding this comment just to confirm whether there was any reason for the previous code.

msbrogli (approved):

jansegre (approved):

msbrogli (approved):

PR #992, opened 2024-04-03
fix(metadata): correctly initialize metadata voided_by

glevco:
### Motivation

### Acceptance Criteria

- Update `TransactionMetadata.__init__()` so it is initialized with `PARTIALLY_VALIDATED_ID` in its `voided_by`.
- Rename `TransactionMetadata.validation` to a private `_validation`, creating a getter and setter called `validation` for compatibility with current code.
- Move metadata logic related to updating its `voided_by` when its `validation` is changed from `BaseTransaction` to `TransactionMetadata`, including it in the newly created property setter.

### Checklist

### DIFF
--- a/tests/wallet/test_wallet.py
+++ b/tests/wallet/test_wallet.py
@@ -119,6 +119,7 @@ class BaseBasicWalletTest(unittest.TestCase):
         out = WalletOutputInfo(decode_address(key.address), BLOCK_REWARD, timelock=None)
         tx = w.prepare_transaction(Transaction, inputs=[], outputs=[out])
         tx.update_hash()
+        tx.get_metadata().voided_by = None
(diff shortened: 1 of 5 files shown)


tests/wallet/test_wallet.py
```diff
@@ -119,6 +119,7 @@ def test_block_increase_balance(self):
         out = WalletOutputInfo(decode_address(key.address), BLOCK_REWARD, timelock=None)
         tx = w.prepare_transaction(Transaction, inputs=[], outputs=[out])
         tx.update_hash()
+        tx.get_metadata().voided_by = None
```
msbrogli, line 122: Why do you need to do this?

msbrogli (changes requested):

PR #998, opened 2024-04-08
chore: update dependencies

glevco:
### Motivation

This PR updates some of our most important dependencies, fixing bugs, improving security, and possibly increasing performance. It resolves https://github.com/HathorNetwork/sec-hathor-core/issues/137.

### Breaking changes

I read the changelog of the most important dependencies to check if there was anything relevant. Here's the list:

#### Cryptography ([Changelog](https://cryptography.io/en/latest/changelog/))

* BACKWARDS INCOMPATIBLE: Support for OpenSSL less than 1.1.1d has been removed. Users on older version of OpenSSL will need to upgrade.
* BACKWARDS INCOMPATIBLE: Dropped support for LibreSSL < 3.7.
* BACKWARDS INCOMPATIBLE: Dropped support for macOS 10.10 and 10.11, macOS users must upgrade to 10.12 or newer.

#### pyOpenSSL ([Changelog](https://pypi.org/project/pyOpenSSL/))

- No relevant changes.

#### Twisted ([Changelog](https://github.com/twisted/twisted/blob/HEAD/NEWS.rst))

- Twisted now officially supports Python 3.11.

### Acceptance Criteria

- Update the following dev-dependencies to their respective latest version:
  - flake8
  - isort
  - mypy
  - mypy-zope
  - pytest
  - pytest-cov
  - flaky
  - pytest-xdist
  - yamllint
- Update the following dependencies to their respective latest version (**ATTENTION:** this includes updates to important dependencies like `twisted` and `cryptography`):
  - twisted
  - autobahn
  - cryptography
  - pyopenssl
  - requests
  - aiohttp
  - setproctitle
  - hathorlib
  - pydantic
  - typing-extensions

### Checklist


pyproject.toml
```diff
@@ -36,50 +36,50 @@ exclude = ["tests", "tests.*"]
...
 rocksdb = {git = "https://github.com/hathornetwork/python-rocksdb.git", markers = "sys_platform != 'win32'"}
-aiohttp = "~3.9.0"
+aiohttp = "~3.9.3"
 idna = "~3.4"
 setproctitle = "^1.2.2"
 sentry-sdk = {version = "^1.5.11", optional = true}
 structlog-sentry = {version = "^1.4.0", optional = true}
-hathorlib = "^0.5.2"
-pydantic = "~1.10.13"
+hathorlib = {git = "https://github.com/HathorNetwork/python-hathorlib.git", branch = "chore/update-deps" } # TODO: Use the latest version
```
glevco, line 79: Waiting on `python-hathorlib` to update this line.

msbrogli: Should we release a new version of `python-hathorlib`? If yes, let's create an internal issue to get this done ASAP.

hathor/p2p/utils.py
```diff
@@ -20,8 +20,8 @@
 import requests
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.backends.openssl.rsa import RSAPrivateKey
 from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPrivateKey
```
msbrogli, line 24: Are you sure this is still using OpenSSL (instead of a custom implementation from the package)?

hathor/wallet/resources/thin_wallet/send_tokens.py
```diff
@@ -155,7 +155,8 @@ def _render_POST_stratum(self, context: _Context) -> None:
 
         # When using stratum to solve pow, we already set timestamp and parents
         stratum_deferred: Deferred[None] = Deferred()
-        stratum_deferred.addCallback(self._stratum_deferred_resolve, request)
+        # FIXME: Skipping mypy on the line below for now, as it looks like it's wrong but we don't have tests for it.
+        stratum_deferred.addCallback(self._stratum_deferred_resolve, request)  # type: ignore
```
msbrogli, line 159: What's mypy error on this line?

hathor/wallet/wallet.py
```diff
@@ -17,9 +17,9 @@
 import os
 from typing import Any, Optional
 
-from cryptography.hazmat.backends.openssl.ec import _EllipticCurvePrivateKey
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey
```
msbrogli, line 22: Same here: Is it still using OpenSSL?

pyproject.toml
```diff
@@ -36,50 +36,50 @@ exclude = ["tests", "tests.*"]
...
 pywin32 = {version = "306", markers = "sys_platform == 'win32'"}
-requests = "=2.28.1"
+requests = "=2.31.0"
 service_identity = "~21.1.0"
 pexpect = "~4.8.0"
 intervaltree = "~3.1.0"
 structlog = "~22.3.0"
 rocksdb = {git = "https://github.com/hathornetwork/python-rocksdb.git", markers = "sys_platform != 'win32'"}
-aiohttp = "~3.9.0"
+aiohttp = "~3.9.3"
```
msbrogli, line 74: Minor version change. Have we checked the changes?

msbrogli (approved):

pyproject.toml
glevco: I already have the PR ready, in which I update the dependencies to the same versions. However for both cases I still have to check the changelogs for the updated dependencies, I'll do that soon.

hathor/p2p/utils.py
glevco: I removed this from this PR.

hathor/wallet/resources/thin_wallet/send_tokens.py
glevco: I removed this from this PR and it seemed to work on the current version.

hathor/wallet/wallet.py
glevco: I removed this from this PR.

hathor/wallet/resources/thin_wallet/send_tokens.py
glevco: The `_stratum_deferred_resolve` callback expects to receive a `_Context` as argument, but in `addCallback` it is set with a `Request` instead. I think this is actually incorrect and exceptions would be raised, but this is never used and will probably be removed soon.

hathor/p2p/utils.py
glevco: Yes, OpenSSL is just the backend. In fact we were already importing the same `cryptography.hazmat.primitives.asymmetric.rsa.RSAPrivateKey`, as a re-export in `cryptography.hazmat.backends.openssl.rsa`. Check it out in the source code for our current version (`38.0.3`):

https://github.com/pyca/cryptography/blob/38.0.3/src/cryptography/hazmat/backends/openssl/rsa.py#L31-L32

And in the source code for the new version (`42.0.5`), check that `RSAPrivateKey` is indeed being registered with an OpenSSL backend:

https://github.com/pyca/cryptography/blob/42.0.5/src/cryptography/hazmat/primitives/asymmetric/rsa.py#L67

hathor/wallet/wallet.py
glevco: Yes. Similarly to the other one, in this case we were importing a private class that inherits from `ec.EllipticCurvePrivateKey`. Check it out in the current version (`38.0.3`):

https://github.com/pyca/cryptography/blob/38.0.3/src/cryptography/hazmat/backends/openssl/ec.py#L125

And in the new version (`42.0.5`) we are just importing that same abstract base class, which is also registered with an OpenSSL backend:

https://github.com/pyca/cryptography/blob/42.0.5/src/cryptography/hazmat/primitives/asymmetric/ec.py#L125

Lastly, from the docs, check the [Use of OpenSSL](https://cryptography.io/en/42.0.5/openssl/) page. It's clear that OpenSSL is the only tool used for all cryptographic operations and that its backend is a private implementation detail.

pyproject.toml
glevco: I checked all changes in sensitive packages (`cryptography`, `pyopenssl`, and `twisted`), specially considering we updated major versions. I added a list of those in the PR description.

pyproject.toml
glevco: I checked all changes in sensitive packages (`cryptography`, `pyopenssl`, and `twisted`), specially considering we updated major versions. I added a list of those in the PR description.

After this is reviewed we can update the `python-hathorlib` and correct the new version here.

jansegre (approved):

jansegre (approved):

hathor/p2p/utils.py
msbrogli: Great! Thanks!

hathor/wallet/resources/thin_wallet/send_tokens.py
msbrogli: I agree. Thanks!

hathor/wallet/wallet.py
msbrogli: Great. Thanks!

pyproject.toml
msbrogli: Ok. Thanks!

msbrogli (approved):

PR #1000, opened 2024-04-11
feat(sync-v2): add support for async on_new_tx

glevco:
### Motivation

### Acceptance Criteria

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/factory.py
+++ b/hathor/p2p/sync_v2/factory.py
@@ -18,8 +18,8 @@ from hathor.p2p.manager import ConnectionsManager
 from hathor.p2p.sync_agent import SyncAgent
 from hathor.p2p.sync_factory import SyncAgentFactory
 from hathor.p2p.sync_v2.agent import NodeBlockSync
-from hathor.p2p.sync_v2.p2p_storage import P2PStorage
-from hathor.p2p.sync_v2.p2p_vertex_handler import P2PVertexHandler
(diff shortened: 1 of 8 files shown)


hathor/p2p/sync_v2/factory.py
```diff
@@ -18,15 +18,34 @@
...
 from hathor.reactor import ReactorProtocol as Reactor
 
 if TYPE_CHECKING:
     from hathor.p2p.protocol import HathorProtocol
 
 
 class SyncV2Factory(SyncAgentFactory):
-    def __init__(self, connections: ConnectionsManager):
+    # TODO: The problem with using a default False for async, is that tests will not run the new async path.
+    def __init__(self, connections: ConnectionsManager, *, use_async: bool = False) -> None:
```
glevco, line 31: Any ideas on how we can easily test both versions?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -1167,6 +1167,27 @@ def get_block(self, block_id: VertexId) -> Block:
...
+        for dep in deps:
+            meta = self.get_metadata(dep)
+            if meta is None:
+                all_exist = False
+                continue
+            if not meta.validation.is_fully_connected():
+                all_valid = False
+            if meta.validation.is_invalid():
+                # or any of them is invalid (which would make this one invalid too)
+                return True  # TODO: ???
```
glevco, line 1188: Shouldn't we return `False` here?

PR #1003, opened 2024-04-17
feat: add first_block to the to_json_extended method of base transaction class

alexruzenhack:
### Motivation

Wallet Mobile needs the ability to show a nano contract transaction "status" in the token's transaction details modal, but this "status" requires to know if a nano contract transaction was already confirmed by a block of not, thus the `first_block` as a first citizen of a transaction.

### Acceptance Criteria

- It should add `first_block` to the `to_json_extended` to BaseTransaction for transactions.
- It should update documentation in the following resources:
  - `/dashboard_tx`
  - `/transaction`
  - `/thin_wallet/address_history`
  - `/thin_wallet/token_history`

### Checklist

### DIFF
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -820,13 +820,12 @@ class BaseTransaction(ABC):
             'parents': [],
         }
 
-        if not self.is_block:
-            # A nano contract tx must be confirmed by one block at least
-            # to be considered "executed"
-            if meta.first_block is not None:
-                ret['first_block'] = meta.first_block.hex()
-            else:
-                ret['first_block'] = None
+        # A nano contract tx must be confirmed by one block at least
+        # to be considered "executed"
+        if meta.first_block is not None:
+            ret['first_block'] = meta.first_block.hex()
+        else:
+            ret['first_block'] = None
 
         for parent in self.parents:
             ret['parents'].append(parent.hex())
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -820,6 +820,14 @@ class BaseTransaction(ABC):
             'parents': [],
         }
 
+        if not self.is_block:
+            # A nano contract tx must be confirmed by one block at least
+            # to be considered "executed"
+            if meta.first_block is not None:
+                ret['first_block'] = meta.first_block.hex()
+            else:
+                ret['first_block'] = None
+
         for parent in self.parents:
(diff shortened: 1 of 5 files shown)


hathor/transaction/base_transaction.py
```diff
@@ -820,6 +820,14 @@ def serialize_output(tx: BaseTransaction, tx_out: TxOutput) -> dict[str, Any]:
             'parents': [],
         }
 
+        if not self.is_block:
```
pedroferreira1, line 823: If you need this information for nano, why don't you add it to the nano class `to_json_extended`?

Even if you prefer to add to all transactions classes, you could add it do the `Transaction` class, instead of doind this `if` here.

alexruzenhack: Oh, I did it in the BaseTransaction class because Transaction doesn't have `to_json_extended` and in all the resources involving transaction we call `to_json_extended`.

In addition I couldn't find the nano class. I believe it is in the other repository.

I agree we can narrow down the scope of `first_block` and remove any block from the equation. However, either we add this property only for nano contract transaction, or to all transactions except blocks. What is best?

On the one side we have a better coherence when `first_block` appears only for nano contract transactions. On the other side, we have a better consistency when `first_block` appears for all transaction except block.

I prefer to keep consistency because it give us a degree of freedom for when the system needs to evolve, for instance a new transaction version that also requires a block confirmation.

alexruzenhack: In the following commit https://github.com/HathorNetwork/hathor-core/pull/1003/commits/37a9515a0e28ef47515e86c1d76ab9552478af0a I removed `first_block` from `to_json_extended` of base transaction and I added `to_json_extended` method to transaction class.

jansegre (approved):

pedroferreira1 (approved):

hathor/transaction/base_transaction.py
```diff
@@ -820,6 +820,13 @@ def serialize_output(tx: BaseTransaction, tx_out: TxOutput) -> dict[str, Any]:
             'parents': [],
         }
 
+        # A nano contract tx must be confirmed by one block at least
+        # to be considered "executed"
+        if meta.first_block is not None:
```
msbrogli, line 825: This is a metadata, why not add it to `ret["meta"]` or equivalent?

msbrogli (changes requested):

msbrogli (approved):

PR #1015, opened 2024-05-01
Refactor/metadata/migrate feature states 3

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1014

### Motivation

What was the motivation for the changes in this PR?

### Acceptance Criteria

- Include here all things that this PR should solve

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

PR #1016, opened 2024-05-03
refactor(metadata): pre-refactors for migrating feature states to static metadata [part 3/12]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1014

---

### Motivation

Continuing with the static metadata refactor introduced in PRs https://github.com/HathorNetwork/hathor-core/pull/1013 and https://github.com/HathorNetwork/hathor-core/pull/1014, the next step is to move the `feature_states` attribute from metadata to static metadata.

This was initially implemented in PR https://github.com/HathorNetwork/hathor-core/pull/1015, which turned out to be too large, so it was divided into 3 PRs: this one, https://github.com/HathorNetwork/hathor-core/pull/1017, and https://github.com/HathorNetwork/hathor-core/pull/1018. The latter is the one that actually moves the `feature_states` attribute, and the first two PRs perform simple refactors that were either necessary or small improvements.

Therefore, this PR simply performs some pre-refactors to debloat the main PR.

### Acceptance Criteria

- **IMPORTANT:** move `verify_aux_pow()` verification from `verify_without_storage()` to `verify()`.
- Rename `FeatureService.get_bits_description()` to `get_feature_infos()`, and `FeatureDescription` to `FeatureInfo`.
- Change `FeatureService._get_ancestor_at_height()` to assert that a block exists in the height index when it must.
- Update the `FeatureState` enum to be a unique string enum, like `Feature` (this will be important for serialization), and implement `is_active()`.
- Change static metadata dataclasses to `BaseModel`s instead, which will be necessary for (de)serialization.
- Other smaller improvements.

### Checklist

### DIFF
--- a/hathor/builder/builder.py
+++ b/hathor/builder/builder.py
@@ -535,10 +535,7 @@ class Builder:
         if self._feature_service is None:
             settings = self._get_or_create_settings()
             tx_storage = self._get_or_create_tx_storage()
-            self._feature_service = FeatureService(
-                feature_settings=settings.FEATURE_ACTIVATION,
-                tx_storage=tx_storage
-            )
+            self._feature_service = FeatureService(settings=settings, tx_storage=tx_storage)
 
         return self._feature_service
 
@@ -549,7 +546,7 @@ class Builder:
             feature_service = self._get_or_create_feature_service()
             feature_storage = self._get_or_create_feature_storage()
             self._bit_signaling_service = BitSignalingService(
-                feature_settings=settings.FEATURE_ACTIVATION,
+                settings=settings,
                 feature_service=feature_service,
                 tx_storage=tx_storage,
                 support_features=self._support_features,
--- a/hathor/transaction/static_metadata.py
+++ b/hathor/transaction/static_metadata.py
@@ -14,9 +14,7 @@
 
 from __future__ import annotations
 
-import dataclasses
 from abc import ABC
-from dataclasses import dataclass
 from itertools import chain, starmap, zip_longest
 from operator import add
 from typing import TYPE_CHECKING, Callable
@@ -26,7 +24,8 @@ from typing_extensions import Self
 from hathor.feature_activation.feature import Feature
 from hathor.feature_activation.model.feature_state import FeatureState
 from hathor.types import VertexId
-from hathor.util import json_dumpb, json_loadb
+from hathor.util import json_loadb
+from hathor.utils.pydantic import BaseModel
 
 if TYPE_CHECKING:
     from hathor.conf.settings import HathorSettings
@@ -34,8 +33,7 @@ if TYPE_CHECKING:
     from hathor.transaction.storage import TransactionStorage
 
 
-@dataclass(slots=True, frozen=True, kw_only=True)
-class VertexStaticMetadata(ABC):
+class VertexStaticMetadata(ABC, BaseModel):
     """
     Static Metadata represents vertex attributes that are not intrinsic to the vertex data, but can be calculated from
     only the vertex itself and its dependencies, and whose values never change.
@@ -48,10 +46,6 @@ class VertexStaticMetadata(ABC):
     # block that confirming this transaction, it is important to always have this set to be able to distinguish an old
     # metadata (that does not have this calculated, from a tx with a new format that does have this calculated)
     min_height: int
-
-    def to_bytes(self) -> bytes:
-        """Convert this static metadata instance to a json bytes representation."""
-        return json_dumpb(dataclasses.asdict(self))
 
     @classmethod
     def from_bytes(cls, data: bytes, *, target: 'BaseTransaction') -> 'VertexStaticMetadata':
@@ -68,7 +62,6 @@ class VertexStaticMetadata(ABC):
         raise NotImplementedError
 
 
-@dataclass(slots=True, frozen=True, kw_only=True)
 class BlockStaticMetadata(VertexStaticMetadata):
     height: int
 
@@ -181,7 +174,6 @@ class BlockStaticMetadata(VertexStaticMetadata):
         return parent_block.static_metadata.feature_activation_bit_counts
 
 
-@dataclass(slots=True, frozen=True, kw_only=True)
 class TransactionStaticMetadata(VertexStaticMetadata):
     @classmethod
     def create_from_storage(cls, tx: 'Transaction', settings: HathorSettings, storage: 'TransactionStorage') -> Self:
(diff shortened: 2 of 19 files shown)


tests/feature_activation/test_feature_service.py
```diff
@@ -445,10 +394,8 @@ def test_get_state_from_active(storage: TransactionStorage, block_height: int) -
 
 
 @pytest.mark.parametrize('block_height', [16, 17, 18, 19])
-def test_caching_mechanism(storage: TransactionStorage, block_height: int) -> None:
```
glevco, line 448: This test was removed because in part 5 the caching mechanism is removed, as it's not necessary anymore.

hathor/transaction/static_metadata.py
```diff
@@ -68,7 +62,6 @@ def from_bytes(cls, data: bytes, *, target: 'BaseTransaction') -> 'VertexStaticM
         raise NotImplementedError
 
 
-@dataclass(slots=True, frozen=True, kw_only=True)
```
msbrogli, line 71: Why are you removing this line?

hathor/transaction/static_metadata.py
```diff
@@ -181,7 +174,6 @@ def _get_previous_feature_activation_bit_counts(
         return parent_block.static_metadata.feature_activation_bit_counts
 
 
-@dataclass(slots=True, frozen=True, kw_only=True)
```
msbrogli, line 184: Why are you removing this line?

hathor/transaction/static_metadata.py
```diff
@@ -26,16 +24,16 @@
...
+from hathor.util import json_loadb
+from hathor.utils.pydantic import BaseModel
 
 if TYPE_CHECKING:
     from hathor.conf.settings import HathorSettings
     from hathor.transaction import BaseTransaction, Block, Transaction
     from hathor.transaction.storage import TransactionStorage
 
 
-@dataclass(slots=True, frozen=True, kw_only=True)
```
msbrogli, line 37: Why are you removing this line?

hathor/verification/verification_service.py
```diff
@@ -274,7 +275,6 @@ def _verify_without_storage_block(self, block: Block) -> None:
         self._verify_without_storage_base_block(block)
 
     def _verify_without_storage_merge_mined_block(self, block: MergeMinedBlock) -> None:
-        self.verifiers.merge_mined_block.verify_aux_pow(block)
```
msbrogli, line 277: Why can't we run this verification without storage? This change seems completely unrelated to the other refactors in this PR.

hathor/builder/builder.py
msbrogli, line 1: I still don't get why you're doing this refactor. It might become clear when reviewing the next PRs.

msbrogli (changes requested):

hathor/transaction/static_metadata.py
glevco: I mentioned it in the Acceptance Criteria:

> Change static metadata dataclasses to `BaseModel`s instead, which will be necessary for (de)serialization.

Specifically, I'm now serializing the `Feature` and `FeatureState` enums. Before, I was using `dataclasses.asdict()` which serializes enums as objects that cannot be converted to json, instead of using their raw values directly. Then, I changed those classes to `BaseModel` which are equivalent and provide the `json_dumpb()` method, so I don't have to do this conversion manually.

hathor/transaction/static_metadata.py
glevco: Please see https://github.com/HathorNetwork/hathor-core/pull/1016#discussion_r1729346188

hathor/transaction/static_metadata.py
glevco: Please see https://github.com/HathorNetwork/hathor-core/pull/1016#discussion_r1729346188

hathor/verification/verification_service.py
glevco: As explained in slack, this was wrong as this verification does in fact use the storage.

hathor/builder/builder.py
glevco: Yes, in the next PR the `FeatureService` becomes just a utility class (that can even be removed later, and converted to functions), and not a singleton anymore. Therefore, it's simpler to use it by passing `settings` as argument instead of `settings.FEATURE_ACTIVATION`

hathor/verification/verification_service.py
msbrogli: Got it. Thanks!

hathor/transaction/static_metadata.py
msbrogli: Is `BaseModel` mutable? If yes, I don't think it's worth losing immutability because of a serialization method.

tests/tx/test_verification.py
```diff
@@ -342,8 +342,6 @@ def test_merge_mined_block_verify_without_storage(self) -> None:
         verify_data_wrapped = Mock(wraps=self.verifiers.block.verify_data)
         verify_sigops_output_wrapped = Mock(wraps=self.verifiers.vertex.verify_sigops_output)
 
-        verify_aux_pow_wrapped = Mock(wraps=self.verifiers.merge_mined_block.verify_aux_pow)
```
msbrogli, line 345: Don't you have to add this mock and `assert_called_once()` to the moved where the call was moved to?

hathor/transaction/static_metadata.py
glevco: It's immutable, as configured here: https://github.com/HathorNetwork/hathor-core/blob/da32f8092a1ae038046864a4edc6e448b9e1b811/hathor/utils/pydantic.py#L33

tests/tx/test_verification.py
glevco: It's already there, since `verify` and `validate_full` called `verify_without_storage`, this was already being tested: https://github.com/HathorNetwork/hathor-core/blob/da32f8092a1ae038046864a4edc6e448b9e1b811/tests/tx/test_verification.py#L409

hathor/transaction/static_metadata.py
msbrogli: Got it. Thanks for clarifying.

tests/tx/test_verification.py
msbrogli: It is actually the `verify_aux_pow` (not the `verify_pow`). But I checked and it's there. Thanks!

msbrogli (approved):

hathor/builder/builder.py
msbrogli: Got it. I'd rather see this change there but it seems harmless to be merged in this PR.

jansegre (approved):

PR #1033, opened 2024-05-09
feat(cli): Add --log-vertex-bytes for debugging

msbrogli:
### Motivation

We faced some difficulties replicating issue reports from developers. So this PR introduces a parameter to make replication easier just by sharing the full node log with us.

### Acceptance Criteria

- Add the `--log-vertex-bytes` parameter in the run_node cli.
- Add the `VertexHandler._log_vertex_bytes: bool` which defaults to false.
- Log the transaction bytes when this new option is enabled.

### Checklist

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -151,6 +151,8 @@ class RunNode:
         # XXX: this is temporary, should be added as a sysctl instead before merging
         parser.add_argument('--x-ipython-kernel', action='store_true',
                             help='Launch embedded IPython kernel for remote debugging')
+        parser.add_argument('--log-vertex-bytes', action='store_true',
+                            help='Log tx bytes for debugging')
         return parser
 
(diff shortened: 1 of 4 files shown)


hathor/cli/run_node.py
```diff
@@ -151,6 +151,8 @@ def create_parser(cls) -> ArgumentParser:
         # XXX: this is temporary, should be added as a sysctl instead before merging
         parser.add_argument('--x-ipython-kernel', action='store_true',
                             help='Launch embedded IPython kernel for remote debugging')
+        parser.add_argument('--x-log-vertex-bytes', action='store_true',
```
glevco, line 154: Why `x`?

glevco (dismissed):

jansegre (approved):
Looks good to me, but I think it's fine for it to be a regular argument instead (no `-x`).

jansegre (approved):

glevco (approved):

PR #1034, opened 2024-05-10
feat(cli): Add load_from_logs cli command to load blocks and transactions as they are found in a log dump

msbrogli:
### Motivation

This new command makes it easier to replay logs generated using `--log-vertex-bytes` (https://github.com/HathorNetwork/hathor-core/pull/1033).

I initially thought that we could replay even the timestamp by using a custom reactor and getting a tighter control of the time. But it was not needed yet for my debugging so I decided to postpone it to another PR.

### Acceptance Criteria

1. Add command `load_from_logs` that loads blocks and transactions as they are found in a log dump generated using `--log-vertex-bytes`. All vertices are processed in the same order as they appear in the log file.
2. The full node exits after the replay is complete.
3. All arguments from `run_node` are available so one can configure the environment where the replay will take place. A new required argument `--log-dump` was added to get the path of the log file that will be replayed.

### Checklist

### DIFF
--- a/hathor/cli/main.py
+++ b/hathor/cli/main.py
@@ -35,6 +35,7 @@ class CliManager:
             db_export,
             db_import,
             generate_valid_words,
+            load_from_logs,
             merged_mining,
             mining,
             multisig_address,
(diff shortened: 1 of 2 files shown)


glevco (approved):

hathor/cli/main.py
```diff
@@ -91,6 +92,7 @@ def __init__(self) -> None:
         self.add_cmd('dev', 'x-export', db_export, 'EXPERIMENTAL: Export database to a simple format.')
         self.add_cmd('dev', 'x-import', db_import, 'EXPERIMENTAL: Import database from exported format.')
         self.add_cmd('dev', 'replay-logs', replay_logs, 'EXPERIMENTAL: re-play json logs as console printted')
+        self.add_cmd('dev', 'run_from_logs', run_from_logs, 'replay the same blocks and transactions in the log')
```
jansegre, line 95: What about naming it `load_from_logs` (or `load-from-logs`, we're not consistent with using `-` or `_`)?

I was initially confused by the name since this isn't replaying the *logs* themselves (that's what `replay-logs` does), this is replaying the *loading* of blocks/transactions as they were found in a log dump. And also because the node doesn't actually *run* after the blocks/transactions are added/loaded.

jansegre (approved):

glevco (approved):

PR #1037, opened 2024-05-13
feat(multiprocess-verification): implement VerificationCoordinator [part 12/12]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1027

---

### Motivation

### Acceptance Criteria

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/blockchain_streaming_client.py
+++ b/hathor/p2p/sync_v2/blockchain_streaming_client.py
@@ -72,7 +72,9 @@ class BlockchainStreamingClient:
 
     def fails(self, reason: 'StreamingError') -> None:
         """Fail the execution by resolving the deferred with an error."""
-        self._deferred.errback(reason)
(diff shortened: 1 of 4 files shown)


hathor/verification/verification_coordinator.py
```diff
@@ -0,0 +1,231 @@
...
+        return basic_deps, deps
+
+    def _get_tx_deps(self, tx: Transaction) -> TransactionDependencies:
+        """
+        Return transaction dependencies for a transaction using dependencies that may or may not have been persisted
+        yet.
+        """
+        parents = {vertex_id: self._get_vertex(vertex_id) for vertex_id in tx.parents}
+        spent_txs = {tx_input.tx_id: self._get_vertex(tx_input.tx_id) for tx_input in tx.inputs}
+        token_info = tx.get_complete_token_info()
```
glevco, line 202: I just noticed this will fail for transactions because this method accesses the storage. We need to make it accept functional dependencies like the other methods.

PR #1039, opened 2024-05-23
feat(side-dag): implement side_dag CLI command

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1040

### Motivation

Implement a new `side_dag` CLI command that is able to run two independent full nodes, each in its own separate process. The command accepts the same options as the `run_node` command, however options can be duplicate with a `--side-dag` prefix. Each full node is configured with the respective set of options. Processes are automatically terminated when one of the full nodes exits.

Here's an example on how this command could be used:

 ```bash
  $ python -m hathor side_dag
      --testnet
      --procname-prefix testnet-
      --memory-storage
      --disable-logs
      --side-dag-config-yaml ./my-side-dag.yml
      --side-dag-procname-prefix my-side-dag-
      --side-dag-memory-storage
      --side-dag-json-logs
```

### Acceptance Criteria

- Implement new `cli.util.process_logging_options()`.
- Change `execute_from_command_line()` so it supports modules opting out of the `setup_logging()` call.
- Update `create_parser()` so it accepts a custom env var prefix.
- Add support to disabling logs.
- Implement `side_dag` CLI command.
- Implement `HathorSettings.GENESIS_TOKENS` calculated property, removing the `GENESIS_TOKENS` constant.

### Checklist

### DIFF
--- a/hathor/cli/side_dag.py
+++ b/hathor/cli/side_dag.py
@@ -1,0 +1,277 @@
+# Copyright 2024 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import argparse
+import os
+import signal
+import sys
+import traceback
+from dataclasses import dataclass
+from enum import Enum
+from multiprocessing import Pipe, Process
+from typing import TYPE_CHECKING
+
+from typing_extensions import assert_never
+
+if TYPE_CHECKING:
+    from hathor.cli.util import LoggingOutput
+
+    # Workaround for a typing issue in Windows
+    if sys.platform == 'win32':
+        from multiprocessing.connection import _ConnectionBase as Connection
+    else:
+        from multiprocessing.connection import Connection
+
+import psutil
+from structlog import get_logger
+
+from hathor.cli.run_node import RunNode
+
+logger = get_logger()
+
+PRE_SETUP_LOGGING: bool = False
+HATHOR_NODE_INIT_TIMEOUT: int = 10
+
+
+@dataclass(frozen=True, slots=True)
+class HathorProcessInitFail:
+    reason: str
+
+
+@dataclass(frozen=True, slots=True)
+class HathorProcessInitSuccess:
+    pass
+
+
+@dataclass(frozen=True, slots=True)
+class HathorProcessTerminated:
+    pass
+
+
+@dataclass(frozen=True, slots=True)
+class SideDagProcessTerminated:
+    pass
+
+
+class SideDagRunNode(RunNode):
+    env_vars_prefix = 'hathor_side_dag_'
+
+
+def main(capture_stdout: bool) -> None:
+    """
+    This command runs two full node instances in separate processes.
+
+    The main process runs a side-dag full node, and it accepts the same options as the `run_node` command. Options
+    with the `--side-dag` prefix are passed to the side-dag full node, while options without this prefix are passed
+    to the non-side-dag full node, which runs in a background process and is commonly just a Hathor full node.
+    Whenever one of the full nodes fail, the other is automatically terminated.
+
+    By default, both full nodes output logs to stdout, but logs can be configured independently. Here's an example:
+
+    ```bash
+    $ python -m hathor side_dag
+        --testnet
+        --procname-prefix testnet-
+        --memory-storage
+        --disable-logs
+        --side-dag-config-yaml ./my-side-dag.yml
+        --side-dag-procname-prefix my-side-dag-
+        --side-dag-memory-storage
+        --side-dag-json-logs
+    ```
+
+    In this example, Hathor testnet logs would be disabled, while side-dag logs would be outputted to stdout as json.
+    """
+    from hathor.cli.util import process_logging_options, setup_logging
+    argv = sys.argv[1:]
+    hathor_logging_output, side_dag_logging_output = _process_logging_output(argv)
+    hathor_node_argv, side_dag_argv = _partition_argv(argv)
+    conn1, conn2 = Pipe()
+    hathor_node_process = _start_hathor_node_process(
+        hathor_node_argv, logging_output=hathor_logging_output, capture_stdout=capture_stdout, conn=conn1
+    )
+
+    log_options = process_logging_options(side_dag_argv)
+    setup_logging(logging_output=side_dag_logging_output, logging_options=log_options, capture_stdout=capture_stdout)
+    logger.info('starting nodes', hathor_node_pid=hathor_node_process.pid, side_dag_pid=os.getpid())
+
+    _run_side_dag_node(side_dag_argv, hathor_node_process=hathor_node_process, conn=conn2)
+
+
+def _process_logging_output(argv: list[str]) -> tuple[LoggingOutput, LoggingOutput]:
+    """Extract logging output before argv parsing."""
+    from hathor.cli.util import LoggingOutput
+
+    class LogOutputConfig(str, Enum):
+        HATHOR = 'hathor'
+        SIDE_DAG = 'side-dag'
+        BOTH = 'both'
+
+    parser = argparse.ArgumentParser()
+    log_args = parser.add_mutually_exclusive_group()
+    log_args.add_argument('--json-logs', nargs='?', const='both', type=LogOutputConfig)
+    log_args.add_argument('--disable-logs', nargs='?', const='both', type=LogOutputConfig)
+
+    args, remaining_argv = parser.parse_known_args(argv)
+    argv.clear()
+    argv.extend(remaining_argv)
+
+    def proces_log_output_config(
+        config: LogOutputConfig,
+        target: LoggingOutput
+    ) -> tuple[LoggingOutput, LoggingOutput]:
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by msbrogli — CHANGES_REQUESTED** [review]

File: hathor/cli/side_dag.py [hunk 1]
```diff
@@ -0,0 +1,210 @@
...
+class HathorProcessTerminated:
+    pass
+
+
+@dataclass(frozen=True, slots=True)
+class SideDagProcessTerminated:
+    pass
+
+
+class SideDag(RunNode):
```
msbrogli, line 61: Maybe rename to `SideDagRunNode`?

hathor/cli/side_dag.py
```diff
@@ -0,0 +1,210 @@
...
+            argv_list = side_dag_argv
+
+        argv_list.append(arg)
+        if value is not None:
+            argv_list.append(value)
+
+    return hathor_node_argv, side_dag_argv
+
+
+def _run_side_dag_node(argv: list[str], *, hathor_node_process: Process, conn: 'Connection') -> None:
```
msbrogli, line 131: If I got it right, the current process will spawn a new process for the mainnet full node and run the side-dag full node in the current process. Is that right? Isn't it better to spawn two new processes, one for each full node and develop a process monitor that gracefully terminates when one of the processes dies or freezes?

About signals, all processes would have to handle signals.

msbrogli (changes requested):

hathor/cli/util.py
```diff
@@ -230,7 +271,12 @@ def setup_logging(
 
     def kwargs_formatter(_, __, event_dict):
         if event_dict and event_dict.get('event') and isinstance(event_dict['event'], str):
-            event_dict['event'] = event_dict['event'].format(**event_dict)
+            try:
```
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/cli/util.py
```diff
@@ -275,10 +321,9 @@ def twisted_structlog_observer(event):
             if failure is not None:
                 kwargs['exc_info'] = (failure.type, failure.value, failure.getTracebackObject())
             twisted_logger.log(level, msg, **kwargs)
-        except Exception as e:
-            print('error when logging event', e)
-            for k, v in event.items():
-                print(k, v)
+        except Exception:
+            logger = structlog.get_logger()
+            logger.exception('error when logging event')
```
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/cli/side_dag.py
```diff
@@ -0,0 +1,210 @@
...
+    hathor_node_process = _start_hathor_node_process(hathor_node_argv, capture_stdout=capture_stdout, conn=conn1)
+
+    debug, logging_output, sentry = process_logging_options(side_dag_argv)
+    setup_logging(debug=debug, capture_stdout=capture_stdout, logging_output=logging_output, sentry=sentry)
+    logger.info('starting nodes', hathor_node_pid=hathor_node_process.pid, side_dag_pid=os.getpid())
+
+    _run_side_dag_node(side_dag_argv, hathor_node_process=hathor_node_process, conn=conn2)
+
+
+def _process_argv(argv: list[str]) -> tuple[list[str], list[str]]:
```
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/cli/side_dag.py
msbrogli: Thanks!

hathor/cli/side_dag.py
msbrogli: As discussed in person, let's do a monitor process in a separate PR so we can skip it for now.

hathor/conf/settings.py
```diff
@@ -55,7 +54,9 @@ class HathorSettings(NamedTuple):
     # Genesis pre-mined tokens
     GENESIS_TOKEN_UNITS: int = GENESIS_TOKEN_UNITS
 
-    GENESIS_TOKENS: int = GENESIS_TOKENS
+    @property
+    def GENESIS_TOKENS(self) -> int:
```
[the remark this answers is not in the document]
msbrogli: I guess we can discuss it separately. We can add a consistency check before initializing a full node, for instance.

hathor/cli/side_dag.py
glevco: Created an issue for this: https://github.com/HathorNetwork/hathor-core/issues/1051

hathor/conf/settings.py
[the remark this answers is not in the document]
glevco: Sure, removed it in https://github.com/HathorNetwork/hathor-core/pull/1039/commits/fbf9caff99ec80ab1786bb2e748f1b4e6bdb8e9e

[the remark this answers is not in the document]

msbrogli (approved):

hathor/cli/side_dag.py
```diff
@@ -0,0 +1,220 @@
...
+    ```bash
+    $ python -m hathor side_dag
+        --testnet
+        --procname-prefix testnet-
+        --memory-storage
+        --disable-logs
+        --side-dag-config-yaml ./my-side-dag.yml
+        --side-dag-procname-prefix my-side-dag-
+        --side-dag-memory-storage
+        --side-dag-json-logs
```
jansegre, line 85: Does `--side-dag-json-logs` need to be a separate option from `--json-logs`? The reason `--json-logs` exist is to make all the logs that the stdout/stderr (as observed by whoever started the process) be written as `JSON` entries (one per line), instead of formatted entries (which can be colored). The option exists independently of which subcommand is being used even for those not based on `RunNode`.

What do you think about using only one argument instead?

hathor/cli/side_dag.py
```diff
@@ -0,0 +1,220 @@
...
+    hathor_node_process.start()
+    return hathor_node_process
+
+
+def _run_hathor_node(argv: list[str], run_node_cmd: type[RunNode], capture_stdout: bool, conn: 'Connection') -> None:
+    """Function to be called by a background process to run the Hathor full node."""
+    from hathor.cli.util import process_logging_options, setup_logging
+
+    # We don't terminate via SIGINT directly, instead the side-dag process will terminate us.
+    signal.signal(signal.SIGINT, lambda _, __: None)
```
jansegre, line 191: Will this suppress the normal behavior of raising a `KeyboardInterrupt` when the process receives a `SIGINT`?

hathor/cli/side_dag.py
```diff
@@ -0,0 +1,220 @@
...
+        message = conn.recv()
+        assert isinstance(message, SideDagProcessTerminated)
+        logger.critical('hathor node terminated because side-dag process was terminated')
+        return
+
+    conn.send(HathorProcessTerminated())
+    logger.critical('terminating side-dag node...')
+    parent_pid = os.getppid()
+    parent_process = psutil.Process(parent_pid)
+    parent_process.terminate()
```
jansegre, line 220: What does `terminate` do? Does it just send a `SIGTERM`?

hathor/cli/side_dag.py
glevco: It could be a single option, the reason I implemented it separately is to be able to configure each full node's logging independently. I created the `LoggingOutput` enum and those arguments simply select one of the enum options. It accepts either `--[side-dag-]disable-logs`, `--[side-dag-]json-logs`, or not set, which respectively map to `NULL`, `JSON`, or `PRETTY` (the default).

So for example, if I'm running locally, I would probably set only `--disable-logs`, which would disable any logs from the hathor node, and the side-dag logs would be pretty outputted to stdout/sderr.

If I'm running on a server, I would probably set `--json-logs --side-dag-json-logs`, to get json outputs from both nodes. But one could easily want to discard hathor logs, for example, in which case they would set `--disable-logs --side-dag-json-logs`.

In summary, this is just giving more flexibility. Do you prefer to keep a single option?

hathor/cli/side_dag.py
glevco: Yes!

hathor/cli/side_dag.py
glevco: > What does `terminate` do? Does it just send a `SIGTERM`? Because that alone might not be enough to reliably terminate the parent process

Yes, it just sends a `SIGTERM`. Why not?

> there's an issue about doing a monitor process in a separate PR, that's fine by me.

Yep, it will be a future improvement.

glevco: I added this in the issue: https://github.com/HathorNetwork/hathor-core/issues/1051#issuecomment-2153518597

hathor/cli/side_dag.py
jansegre: I see, I'm not against the flexibility, but I'm not a fan of having the behavior of `--json-logs` sort of change (from having all logs in JSON, to some of the logs in JSON), I understand that production would just need to use `--json-logs --side-dag-json-logs` but I think that's a bit counter-intuitive.

I'd rather have the `side-dag` process use the same behavior (`NULL`, `JSON`, `PRETTY`) as the main log, unless there's an argument that changes it.

What about making the options `--json-logs` and `--disable-logs` take an optional argument: `both`, `main`, `side-dag`? And the default when the argument is present would be `both`. Since these two are special options that are processed before the `RunNode` ones, I think it would be an acceptable compromise.

jansegre: This can be implemented with [ArgumentParser.parse_known_args](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args) instead of doing it manually like is was before.

jansegre (approved):

hathor/cli/side_dag.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1039/commits/90a0606cb53cc1939157c4b172cf5ad8d380e6f0

jansegre (approved):

hathor/cli/side_dag.py
msbrogli: @jansegre I had a similar conversation about this topic with @glevco during my review.

msbrogli (approved):

PR #1040, opened 2024-05-23
fix(log): correctly handle errors when logging events

glevco:
### Motivation

This PR contains 2 separate but related fixes that will improve our on-call overhead. They would cause some errors to be logged in multiple unstructured lines, generating several unnecessary alerts.

The first problem was that a bug in string formatting would generate an exception during log event processing, triggering the "error when logging event" path. The second problem was with the logging of this error itself, that was using print statements, and therefore generating the multiple unstructured lines.

This finally resolves https://github.com/HathorNetwork/on-call-incidents/issues/153.

### Acceptance Criteria

- Fix incorrect string formatting that caused a `KeyError` to be raised during logging.
- Remove `print()` statements when errors occur during logging.

### Checklist

### DIFF
--- a/hathor/cli/util.py
+++ b/hathor/cli/util.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import sys
+import traceback
 from argparse import ArgumentParser
 from collections import OrderedDict
 from datetime import datetime
@@ -230,7 +232,12 @@ def setup_logging(
 
     def kwargs_formatter(_, __, event_dict):
         if event_dict and event_dict.get('event') and isinstance(event_dict['event'], str):
-            event_dict['event'] = event_dict['event'].format(**event_dict)
+            try:
+                event_dict['event'] = event_dict['event'].format(**event_dict)
+            except KeyError:
+                # The event string may contain '{}'s that are not used for formatting, resulting in a KeyError in the
+                # event_dict. In this case, we don't format it.
+                pass
         return event_dict
 
     processors: list[Any] = [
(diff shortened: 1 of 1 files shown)


hathor/cli/util.py
```diff
@@ -230,7 +230,12 @@ def setup_logging(
...
     def kwargs_formatter(_, __, event_dict):
         if event_dict and event_dict.get('event') and isinstance(event_dict['event'], str):
-            event_dict['event'] = event_dict['event'].format(**event_dict)
+            try:
+                event_dict['event'] = event_dict['event'].format(**event_dict)
+            except KeyError:
+                # The event string may contain '{}'s that are not used for formatting, resulting in a KeyError in the
+                # event_dict. In this case, we don't format it.
+                pass
         return event_dict
```
glevco, line 239: I would prefer to simply remove this `kwargs_formatter()` function instead, then we would stop supporting log calls like this:

```python
log.info('logging {something}', something=123)
```

We can always use f-strings or `str.format()` directly instead. But that would require searching through the project to convert those calls, and it may not be trivial to find all occurrences.

jansegre: From what I remember, when I initially wrote this function I tested removing support for the `kwargs_formatter` but it broke log calls by twisted itself, otherwise those logs would be literally displayed as `logging {something}` with `something=123` shown separately.

jansegre (approved):

hathor/cli/util.py
glevco: Ahh ok. It's weird that twisted depends on us providing this function, there's probably another way to do it, but I'm fine keeping it for now, at least the problem was fixed.

jansegre: It's not that twisted depends on us providing this functions, it's more that twisted's logging system implements this and if we are to replace it we have to keep this behavior.

glevco: @jansegre maybe we should copy their implementation, then?

hathor/cli/util.py
```diff
@@ -261,6 +268,7 @@ def kwargs_formatter(_, __, event_dict):
     )
 
     twisted_logger = structlog.get_logger('twisted')
+    logger = structlog.get_logger()
```
jansegre, line 271: We don't need anymore, right?

glevco: Done

jansegre (approved):

msbrogli (approved):

PR #1042, opened 2024-05-28
fix(resources): regression in address history endpoint

jansegre:
### Motivation

After the release o v0.60.0 [a bug was reported](https://github.com/HathorNetwork/on-call-incidents/issues/173) in the address-history API. The bug was introduced after the optimizations in #978, reverting the PR makes the bug disappear.

### Acceptance Criteria

- Restore the correct behavior without undoing the optimizations
- Add a test that reproduces the issue 

### Checklist

### DIFF
--- a/hathor/wallet/resources/thin_wallet/address_history.py
+++ b/hathor/wallet/resources/thin_wallet/address_history.py
@@ -35,8 +35,11 @@ class AddressHistoryResource(Resource):
     isLeaf = True
 
     def __init__(self, manager):
-        self._settings = get_global_settings()
         self.manager = manager
+        settings = get_global_settings()
+        # XXX: copy the parameters that are needed so tests can more easily tweak them
+        self.max_tx_addresses_history = settings.MAX_TX_ADDRESSES_HISTORY
+        self.max_inputs_outputs_address_history = settings.MAX_INPUTS_OUTPUTS_ADDRESS_HISTORY
 
     # TODO add openapi docs for this API
     def render_POST(self, request: Request) -> bytes:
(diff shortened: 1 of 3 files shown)


hathor/wallet/resources/thin_wallet/address_history.py
```diff
@@ -166,18 +166,20 @@ def get_address_history(self, addresses: list[str], ref_hash: Optional[str]) ->
...
-                    return json_dumpb({
-                        'success': False,
-                        'message': 'Hash {} is not a transaction hash.'.format(ref_hash)
-                    })
-
             # The address index returns an iterable that starts at `tx`.
-            hashes = addresses_index.get_sorted_from_address(address, tx)
+            if idx == 0:
+                ref_tx = None
+                if ref_hash_bytes:
```
msbrogli, line 172: Maybe move this fetch to a separate method to clean up the code?

msbrogli (approved):

msbrogli (dismissed):

pedroferreira1 (dismissed):

hathor/wallet/resources/thin_wallet/address_history.py
jansegre: I'd like to also refactor other parts of the code to make it easier to read and to handle errors and I think I'd rather do it all in a separate PR, what do you think?

msbrogli: Sure!

msbrogli (approved):

pedroferreira1 (approved):

PR #1046, opened 2024-06-03
feat(api): Add include_transactions parameter to the BlockAtHeightResource

msbrogli:
### Motivation

### Acceptance Criteria

### Checklist


hathor/transaction/resources/block_at_height.py
```diff
@@ -75,8 +75,29 @@ def render_GET(self, request: 'Request') -> bytes:
             })
 
         block = self.manager.tx_storage.get_transaction(block_hash)
-
         data = {'success': True, 'block': block.to_json_extended()}
+
+        # include_transactions parameter is optional and must be an integer
+        try:
+            include_transactions = parse_int(args.get('include_transactions', 0))
```
jansegre, line 82: I think we should add a query with this parameter to the OpenAPI definition.

hathor/transaction/resources/block_at_height.py
```diff
@@ -75,8 +75,29 @@ def render_GET(self, request: 'Request') -> bytes:
...
+        except ValueError as e:
+            return json_dumpb({
+                'success': False,
+                'message': f'Failed to parse \'include_transactions\': {e}'
+            })
+
+        if include_transactions == 1:
+            tx_ids: list[str] = []
```
obiyankenobi, line 90: Add comment to explain difference between 1 and 2.

hathor/transaction/resources/block_at_height.py
```diff
@@ -75,8 +75,29 @@ def render_GET(self, request: 'Request') -> bytes:
...
+            for tx in block.iter_transactions_in_this_block():
+                tx_ids.append(tx.hash.hex())
+            data['tx_ids'] = tx_ids
```
obiyankenobi, line 93: Should we keep the fields the same? So having `transactions` for both, instead of `tx_ids` in case it's just the ids.

hathor/transaction/resources/block_at_height.py
jansegre: Also, I guess we can keep 1 and 2 for compatibility, but wouldn't it be better to use something like `include_transactions=txid` (or id/hash/...) and `include_transactions=full` (or metadata/data/meta/...)

msbrogli: I just did it. And I guess we don't have to keep compatibility since this is a brand new parameter.

jansegre (dismissed):

hathor/transaction/resources/block_at_height.py
msbrogli: I used separate keys to facilitate parsing since they have different formas.

hathor/transaction/resources/block_at_height.py
```diff
@@ -74,9 +74,30 @@ def render_GET(self, request: 'Request') -> bytes:
                 'message': 'No block with height {}.'.format(height)
             })
 
-        block = self.manager.tx_storage.get_transaction(block_hash)
-
+        block = self.manager.tx_storage.get_block(block_hash)
         data = {'success': True, 'block': block.to_json_extended()}
+
+        # include_transactions parameter is optional
+        include_transactions: str = args.get('include_transactions', None)
```
pedroferreira1, line 81:
```suggestion
        include_transactions: str | None = args.get('include_transactions', None)
```

obiyankenobi (approved):

pedroferreira1 (approved):

hathor/transaction/resources/block_at_height.py
msbrogli: Fixed. Thanks!

hathor/transaction/resources/block_at_height.py
msbrogli: Fixed. Thanks!

jansegre (approved):

PR #1049, opened 2024-06-04
fix: improve stratum logs

glevco:
### Motivation

We're receiving some calls to the stratum API with wrong arguments, but current code doesn't allow us to investigate it. This PRs adds logs that will help debugging this. See https://github.com/HathorNetwork/on-call-incidents/issues/176.

### Acceptance Criteria

- Add logs and try to convert stratum `params` to a `dict`.

### Checklist

### DIFF
--- a/hathor/stratum/stratum.py
+++ b/hathor/stratum/stratum.py
@@ -418,11 +418,13 @@ class StratumProtocol(JSONRPC):
             if not self.manager.can_start_mining():
                 return self.send_error(NODE_SYNCING, msgid)
 
-        if method in ['mining.subscribe', 'subscribe']:
+        if not isinstance(params, dict):
+            self.log.error(f'expected dict params, received: {params}')
             params = cast(dict, params)
+
+        if method in ['mining.subscribe', 'subscribe']:
             return self.handle_subscribe(params, msgid)
         if method in ['mining.submit', 'submit']:
-            params = cast(dict, params)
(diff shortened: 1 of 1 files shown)


hathor/stratum/stratum.py
```diff
@@ -418,11 +418,17 @@ def handle_request(self, method: str, params: Optional[Union[list, dict]], msgid
             if not self.manager.can_start_mining():
                 return self.send_error(NODE_SYNCING, msgid)
 
+        if not isinstance(params, dict):
+            self.log.error(f'expected dict, received {params}. Trying to convert...')
+            try:
+                params = dict(params)  # type: ignore[arg-type]
```
jansegre, line 424: This is a behavior change, is this actually fixing anything if it succeeds? For example `dict([[1,2]])` will result in `{1: 2}', so sometimes `dict(a: list)` successfully returns a dict. Before this change we didn't have this behavior.

If we don't have this `try` block and instead only have the cast that is inside the `except` block, the behavior is exactly the same but we'd have better logging.

msbrogli: Nice catch!

glevco: Removed in https://github.com/HathorNetwork/hathor-core/pull/1049/commits/1f595d81b608578dae609479fe62b02ee77d1eab

jansegre (approved):

msbrogli (approved):

PR #1060, opened 2024-06-11
feat(side-dag): implement Proof-of-Authority

glevco:
### Motivation

### Acceptance Criteria

### Checklist


jansegre (dismissed):

hathor/transaction/poa/poa_block.py
```diff
@@ -0,0 +1,78 @@
...
+        storage: TransactionStorage | None = None,
+        signer_id: bytes = b'',
+        signature: bytes = b'',
+    ) -> None:
+        super().__init__(
+            nonce=0,
+            timestamp=timestamp,
+            signal_bits=signal_bits,
+            version=TxVersion.POA_BLOCK,
+            weight=weight,
```
msbrogli, line 44: Should we remove weight?

hathor/verification/verification_service.py
```diff
@@ -226,24 +248,31 @@ def verify_without_storage(self, vertex: BaseTransaction) -> None:
...
         self._verify_without_storage_block(block)
 
+    def _verify_without_storage_poa_block(self, block: PoaBlock) -> None:
+        self._verify_without_storage_base_block(block)
+
     def _verify_without_storage_tx(self, tx: Transaction) -> None:
         """ Run all verifications that do not need a storage.
         """
-        self.verifiers.vertex.verify_pow(tx)
+        if self._settings.CONSENSUS_ALGORITHM.is_pow():
```
msbrogli, line 274: Why don't we keep the tx weight verification?

hathor/verification/vertex_verifier.py
```diff
@@ -44,6 +45,25 @@ class VertexVerifier:
     def __init__(self, *, settings: HathorSettings) -> None:
         self._settings = settings
 
+    def verify_version(self, vertex: BaseTransaction) -> None:
+        """Verify that the vertex version is valid."""
+        valid_versions = [
```
msbrogli, line 50: You can create global sets with these versions and just use them here. You might even refactor it to `self._settings.CONSENSUS_ALGORITHM.get_valid_versions()`.

Note: Remember to use a `set()` instead of a `list()` so we can check if an element exists in O(1).

hathor/transaction/poa/poa_block.py
msbrogli, line 1: We can't allow `PoaBlock` to be sent to PoW networks (e.g., testnet and mainnet). I guess we should forbid them from being instantiated in the `hathor.transaction.base_transaction.tx_or_block_from_bytes()` or somewhere else.

msbrogli (changes requested):

hathor/transaction/poa/poa_block.py
msbrogli: Should we have a list of valid versions in the builder? We might refactor `tx_or_block_from_bytes()` to a `ParseVertexFromBytes` singleton.

hathor/transaction/poa/poa_block.py
glevco: The `weight` is used to keep compatibility with the current `BlockConsensusAlgorithm`. It is used to signal whether a block is in turn or out of turn.

hathor/verification/vertex_verifier.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1060/commits/157044e8acb40e2ed85ee1a35b983d1c8df46e3f

hathor/transaction/poa/poa_block.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1060/commits/157044e8acb40e2ed85ee1a35b983d1c8df46e3f

hathor/verification/verification_service.py
glevco: Since there's no block mining in a PoA network, the idea was to remove transaction mining, too. To do that, we use `MIN_TX_WEIGHT=0` in settings and disable tx weight verification.

By doing this, we also remove the requirement to run a TxMiningService, simplifying use cases — at least from the full node perspective, but in practice the wallet-lib still requires a TxMiningService. This dependency can be removed from the wallet in the future so the use case would not need a TxMiningService.

jansegre (dismissed):

hathor/verification/vertex_verifier.py
msbrogli: Great! Thanks!

hathor/verification/verification_service.py
msbrogli: Got it. Thanks for clarifying.

hathor/transaction/poa/poa_block.py
msbrogli: Got it. I guess it's ok to move forward like this. But shouldn't we write a new PoA Consensus Algorithm?

hathor/transaction/poa/poa_block.py
msbrogli: Thanks!

jansegre (approved):

hathor/transaction/poa/poa_block.py
msbrogli: Thanks!

msbrogli (approved):

PR #1061, opened 2024-06-11
feat(side-dag): implement PoaBlockProducer

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1060

### Motivation

Finish the Proof-of-Authority side-dag implementation by implementing the `PoaBlockProducer` which is analogous to a mining service, in the full node itself.

### Acceptance Criteria

- Implement `poa` module and `PoaBlockProducer`.
- Implement `gen_genesis` CLI command.

### Checklist

### DIFF
--- a/hathor/cli/generate_genesis.py
+++ b/hathor/cli/generate_genesis.py
@@ -1,0 +1,79 @@
+# Copyright 2024 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+
+import base58
+
+
+def main():
+    from hathor.cli.util import create_parser
+    from hathor.conf.get_settings import get_global_settings
+    from hathor.mining.cpu_mining_service import CpuMiningService
+    from hathor.transaction import Block, Transaction, TxOutput
+    from hathor.transaction.scripts import P2PKH
+
+    parser = create_parser()
+    parser.add_argument('--config-yaml', type=str, help='Configuration yaml filepath')
+    parser.add_argument('--genesis-address', type=str, help='Address for genesis tokens')
+    parser.add_argument('--genesis-block-timestamp', type=int, help='Timestamp for the genesis block')
+
+    args = parser.parse_args(sys.argv[1:])
+    if not args.config_yaml:
+        raise Exception('`--config-yaml` is required')
+    if not args.genesis_address:
+        raise Exception('`--genesis-address` is required')
+    if not args.genesis_block_timestamp:
+        raise Exception('`--genesis-block-timestamp` is required')
+
+    os.environ['HATHOR_CONFIG_YAML'] = args.config_yaml
+    settings = get_global_settings()
+    output_script = P2PKH.create_output_script(address=base58.b58decode(args.genesis_address))
+    block_timestamp = int(args.genesis_block_timestamp)
+    mining_service = CpuMiningService()
+
+    block = Block(
+        timestamp=block_timestamp,
+        weight=settings.MIN_BLOCK_WEIGHT,
+        outputs=[
+            TxOutput(settings.GENESIS_TOKENS, output_script),
+        ],
+    )
+    mining_service.start_mining(block)
+    block.update_hash()
+
+    tx1 = Transaction(
+        timestamp=settings.GENESIS_TX1_TIMESTAMP,
+        weight=settings.MIN_TX_WEIGHT,
+    )
+    mining_service.start_mining(tx1)
+    tx1.update_hash()
+
+    tx2 = Transaction(
+        timestamp=settings.GENESIS_TX2_TIMESTAMP,
+        weight=settings.MIN_TX_WEIGHT,
+    )
+    mining_service.start_mining(tx2)
+    tx2.update_hash()
+
+    # The output format is compatible with the yaml config file
+    print('GENESIS_OUTPUT_SCRIPT:', output_script.hex())
+    print('GENESIS_BLOCK_TIMESTAMP:', block.timestamp)
+    print('GENESIS_BLOCK_HASH:', block.hash_hex)
+    print('GENESIS_BLOCK_NONCE:', block.nonce)
+    print('GENESIS_TX1_HASH:', tx1.hash_hex)
+    print('GENESIS_TX1_NONCE:', tx1.nonce)
+    print('GENESIS_TX2_HASH:', tx2.hash_hex)
+    print('GENESIS_TX2_NONCE:', tx2.nonce)
--- a/hathor/consensus/poa/poa_block_producer.py
+++ b/hathor/consensus/poa/poa_block_producer.py
@@ -1,0 +1,173 @@
+#  Copyright 2024 Hathor Labs
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from structlog import get_logger
+from twisted.internet.task import LoopingCall
+
+from hathor.conf.settings import HathorSettings
+from hathor.consensus import poa
+from hathor.consensus.consensus_settings import PoaSettings
+from hathor.crypto.util import get_public_key_bytes_compressed
+from hathor.reactor import ReactorProtocol
+
+if TYPE_CHECKING:
+    from hathor.consensus.poa import PoaSigner
+    from hathor.manager import HathorManager
+    from hathor.transaction import Block
+    from hathor.transaction.poa import PoaBlock
+
+logger = get_logger()
+
+# Number of seconds to wait for a sync to finish before trying to produce blocks
+_WAIT_SYNC_DELAY: int = 30
+
+# Number of seconds used in random delay calculation
+_RANDOM_DELAY_MULTIPLIER: int = 1
+
+
+class PoaBlockProducer:
+    """
+    This class is analogous to mining classes, but for Proof-of-Authority networks.
+    It waits for blocks to arrive, gets templates, and propagates new blocks accordingly.
+    """
+    __slots__ = (
+        '_log',
+        '_settings',
+        '_poa_settings',
+        '_reactor',
+        '_manager',
+        '_poa_signer',
+        '_signer_index',
+        '_started_producing',
+        '_start_producing_lc',
+        '_schedule_block_lc',
+        '_last_seen_best_block',
+    )
+
+    def __init__(
+        self,
+        *,
+        settings: HathorSettings,
+        reactor: ReactorProtocol,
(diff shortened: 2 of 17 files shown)


hathor/consensus/poa/poa_block_producer.py
```diff
@@ -0,0 +1,172 @@
...
+        """Calculate the expected timestamp for a new block."""
+        height = previous_block.get_height() + 1
+        is_in_turn = poa.is_in_turn(settings=self._poa_settings, height=height, signer_index=self._signer_index)
+        timestamp = previous_block.timestamp + self._settings.AVG_TIME_BETWEEN_BLOCKS
+        if is_in_turn:
+            return timestamp
+
+        signer_count = len(self._poa_settings.signers)
+        assert signer_count >= 1
+        random_offset = self._manager.rng.choice(range(signer_count * _RANDOM_DELAY_MULTIPLIER)) + 1
```
glevco, line 171: @msbrogli commented on https://github.com/HathorNetwork/hathor-core-side-dag/pull/11#discussion_r1618051818:

> This approach could potentially lead to orphan blocks. If the designated signer fails to propagate a new block promptly, other signers might simultaneously broadcast their blocks, resulting in conflicts.

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -0,0 +1,172 @@
...
+        self._log.debug(
+            'scheduling block production',
+            previous_block=previous_block.hash_hex,
+            previous_block_height=previous_block.get_height(),
+            delay=propagation_delay,
+        )
+
+    def _produce_block(self, previous_block: PoaBlock) -> None:
+        """Create and propagate a new block."""
+        block_templates = self._manager.get_block_templates(parent_block_hash=previous_block.hash)
```
glevco, line 141: @msbrogli commented on https://github.com/HathorNetwork/hathor-core-side-dag/pull/11#discussion_r1634979033:

> Why don't you get the previous block from `self._manager.tx_storage.get_best_block()`? Why are you getting this information when a block is scheduled (instead of when it is generated)?

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -0,0 +1,172 @@
...
+        block = block_templates.generate_mining_block(self._manager.rng, poa_block=True)
+        assert isinstance(block, PoaBlock)
+        block.weight = poa.calculate_weight(self._poa_settings, block, self._signer_index)
+        self._poa_signer.sign_block(block)
+        block.update_hash()
+
+        self._manager.on_new_tx(block, propagate_to_peers=False, fails_silently=False)
+        if not block.get_metadata().voided_by:
```
glevco, line 149: @msbrogli commented on https://github.com/HathorNetwork/hathor-core-side-dag/pull/11#discussion_r1634980005:

> Nice rule. I guess we can add it to the `on_new_tx()` propagation part.

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -0,0 +1,172 @@
...
+        self._settings = settings
+        self._poa_settings = settings.CONSENSUS_ALGORITHM
+        self._reactor = reactor
+        self._manager = manager
+        self._poa_signer = poa_signer
+        self._signer_index = self._calculate_signer_index(self._poa_settings, self._poa_signer)
+        self._last_seen_best_block: PoaBlock | None = None
+
+        self._started_producing = False
+        self._start_producing_lc = LoopingCall(self._start_producing)
```
glevco, line 80: @msbrogli commented on https://github.com/HathorNetwork/hathor-core-side-dag/pull/11#discussion_r1634983595:

> As it is a single execution to flip the `self._started_producing` flag, it seems safe to use `callLater()`.

hathor/consensus/poa/poa_block_producer.py
glevco: @msbrogli maybe, but I would rather not change it on this PR, as it may affect PoW block propagation on mainnet so it deserves more thought. For instance, could it be possible that a node observes a block as voided, propagates it, an then other nodes (which may be synced to different best chains), observe it as non-voided?

hathor/consensus/poa/poa_block_producer.py
glevco: @msbrogli I would still need to recurse using `callLater()` until the node is synced, and using the same argument, the loop would be stopped if for any reason `manager.can_start_mining()` raises an exception.

hathor/consensus/poa/poa_block_producer.py
msbrogli: But won't peers reject out-of-turn blocks in most cases? I mean, can a misconfigured peer generate and propagate out-of-turn blocks?

hathor/consensus/poa/poa_block_producer.py
msbrogli: I agree to do it in a separate PR. Let's discuss it there. As a quick answer: I guess your situation would be fine because the sync algorithm would get these peers sync'ed anyway.

hathor/consensus/poa/poa_block_producer.py
msbrogli: You're right. Ok. Thanks!

hathor/mining/block_template.py
```diff
@@ -60,11 +68,25 @@ def generate_mining_block(self, rng: Random, merge_mined: bool = False, address:
...
         block_timestamp = min(max(base_timestamp, self.timestamp_min), self.timestamp_max)
-        tx_outputs = [TxOutput(self.reward, output_script)]
-        cls: Union[type['Block'], type['MergeMinedBlock']] = MergeMinedBlock if merge_mined else Block
+        tx_outputs = []
+        if self.reward:
+            output_script = create_output_script(address) if address is not None else b''
+            tx_outputs = [TxOutput(self.reward, output_script)]
+
+        # TODO: Change to a generic function
+        cls: type[Block] | type[MergeMinedBlock] | type[PoaBlock]
```
msbrogli, line 79: I guess this method should receive the class itself or an enum (instead of `merge_mined: bool`).

hathor/consensus/poa/poa_block_producer.py
msbrogli: What if `previous_block != self._manager.tx_storage.get_best_block()`? Should we keep producing our block? Or should we wait for our next turn?

msbrogli (changes requested):

hathor/consensus/poa/poa_block_producer.py
glevco: Created issue https://github.com/HathorNetwork/hathor-core/issues/1066

hathor/mining/block_template.py
glevco: I agree, that's why I added the `TODO`, but initially I wanted to to it on a separate PR because I thought I would have to change a lot of call sites.

However I checked it and it wasn't a lot, so I did it in https://github.com/HathorNetwork/hathor-core/pull/1061/commits/2885ed533ccfaf10c646950817971e2d6e476350

hathor/cli/generate_genesis.py
```diff
@@ -0,0 +1,79 @@
+# Copyright 2021 Hathor Labs
```
jansegre, line 1: Outdated year.

hathor/consensus/poa/poa_block_producer.py
jansegre: Wouldn't it be better to accept out-of-turn blocks, even if the recovery strategy is inefficient it should at least be correct and we can implement a better one in the future, very possibly in a backwards compatible way. At least accepting out-of-turn blocks would make it easier to make a backwards compatible improvement.

jansegre (approved):

hathor/cli/generate_genesis.py
glevco: Fixed in https://github.com/HathorNetwork/hathor-core/pull/1061/commits/a750cbd94afddebb8e1d42494c61196e8b1cf5e4

hathor/consensus/poa/poa_block_producer.py
msbrogli: Thanks!

hathor/mining/block_template.py
msbrogli: Great! Thanks!

hathor/consensus/poa/poa_block_producer.py
glevco: @msbrogli I changed it to use a round-robin increasing delay instead of a random delay in a separate PR, https://github.com/HathorNetwork/hathor-core/pull/1074

hathor/consensus/poa/poa_block_producer.py
glevco: @msbrogli 

> But won't peers reject out-of-turn blocks in most cases?

What do you mean by "reject"? The blocks will be accepted, but voided if we receive an in-turn block for the same height.

> I mean, can a misconfigured peer generate and propagate out-of-turn blocks?

A well-configured peer will generate and propagate out-of-turn blocks when the in-turn signer is not online. A misconfigured peer will only do something different in a modified full node.

> What if previous_block != self._manager.tx_storage.get_best_block()? Should we keep producing our block? Or should we wait for our next turn?

As explained in the previous comment, in that case it'll simply be voided and not propagated.

glevco: @jansegre 

> Wouldn't it be better to accept out-of-turn blocks?

We _do_ accept out-of-turn blocks

hathor/consensus/poa/poa_block_producer.py
msbrogli: Ok. Thanks!

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -0,0 +1,173 @@
...
+        """Return the signer index for the given private key."""
+        public_key = poa_signer.get_public_key()
+        public_key_bytes = get_public_key_bytes_compressed(public_key)
+        sorted_signers = sorted(settings.signers)
+        try:
+            return sorted_signers.index(public_key_bytes)
+        except ValueError:
+            raise ValueError(f'Public key "{public_key_bytes.hex()}" not in list of PoA signers')
+
+    def _start_producing(self) -> None:
```
msbrogli, line 109: You can remove this method and the `_schedule_block_lc` and replace it by simply calling `self._manager.can_start_mining()` at the beginning of the `_schedule_block` method.

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -0,0 +1,173 @@
...
+        previous_block = self._manager.tx_storage.get_best_block()
+        if not self._started_producing or previous_block == self._last_seen_best_block:
+            return
+
+        self._last_seen_best_block = previous_block
+        now = self._reactor.seconds()
+        expected_timestamp = self._expected_block_timestamp(previous_block)
+        propagation_delay = 0 if expected_timestamp < now else expected_timestamp - now
+
+        self._reactor.callLater(propagation_delay, self._produce_block, previous_block)
```
msbrogli, line 131: You don't have any safety check whether there are two or more scheduled calls to `self._produce_block`. This might happen if the `propagation_delay` gets too large and this method is called again by the looping call. You can easily do it saving the `DelayedCall` returned by the `callLater()` call and set it back to none in the `_produce_block()` method.

hathor/consensus/poa/poa_block_producer.py
msbrogli: Thanks!

hathor/consensus/poa/poa_block_producer.py
glevco: Done in a separate PR to make it easy to update future tests: https://github.com/HathorNetwork/hathor-core/pull/1080

hathor/consensus/poa/poa_block_producer.py
glevco: Done in a separate PR to make it easy to update future tests: https://github.com/HathorNetwork/hathor-core/pull/1080

hathor/consensus/poa/poa_block_producer.py
msbrogli: Ok! Thanks!

hathor/consensus/poa/poa_block_producer.py
msbrogli: Thanks!

msbrogli (approved):

jansegre (approved):

Issue #1051, opened 2024-06-04
Refactor the `side_dag` command so it has a monitor process


glevco:
As discussed in person, let's do a monitor process in a separate PR so we can skip it for now.

_Originally posted by @msbrogli in https://github.com/HathorNetwork/hathor-core/pull/1039#discussion_r1626125585_

glevco:
- [ ] Correctly handle termination: https://github.com/HathorNetwork/hathor-core/pull/1039#discussion_r1630191356

PR #1070, opened 2024-06-17
fix(side-dag): remove init timeout

glevco:
### Motivation

The new `run_node_with_side_dag` command timed out in 10 seconds if the Hathor node did not start. This was incorrect because when loading from an existing storage, the Hathor node will always take more than 10 seconds to initialize.

This PR removes the timeout and simply waits indefinitely until the Hathor node is either ready or errored. This will be further improved when the monitor process is implemented (see https://github.com/HathorNetwork/hathor-core/issues/1051).

### Acceptance Criteria

- Remove Hathor node init timeout in favor of waiting.

### Checklist

### DIFF
--- a/hathor/cli/side_dag.py
+++ b/hathor/cli/side_dag.py
@@ -43,7 +43,7 @@ from hathor.cli.run_node import RunNode
 logger = get_logger()
 
 PRE_SETUP_LOGGING: bool = False
-HATHOR_NODE_INIT_TIMEOUT: int = 10
+HATHOR_NODE_INIT_WAIT_PERIOD: int = 10
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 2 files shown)


hathor/cli/side_dag.py
```diff
@@ -186,12 +186,8 @@ def is_option(arg_: str) -> bool:
...
-    if not conn.poll(HATHOR_NODE_INIT_TIMEOUT):
-        logger.critical(
-            f'side-dag node not started because hathor node failed to initialize before {HATHOR_NODE_INIT_TIMEOUT} '
-            f'seconds timeout'
-        )
-        return
+    while not conn.poll(HATHOR_NODE_INIT_WAIT_PERIOD):
+        logger.info('still waiting for hathor node to initialize...')
 
     message = conn.recv()
```
msbrogli, line 192: Should we add a timeout to this `recv()` call and log `still waiting for hathor node to initialize...`?

msbrogli (approved):

hathor/cli/side_dag.py
glevco: That's the exact purpose of the `while` loop above, when `conn.poll` returns `True` and it stops, it's guaranteed that there's a message available and therefore `recv()` will return immediately.

glevco: This has been removed in https://github.com/HathorNetwork/hathor-core/pull/1079

msbrogli: Got it. Make sense. Thanks!

PR #1075, opened 2024-06-25
feat(side-dag): add ability to update signers list

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1074

### Motivation

Implement ability to update the signers list on PoA networks by configuring start and end block heights where each signer is active. By the default, a signer is active from height 0 to infinity. By setting the start height on a new signer, the network can coordinate to add it without disrupting existing blocks.

### Acceptance Criteria

- Update PoA settings to be able to configure `start_height` and `end_height` for signers.
- Implement `poa.get_active_signers()` and update all usages of `poa_settings.signers` to use it instead.
- Update `PoaBlockProducer` to calculate the signer index depending on active signers.

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

PR #1078, opened 2024-07-01
feat(side-dag): add signers validation on peer hello

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1075

### Motivation

This PR adds a validation of PoA signers list when peers connect.

### Acceptance Criteria

- Add consensus algorithm validation on peer hello.
- Change `Builder` so it requires settings to be set.
- Update P2P related code to use injected settings.

### Checklist

### DIFF
--- a/hathor/consensus/consensus_settings.py
+++ b/hathor/consensus/consensus_settings.py
@@ -12,14 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import hashlib
 from abc import ABC, abstractmethod
 from enum import Enum, unique
 from typing import Annotated, Any, Literal, TypeAlias
 
-from pydantic import Field, NonNegativeInt, validator
+from pydantic import Field, NonNegativeInt, PrivateAttr, validator
 from typing_extensions import override
 
 from hathor.transaction import TxVersion
+from hathor.util import json_dumpb
 from hathor.utils.pydantic import BaseModel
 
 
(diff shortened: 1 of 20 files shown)


jansegre (approved):

hathor/consensus/consensus_settings.py
```diff
@@ -112,5 +130,12 @@ def _get_valid_vertex_versions(self, include_genesis: bool) -> set[TxVersion]:
 
         return versions
 
+    @override
+    def get_peer_hello_hash(self) -> str | None:
+        data = b''
+        for signer in self.signers:
```
msbrogli, line 136: This code enforces the order of the signers and I think this is a good verification. I think you can safely remove the `sorted()` from the verification so we ensure that the signed id is the same for all peers and that we can check the signer id simply by checking the settings file.

hathor/consensus/consensus_settings.py
```diff
@@ -112,5 +130,12 @@ def _get_valid_vertex_versions(self, include_genesis: bool) -> set[TxVersion]:
 
         return versions
 
+    @override
+    def get_peer_hello_hash(self) -> str | None:
```
msbrogli, line 134: This is immutable. Should we calculate it once and just cache the digest?

hathor/consensus/consensus_settings.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1078/commits/b58b125dae7da71d39f750fa53d8729f4efe3463

hathor/consensus/consensus_settings.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1078/commits/b58b125dae7da71d39f750fa53d8729f4efe3463

msbrogli: Thanks!

msbrogli (dismissed):

msbrogli (approved):

msbrogli (approved):

jansegre (approved):

PR #1080, opened 2024-07-08
refactor(side-dag): general improvements

glevco:
### Motivation

Implement some of the requested changes in previous PRs.

### Acceptance Criteria

- Remove `poa.in_turn_signer_index()` and add `poa.get_signer_index_distance()`.
- Update `poa.calculate_weight()` to decrease the block weight inversely proportional to the index distance.
- Changes in `PoaBlockProducer`:
  - Merge both looping calls into one.
  - Add delayed block production cancellation.
  - Update logs.
  - Change block delay function to include a constant offset for out of turn blocks, and increase the interval between each signer.
- Update tests accordingly.

### Checklist

### DIFF
--- a/hathor/consensus/poa/poa_block_producer.py
+++ b/hathor/consensus/poa/poa_block_producer.py
@@ -24,6 +24,7 @@ from hathor.conf.settings import HathorSettings
 from hathor.consensus import poa
 from hathor.consensus.consensus_settings import PoaSettings
 from hathor.crypto.util import get_public_key_bytes_compressed
+from hathor.pubsub import EventArguments, HathorEvents
 from hathor.reactor import ReactorProtocol
 from hathor.util import not_none
 
@@ -35,11 +36,8 @@ if TYPE_CHECKING:
 
 logger = get_logger()
 
-# Number of seconds to wait for a sync to finish before trying to produce blocks
-_WAIT_SYNC_DELAY: int = 30
-
 # Number of seconds used between each signer depending on its distance to the expected signer
-_SIGNER_TURN_INTERVAL: int = 1
+_SIGNER_TURN_INTERVAL: int = 10
 
 
 class PoaBlockProducer:
@@ -54,11 +52,9 @@ class PoaBlockProducer:
         '_reactor',
         '_manager',
         '_poa_signer',
-        '_started_producing',
-        '_start_producing_lc',
-        '_schedule_block_lc',
         '_last_seen_best_block',
         '_delayed_call',
+        '_start_producing_lc',
     )
 
     def __init__(self, *, settings: HathorSettings, reactor: ReactorProtocol, poa_signer: PoaSigner) -> None:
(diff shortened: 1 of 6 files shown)


msbrogli (approved):

jansegre (approved):

glevco:
I had to make some changes because I noticed the behavior was not working as expected, in relation to cancellation of block production. Before these changes,

- Since delayed calls were not cancelled when the delayed block was in turn, that meant we could have more than one active delayed call, and only one of them would have a pointer (in `self._delayed_call`).
- In fact, the block used to check the turn was incorrect, as it was using the delayed call argument which is the `previous_block`, not the new one being generated.
- We were not cancelling block production when a new best block arrived from another peer.

All these issues have been addressed in a separate commit, https://github.com/HathorNetwork/hathor-core/pull/1080/commits/452b00e0e2957d2e1b0d7dfe755c0e04d180f108.

jansegre (approved):

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -113,21 +103,49 @@ def _get_signer_index(self, previous_block: Block) -> int | None:
...
+        if isinstance(block, PoaBlock) and not block.weight == poa.BLOCK_WEIGHT_IN_TURN:
+            self._log.info('received out of turn block', block=block.hash_hex, signer_id=block.signer_id)
+
+        self._schedule_block()
 
     def _schedule_block(self) -> None:
         """Schedule propagation of a new block."""
+        if not self.manager.can_start_mining():
+            # We're syncing, so we'll try again later
+            self._log.info('cannot produce new block, node not synced')
```
msbrogli, line 141: I feel that the full node would not recover from getting out of sync after it was already synced. From recovery, I mean that this full node would stop producing blocks.

hathor/consensus/poa/poa_block_producer.py
```diff
@@ -70,14 +66,9 @@ def __init__(self, *, settings: HathorSettings, reactor: ReactorProtocol, poa_si
...
         self._last_seen_best_block: Block | None = None
-
-        self._started_producing = False
-        self._start_producing_lc = LoopingCall(self._start_producing)
-        self._start_producing_lc.clock = self._reactor
-
-        self._schedule_block_lc = LoopingCall(self._schedule_block)
-        self._schedule_block_lc.clock = self._reactor
         self._delayed_call: IDelayedCall | None = None
+        self._start_producing_lc: LoopingCall = LoopingCall(self._safe_start_producing)
```
msbrogli, line 70: I still don't understand why we need this `LoopingCall`. Anyway, I'm ok to get this PR merged and get this changed later because it has no impact on mainnet and testnet.

msbrogli (dismissed):

hathor/consensus/poa/poa_block_producer.py
glevco: The full node recovers as long as it receives a block from another peer. I added an assertion in an existing test to cover this, in https://github.com/HathorNetwork/hathor-core/pull/1080/commits/0d4fea63306a70d5d0d4b347fa2d0b3ef3a1b97b

hathor/consensus/poa/poa_block_producer.py
glevco: Using either a `LoopingCall` or a recursive `callLater`, separate from the main `callLater` logic (triggered reactively via pubsub) were the only good ways I found to implement this while covering the following requirements:

- A single full node should be able to produce blocks by itself.
- If block production throws during the "start producing" phase, production should be tried again. However, if it fails after a call via pubsub, it should **not** be retried.
- We have to consider that the call to `can_start_mining()` may throw an exception.

I created an issue to review this later: https://github.com/HathorNetwork/hathor-core/issues/1097

jansegre (approved):

PR #1083, opened 2024-07-12
feat: add sync-v2 benchmark to CI

glevco:
### Motivation

This PR implements the basic infrastructure for running benchmarks in our CI through GitHub Actions and the [Bencher](https://bencher.dev) tool, which automatically tracks changes in performance across PRs (though it is more flexible than that). Bencher is [open source](https://github.com/bencherdev/bencher) and provides their cloud to run the Bencher service, which we're using through a free tier plan. We can consider upgrading our plan or running our own self-hosted service if necessary, but I don't see any reasons for now. This means our benchmarking results are public on the [Bencher Console](https://bencher.dev/console/projects/hathor-core/).

We setup a benchmark for tracking sync-v2 performance. It works by:

1. Setting up a temporary node connected to testnet that exits after downloading 20k blocks.
2. Running a server node in the background, connected only on localhost, and using the database with the previously downloaded blocks.
3. Running a [Hyperfine](https://github.com/sharkdp/hyperfine) benchmark integrated with Bencher, which starts a new full node that syncs 20k blocks with the server node, and exits.

By doing this we isolate the network and test only the node performance itself. Of course there may be unpredictable variations in GitHub machines, but we'll only understand this impact by observing the metrics over time.

The job described above runs in two workflows, one for every push on `master`, which sets the baseline for our metrics, and one for each PR to `master`, which compares the metric value to the baseline and triggers a CI error if it's above a certain threshold. This is configured though the Bencher console directly. I configured both upper and lower thresholds of 10% (it's useful to set a lower threshold to track performance improvements, as an unpredicted drop in time may indicate new bugs).

There are improvements to be made in the future:

- [ ] Instead of using a node connected to testnet and syncing 20k blocks every time, we could have a pre-downloaded database and save it in S3, for example. Currently, this downloading step is way slower than the benchmark itself (to be clear: it does not affect the benchmark time, but it does affect the CI job duration).
- [ ] We currently only sync 20k blocks from testnet, and that doesn't include any transactions. We could test the improvement above with different database sizes and find a tradeoff to include transactions. We could even create a fake database with specific blocks/transactions ratios and benchmark them.
- [ ] We can add other benchmarks such as node initialization time, vertex (de)serialization time, API response times, etc.
- [ ] We could add this to other projects, not only hathor-core.

### Acceptance Criteria

- Create a new reusable action `setup-hathor-env`, used by our main CI and the new benchmarking workflows.
- Add new `Continuous Benchmarking base branch` and `Continuous Benchmarking PRs` jobs.

### Checklist

### DIFF
--- a/extras/benchmarking/benchmark_sync_v2.sh
+++ b/extras/benchmarking/benchmark_sync_v2.sh
@@ -1,0 +1,45 @@
+N_BLOCKS=20000
+CACHE_SIZE=100000
+TESTNET_DATA_DIR=server-data
+TCP_PORT=40403
+AWAIT_INIT_DELAY=10
+N_RUNS=2
+BENCH_FILE=bench_results.json
+BENCH_DATA_DIR=bench-data
+
+BLUE='\033[0;34m'
+NO_COLOR='\033[0m'
+
+echo "${BLUE}Downloading testnet data...${NO_COLOR}"
+mkdir $TESTNET_DATA_DIR
+poetry run hathor-cli quick_test --testnet --data $TESTNET_DATA_DIR --quit-after-n-blocks $N_BLOCKS > /dev/null 2>&1
+
+echo "${BLUE}Running server node in the background...${NO_COLOR}"
+poetry run hathor-cli run_node \
+  --testnet \
+  --data $TESTNET_DATA_DIR \
+  --cache \
+  --cache-size $CACHE_SIZE \
+  --x-localhost-only \
+  --listen tcp:$TCP_PORT \
+  > /dev/null 2>&1 &
+
+# Await initialization
+sleep $AWAIT_INIT_DELAY
+
+echo "${BLUE}Running benchmark...${NO_COLOR}"
+hyperfine \
+  --runs $N_RUNS \
+  --export-json $BENCH_FILE \
+  --command-name "sync-v2 (up to $N_BLOCKS blocks)" \
+  --prepare "rm -rf $BENCH_DATA_DIR && mkdir $BENCH_DATA_DIR" \
+  "
+    poetry run hathor-cli quick_test \
+      --testnet \
+      --data $BENCH_DATA_DIR \
+      --cache \
+      --cache-size $CACHE_SIZE \
+      --x-localhost-only \
+      --bootstrap tcp://localhost:$TCP_PORT \
+      --quit-after-n-blocks $N_BLOCKS
+  "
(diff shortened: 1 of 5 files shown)


extras/benchmarking/benchmark_sync_v2.sh
```diff
@@ -0,0 +1,45 @@
+N_BLOCKS=100
```
glevco, line 1: Change this back to 20k. I only changed it to 100 to run the CI quickly and generate a fake result to demonstrate Bencher plots, as the real plot will only exist after this PR is merged to master.

extras/benchmarking/benchmark_sync_v2.sh
```diff
@@ -0,0 +1,45 @@
...
+CACHE_SIZE=100000
+TESTNET_DATA_DIR=server-data
+TCP_PORT=40403
+AWAIT_INIT_DELAY=10
+N_RUNS=2
+BENCH_FILE=bench_results.json
+BENCH_DATA_DIR=bench-data
+
+BLUE='\033[0;34m'
+NO_COLOR='\033[0m'
```
glevco, line 11: Fix: code below is using `NC`, change to `NO_COLOR`

jansegre (dismissed):
Looks good to me.

extras/benchmarking/benchmark_sync_v2.sh
msbrogli, line 1: We can also use the `load_from_logs` cli command. This way, we can use the simulator to create scenarios (or even extract the logs from tests that already exists) and upload a gzip log to S3. Thus, the CI just downloads the log, unzip it and run the `load_from_logs`.

msbrogli (approved):

extras/benchmarking/benchmark_sync_v2.sh
glevco: I created an issue to implement this in a separate PR: https://github.com/HathorNetwork/hathor-core/issues/1105

extras/benchmarking/benchmark_sync_v2.sh
```diff
@@ -0,0 +1,45 @@
+N_BLOCKS=100 # TODO: Change to 20000
```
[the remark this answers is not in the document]
glevco: Done

jansegre (approved):

extras/benchmarking/benchmark_sync_v2.sh
msbrogli: Great! Thanks!

msbrogli (approved):

PR #1085, opened 2024-07-15
refactor(settings): more settings injection refactors

glevco:
### Motivation

Continue refactoring `HathorSettings` usages to use an injected instance rather than global.

### Acceptance Criteria

- Updated to use injected settings:
  - `BaseIndex`, `MemoryIndexesManager`, and `MemoryHeightIndex`.
  - `BaseTransaction` and all its subclasses, and `TransactionMetadata`.
  - `get_deposit_amount()` and `get_withdraw_amount()` functions.
- Updated `TransactionStorage` to inject its settings into constructed genesis.

### Checklist

### DIFF
--- a/hathor/indexes/base_index.py
+++ b/hathor/indexes/base_index.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Optional
 
@@ -22,6 +24,7 @@ from hathor.indexes.scope import Scope
 from hathor.transaction.base_transaction import BaseTransaction
 
 if TYPE_CHECKING:  # pragma: no cover
+    from hathor.conf.settings import HathorSettings
     from hathor.indexes.manager import IndexesManager
 
 logger = get_logger()
@@ -33,8 +36,8 @@ class BaseIndex(ABC):
     This class exists so we can interact with indexes without knowing anything specific to its implemented. It was
     created to generalize how we initialize indexes and keep track of which ones are up-to-date.
     """
-    def __init__(self) -> None:
-        self._settings = get_global_settings()
+    def __init__(self, *, settings: HathorSettings | None = None) -> None:
+        self._settings = settings or get_global_settings()
         self.log = logger.new()
 
     def init_start(self, indexes_manager: 'IndexesManager') -> None:
--- a/hathor/transaction/transaction_metadata.py
+++ b/hathor/transaction/transaction_metadata.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Optional
 
@@ -24,6 +26,7 @@ from hathor.util import practically_equal
 if TYPE_CHECKING:
     from weakref import ReferenceType  # noqa: F401
 
+    from hathor.conf.settings import HathorSettings
     from hathor.transaction import BaseTransaction
     from hathor.transaction.storage import TransactionStorage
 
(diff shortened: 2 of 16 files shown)


jansegre (approved):

hathor/indexes/base_index.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
```
msbrogli, line 15: Why is it needed?

hathor/transaction/transaction_metadata.py
```diff
@@ -129,7 +133,7 @@ def __init__(
 
         self.feature_activation_bit_counts = feature_activation_bit_counts
 
-        settings = get_global_settings()
+        settings = settings or get_global_settings()
 
         # Genesis specific:
         if hash is not None and is_genesis(hash, settings=settings):
```
msbrogli, line 139: Is this the only use for settings? Should we refactor this `__init__()` method to receive a `is_genesis: bool` attribute (instead of settings)? Or check with the storage?

hathor/transaction/util.py
```diff
@@ -48,13 +51,11 @@ def unpack_len(n: int, buf: bytes) -> tuple[bytes, bytes]:
     return buf[:n], buf[n:]
 
 
-def get_deposit_amount(mint_amount: int) -> int:
-    settings = get_global_settings()
+def get_deposit_amount(settings: HathorSettings, mint_amount: int) -> int:
```
msbrogli, line 54: Should this method receive the percentage and let the caller pass it?

hathor/transaction/util.py
```diff
@@ -48,13 +51,11 @@ def unpack_len(n: int, buf: bytes) -> tuple[bytes, bytes]:
...
     return ceil(abs(settings.TOKEN_DEPOSIT_PERCENTAGE * mint_amount))
 
 
-def get_withdraw_amount(melt_amount: int) -> int:
-    settings = get_global_settings()
+def get_withdraw_amount(settings: HathorSettings, melt_amount: int) -> int:
```
msbrogli, line 58: Should this method receive the percentage and let the caller pass it?

msbrogli (approved):

hathor/indexes/base_index.py
glevco: To import `HathorSettings` only for type checking. Before, we would do `"HathorSettings"`, but I'm now pushing into using this which allows using the type name normally, without quotes.

hathor/transaction/util.py
glevco: What's your reasoning? In my opinion, we shouldn't. This would transfer the responsibility of passing the correct percentage to the caller, who would have to know to always pass `settings.TOKEN_DEPOSIT_PERCENTAGE` anyway, being prone to errors as users could call this with incorrect values. Receiving settings instead, leaves this logic internal to the function and impossible to mistake.

hathor/transaction/util.py
glevco: Same as https://github.com/HathorNetwork/hathor-core/pull/1085#discussion_r1679874120

hathor/transaction/transaction_metadata.py
glevco: Similarly to what I described in https://github.com/HathorNetwork/hathor-core/pull/1085#discussion_r1679874120, I don't think there's any advantage in changing this. There's no benefit in leaving the responsibility of passing the correct value of `is_genesis` to the caller, as this is an internal logic to this class. About the storage, it's not available in this context, so it would also have to be passed as an argument anyway. So it's better to just pass the settings.

hathor/indexes/base_index.py
msbrogli: Got it. Thanks!

hathor/transaction/transaction_metadata.py
msbrogli: Got it. Ok. We can do it later anyways.

hathor/transaction/util.py
msbrogli: Got it. Ok. We can do it later anyways.

hathor/transaction/util.py
msbrogli: Agreed. Thanks!

PR #1106, opened 2024-08-06
feat(event-queue): add tx data to VERTEX_REMOVED

glevco:
### Motivation

Our main Event Queue use case (wallet-service) asked us to provide the whole vertex in the `VERTEX_REMOVED` event data, instead of just the vertex ID.

### Acceptance Criteria

- Change `VERTEX_REMOVED` event to use `TxDataWithoutMeta` instead of `VertexIdData`.
- Delete `VertexIdData` as it's not used anymore.

### Checklist


hathor/pubsub.py
```diff
@@ -62,7 +61,7 @@ class HathorEvents(Enum):
 
         CONSENSUS_TX_REMOVED:
             Triggered when a tx is removed because it became invalid (due to a reward lock check)
-            Publishes the tx hash
+            Publishes the tx object
```
msbrogli, line 64: Should we emphasize that the metadata will have the values just before the removal? I'm wondering whether we should just skip the metadata information.

msbrogli (dismissed):

hathor/pubsub.py
glevco: I think it makes sense to remove the metadata in this case. I'll check with Abadesso if it would complicate things on the wallet service side.

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1106/commits/9a42b1f050682e28b5bc61bb85ef55d9ab0f3f4f

jansegre (approved):

hathor/pubsub.py
msbrogli: Great! Thanks!

msbrogli (approved):

PR #1107, opened 2024-08-07
feat(cli): early protection for invalid prometheus prefix

jansegre:
### Motivation

We had a report from someone accidentally using an invalid prefix through `HATHOR_PROMETHEUS_PREFIX` which had a message that was hard to understand.

### Acceptance Criteria

- Check for invalid Prometheus prefix when building and give a more helpful message.

### Checklist

### DIFF
--- a/hathor/builder/resources_builder.py
+++ b/hathor/builder/resources_builder.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import re
 from typing import TYPE_CHECKING, Any, Optional
 
 from autobahn.twisted.resource import WebSocketResource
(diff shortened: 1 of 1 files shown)


msbrogli (approved):

hathor/builder/resources_builder.py
```diff
@@ -60,9 +86,14 @@ def build(self) -> Optional[server.Site]:
         return None
 
     def create_prometheus(self) -> PrometheusMetricsExporter:
+        prometheus_prefix = self._args.prometheus_prefix
+        if self._args.prometheus_prefix and is_prometheus_metric_name_valid(prometheus_prefix):
```
glevco, line 90: Missing `not`?

glevco (changes requested):

hathor/builder/resources_builder.py
jansegre: True! I missed it after refactoring.

jansegre: Fixed.

glevco (approved):

hathor/builder/resources_builder.py
glevco: Thanks!

msbrogli (approved):

PR #1125, opened 2024-08-28
feat: enable cache by default

glevco:
### Motivation

### Acceptance Criteria

### Checklist


jansegre (approved):

jansegre (dismissed):

hathor/builder/cli_builder.py
```diff
@@ -162,7 +162,8 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
 
         self.log.info('with storage', storage_class=type(tx_storage).__name__, path=self._args.data)
         if self._args.cache:
-            self.check_or_raise(not self._args.memory_storage, '--cache should not be used with --memory-storage')
```
msbrogli, line 165: Should we keep this verification? We might modify the error message to suggest using `--disable-cache`.

msbrogli (approved):

hathor/builder/cli_builder.py
glevco: I don't think so, because it's inconvenient. Usually `--memory-storage` is only used for development, and keeping this check would mean we would have to pass `--memory-storage --disable-cache` every time. Instead, I think it's obvious that the memory storage is indeed using cache because well, it is "cache" _itself_ . What do you think?

jansegre: I'd rather have `--memory-storage` imply that cache is disabled. But it might make sense to do this check with `--cache-size`, maybe error when used with `--disable-cache` and warn that it has no effect when used with `--memory-storage`.

glevco: I added those checks/warnings in https://github.com/HathorNetwork/hathor-core/pull/1125/commits/23fb591f832684a8705251d196ffbdf0481ebf85. I had to move the default cache size to the CliBuilder because otherwise it would trigger the check when `--disable-cache` was used.

msbrogli: Great! Thanks!

msbrogli (approved):

jansegre (approved):

PR #1126, opened 2024-08-29
feat(ws): Add flow control to the tx history streamer

msbrogli:
### Motivation

### Acceptance Criteria

### Checklist


hathor/websocket/protocol.py
```diff
@@ -140,9 +142,11 @@ def fail_if_history_streaming_is_disabled(self) -> bool:
         ))
         return True
 
-    def _create_streamer(self, stream_id: str, search: AddressSearch) -> None:
+    def _create_streamer(self, stream_id: str, search: AddressSearch, window_size: Optional[int]) -> None:
         """Create the streamer and handle its callbacks."""
         self._history_streamer = HistoryStreamer(protocol=self, stream_id=stream_id, search=search)
+        if window_size:
+            self._history_streamer.set_sliding_window_size(window_size)
```
msbrogli, line 149: Should we add a minimum and maximum value for the window size?

jansegre (dismissed):

tests/websocket/test_streamer.py
```diff
@@ -39,7 +39,11 @@ def test_streamer(self) -> None:
             addresses.append(AddressItem(idx, wallet.get_address(wallet.get_key_at_index(idx))))
 
         # Create the expected result.
-        expected_result: list[dict[str, Any]] = [{'type': 'stream:history:begin', 'id': stream_id}]
+        expected_result: list[dict[str, Any]] = [{
+            'type': 'stream:history:begin',
+            'id': stream_id,
+            'sliding_window_size': 10000
```
r4mmer, line 45: issue(blocking): This should be `window_size`.

`sliding_window_size` is the internal name, but the protocol extracts `window_size` from the client's message

r4mmer (changes requested):

hathor/websocket/protocol.py
```diff
@@ -140,9 +142,17 @@ def fail_if_history_streaming_is_disabled(self) -> bool:
...
+    def _create_streamer(self, stream_id: str, search: AddressSearch, window_size: int | str | None) -> None:
         """Create the streamer and handle its callbacks."""
         self._history_streamer = HistoryStreamer(protocol=self, stream_id=stream_id, search=search)
+        if window_size is not None:
+            if isinstance(window_size, int):
+                self._history_streamer.set_sliding_window_size(window_size)
+            elif window_size == 'inf':
+                self._history_streamer.set_sliding_window_size(None)
+            else:
+                # Raise an exception?
```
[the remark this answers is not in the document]
jansegre: I'm not sure, but it should never happen unless there's a coding error that mypy did not get.

hathor/websocket/streamer.py
```diff
@@ -69,17 +70,51 @@ def __init__(self,
...
+        self.resume_if_possible()
+
+    def resume_if_possible(self) -> None:
+        if not self.should_pause_streaming() and not self._is_running:
+            self.resumeProducing()
+
     def start(self) -> Deferred[bool]:
         """Start streaming items."""
-        self.send_message(StreamBeginMessage(id=self.stream_id))
+        self.send_message(StreamBeginMessage(id=self.stream_id, sliding_window_size=self._sliding_window_size))
```
r4mmer, line 117: The client sends `window_size`, why not send `window_size` back?

hathor/websocket/protocol.py
```diff
@@ -140,9 +142,14 @@ def fail_if_history_streaming_is_disabled(self) -> bool:
...
 
-    def _create_streamer(self, stream_id: str, search: AddressSearch) -> None:
+    def _create_streamer(self, stream_id: str, search: AddressSearch, window_size: int | None) -> None:
         """Create the streamer and handle its callbacks."""
         self._history_streamer = HistoryStreamer(protocol=self, stream_id=stream_id, search=search)
+        if window_size is not None:
+            if window_size < 0:
+                self._history_streamer.set_sliding_window_size(None)
+            else:
+                self._history_streamer.set_sliding_window_size(window_size)
```
r4mmer, line 152: question(if-minor): Why not pass the `window_size` on the constructor?

tests/websocket/test_streamer.py
r4mmer, line 45: Currently the `window_size` is configured into the history streamer after the instance is constructed, so if we want this window size of 10000 to be valid we should configure it manually before line 82 (this is the reason for the tests failing I believe).
The alternative would be (as suggested on another comment) to move the window_size to the contructor and also add this 10000 on line 81.

r4mmer (changes requested):

jansegre (approved):
Looks good to me.

hathor/websocket/protocol.py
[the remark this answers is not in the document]
msbrogli: This part was changed. So I'm closing this thread.

tests/websocket/test_streamer.py
msbrogli: There were other outdated things in this test.

hathor/websocket/protocol.py
msbrogli: I guess we don't need it since the flow control can be disabled.

hathor/websocket/streamer.py
```diff
@@ -69,17 +70,45 @@ def __init__(self,
...
+    def set_ack(self, ack: int) -> None:
+        """Set the ack value for flow control.
+
+        If the new value is bigger than the previous value, the streaming might be resumed.
+        """
+        if ack <= self._last_ack:
+            # We might receive outdated or duplicate ACKs, and we can safely ignore them.
+            return
+        if ack >= self._next_sequence_number:
+            # Should we raise an exception here?
```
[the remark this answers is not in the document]
msbrogli: Done! I just kept the streaming going so the client can decide whether it should be stopped or not.

tests/websocket/test_streamer.py
msbrogli: Done!

hathor/websocket/streamer.py
msbrogli: Done!

hathor/websocket/protocol.py
msbrogli: I'd rather have a separate call for now because the constructor would have to be able to differentiate between Enabled, Disabled, and Default. I could create a `_NOT_PROVIDED` object to do it but I guess it's not worth the effort.

tests/websocket/test_streamer.py
```diff
@@ -39,7 +39,11 @@ def test_streamer(self) -> None:
             addresses.append(AddressItem(idx, wallet.get_address(wallet.get_key_at_index(idx))))
 
         # Create the expected result.
-        expected_result: list[dict[str, Any]] = [{'type': 'stream:history:begin', 'id': stream_id}]
+        expected_result: list[dict[str, Any]] = [{
+            'type': 'stream:history:begin',
+            'id': stream_id,
+            'sliding_window_size': None,
```
r4mmer, line 45:
```suggestion
            'window_size': None,
```

r4mmer (changes requested):

jansegre (approved):

r4mmer (approved):

PR #1128, opened 2024-09-02
perf(verification): improve tx verification speed

glevco:
### Motivation

### Acceptance Criteria

### Checklist


jansegre:
What about tests with 1 or 2 inputs? Since 255 is the worst case for verification it's also probably the best case for optimization gains, but it's not the most common case by far. My reasoning is this:

- If tests with few inputs show no significant slowdown, then it's a no-brainer and this is an absolute improvement (I don't expect this)
- If tests with few inputs are significantly slower because of any added overhead, then we might want to consider keeping both methods and look for the N for which N inputs seem to start benefiting (this might not be the case, but I'm not sure)
- If tests with few inputs are only slightly slower, then I think it's probably not worth any additional complexity, I'm not sure what "slightly slower" would be, but I guess we can find out (this is more what I would expect, but I think it's worth finding out)

PR #1129, opened 2024-09-04
fix(ws): Add a graceful close mechanism to handle late messages and prevent errors

msbrogli:
### Motivation

When the stream ended, an ACK message arrived late, causing an error response to be generated.

Check out the state diagram in the Reference-level explanation here: https://github.com/HathorNetwork/internal-rfcs/pull/25

### Acceptance Criteria

1. Do nothing if the ACK received from the client is duplicated.
2. Stop the streamer if the ACK received from the client decreases.
3. Stop the streamer if the ACK received from the client is higher than the last sent message sequence number.
4. Add some protection to the `pauseProducing`, `stopProducing`, and `resumeProducing` methods.
5. Replace flags `_started`, `_paused`, and `_stop` by a state machine.
6. Introduce the new state `CLOSING` which means the streaming has ended and it is just waiting for the last ACK.
7. Modify streamer to expect ack for the stream end message.

### Checklist

### DIFF
--- a/hathor/websocket/streamer.py
+++ b/hathor/websocket/streamer.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from enum import Enum, auto
 from typing import TYPE_CHECKING, Optional
 
 from twisted.internet.defer import Deferred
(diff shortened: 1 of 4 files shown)


hathor/websocket/streamer.py
```diff
@@ -72,23 +94,34 @@ def __init__(self,
...
-        self._is_running = False
-        self._paused = False
-        self._stop = False
+        # Execution control
+        # -----------------
+        self._state = StreamerState.NOT_STARTED
+        # Used to mark that the streamer is currently running its main loop and sending messages.
+        self._is_main_loop_running = False
+        # Used to mark that the streamer was paused by the transport layer.
+        self._is_paused_by_transport = False
```
jansegre, line 110: What's the difference between those and the `ACTIVE` and `PAUSED` states?

hathor/websocket/streamer.py
```diff
@@ -102,87 +135,130 @@ def set_ack(self, ack: int) -> None:
...
-        assert not self._is_running
-        self._is_running = True
+        assert not self._is_main_loop_running
+        self._state = StreamerState.ACTIVE
+        self._is_main_loop_running = True
         try:
             await self._async_run_unsafe()
         finally:
-            self._is_running = False
+            self._is_main_loop_running = False
```
jansegre, line 261: Should this mean the state is not `ACTIVE` anymore?

jansegre (dismissed):

r4mmer (dismissed):

hathor/websocket/streamer.py
msbrogli: `ACTIVE` indicates that the streamer is actively attempting to send messages. However, the execution of the main loop is somewhat independent of this state. The main loop can stop due to flow control mechanisms, meaning that the streamer may be in the `ACTIVE` state without the main loop actually running.

Although I considered introducing separate states like `ACTIVE_NOT_RUNNING` and `ACTIVE_RUNNING` to differentiate these scenarios, I ultimately decided to maintain a single `ACTIVE` state, using a supporting flag to track whether the main loop is currently running.

hathor/websocket/streamer.py
msbrogli: You're right about the `PAUSED` state. I just refactored out the `self._is_paused_by_transport`. Nice catch! Thanks!

I answered about the `ACTIVE` in this thread: https://github.com/HathorNetwork/hathor-core/pull/1129#discussion_r1747376634

r4mmer (approved):

jansegre (approved):

r4mmer (approved):

PR #1133, opened 2024-09-09
refactor(benchmark): change benchmark CI to download data from logs in AWS

glevco:
### Motivation

Currently, CI benchmark directly connects to testnet to download benchmark data (it performs a sync all the way up to 20k blocks). This PR changes it to use the latest load-from-logs CLI tool update, making the CI run faster and more reliably. Then, we'll also be able to introduce new CI benchmarks in other PRs, including for example syncing a "dummy" network with more transactions than blocks.

### Acceptance Criteria

- Change Sync-v2 benchmark CI to download benchmark data from a logs file in AWS instead of directly connecting to the network.
- Update config files accordingly and improve organization.

### Checklist


jansegre (approved):
Should we call it something else other than "sync_v2" now? Using load-from-logs shouldn't trigger any sync code, right? Maybe "load_sample". Other than that it looks good to me.

glevco:
@jansegre 

> Should we call it something else other than "sync_v2" now? Using load-from-logs shouldn't trigger any sync code, right? Maybe "load_sample". Other than that it looks good to me.

The benchmark is still the same, it uses Sync-v2. I only changed the way the server node boots up (from syncing with the network vs from logs). Then, a second node connects to the server node and the measurement is made on the sync of that second node.

jansegre (approved):

msbrogli (approved):

PR #1139, opened 2024-09-30
fix: removed txs from test wallet

glevco:
### Motivation

We noticed that the `wallet.get_balance_per_address()` method was not working correctly in the end of the reward lock event simulation — it was not considering the removed tx. This PR fixes this.

### Acceptance Criteria

- Add voidance assertion in `_remove_transactions()`.
- Prevent event processing from reloading metadata.
- Add `CONSENSUS_TX_REMOVED` event handling on test wallet.
- Add typing to `test_reward_lock.py`.
- Add balance and voidance assertions in reward lock test and event simulation scenario.

### Checklist

### DIFF
--- a/hathor/consensus/consensus.py
+++ b/hathor/consensus/consensus.py
@@ -210,6 +210,7 @@ class ConsensusAlgorithm:
         for tx in txs:
             tx_meta = tx.get_metadata()
             assert not tx_meta.validation.is_checkpoint()
+            assert bool(tx_meta.voided_by), 'removed txs must be voided'
             for parent in set(tx.parents) - txset:
                 parents_to_update[parent].append(tx.hash)
             dangling_children.update(set(tx_meta.children) - txset)
--- a/hathor/wallet/base_wallet.py
+++ b/hathor/wallet/base_wallet.py
@@ -126,6 +126,7 @@ class BaseWallet:
 
         self.pubsub_events = [
             HathorEvents.CONSENSUS_TX_UPDATE,
+            HathorEvents.CONSENSUS_TX_REMOVED,
         ]
 
         if reactor is None:
@@ -176,6 +177,12 @@ class BaseWallet:
         data = args.__dict__
         if key == HathorEvents.CONSENSUS_TX_UPDATE:
             self.on_tx_update(data['tx'])
+        elif key == HathorEvents.CONSENSUS_TX_REMOVED:
+            # we use the same method as above because a removed tx is also voided
+            tx = data['tx']
+            assert isinstance(tx, Transaction)
+            assert bool(tx.get_metadata().voided_by)
+            self.on_tx_update(tx)
         else:
             raise NotImplementedError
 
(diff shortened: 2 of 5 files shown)


hathor/consensus/consensus.py
```diff
@@ -210,6 +210,7 @@ def _remove_transactions(
         for tx in txs:
             tx_meta = tx.get_metadata()
             assert not tx_meta.validation.is_checkpoint()
+            assert bool(tx_meta.voided_by), 'removed txs must be voided'
```
msbrogli, line 213: Why?

hathor/wallet/base_wallet.py
```diff
@@ -176,6 +177,9 @@ def handle_publish(self, key: HathorEvents, args: EventArguments) -> None:
         data = args.__dict__
         if key == HathorEvents.CONSENSUS_TX_UPDATE:
             self.on_tx_update(data['tx'])
+        elif key == HathorEvents.CONSENSUS_TX_REMOVED:
+            # we use the same method as above because a removed tx is also voided
+            self.on_tx_update(data['tx'])
```
msbrogli, line 182: Should we assert here that this tx is voided?

msbrogli (dismissed):

hathor/wallet/base_wallet.py
glevco: I guess it's a bit redundant, but no harm. Done in https://github.com/HathorNetwork/hathor-core/pull/1139/commits/6b95cb0b3e75f7551cb3c4f3f24f455de85ffdeb

hathor/consensus/consensus.py
glevco: This is an invariant that we currently uphold in our code and if we ever introduce something that fails this assertion, we must know.

The `_remove_transactions()` method is called with the results from `storage.compute_transactions_that_became_invalid()`, which calls `tx.set_validation(ValidationState.INVALID)` for each tx to be removed. This setter calls `_mark_partially_validated()` which sets `b'pending-validation'` in the `voided_by`.

glevco: And most importantly, semantically it makes sense: we only ever remove transactions that became invalid, and by definition if transactions are invalid they must have also be voided.

jansegre (dismissed):

hathor/wallet/base_wallet.py
msbrogli: Thanks!

hathor/consensus/consensus.py
msbrogli: It was not that clear to me that invalid transactions must be voided. I guess the consensus is the only part of the code that can void a transaction while the transaction can get invalid for other reasons.

I feel that this assert will fail when a transaction fails the reward lock verification after a reorg. In this case, `TransactionStorage.compute_transactions_that_became_invalid()` will set the validation to `INVALID` but it will not necessarily be voided.

@jansegre What do you think?

msbrogli: After a quick conversation, I saw that the `set_validation()` method will add a marker to the `meta.voided_by`. Thanks!

msbrogli (dismissed):

jansegre (approved):

PR #1140, opened 2024-10-01
fix: event handling of txs with custom scripts

glevco:
### Motivation

There's a bug in event handling when a tx spends another tx with a custom (or empty) script. This PR fixes this.

### Acceptance Criteria

- Add two new event simulation scenarios, `EMPTY_SCRIPT` and `CUSTOM_SCRIPT`, with their respective tests.

### Checklist


hathor/event/model/event_data.py
```diff
@@ -125,15 +125,22 @@ def from_event_arguments(cls, args: EventArguments) -> Self:
...
-        tx_json['inputs'] = [
-            dict(
-                tx_id=input_['tx_id'],
-                index=input_['index'],
-                spent_output=input_
+
+        inputs = []
+        for tx_input in tx_json['inputs']:
+            decoded = tx_input.get('decoded')
+            if decoded and decoded.get('address') is None:
```
msbrogli, line 132: Why don't you add whatever data is available in `decoded`?

msbrogli (changes requested):

hathor/event/model/event_data.py
glevco: Because we have an event schema that the data must adhere to. Here's the decoded output schema:

```python
class DecodedTxOutput(BaseModel, extra=Extra.ignore):
    type: str
    address: str
    timelock: Optional[int]
```

Those 3 properties are the ones used by our clients, and when `address` is `None`, it means this is a custom script and therefore all of those properties are also `None` (because we don't know how to parse them). Instead of setting them individually, we just set `decoded: None`.

msbrogli (approved):

PR #1144, opened 2024-10-04
feat(p2p): support IPv6

luislhl:
### Motivation

We want to support IPv6 connections between fullnodes and support exposing their API through IPv6

### Acceptance Criteria

- We should be able to listen the API, the Stratum and the P2P port in IPv6 interfaces. This will be activated by a new CLI argument deemed initially as experimental: `--x-enable-ipv6`
- We will also have a `--x-disable-ipv4`, to make nodes stop connecting to IPv4 endpoints. For now this will be used mostly for tests though, to force a fullnode to behave as IPv6-only.
- The peers should relay IPv6 endpoints only to other peers that also have the IPv6 capability. Relaying them to old fullnodes would cause errors when they try to parse IPv6.
- For simplicity, this PR only implements support for IPv6 addresses in endpoints, not domains that contain AAAA records

### TODO

- [ ] After merging, we should think about how to have permanent nodes that are dualstack and ipv6-only so we always test this is working

### Checklist

### DIFF
--- a/hathor/cli/nginx_config.py
+++ b/hathor/cli/nginx_config.py
@@ -240,11 +240,12 @@ limit_conn_zone $per_ip_key zone=per_ip__event_ws:10m;
 
     server_open = f'''
 upstream backend {{
-    server fullnode:8080;
+    server 127.0.0.1:8080;
 }}
 
 server {{
     listen 80;
+    listen [::]:80;
     server_name localhost;
 
     # Look for client IP in the X-Forwarded-For header
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -93,7 +93,9 @@ class RunNode:
                             help='Address to listen for new connections (eg: tcp:8000)')
         parser.add_argument('--bootstrap', action='append', help='Address to connect to (eg: tcp:127.0.0.1:8000')
         parser.add_argument('--status', type=int, help='Port to run status server')
+        parser.add_argument('--x-status-ipv6-interface', help='IPv6 interface to bind the status server')
         parser.add_argument('--stratum', type=int, help='Port to run stratum server')
+        parser.add_argument('--x-stratum-ipv6-interface', help='IPv6 interface to bind the stratum server')
         parser.add_argument('--data', help='Data directory')
         storage = parser.add_mutually_exclusive_group()
         storage.add_argument('--rocksdb-storage', action='store_true', help='Use RocksDB storage backend (default)')
@@ -162,6 +164,10 @@ class RunNode:
                             help='Log tx bytes for debugging')
         parser.add_argument('--disable-ws-history-streaming', action='store_true',
                             help='Disable websocket history streaming API')
+        parser.add_argument('--x-enable-ipv6', action='store_true',
+                            help='Enables listening on IPv6 interface and connecting to IPv6 peers')
+        parser.add_argument('--x-disable-ipv4', action='store_true',
+                            help='Disables connecting to IPv4 peers')
         return parser
 
     def prepare(self, *, register_resources: bool = True) -> None:
@@ -181,6 +187,7 @@ class RunNode:
             print('Maximum number of open file descriptors is too low. Minimum required is 256.')
             sys.exit(-2)
 
+        self.validate_args()
         self.check_unsafe_arguments()
         self.check_python_version()
 
@@ -202,7 +209,15 @@ class RunNode:
 
         if self._args.stratum:
             assert self.manager.stratum_factory is not None
-            self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory)
+
+            if self._args.x_enable_ipv6:
+                interface = self._args.x_stratum_ipv6_interface or '::0'
+                # Linux by default will map IPv4 to IPv6, so listening only in the IPv6 interface will be
+                # enough to handle IPv4 connections. There is a kernel parameter that controls this behavior:
+                # https://sysctl-explorer.net/net/ipv6/bindv6only/
+                self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory, interface=interface)
+            else:
+                self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory)
 
         from hathor.conf.get_settings import get_global_settings
         settings = get_global_settings()
@@ -217,7 +232,12 @@ class RunNode:
             status_server = resources_builder.build()
             if self._args.status:
                 assert status_server is not None
-                self.reactor.listenTCP(self._args.status, status_server)
+
+                if self._args.x_enable_ipv6:
+                    interface = self._args.x_status_ipv6_interface or '::0'
+                    self.reactor.listenTCP(self._args.status, status_server, interface=interface)
+                else:
+                    self.reactor.listenTCP(self._args.status, status_server)
 
         self.start_manager()
 
@@ -350,6 +370,11 @@ class RunNode:
                     self.log.warn('[USR2] Error', errmsg=str(e))
                 except SysctlRunnerException as e:
                     self.log.warn('[USR2] Error', errmsg=str(e))
+
+    def validate_args(self) -> None:
+        if self._args.x_disable_ipv4 and not self._args.x_enable_ipv6:
+            self.log.critical('You must enable IPv6 if you disable IPv4.')
+            sys.exit(-1)
 
     def check_unsafe_arguments(self) -> None:
         unsafe_args_found = []
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -100,6 +100,8 @@ class ConnectionsManager:
         ssl: bool,
         rng: Random,
         whitelist_only: bool,
+        enable_ipv6: bool,
+        disable_ipv4: bool,
     ) -> None:
         self.log = logger.new()
         self._settings = settings
@@ -189,6 +191,12 @@ class ConnectionsManager:
 
         # Parameter to explicitly enable whitelist-only mode, when False it will still check the whitelist for sync-v1
         self.whitelist_only = whitelist_only
+
+        # Parameter to enable IPv6 connections
+        self.enable_ipv6 = enable_ipv6
+
+        # Parameter to disable IPv4 connections
+        self.disable_ipv4 = disable_ipv4
 
         # Timestamp when the last discovery ran
         self._last_discovery: float = 0.
@@ -577,7 +585,11 @@ class ConnectionsManager:
     def connect_to_if_not_connected(self, peer: UnverifiedPeer | PublicPeer, now: int) -> None:
         """ Attempts to connect if it is not connected to the peer.
         """
-        if not peer.info.entrypoints:
+        if not peer.info.entrypoints or (
+            not self.enable_ipv6 and not peer.info.get_ipv4_only_entrypoints()
+        ) or (
+            self.disable_ipv4 and not peer.info.get_ipv6_only_entrypoints()
+        ):
             # It makes no sense to keep storing peers that have disconnected and have no entrypoints
             # We will never be able to connect to them anymore and they will only keep spending memory
             # and other resources when used in APIs, so we are removing them here
@@ -589,7 +601,15 @@ class ConnectionsManager:
 
         assert peer.id is not None
         if peer.info.can_retry(now):
-            addr = self.rng.choice(peer.info.entrypoints)
+            if self.enable_ipv6 and not self.disable_ipv4:
+                addr = self.rng.choice(peer.info.entrypoints)
+            elif self.enable_ipv6 and self.disable_ipv4:
+                addr = self.rng.choice(peer.info.get_ipv6_only_entrypoints())
+            elif not self.enable_ipv6 and not self.disable_ipv4:
+                addr = self.rng.choice(peer.info.get_ipv4_only_entrypoints())
+            else:
+                raise ValueError('IPv4 is disabled and IPv6 is not enabled')
+
             self.connect_to(addr.with_id(peer.id), peer)
 
     def _connect_to_callback(
@@ -634,6 +654,14 @@ class ConnectionsManager:
 
         if self.localhost_only and not entrypoint.addr.is_localhost():
             self.log.debug('skip because of simple localhost check', entrypoint=str(entrypoint))
+            return
+
+        if not self.enable_ipv6 and entrypoint.addr.is_ipv6():
+            self.log.info('skip because IPv6 is disabled', entrypoint=entrypoint)
+            return
+
+        if self.disable_ipv4 and entrypoint.addr.is_ipv4():
+            self.log.info('skip because IPv4 is disabled', entrypoint=entrypoint)
             return
 
         if use_ssl is None:
--- a/hathor/p2p/peer.py
+++ b/hathor/p2p/peer.py
@@ -114,6 +114,18 @@ class PeerInfo:
     flags: set[str] = field(default_factory=set)
     _settings: HathorSettings = field(default_factory=get_global_settings, repr=False)
 
+    def get_ipv4_only_entrypoints(self) -> list[PeerAddress]:
+        return list(filter(lambda e: not e.is_ipv6(), self.entrypoints))
+
+    def get_ipv6_only_entrypoints(self) -> list[PeerAddress]:
+        return list(filter(lambda e: e.is_ipv6(), self.entrypoints))
+
+    def ipv4_entrypoints_as_str(self) -> list[str]:
+        return list(map(str, self.get_ipv4_only_entrypoints()))
+
+    def ipv6_entrypoints_as_str(self) -> list[str]:
+        return list(map(str, self.get_ipv6_only_entrypoints()))
+
     def entrypoints_as_str(self) -> list[str]:
         """Return a list of entrypoints serialized as str"""
         return list(map(str, self.entrypoints))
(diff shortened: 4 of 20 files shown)


hathor/metrics.py
jansegre, line 1: I'm assuming this change will disappear after the last #1145 update.

hathor/p2p/peer.py
jansegre, line 1: This module has been refactored a lot and the PR was just merged. Let me know if you need help rebasing.

hathor/p2p/states/peer_id.py
```diff
@@ -68,10 +74,16 @@ def send_peer_id(self) -> None:
         """
         protocol = self.protocol
         my_peer = protocol.my_peer
+
+        if not self.should_relay_ipv6_entrypoints:
+            entrypoints_as_str = my_peer.ipv4_entrypoints_as_str()
+        else:
+            entrypoints_as_str = my_peer.entrypoints_as_str()
```
jansegre, line 81: Should we restrict `my_peer.entrypoints_as_str()` if IPv4 is disabled? It could make sense to at least issue warnings if there are IPv4 addresses in `my_peer.entrypoints`.

hathor/p2p/states/peer_id.py
```diff
@@ -89,7 +101,6 @@ async def handle_peer_id(self, payload: str) -> None:
         data = json_loads(payload)
 
         peer = Peer.create_from_json(data)
-        peer.validate()
```
jansegre, line 92: Why was this removed? Either way this will conflict with the latest refactor.

jansegre (approved):

hathor/p2p/states/peer_id.py
luislhl: Because I had added this validation to be done inside of `create_from_json`. It seems you did this too in the PR merged to master, so it still made sense to remove this while I was fixing conflicts.

hathor/metrics.py
luislhl: It did

hathor/p2p/peer.py
luislhl: I've rebased it already. Can you check?

hathor/p2p/states/peer_id.py
luislhl: I did this way for simplicity. My objective with this check was to avoid relaying IPv6 to nodes that aren't ready to parse them.

All nodes are ready to parse IPv4, so it didn't seem necessary to limit them.

However, nodes that have IPv4 disabled will not try to connect to IPv4 entrypoints they receive.

But, thinking about your suggestion, isn't it better to let the IPv4 addresses be relayed even by nodes that have it disabled, so it could pass them forward to others that might use them? In some very specific (and unlikely) cases this could be useful, like if some node needs to first connect to some other that is IPv6-only as a bootstrap.

jansegre (approved):

hathor/p2p/states/peer_id.py
jansegre: I think it makes sense to try to propagate all IPs as much as possible (and just avoid sending IPv6 addresses to peers that can't parse them).

hathor/cli/nginx_config.py
```diff
@@ -240,11 +240,12 @@ def generate_nginx_config(openapi: dict[str, Any], *, out_file: TextIO, rate_k:
 
     server_open = f'''
 upstream backend {{
-    server fullnode:8080;
+    server 127.0.0.1:8080;
 }}
 
 server {{
     listen 80;
+    listen [::]:80;
```
msbrogli, line 248: Should we add an `--enable-ipv6` here too?

hathor/cli/run_node.py
```diff
@@ -201,7 +206,11 @@ def prepare(self, *, register_resources: bool = True) -> None:
 
         if self._args.stratum:
             assert self.manager.stratum_factory is not None
-            self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory)
+
+            if self._args.x_enable_ipv6:
+                self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory, interface='::0')
+            else:
+                self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory)
```
msbrogli, line 213: Shouldn't we listen in both interfaces (ipv4 and ipv6) when ipv6 is enabled?

hathor/cli/run_node.py
```diff
@@ -201,7 +206,11 @@ def prepare(self, *, register_resources: bool = True) -> None:
 
         if self._args.stratum:
             assert self.manager.stratum_factory is not None
-            self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory)
+
+            if self._args.x_enable_ipv6:
+                self.reactor.listenTCP(self._args.stratum, self.manager.stratum_factory, interface='::0')
```
msbrogli, line 211: Why `::0`? I understand this might be default but there should be a way to choose one specific interface.

hathor/cli/run_node.py
```diff
@@ -216,7 +225,11 @@ def prepare(self, *, register_resources: bool = True) -> None:
             status_server = resources_builder.build()
             if self._args.status:
                 assert status_server is not None
-                self.reactor.listenTCP(self._args.status, status_server)
+
+                if self._args.x_enable_ipv6:
+                    self.reactor.listenTCP(self._args.status, status_server, interface='::0')
+                else:
+                    self.reactor.listenTCP(self._args.status, status_server)
```
msbrogli, line 232: Same here. Shouldn't we listen in both ipv4 and ipv6 interfaces?

hathor/metrics.py
msbrogli, line 1: These changes seem unrelated to this PR.

hathor/p2p/entrypoint.py
```diff
@@ -12,18 +12,21 @@
...
 from twisted.internet.address import IPv4Address, IPv6Address
-from twisted.internet.endpoints import TCP4ClientEndpoint
+from twisted.internet.endpoints import TCP4ClientEndpoint, TCP6ClientEndpoint
 from twisted.internet.interfaces import IStreamClientEndpoint
 from typing_extensions import Self
 
 from hathor.p2p.peer_id import PeerId
 from hathor.reactor import ReactorProtocol as Reactor
 
+IPV6_REGEX = re.compile(r'''^(([0-9a-fA-F]{1,4}:){7}([0-9a-fA-F]{1,4}|:)|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))$''')  # noqa: E501
```
msbrogli, line 28: Can you add a docstring and some examples that match and do not match this regex, please?

hathor/p2p/entrypoint.py
```diff
@@ -39,10 +42,14 @@ class Entrypoint:
     peer_id: PeerId | None = None
```
msbrogli, line 42: Should we add a `tcp6` protocol instead of trying to match a regex? Alternatively, we could use the `[host]` pattern for all ipv6 hosts.

hathor/p2p/manager.py
```diff
@@ -580,7 +588,11 @@ def _update_whitelist_cb(self, body: bytes) -> None:
     def connect_to_if_not_connected(self, peer: UnverifiedPeer | PublicPeer, now: int) -> None:
         """ Attempts to connect if it is not connected to the peer.
         """
-        if not peer.info.entrypoints:
+        if not peer.info.entrypoints or (
```
msbrogli, line 591: I didn't get it. What is this change doing?

hathor/p2p/manager.py
```diff
@@ -592,7 +604,14 @@ def connect_to_if_not_connected(self, peer: UnverifiedPeer | PublicPeer, now: in
...
         if peer.info.can_retry(now):
-            self.connect_to(self.rng.choice(peer.info.entrypoints), peer)
+            if self.enable_ipv6 and not self.disable_ipv4:
+                self.connect_to(self.rng.choice(peer.info.entrypoints), peer)
+            elif self.enable_ipv6 and self.disable_ipv4:
+                self.connect_to(self.rng.choice(peer.info.ipv6_entrypoints), peer)
+            elif not self.enable_ipv6 and not self.disable_ipv4:
+                self.connect_to(self.rng.choice(peer.info.ipv4_entrypoints), peer)
+            else:
+                raise ValueError('IPv4 is disabled and IPv6 is not enabled')
```
msbrogli, line 614: You can use `assert_never()` here.

hathor/p2p/peer.py
```diff
@@ -122,6 +122,20 @@ class PeerInfo:
     flags: set[str] = field(default_factory=set)
     _settings: HathorSettings = field(default_factory=get_global_settings, repr=False)
 
+    @property
+    def ipv4_entrypoints(self) -> list[Entrypoint]:
```
msbrogli, line 126: I'd rather rename it to `get_ipv4_only_entrypoints()`.

hathor/p2p/peer.py
```diff
@@ -122,6 +122,20 @@ class PeerInfo:
...
+        return list(filter(lambda e: not e.is_ipv6(), self.entrypoints))
+
+    @property
+    def ipv6_entrypoints(self) -> list[Entrypoint]:
```
msbrogli, line 130: I'd rather rename it to `get_ipv6_only_entrypoints()`.

hathor/p2p/peer.py
```diff
@@ -226,14 +240,19 @@ class UnverifiedPeer:
     id: PeerId
     info: PeerInfo = field(default_factory=PeerInfo)
 
-    def to_json(self) -> dict[str, Any]:
+    def to_json(self, only_ipv4_entrypoints: bool = True) -> dict[str, Any]:
```
msbrogli, line 243: Why don't we add all entrypoints regardless of what's enabled/disabled?

msbrogli (changes requested):

hathor/cli/nginx_config.py
luislhl: Actually, now that you mentioned, I'm wondering whether we actually need to add this here.

I had added it mostly to test IPv6 support in the API. But in our deploys, we won't really need it, because the API is server through a Load Balancer. The LB will still talk IPv4 with Nginx.

I think we could even remove it and it would make no difference.

However, I guess there is no harm in leaving it and always making it listen in the IPv6 interface, is there? The image would be ready for anyone that needs it eventually over IPv6, for any reason. And if it's not needed, it just won't be used.

Another thing is that we would be creating more variations of the Nginx Docker image that we generate. We already have the `latest` and `no-rate-limit-latest`.

We would probably need to generate additionally the `latest-with-ipv6` and `no-rate-limit-latest-with-ipv6`.

I would rather avoid these variations unless we have a good reason to not always listen on IPv6 in Nginx.

hathor/cli/run_node.py
luislhl: Initially I had this impression too. But it turns out Linux by default will map IPv4 to IPv6, and listening on `::0` is enough to handle IPv4 traffic as well.

I was actually getting an error when trying to listen on IPv4 simultaneously on the same port.

You can reproduce this behavior by running a simple server like this listening on `::`:

```
python3 -m http.server --bind :: 8080
```

And then connecting to it both over IPv6 or IPv4:

```
nc 127.0.0.1 8080
nc ::1 8080
```

There is a kernel paramenter to control this behavior, but the default is to translate IPv4 to IPv6 when bound to `::0`: https://sysctl-explorer.net/net/ipv6/bindv6only/

hathor/cli/run_node.py
luislhl: This is a good point. I'll add new parameters to allow changing this default.

hathor/cli/run_node.py
luislhl: Answered in the other thread.

hathor/p2p/peer.py
luislhl: This method was being used to build the JSON that is sent to other peers in `ReadyState.send_peers`.

I needed to avoid sending IPv6 endpoints to peers that can't handle it.

I could also just build the response manually there as an alternative, so I don't have to use this `.to_json` method. What do you think?

hathor/p2p/manager.py
luislhl: This was introduced in Python 3.11, right? 

We are still supporting 3.10, AFAIK, so I guess we should avoid this?

hathor/p2p/manager.py
luislhl: This code tries to remove from the list of peers the ones that disconnected from us and to which we surely won't be able to connect anymore.

Previously, to detect this, we only had to check whether they had no entrypoints.

But now, with IPv6, new cases could happen that wouldn't be caught:
- We don't have IPv6 enabled, and the peer only has IPv6 entrypoints (no IPv4)
- We have IPv4 disabled, and the peer only has IPv4 entrypoints (no IPv6)

It seemed like we would want to remove them from the list as well.

Does it make sense?

hathor/p2p/peer.py
luislhl: Will do

hathor/p2p/peer.py
luislhl: Will do

hathor/p2p/entrypoint.py
luislhl: Will do

hathor/p2p/entrypoint.py
luislhl: This is a good question, I thought about this at some point.

I chose not to rely so much in user input for this and just detect whether an IPv6 was provided.

Otherwise, someone could wrongly use `tcp` instead of `tcp6` as an entrypoint, and it would be relayed as if it was an IPv4 to others, that would error out while trying to parse it, or at least keep trying to connect to it and getting timeouts.

If we do start using `tcp6`, I think it would be better to keep the regex anyway for validation purposes to avoid the scenario above.

What do you think?

One of the advantages of using `tcp6` I thought about initially was that it could be used to also inform about DNS names holding AAAA records. For instance, `tcp6://abc.xyz` would be a way for the user to inform that this domain should be treated as IPv6. This something not supported in the current implementation with the Regex.

We would still have the validation issue, though, which is even harder in this case because it needs a DNS query.

hathor/metrics.py
luislhl: Removed

hathor/p2p/entrypoint.py
luislhl: Done

hathor/p2p/peer.py
luislhl: Done

hathor/p2p/peer.py
luislhl: Done

hathor/cli/run_node.py
luislhl: Done

hathor/p2p/manager.py
jansegre: While it was added to the built-in `typing` in Python 3.11, it is available in `typing_extensions` for Python 3.10 too (as a backport), we've been using it already.

jansegre (approved):

hathor/cli/run_node.py
msbrogli: Thanks!

hathor/metrics.py
msbrogli: Thanks!

hathor/p2p/peer.py
msbrogli: Thanks!

hathor/p2p/peer.py
msbrogli: Thanks!

hathor/cli/run_node.py
luislhl: Done

msbrogli: Thanks!

hathor/p2p/manager.py
msbrogli: Got it. I guess the following implementation could be more straightforward:

1. Create an enum of types (IPv4=1 and IPv6=2).
2. Create a list of supported addresses (IPv4, IPv6 ou IPv4 | IPv6).
3. Filter all valid addresses. If empty, skip it.

I guess it would prevent some of the most complex logic you are adding.

hathor/cli/nginx_config.py
msbrogli: I agree that there's no harm in enabling it. Let's keep it.

hathor/p2p/entrypoint.py
msbrogli: Thanks!

hathor/p2p/entrypoint.py
msbrogli: Got it. I agree with you. The current implementation seems good to me.

hathor/p2p/manager.py
msbrogli: Got it. Make sense. You can keep raising the exception.

msbrogli (approved):

hathor/p2p/manager.py
jansegre: > Got it. I guess the following implementation could be more straightforward:
> 
>     1. Create an enum of types (IPv4=1 and IPv6=2).
> 
>     2. Create a list of supported addresses (IPv4, IPv6 ou IPv4 | IPv6).
> 
>     3. Filter all valid addresses. If empty, skip it.
> 
> 
> I guess it would prevent some of the most complex logic you are adding.

This is what is done for `sync-v1` / `sync-v2`. I think it's a good strategy, but I'm OK doing the refactor in a new PR.

jansegre (approved):

hathor/p2p/manager.py
luislhl: Are you ok with doing this in another PR, @msbrogli ? 

Then I'll merge this one already

msbrogli: Ok. Let's do it separately.

PR #1145, opened 2024-10-04
refactor[websocket]: improve websocket factory start

luislhl:
### Motivation

I had a problem while trying to write a test for `run_node.py` in another PR. The reason was that the `HathorAdminWebsocketFactory` was started and never stopped.

I noticed that its stop method was never called, and its start was called inside `ResourcesBuilder.create_resources`, which was unlike any other resource we start.

I've included a test similar to the one I had problems with to make my difficulties clearer.

### Acceptance Criteria

- The `HathorAdminWebsocketFactory` should be started as a consequence of starting the Manager, and should be stopped when the Manager is stopped
- There should be no impact in the overall logic of hathor-core. Everything should work like before.

### Checklist

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -198,7 +198,6 @@ class RunNode:
 
         self.tx_storage = self.manager.tx_storage
         self.wallet = self.manager.wallet
-        self.start_manager()
 
         if self._args.stratum:
             assert self.manager.stratum_factory is not None
@@ -218,6 +217,8 @@ class RunNode:
             if self._args.status:
                 assert status_server is not None
                 self.reactor.listenTCP(self._args.status, status_server)
+
+        self.start_manager()
 
         from hathor.builder.builder import BuildArtifacts
         self.artifacts = BuildArtifacts(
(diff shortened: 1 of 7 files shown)


jansegre (dismissed):

hathor/cli/run_node.py
```diff
@@ -219,6 +218,8 @@ def prepare(self, *, register_resources: bool = True) -> None:
                 assert status_server is not None
                 self.reactor.listenTCP(self._args.status, status_server)
 
+        self.start_manager()
```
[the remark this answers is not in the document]
msbrogli: I guess it won't be a problem.

hathor/metrics.py
```diff
@@ -159,11 +159,17 @@ def start(self) -> None:
         self.is_running = True
         self._lc_collect_data.start(self.collect_data_interval, now=False)
 
+        if self.websocket_factory:
```
msbrogli, line 162: Why is it being initialized by the metrics singleton? I'd say the best place to start/stop is in the `HathorManager`.

msbrogli (changes requested):

hathor/metrics.py
luislhl: Fixed. Can you take another look?

msbrogli (approved):

jansegre (approved):

PR #1148, opened 2024-10-08
refactor(settings): change wallets and reward lock to use injected settings

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1146

### Motivation

Remove global settings from reward lock and test wallets. This will be useful for Feature Activation for Transaction tests.

### Acceptance Criteria

- Move `HathorSettings.HATHOR_TOKEN_UID` to a constant.
- Change reward lock functions to use injected settings.
- Change test wallets to use injected settings.

### Checklist

### DIFF
--- a/hathor/conf/settings.py
+++ b/hathor/conf/settings.py
@@ -30,6 +30,8 @@ DECIMAL_PLACES = 2
 GENESIS_TOKEN_UNITS = 1 * (10**9)  # 1B
 GENESIS_TOKENS = GENESIS_TOKEN_UNITS * (10**DECIMAL_PLACES)  # 100B
 
+HATHOR_TOKEN_UID = b'\x00'
+
 
 class HathorSettings(NamedTuple):
     # Version byte of the address in P2PKH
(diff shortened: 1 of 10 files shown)


hathor/conf/settings.py
```diff
@@ -30,6 +30,8 @@
 GENESIS_TOKEN_UNITS = 1 * (10**9)  # 1B
 GENESIS_TOKENS = GENESIS_TOKEN_UNITS * (10**DECIMAL_PLACES)  # 100B
 
+HATHOR_TOKEN_UID = b'\x00'
```
msbrogli, line 33: Isn't this essentially hiding a dependency? Some files are importing this constant directly, obscuring their dependency on the settings.

msbrogli (dismissed):

hathor/conf/settings.py
glevco: I had to move it out of settings so it can be used as a default value in a named tuple definition. In my opinion this shouldn't be part of settings, it should be just the constant and completely removed from settings. But this requires refactoring its multiple uses which I didn't do in this PR, but should be done in the future.

jansegre (dismissed):

jansegre (approved):

msbrogli (approved):

hathor/conf/settings.py
msbrogli: I understand but another way of handling it is by using factories that carries the default value. I mean, it can be either a constant or a settings and each one would have a different implementation.

Anyway, I'm good with moving forward with this PR.

jansegre: I think this is a "constant constant", as in it might not make sense to have the option to customize this (which is basically what the settings are). So I think it's good to move this out of the `HathorSettings`, we could leave a comment that the move is underway and prepare an issue or a PR to finish the refactor.

glevco: I created an issue for this, I can do it in this sprint: https://github.com/HathorNetwork/hathor-core/issues/1160

PR #1154, opened 2024-10-10
refactor(vertex-handler): remove p2p_manager dependency [part 1/8]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1153

### Motivation

As part of the P2P Multiprocess project, this PR refactors the `VertexHandler` so it does not depend on the `p2p_manager` anymore.

### Acceptance Criteria

- Remove the `p2p_manager` dependency from `VertexHandler`. This means it does not propagate vertices anymore, and it's the caller's responsibility to do it.
- Remove duplicate `partial_vertex_exists` methods from sync classes and move it to the storage.
- Update all calls to `VertexHandler.on_new_vertex` to propagate vertices accordingly:
  - If the call explicitly set `propagate_to_peers=False`, no change was necessary.
  - If the call didn't provide the `propagate_to_peers` argument or set it as `True`, add a follow-up call to `p2p_manager.send_tx_to_peers`.
- Move CPU profiler from `HathorManager.on_new_tx` to `VertexHandler.on_new_vertex`.
- Fix CPU profiler typing.
- No changes in behavior. 

### Checklist

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -48,7 +48,6 @@ from hathor.mining.cpu_mining_service import CpuMiningService
 from hathor.p2p.manager import ConnectionsManager
 from hathor.p2p.peer import PrivatePeer
 from hathor.p2p.peer_id import PeerId
-from hathor.profiler import get_cpu_profiler
 from hathor.pubsub import HathorEvents, PubSubManager
 from hathor.reactor import ReactorProtocol as Reactor
 from hathor.reward_lock import is_spent_reward_locked
@@ -70,7 +69,6 @@ if TYPE_CHECKING:
     from hathor.websocket.factory import HathorAdminWebsocketFactory
 
 logger = get_logger()
-cpu = get_cpu_profiler()
 
 
 class HathorManager:
@@ -171,8 +169,6 @@ class HathorManager:
         self.network = settings.NETWORK_NAME
 
         self.is_started: bool = False
-
-        self.cpu = cpu
 
         # XXX: first checkpoint must be genesis (height=0)
         self.checkpoints: list[Checkpoint] = checkpoints or []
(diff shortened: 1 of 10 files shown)


hathor/manager.py
```diff
@@ -977,14 +972,18 @@ def on_new_tx(
...
-        return self.vertex_handler.on_new_vertex(
+        success = self.vertex_handler.on_new_vertex(
             tx,
             quiet=quiet,
             fails_silently=fails_silently,
-            propagate_to_peers=propagate_to_peers,
             reject_locked_reward=reject_locked_reward,
         )
 
+        if propagate_to_peers and success:
```
msbrogli, line 982: I've seen this same block of code over and over in this PR. I'm wondering how we can remove this duplicated code.

msbrogli (approved):

jansegre (approved):

hathor/manager.py
glevco: This code is duplicate in the following places:

- 2 times in sync-v1 agent.
- 1 time in sync-v2 agent.
- 1 time in sync-v2 mempool.
- 1 time in `HathorManager.on_new_tx`.

Considering sync-v1 will be deprecated and removed soon, and `HathorManager.on_new_tx` will also be removed, we'll be left with only two occurrences of this simple `if`. I would rather keep it like this for now, even if temporarily, because my next PRs need this dependency removed — sync-v2 will no longer call `ConnectionManager.send_tx_to_peers()`, and instead will perform this though IPC.

After those refactors are complete, we may remove this code and improve it by making the `ConnectionsManager` read the PubSub message for when a new tx is accepted, and propagate it automatically to the interested peers, as suggested by you.

I created an issue so we can do that in the future: https://github.com/HathorNetwork/hathor-core/issues/1168

msbrogli: Ok. Thanks for the explanation and for creating this issue to refactor it later.

msbrogli (approved):

PR #1163, opened 2024-10-22
feat(reactor): add IReactorProcess and IReactorSocket support

glevco:
### Motivation

### Acceptance Criteria

### Checklist


hathor/simulator/heap_clock.py
```diff
@@ -94,10 +95,3 @@ class MemoryReactorHeapClock(MemoryReactor, HeapClock):
     def __init__(self):
         MemoryReactor.__init__(self)
         HeapClock.__init__(self)
-
-    def run(self):
```
msbrogli, line 98: Why did you remove this method?

msbrogli (approved):

hathor/simulator/heap_clock.py
glevco: It was not removed, it was moved to `MemoryReactor` in `hathor/reactor/memory_reactor.py` and is available in this class via inheritance. Before, we had `TestMemoryReactorClock` with a duplicate method. I did this to unify them.

msbrogli: Got it. Thanks!

jansegre (approved):

PR #1165, opened 2024-10-23
refactor(p2p): minor pre-IPC refactors

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1164

### Motivation

### Acceptance Criteria

### Checklist


hathor/conf/settings.py
```diff
@@ -458,6 +458,15 @@ def from_yaml(cls, *, filepath: str) -> 'HathorSettings':
             validators=_VALIDATORS
         )
 
+    def get_default_capabilities(self) -> list[str]:
+        """Return the default capabilities."""
+        return [
```
msbrogli, line 463: Should we make it a tuple so it's immutable?

hathor/p2p/manager.py
```diff
@@ -795,3 +790,17 @@ def reload_entrypoints_and_connections(self) -> None:
         self.log.warn('Killing all connections and resetting entrypoints...')
         self.disconnect_all_peers(force=True)
         self.my_peer.reload_entrypoints_from_source_file()
+
+    def get_peers_whitelist(self) -> list[PeerId]:
```
msbrogli, line 794: Missing docstring in all new methods in this file.

hathor/p2p/protocol.py
```diff
@@ -410,6 +411,27 @@ def disable_sync(self) -> None:
         self.log.info('disable sync')
         self.state.sync_agent.disable_sync()
 
+    def is_synced(self) -> bool:
```
msbrogli, line 414: Missing docstring in all new methods in this file.

hathor/p2p/states/ready.py
```diff
@@ -206,8 +206,7 @@ def send_ping(self) -> None:
         """
         # Add a salt number to prevent peers from faking rtt.
         self.ping_start_time = self.reactor.seconds()
-        rng = self.protocol.connections.rng
-        self.ping_salt = rng.randbytes(self.ping_salt_size).hex()
+        self.ping_salt = self.protocol.connections.get_randbytes(self.ping_salt_size).hex()
```
msbrogli, line 209: It seems better to simply pass an rng to the protocol. Isn't it?

hathor/p2p/sync_v2/blockchain_streaming_client.py
```diff
@@ -102,7 +102,6 @@ def handle_blocks(self, blk: Block) -> None:
         if self.tx_storage.partial_vertex_exists(blk.hash):
             # We reached a block we already have. Skip it.
             self._blk_repeated += 1
-            is_duplicated = True
```
msbrogli, line 105: Why did you remove this? It seems unrelated to the PR and it would change the behavior of sync-v2.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -541,12 +541,14 @@ def get_transaction(self, hash_bytes: bytes) -> BaseTransaction:
         self.post_get_validation(tx)
         return tx
 
-    def get_block_by_height(self, height: int) -> Optional[Block]:
-        """Return a block in the best blockchain from the height index. This is fast."""
+    def get_block_id_by_height(self, height: int) -> VertexId | None:
```
msbrogli, line 544: Missing docstring.

hathor/transaction/storage/transaction_storage.py
```diff
@@ -1146,6 +1148,11 @@ def partial_vertex_exists(self, vertex_id: VertexId) -> bool:
         with self.allow_partially_validated_context():
             return self.transaction_exists(vertex_id)
 
+    def get_mempool_tips(self) -> set[VertexId]:
```
msbrogli, line 1151: Missing docstring.

msbrogli (approved):

hathor/p2p/states/ready.py
glevco: The RNG should be the same in the `ConnectionsManager` and in `HathorProtocol`, right? At least for simulation purposes.

If that's the case, we can't simply pass the RNG between them because now the protocol will live in a different process. Otherwise, if we can have different RNGs, then yes, it would be simpler to just have its own RNG here.

hathor/conf/settings.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1165/commits/b9b84ea0802ea3ae98dbc324d8ab07434a2f3575

hathor/p2p/manager.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1165/commits/b9b84ea0802ea3ae98dbc324d8ab07434a2f3575

hathor/p2p/protocol.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1165/commits/b9b84ea0802ea3ae98dbc324d8ab07434a2f3575

hathor/p2p/sync_v2/blockchain_streaming_client.py
glevco: This variable is unused in this context. It's just a local variable that we set here, and the method returns just below.

In fact, in the latest commit (https://github.com/HathorNetwork/hathor-core/pull/1165/commits/b9b84ea0802ea3ae98dbc324d8ab07434a2f3575) I completely removed as it's just used to toggle a debug message that would never change.

hathor/transaction/storage/transaction_storage.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1165/commits/b9b84ea0802ea3ae98dbc324d8ab07434a2f3575

hathor/transaction/storage/transaction_storage.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1165/commits/b9b84ea0802ea3ae98dbc324d8ab07434a2f3575

hathor/p2p/protocol.py
msbrogli: Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Thanks!

hathor/transaction/storage/transaction_storage.py
msbrogli: Thanks!

hathor/conf/settings.py

hathor/p2p/states/ready.py
jansegre: We could use a "derived" rng instance, where rng in the main process just generates a seed that is used to instantiate the rng in the subprocess.

There's really only one reason for using an rng instance instead of a global rng, and that is using seeds to reproduce test cases. It's an important for reproducing rare cases that happen randomly in tests.

Considering there's parallelism, some events can happen in different order regardless of the initial seed, and this will affect how we generate random bytes regardless if we query the main thread for random bytes or generate them locally, so in that sense I think generating them locally is better because it avoids an IPC call. And if we really need to we can focus on avoiding that parallelism in tests or simulating it in a reproducible way.

PR #1174, opened 2024-11-06
refactor(p2p): implement PeerConnections [part 2/10]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1173

### Motivation

This PR is mostly a refactor for the P2P Multiprocess project, but it contains several general improvements and also a small bug fix.

Currently, the `ConnectionsManager` class keeps track of existing connections in their different states in multiple attributes with heterogeneous types. Connecting peers are identified by `IStreamClientEndpoints`, handshaking peers are identified by protocol instances themselves, ready peers are identified by `PeerIds`, and there's also a `connections` attribute that holds all of them as protocol instances too.

The main refactor of this PR is moving, renaming, and changing all those attributes. They're moved to the new `PeerConnections` class, that now holds all logic related to handling the connections as they advance in their state machine.

Now, every connection is uniquely identified by its connection address, which is present in both inbound and outbound connections (contrary to the `entrypoint`, which is removed). This is less error prone as its now impossible to connect to the same address twice (there was a small bug that briefly allowed bootstrap connections to the same address). Renames are also done to reduce confusion, specifically "ready peers" was called "connected peers" and is updated.

Here's the list of `ConnectionsManager` attributes that are moved to `PeerConnections` and updated to use `PeerAddress`:

- `connecting_peers: dict[IStreamClientEndpoint, _ConnectingPeer]` → `_connecting_outbound: set[PeerAddress]`.
- `handshaking_peers: set[HathorProtocol]` → `_handshaking: dict[PeerAddress, HathorProtocol]`.
- `connected_peers: dict[PeerId, HathorProtocol]` → `_ready: dict[PeerAddress, HathorProtocol]`.
- `connections: set[HathorProtocol]` → removed, but equivalent methods are provided as `get_connected_peers` or similar.

### Acceptance Criteria

- Introduce new `PeerConnections` class, moving some connection logic from the `ConnectionsManager` to it.
- Peer connections, both inbound and outbound, now use an `addr: PeerAddress` attribute to be uniquely identified. The `entrypoint` attribute is removed.
- Remove `use_ssl` parameter from `ConnectionsManager.connect_to` and `listen` — this was unused, the instance attribute is used instead.
- Remove `"deferred"` key from `status` API and update its address strings.
- Add more tests.

### Checklist


hathor/p2p/manager.py
```diff
@@ -581,93 +553,68 @@ def connect_to_if_not_connected(self, peer: UnverifiedPeer | PublicPeer, now: in
...
+        endpoint: PeerEndpoint,
         peer: UnverifiedPeer | PublicPeer | None = None,
-        use_ssl: bool | None = None,
-    ) -> None:
-        """ Attempt to connect to a peer, even if a connection already exists.
-        Usually you should call `connect_to_if_not_connected`.
+    ) -> Deferred[IProtocol] | None:
+        """Attempt to connect to a peer though a specific endpoint."""
+        if endpoint.peer_id is not None and peer is not None:
+            assert endpoint.peer_id == peer.id, 'the entrypoint peer_id does not match the actual peer_id'
```
glevco, line 582: This is an assertion because the only caller that passes a `peer` constructs the `endpoint` using the `peer.id`. In another PR this may be further improved so this assertion can be removed.

hathor/p2p/manager.py
```diff
@@ -442,54 +413,55 @@ def relay_peer_to_ready_connections(self, peer: PublicPeer) -> None:
...
-            if existing_protocol is None:
-                # in this case, the connection was closed before it got to READY state
-                return
-            if existing_protocol != protocol:
-                # this is the case we're closing a duplicate connection. We need to set the
-                # existing protocol object back to connected_peers, as that connection is still ongoing.
-                # A check for duplicate connections is done during PEER_ID state, but there's still a
-                # chance it can happen if both connections start at the same time and none of them has
-                # reached READY state while the other is on PEER_ID state
-                self.connected_peers[protocol.peer.id] = existing_protocol
```
glevco, line 461: This logic was simplified by calling a specific method for each state.

hathor/p2p/manager.py
```diff
@@ -721,42 +663,17 @@ def _add_hostname_entrypoint(self, hostname: str, address: IPv4Address | IPv6Add
...
-            else:
-                # this was started by peer, so drop it
-                return protocol
-        else:
-            # connection started by peer is kept
-            if not protocol.inbound:
-                return protocol
-            else:
-                return other_connection
-
```
glevco, line 748: This was moved to `PeerConnections` and simplified.

hathor/p2p/peer.py
```diff
@@ -137,31 +137,15 @@ async def validate_entrypoint(self, protocol: HathorProtocol) -> bool:
...
-                connection_host = getattr(connection_remote, 'host', None)
-                if connection_host is None:
-                    continue
-                # Connection host has only the IP
-                # So we must consider that the entrypoint could be in name format and we just validate the host
-                if connection_host == entrypoint.host:
-                    return True
-                test_mode = not_none(DifficultyAdjustmentAlgorithm.singleton).TEST_MODE
-                result = await discover_dns(entrypoint.host, test_mode)
-                if connection_host in [entrypoint.addr.host for entrypoint in result]:
```
glevco, line 164: I was able to remove this because before we only had `entrypoint` for outbound peers, but now we have `addr` for both inbound and outbound peers.

hathor/p2p/states/peer_id.py
```diff
@@ -120,9 +120,6 @@ async def handle_peer_id(self, payload: str) -> None:
             protocol.send_error_and_close_connection('Connection string is not in the entrypoints.')
             return
 
-        if protocol.entrypoint is not None and protocol.entrypoint.peer_id is not None:
-            assert protocol.entrypoint.peer_id == peer.id
-
```
glevco, line 125: The protocol does not contain an `entrypoint` anymore, so this can be removed.

tests/p2p/test_protocol.py
```diff
@@ -251,21 +293,24 @@ def test_invalid_same_peer_id2(self) -> None:
...
-        self.assertIn(True, [conn_dead.tr1.disconnecting, conn_dead.tr2.disconnecting])
-        # check connected_peers
-        connected_peers = list(self.manager1.connections.connected_peers.values())
-        self.assertEquals(1, len(connected_peers))
-        self.assertIn(connected_peers[0], [conn_alive.proto1, conn_alive.proto2])
+        self.assertTrue(tr_dead.disconnecting)
+        # check ready_peers
+        ready_peers = list(self.manager1.connections.iter_ready_connections())
+        self.assertEquals(1, len(ready_peers))
+        self.assertEquals(ready_peers[0], proto_alive)
```
glevco, line 313: Since I refactored the method that chooses which connection to drop, I updated this test to actually test the specific connection that is dropped, which is important.

Before, it would just test that _some_ connection is dropped, but not _which exact_ connection is dropped.

hathor/p2p/peer_connections.py
```diff
@@ -0,0 +1,200 @@
...
+    @staticmethod
+    def _should_drop_new_connection(new_conn: HathorProtocol) -> bool:
+        """
+        When there are connections with duplicate PeerIds, determine which one should be dropped, the old or the new.
+        Return True if we should drop the new connection, and False otherwise.
+
+        The logic to determine this is `(my_peer_id > other_peer_id) XNOR new_conn.inbound`.
+        """
+        my_peer_is_larger = bytes(new_conn.my_peer.id) > bytes(new_conn.peer.id)
+        return my_peer_is_larger == new_conn.inbound
```
glevco, line 200: I refactored this from `ConnectionsManager`, simplifying its logic to a simple boolean operation. The old docstring and comments were incorrect, because it stated a decision to drop either the connection initiated by us, or the one initiated by the other peer.

This was incorrect because it's possible that both duplicate connections are initiated by the other peer, for example. Therefore the correct description is just choosing whether to drop the old connection or the new, independent of who initiated them. The `new_conn.inbound` is used in the boolean logic but doesn't actually convey any meaning.

PR #1176, opened 2024-11-07
feat(consensus): Assert that consensus ended in a valid state after p…

msbrogli:
…rocessing a new block

### Motivation

### Acceptance Criteria

### Checklist


jansegre (approved):

hathor/consensus/block_consensus.py
```diff
@@ -47,6 +47,19 @@ def log(cls) -> Any:
...
+        self.assert_valid_consensus(block)
+
+    def assert_valid_consensus(self, block: Block) -> None:
+        """Assert that all transactions confirmed by the block are valid.
+
+        This assertion might be slow because it runs in O(n) where n is number of transactions confirmed
+        by the block."""
+        for tx in block.iter_transactions_in_this_block():
+            # Since this is the last step in the consensus, the conflict resolution has already been resolved.
+            # Hence, at this point, all transactins confirmed by a block must be non-voided.
```
glevco, line 59: Typo: `transactins -> transactions`

hathor/consensus/block_consensus.py
```diff
@@ -47,6 +47,19 @@ def log(cls) -> Any:
...
+            if tx_meta.voided_by and bool(self.context.consensus.soft_voided_tx_ids & tx_meta.voided_by):
```
glevco, line 60: What if there are other IDs in `voided_by` that are not `soft_voided`? Shouldn't it be invalid?

hathor/consensus/block_consensus.py
```diff
@@ -47,6 +47,19 @@ def log(cls) -> Any:
 
     def update_consensus(self, block: Block) -> None:
         self.update_voided_info(block)
+        self.assert_valid_consensus(block)
+
+    def assert_valid_consensus(self, block: Block) -> None:
+        """Assert that all transactions confirmed by the block are valid.
+
+        This assertion might be slow because it runs in O(n) where n is number of transactions confirmed
+        by the block."""
```
glevco, line 56: Maybe we should start running some nodes with [`-O`](https://docs.python.org/3/using/cmdline.html#cmdoption-O). It removes `assert` statements, so we could have more liberty in adding costly assertions. We should keep some nodes without `-O` so we do catch bugs when those asserts are reached, but some other nodes could run without them so they don't have a performance penalty.

*The loop, or the whole function, would have to be moved to `if __debug__`.

PR #1193, opened 2024-12-11
fix: acc weight regression

glevco:
### Motivation

The recent typing change in the `meta.accumulated_weight` introduced a regression in the `BaseTransaction.update_accumulated_weight()` method. Its implementation was not updated to consider the change. Also, the `/transaction_acc_weight` API was changed to incorrectly return work instead of weight, making it a breaking change for clients.

### Acceptance Criteria

- Fix `BaseTransaction.update_accumulated_weight()` so it correctly compares work to the provided `stop_value`.
- Fix `update_accumulated_weight()` call in the consensus which was passing the incorrect `stop_value`.
- Fix `/transaction_acc_weight` API so it returns the correct values (weight instead of work).

### Checklist

### DIFF
--- a/hathor/transaction/base_transaction.py
+++ b/hathor/transaction/base_transaction.py
@@ -22,7 +22,7 @@ import weakref
 from abc import ABC, abstractmethod
 from enum import IntEnum
 from itertools import chain
-from math import inf, isfinite, log
+from math import isfinite, log
 from struct import error as StructError, pack
 from typing import TYPE_CHECKING, Any, ClassVar, Generic, Iterator, Optional, TypeAlias, TypeVar
 
@@ -638,7 +638,12 @@ class GenericVertex(ABC, Generic[StaticMetadataT]):
 
         self.storage.save_transaction(self, only_metadata=True)
 
-    def update_accumulated_weight(self, *, stop_value: float = inf, save_file: bool = True) -> TransactionMetadata:
+    def update_accumulated_weight(
+        self,
+        *,
+        stop_value: int | None = None,
+        save_file: bool = True,
+    ) -> TransactionMetadata:
         """Calculates the tx's accumulated weight and update its metadata.
 
         It starts at the current transaction and does a BFS to the tips. In the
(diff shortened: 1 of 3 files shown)


hathor/consensus/transaction_consensus.py
```diff
@@ -296,7 +296,7 @@ def check_conflicts(self, tx: Transaction) -> None:
         for candidate in candidates:
             tx_meta = candidate.get_metadata()
             if not tx_meta.voided_by:
-                candidate.update_accumulated_weight(stop_value=meta.accumulated_weight)
+                candidate.update_accumulated_weight(stop_value=work_to_weight(meta.accumulated_weight))
```
glevco, line 299: We were passing an `int` to a method that takes `float`, why didn't mypy catch it??

jansegre: Apparently it's not an error, maybe because integers coerce to float, I tried this and it passes:

```
def foo(i: float) -> float:
    return i * 0.99

foo(1)
```

So I think we would be better off having `Weight` and `Work` new-types.

jansegre (dismissed):

hathor/consensus/transaction_consensus.py
glevco: Right, https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html

msbrogli, line 299: I think we should keep passing work (instead of weight) because it is the actual way we are calculating the accumulated weight and there's no reason to convert to weight with loss of precision.

hathor/transaction/base_transaction.py
```diff
@@ -638,7 +638,12 @@ def reset_metadata(self) -> None:
 
         self.storage.save_transaction(self, only_metadata=True)
 
-    def update_accumulated_weight(self, *, stop_value: float = inf, save_file: bool = True) -> TransactionMetadata:
+    def update_accumulated_weight(
+        self,
+        *,
+        stop_value: float | None = None,
```
msbrogli, line 644: I'd change it to work, i.e., `stop_value: int | None`.

msbrogli (approved):

hathor/consensus/transaction_consensus.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1193/commits/f42c4a7e07adfbdd7a8fce8fb73a488a2a5fa8f2

hathor/transaction/base_transaction.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1193/commits/f42c4a7e07adfbdd7a8fce8fb73a488a2a5fa8f2

jansegre (approved):

hathor/consensus/transaction_consensus.py
msbrogli: Thanks!

hathor/transaction/base_transaction.py
msbrogli: Thanks!

PR #1200, opened 2025-01-08
feat(indexes): remove sync-v1 indexes

jansegre:
### Motivation

After removing sync-v1 (#1197) and removing block ties (#673) there is no need to have the sync-v1 indexes (tx-tips, block-tips and all-tips). There should be a slight runtime benefit (no measurement has been made), but more noticeably, there will be a substantial improvement in startup time.

### Acceptance Criteria

- Remove tx_tips, block_tips and all_tips indexes
- Refactor relevant methods
- Refactor tests

### Checklist


glevco:
This PR greatly reduces memory usage, see https://github.com/HathorNetwork/on-call-incidents/issues/196.

hathor/manager.py
```diff
@@ -550,58 +550,76 @@ def _verify_checkpoints(self) -> None:
                     f'hash {block_hash.hex()} was found'
                 )
 
+    def get_new_timestamp(self) -> int:
+        """Generate a timestamp appropriate for a new transaction."""
+        timestamp_now = int(self.reactor.seconds())
+        best_block = self.tx_storage.get_best_block()
+        return max(timestamp_now, best_block.timestamp)
+
```
jansegre: I added this method to help transaction generators pick a consistent timestamp. Because tests don't generate blocks in real-time, there are many cases where the timestamp of the latest block is ahead of the reactor (because blocks need strictly increasing timestamps) and with this PR the `generate_parent_txs` does not support picking tx-parents for _before_ the best block, so many places needed to pick an appropriate timestamp using this exact logic.

Although I've only made use of `get_new_timestamp` in places that are in the error path of some test, not in every place that needs a timestamp for a new transaction (which would be the ideal).

hathor/manager.py
```diff
@@ -550,58 +550,76 @@ def _verify_checkpoints(self) -> None:
...
-
+                if unconfirmed_extras:
+                    self.log.debug('generate_parent_txs: one tx tip and at least one other mempool tx')
+                    return ParentTxs.from_txs(can_include=unconfirmed_extras[-1:], must_include=(tip_tx,))
+                else:
+                    self.log.debug('generate_parent_txs: one tx in mempool, fill with one repeated parent')
+                    return ParentTxs.from_txs(can_include=confirmed_tips[-1:], must_include=(tip_tx,))
             case _:
+                self.log.debug('generate_parent_txs: multiple unconfirmed mempool tips')
                 return ParentTxs.from_txs(can_include=unconfirmed_tips, must_include=())
```
jansegre: I've kept the previous refactored structure as much as sensible, the main difference is that now `ParentTxs` will only have more than 2 transactions in the last case, when there are more than 2 unconfirmed tips, every other case should lead to `len(can_include) + len(must_include) == 2`, highest timestamp are prioritized when possible (sorting by timestamp and using `[-1:]` and `[-2:]`).

pyproject.toml
```diff
@@ -70,7 +70,7 @@ pycoin = "~0.92.20230326"
 requests = "=2.32.3"
 service_identity = "~21.1.0"
 pexpect = "~4.8.0"
-intervaltree = "~3.1.0"
+sortedcontainers = "~2.4.0"
```
jansegre: We don't need `intervaltree` anymore, but we do still use `sortedcontainers` which is an indirect dependency that should be made explicit regardless.

hathor/indexes/mempool_tips_index.py
```diff
@@ -98,6 +96,9 @@ def init_loop_step(self, tx: BaseTransaction) -> None:
         assert tx.hash is not None
         assert tx.storage is not None
         tx_meta = tx.get_metadata()
+        # do not include voided transactions
+        if tx_meta.voided_by:
```
msbrogli: Why this change in this PR?

hathor/p2p/resources/status.py
```diff
@@ -90,14 +90,10 @@ def render_GET(self, request):
...
 
-        best_block_tips = []
-        for tip in self.manager.tx_storage.get_best_block_tips():
-            block = self.manager.tx_storage.get_block(tip)
-            best_block_tips.append({'hash': block.hash_hex, 'height': block.static_metadata.height})
-
         best_block = self.manager.tx_storage.get_best_block()
         raw_best_blockchain = self.manager.tx_storage.get_n_height_tips(self._settings.DEFAULT_BEST_BLOCKCHAIN_BLOCKS)
         best_blockchain = to_serializable_best_blockchain(raw_best_blockchain)
+        best_block_tips = [{'hash': best_block.hash_hex, 'height': best_block.static_metadata.height}]
```
msbrogli: We still have candidate chains. Should we add them too?

hathor/reward_lock/reward_lock.py
```diff
@@ -58,17 +58,9 @@ def get_spent_reward_locked_info(
...
-    # omitting timestamp to get the current best block, this will usually hit the cache instead of being slow
-    tips = storage.get_best_block_tips()
-    assert len(tips) > 0
-    best_height = math.inf
-    for tip in tips:
-        blk = storage.get_block(tip)
-        best_height = min(best_height, blk.get_height())
-    assert isinstance(best_height, int)
-    return best_height
+    # XXX: only use methods available in VertexStorageProtocol, otherwise TransactionStorage.get_height_best_block
```
msbrogli: I couldn't understand this comment. Can you help me, please?

hathor/manager.py
```diff
@@ -550,58 +550,76 @@ def _verify_checkpoints(self) -> None:
                     f'hash {block_hash.hex()} was found'
                 )
 
+    def get_new_timestamp(self) -> int:
```
msbrogli: Maybe rename it to `get_timestamp_for_new_vertex()`?

hathor_tests/poa/test_poa_simulation.py
```diff
@@ -158,7 +158,7 @@ def test_two_producers(self) -> None:
         assert self.simulator.run(200, trigger=trigger)
         assert manager1.tx_storage.get_block_count() == 12
         assert manager2.tx_storage.get_block_count() == 12
-        assert manager1.tx_storage.get_best_block_tips() == manager2.tx_storage.get_best_block_tips()
+        assert manager1.tx_storage.get_best_block_hash() == manager2.tx_storage.get_best_block_hash()
```
msbrogli: Should we keep the `get_best_block_tips()` that should return the best chain together with all candidate chains?

hathor/transaction/storage/transaction_storage.py
```diff
@@ -635,108 +606,24 @@ def latest_timestamp(self) -> int:
...
-                best_tip_blocks = [block_hash]
-
-        # XXX: if there's more than one we filter it so it's the smallest hash
-        if len(best_tip_blocks) > 1:
-            best_tip_blocks = [min(best_tip_blocks)]
-
-        if timestamp is None:
-            self._best_block_tips_cache = best_tip_blocks[:]
-        return best_tip_blocks
+    def get_best_block_hash(self, timestamp: Optional[float] = None, *, skip_cache: bool = False) -> VertexId:
```
msbrogli: I guess we should remove the `timestamp` and the `skip_cache` arguments.

msbrogli (changes requested):

hathor/p2p/resources/status.py
jansegre: How do we find the candidate chains without a costly search?

hathor/indexes/memory_timestamp_index.py
```diff
@@ -89,3 +91,6 @@ def get_hashes_and_next_idx(self, from_idx: RangeIdx, count: int) -> tuple[list[
     def iter(self) -> Iterator[bytes]:
         for element in self._index:
             yield element.hash
+
+    def __contains__(self, elem: tuple[int, bytes]) -> bool:
+        return TransactionIndexElement(*elem) in self._index
```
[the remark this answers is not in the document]
jansegre: This code has been removed.

hathor/p2p/resources/status.py
msbrogli: We need to keep track of them, only the tips. This should be in a separate list in the storage.

hathor/manager.py

hathor/transaction/storage/transaction_storage.py

msbrogli (dismissed):

hathor/transaction/storage/transaction_storage.py
```diff
@@ -635,108 +606,24 @@ def latest_timestamp(self) -> int:
     def first_timestamp(self) -> int:
         raise NotImplementedError
 
-    @abstractmethod
-    def get_best_block_tips(self, timestamp: Optional[float] = None, *, skip_cache: bool = False) -> list[bytes]:
```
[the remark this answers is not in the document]
jansegre: I see. Maybe it makes sense to have an orphan-tips index, not just for blocks, but for transactions too. I can add a task to implement that, which would be how the starting point for the clean up.

hathor/p2p/resources/status.py
jansegre: Same applies from https://github.com/HathorNetwork/hathor-core/pull/1200#discussion_r2610547559

msbrogli (dismissed):

hathor/indexes/manager.py
```diff
@@ -390,10 +375,10 @@ def add_tx(self, tx: BaseTransaction) -> bool:
 
         # We need to check r1 as well to make sure we don't count twice the transactions/blocks that are
         # just changing from voided to executed or vice-versa
-        if r1 and r3:
+        if r1:
```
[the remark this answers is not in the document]
jansegre: Fixed.

glevco (approved):

msbrogli (approved):

PR #1218, opened 2025-02-04
refactor: cleanup full verification

glevco:
### Motivation

Recent PR https://github.com/HathorNetwork/hathor-core/pull/1206 deprecated `--x-full-verification`, but its code was not cleaned up. This PR does this.

### Acceptance Criteria

- Remove all mentions to full verification.
- Rename `HathorManager._initialize_components_new()` to `_initialize_components()`.

### Checklist

### DIFF
--- a/tests/others/test_init_manager.py
+++ b/tests/others/test_init_manager.py
@@ -1,18 +1,26 @@
+import tempfile
 from typing import Iterator
 
+from hathor.conf.settings import HathorSettings
 from hathor.pubsub import PubSubManager
 from hathor.simulator.utils import add_new_block, add_new_blocks
+from hathor.storage import RocksDBStorage
 from hathor.transaction import BaseTransaction
-from hathor.transaction.storage import TransactionMemoryStorage
+from hathor.transaction.storage import TransactionRocksDBStorage
+from hathor.transaction.vertex_parser import VertexParser
 from tests import unittest
 from tests.unittest import TestBuilder
 from tests.utils import add_blocks_unlock_reward, add_new_double_spending, add_new_transactions
 
 
-class ModifiedTransactionMemoryStorage(TransactionMemoryStorage):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._first_tx = None
+class ModifiedTransactionRocksDBStorage(TransactionRocksDBStorage):
+    def __init__(self, path: str, settings: HathorSettings):
+        super().__init__(
+            rocksdb_storage=RocksDBStorage(path=path),
+            settings=settings,
+            vertex_parser=VertexParser(settings=settings),
+        )
+        self._first_tx: BaseTransaction | None = None
 
     def set_first_tx(self, tx: BaseTransaction) -> None:
         self._first_tx = tx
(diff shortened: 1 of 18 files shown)


jansegre (dismissed):

tests/others/test_init_manager.py
```diff
@@ -30,7 +38,8 @@ def _get_all_transactions(self) -> Iterator[BaseTransaction]:
 class SimpleManagerInitializationTestCase(unittest.TestCase):
     def setUp(self):
         super().setUp()
-        self.tx_storage = ModifiedTransactionMemoryStorage(settings=self._settings)
+        self.temp_dir = tempfile.TemporaryDirectory()
```
msbrogli, line 41: Maybe use `self.mkdtemp()` so it is automatically clean up?

tests/others/test_init_manager.py
```diff
@@ -89,7 +98,8 @@ def test_wrong_stop(self):
 class ManagerInitializationTestCase(unittest.TestCase):
     def setUp(self):
         super().setUp()
-        self.tx_storage = ModifiedTransactionMemoryStorage(settings=self._settings)
+        self.temp_dir = tempfile.TemporaryDirectory()
```
msbrogli, line 101: Maybe use `self.mkdtemp()` so it is automatically clean up?

msbrogli (dismissed):

tests/others/test_init_manager.py
glevco: `tempfile.TemporaryDirectory()` is automatically cleaned up:

> This class securely creates a temporary directory using the same rules as [mkdtemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp). The resulting object can be used as a [context manager](https://docs.python.org/3/glossary.html#term-context-manager) (see [Examples](https://docs.python.org/3/library/tempfile.html#tempfile-examples)). On completion of the context or destruction of the temporary directory object, the newly created temporary directory and all its contents are removed from the filesystem.

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

I actually want to remove `self.mkdtemp()` in a future PR to simplify and unify.

tests/others/test_init_manager.py
glevco: Please see https://github.com/HathorNetwork/hathor-core/pull/1218#discussion_r1951592626

msbrogli: Ok. Thanks!

tests/others/test_init_manager.py
msbrogli: I'm afraid we will lose control of the moment the directory will be removed because it might depend on the garbage collector. Is that correct? So we need to keep a reference to the returned object until we are done using the file system. Otherwise, it might clean up the directory during the execution and cause random errors.

glevco: > I'm afraid we will lose control of the moment the directory will be removed because it might depend on the garbage collector. Is that correct?

Correct, do you think that could be a problem? When tests finish running and python exits, everything will be automatically cleaned up.

> So we need to keep a reference to the returned object until we are done using the file system. Otherwise, it might clean up the directory during the execution and cause random errors.

Yes, that's why I assign it to `self.temp_dir`.

msbrogli: Since we already have a structure that does exactly this in a very explicit way, I don't see any benefit from using this class. I usually prefer explicit things than implicit ones.

glevco: I believe using `tearDown` is also pretty implicit, and we're reimplementing something already provided by the stdlib. But I don't think this is too important, changed in https://github.com/HathorNetwork/hathor-core/pull/1218/commits/42136600178ded7cf70d3b796b85a45668cf9520

jansegre (approved):

tests/others/test_init_manager.py
msbrogli: Thanks!

msbrogli (approved):

PR #1219, opened 2025-02-05
refactor: remove memory storage and indexes

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1218

### Motivation

The exclusively in-memory storage and indexes haven't been very useful for a while, and generate maintenance overhead. Its implementation has to be fully compatible with its RocksDB counterpart, which is not true for some cases. For this and for simplicity, this PR completely removes support for in-memory storage/indexes, and updates all tests to run using RocksDB instead. This also allows us to have better guarantees that we're testing the same behavior that we observe in production.

A new CLI option `--temp-data` is provided to be used instead of `--memory-storage`. It requires no arguments and creates a temporary directory that is automatically cleaned up when the full node exits.

### Acceptance Criteria

- Changes in CLI args:
  - Deprecate `--memory-storage` and `--memory-indexes`.
  - Remove `--rocksdb-storage` and `--x-rocksdb-indexes` (because they're now the only possibility).
- Remove all classes that were exclusively being used by memory storage or indexes.
- Update `Builder` and `CLIBuilder` accordingly.
- Refactor settings management on indexes so they all use injected settings.
- Remove `TimestampIndex.get_hashes_and_next_idx()` as it's unused (this was a Sync-v1 only method).
- All tests are changed to use RocksDB instead of memory.
  - This surfaced a few bugs/incompatible implementations that were also fixed in this PR. See the comments in the respective test files.

### Checklist


tests/resources/transaction/test_tx.py
```diff
@@ -290,34 +278,29 @@ def test_get_many(self):
...
             self.assertEqual(expected.hash.hex(), result['tx_id'])
 
         self.assertTrue(data2['has_more'])
 
         # Get older blocks with hash reference
-        expected3 = blocks[:2]
-        expected3.reverse()
-
+        genesis_block = self.manager.tx_storage.get_genesis(self._settings.GENESIS_BLOCK_HASH)
+        expected3 = [genesis_block, *blocks[:2]]
```
glevco, line 303: This test was actually incorrect. Since it was not using the `strict` option of the `zip()` function, the iteration was shorter than it should be and it didn't assert everything that should be asserted. I added the `strict=True` option to all `zip()` calls and fixed the expected values accordingly.

tests/resources/transaction/test_tx.py
```diff
@@ -345,36 +328,32 @@ def test_get_many(self):
             })
         data4 = response4.json_value()
 
-        for expected, result in zip(expected2, data4['transactions']):
+        for expected, result in zip(expected2, data4['transactions'], strict=True):
```
glevco, line 331: Here I'm using the corrected `expected2` instead of `reversed(expected2)`, which would have been equivalent to the previous code. This is because the memory storage's implementation was incompatible with the RocksDB implementation.

tests/resources/transaction/test_tx.py
```diff
@@ -345,36 +328,32 @@ def test_get_many(self):
...
-        expected5.reverse()
-
         response5 = yield self.web.get(
             "transaction", {
                 b'count': b'3',
                 b'type': b'block',
-                b'timestamp': bytes(str(expected1[-1].timestamp), 'utf-8'),
-                b'hash': bytes(expected1[-1].hash.hex(), 'utf-8'),
+                b'timestamp': bytes(str(blocks[-3].timestamp), 'utf-8'),
+                b'hash': bytes(blocks[-3].hash.hex(), 'utf-8'),
```
glevco, line 344: `expected1[-1]` from the previous code was actually equivalent to `blocks[-2]`, but that meant only a single block was returned by the API call, which is not useful to test ordering. Therefore, I purposefully increased the window here so at least 2 blocks are returned.

tests/resources/transaction/test_tx.py
```diff
@@ -345,36 +328,32 @@ def test_get_many(self):
...
                 b'page': b'previous'
             })
         data5 = response5.json_value()
 
-        for expected, result in zip(expected5, data5['transactions']):
+        for expected, result in zip(expected5, data5['transactions'], strict=True):
```
glevco, line 349: Same here, about not using `reversed(expected5)`.

tests/resources/transaction/test_tx.py
```diff
@@ -462,12 +441,12 @@ def test_zero_count(self):
...
         data = response.json_value()
         self.assertEqual(0, len(data['transactions']))
-        self.assertFalse(data['has_more'])
+        self.assertTrue(data['has_more'])
 
         response = yield self.web.get("transaction", {b'count': b'0', b'type': b'tx'})
         data = response.json_value()
         self.assertEqual(0, len(data['transactions']))
-        self.assertFalse(data['has_more'])
+        self.assertTrue(data['has_more'])
```
glevco, line 449: The memory storage's implementation incorrectly always returned `has_more=False` when `count=0`. In reality, the genesis vertices cause `has_more=True` which is correctly implemented in RocksDB.

tests/tx/test_indexes2.py
glevco, line 1: From my understanding this test was only checking whether `MemoryTimestampIndex` and `RocksDBTimestampIndex` had compatible implementations, nothing more, so I removed it.

hathor/cli/run_node_args.py
```diff
@@ -40,9 +40,9 @@ class RunNodeArgs(BaseModel, extra=Extra.allow):
     stratum: Optional[int]
     x_stratum_ipv6_interface: Optional[str]
     data: Optional[str]
-    rocksdb_storage: bool
     memory_storage: bool
     memory_indexes: bool
+    temp_data: bool
```
msbrogli, line 45: Why are we adding this choice? What happens when we use it? What happens after the full node is stopped?

glevco: As stated in the PR description:

> A new CLI option `--temp-data` is provided to be used instead of `--memory-storage`. It requires no arguments and creates a temporary directory that is automatically cleaned up when the full node exits.

hathor/indexes/manager.py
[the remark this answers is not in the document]
glevco: Yes. We have to check with them about using the wallet streaming API and verify it works before merging this PR.

hathor/cli/run_node_args.py
glevco: While I obviously agree it shouldn't be used in production, the worse that could happen if it was used is that a resync/load from snapshot would be necessary. I don't think the feature is _unsafe_ per se, so I wouldn't mind keeping it like it is. I like the convenience of simply running `run_node --temp-data` during development, just like we did with `--memory-storage` which was also unsuited for production, but was not unsafe.

jansegre (approved):
The proposal looks good to me. I'm only a bit worried about deploys that still use `--memory-indexes`, but I guess they can wait to update when it works well with rocksdb indexes, which we might have to prioritize for the next releases.

hathor/indexes/manager.py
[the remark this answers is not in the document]
glevco: We have checked and verified that we don't have any instances using this anymore.

hathor/cli/run_node_args.py
glevco: This has been already merged in another PR.

hathor/indexes/rocksdb_timestamp_index.py
```diff
@@ -135,30 +136,6 @@ def get_newer(self, timestamp: int, hash_bytes: bytes, count: int) -> tuple[list
         it = (x for _, x in self._iter(timestamp, hash_bytes))
         return collect_n(it, count)
 
-    def get_hashes_and_next_idx(self, from_idx: RangeIdx, count: int) -> tuple[list[bytes], Optional[RangeIdx]]:
```
msbrogli, line 138: Why are you removing this?

msbrogli (approved):

hathor/indexes/rocksdb_timestamp_index.py
msbrogli: Just saw the explanation in the acceptance criteria. My bad!

hathor/cli/run_node_args.py
msbrogli: Ok!

hathor/indexes/manager.py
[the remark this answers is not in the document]
msbrogli: Ok! Thanks!

PR #1220, opened 2025-02-05
fix(cli): regression on `run_node --help`

jansegre:
### Motivation

Doing `python -m hathor run_node --help` broke in some cases due to a bug in some versions of argparse.

### Acceptance Criteria

- Remove all mutually exclusive groups where all arguments have `help=SUPRESS`, but keep the arguments in the base parser
- Add a test that runs the CLI's help and have a basic sanity check (exit code 0 and at least 80 lines)

### Checklist

### DIFF
--- a/hathor/cli/run_node.py
+++ b/hathor/cli/run_node.py
@@ -124,8 +124,7 @@ class RunNode:
         parser.add_argument('--cache-interval', type=int, help='Cache flush interval')
         parser.add_argument('--recursion-limit', type=int, help='Set python recursion limit')
         parser.add_argument('--allow-mining-without-peers', action='store_true', help='Allow mining without peers')
-        fvargs = parser.add_mutually_exclusive_group()
-        fvargs.add_argument('--x-full-verification', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--x-full-verification', action='store_true', help=SUPPRESS)  # deprecated
         parser.add_argument('--procname-prefix', help='Add a prefix to the process name', default='')
         parser.add_argument('--allow-non-standard-script', action='store_true', help='Accept non-standard scripts on '
                             '/push-tx API')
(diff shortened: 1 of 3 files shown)


msbrogli (dismissed):

hathor/cli/run_node.py
```diff
@@ -134,14 +133,13 @@ def create_parser(cls) -> ArgumentParser:
...
-        sync_args.add_argument('--x-sync-v1-only', action='store_true', help=SUPPRESS)  # deprecated
-        sync_args.add_argument('--x-sync-v2-only', action='store_true', help=SUPPRESS)  # deprecated
-        sync_args.add_argument('--x-sync-bridge', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--sync-bridge', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--sync-v1-only', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--sync-v2-only', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--x-remove-sync-v1', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--x-sync-v1-only', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--x-sync-v2-only', action='store_true', help=SUPPRESS)  # deprecated
+        parser.add_argument('--x-sync-bridge', action='store_true', help=SUPPRESS)  # deprecated
```
glevco, line 142: There are still some wrong combinations that are allowed by removing this, such as `python -m hathor run_node --unsafe testnet-golf --testnet --x-sync-v1-only --sync-v2-only --memory-storage`. We must add exceptions for all deprecated options

jansegre: Fixed, the previous checks in `hathor/builder/cli_builder.py` were relying on the fact that the arguments were mutually exclusive.

Now all the checks will happen independently and the only practical difference will be that if two sync arguments are used the error will be that the argument was removed, not that the arguments are mutually exclusive. Which is totally fine.

glevco (approved):

msbrogli (approved):

PR #1228, opened 2025-02-25
feat: add --temp-data CLI option

glevco:
### Motivation

In preparation for removing memory storage and indexes, this PR adds the `--temp-data` CLI option to substitute `--memory-storage`

### Acceptance Criteria

- Implement `--temp-data`.
- Deprecate `--memory-storage` and `--memory-indexes`.

### Checklist

### DIFF
--- a/hathor/storage/rocksdb_storage.py
+++ b/hathor/storage/rocksdb_storage.py
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+
 import os
-from typing import TYPE_CHECKING, Optional
+import tempfile
 
-if TYPE_CHECKING:
-    import rocksdb
-
+import rocksdb
 from structlog import get_logger
+from typing_extensions import assert_never
 
 logger = get_logger()
 _DB_NAME = 'data_v2.db'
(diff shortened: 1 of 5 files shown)


jansegre (dismissed):

jansegre (approved):

hathor/storage/rocksdb_storage.py
```diff
@@ -28,12 +29,16 @@ class RocksDBStorage:
     """ Creates a RocksDB database
         Give clients the option to create column families
     """
-    def __init__(self, path: str = './', cache_capacity: Optional[int] = None):
-        import rocksdb
+    def __init__(
+        self,
+        path: str | tempfile.TemporaryDirectory | None = None,
```
msbrogli, line 34: I don't like this model where a temporary directory will be created if we simply instantiate a new RocksDBStorage. I'd rather explicitly receive the path as before. I'm ok to receive both an `str` or `tempfile.TemporaryDirectory`.

msbrogli (changes requested):

hathor/storage/rocksdb_storage.py
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/1228/commits/6f5c3ad6229a1fd3dcb74b0dacf7a276036f9576

msbrogli: Ok. Thanks!

msbrogli (approved):

PR #1242, opened 2025-03-19
Feat: Pool Slots, Connection Queues and States

LFRezende:
### Motivation

This PR implements the connection pool slotting as required, separating the connection pool into four different slots: outgoing, incoming, discovered, and check_entrypoint connections. 

Each slot will have a queue to store additional connections (size 10) that exceed the total amount allowed at a given instant. 

Each connection will also have five different states: Connecting, Ready, Queue_Connecting, Queue_Ready, and Out_queued.

Connections accepted into the slot will be either in the handshaking process or ready. Connections in Queue might be in the same condition, however since they are in the Queue, they are not accepted into the pool set (self.connections), nor in any of the slots. Only when one of the active peers disconnect will then the queue pop one of its connections in standby, and then be appended to the connection pool.  Connections tagged as "Out_queued" are the ones that come from an attempt to connect while both slot and queue, for that particular type of connection (outgoing, incoming, etc.), are full, so the action is to disconnect. 

### Acceptance Criteria

- Limits the accepted outgoing connections to M connections (30% max_connections, arbitrary).
- Limits the accepted incoming connections to N connecitons (30% max_connections, arbitrary).
- Limits the accepted discovered connections in bootstrap to Q connections (30% max connections, arbitrary).
- Leaves some space to connections to check for other peers entrypoints. (10% max connections)
- Associates to each connection a "connectionState" enum class (CONNECTING, READY, Q_C, Q_R and OUT_QUEUED).
### Checklist

- This is the first complete draft of the PR, see below.
-  **Important Note:** The whole structure coded for slotting seems to be working, HOWEVER, keep in mind the DISCOVERED and CHECK_ENTRYPOINT slots, while functional, still need to be triggered. This will be left for a future/soon-to-be-released PR, but it is said to be a "complete draft" as the implementation, once realized how to be done, is fairly straight forward and should not amount to more than 5 or 6 lines of code.


hathor/conf/settings.py
```diff
@@ -285,11 +286,41 @@ def GENESIS_TX2_TIMESTAMP(self) -> int:
     # Number max of connections in the p2p network
     PEER_MAX_CONNECTIONS: int = 125
 
+    # Percentages to each connection slot (int):
```
msbrogli, line 289: I guess we don't need the percentages; we can simply have the constants below.

hathor/conf/settings.py
```diff
@@ -285,11 +286,41 @@ def GENESIS_TX2_TIMESTAMP(self) -> int:
...
+    # The discovered peers slot is left bound by PEER_MAX_CONNECTIONS.
+    PEER_MAX_CHECK_PEER_CONNECTIONS: int = PERCENTAGE_MAX_CHECK_ENTRYPOINTS_CONNECTIONS*PEER_MAX_CONNECTIONS/100
+
+    # Maximum number of connections for discovered peers after bootstrap.
+    PEER_MAX_DISCOVERED_PEERS_CONNECTIONS: int = PERCENTAGE_MAX_DISCOVERED_CONNECTIONS*PEER_MAX_CONNECTIONS/100
+    # Safeguard check
+    if PEER_MAX_DISCOVERED_PEERS_CONNECTIONS <= 0:
+        raise Exception("PEER_MAX_DISCOVERED_PEERS_CONNECTIONS must be bigger than zero.")
+
+    class ConnectionType(Enum):
```
msbrogli, line 318: Why is it in the settings? We try to keep the settings very simple since we also have yaml format. Maybe move it somewhere in the `hathor.p2p` module?

hathor/p2p/factory.py
```diff
@@ -24,7 +24,7 @@
 
 
 class _HathorLineReceiverFactory(ABC, protocol.Factory):
-    inbound: bool
+    inbound: HathorSettings.ConnectionType
```
msbrogli, line 27: Rename to `connection_type`?

hathor/p2p/factory.py
```diff
@@ -55,10 +55,10 @@ def buildProtocol(self, addr: IAddress) -> HathorLineReceiver:
 class HathorServerFactory(_HathorLineReceiverFactory, protocol.ServerFactory):
     """ HathorServerFactory is used to generate HathorProtocol objects when a new connection arrives.
     """
-    inbound = True
+    inbound = HathorSettings.ConnectionType.INCOMING
```
msbrogli, line 58: Rename to `connection_type`?

hathor/p2p/factory.py
```diff
@@ -55,10 +55,10 @@ def buildProtocol(self, addr: IAddress) -> HathorLineReceiver:
...
 
 
 class HathorClientFactory(_HathorLineReceiverFactory, protocol.ClientFactory):
     """ HathorClientFactory is used to generate HathorProtocol objects when we connected to another peer.
     """
-    inbound = False
+    inbound = HathorSettings.ConnectionType.OUTGOING
```
msbrogli, line 64: Rename to `connection_type`?

hathor/p2p/manager.py
```diff
@@ -70,7 +70,11 @@ class PeerConnectionsMetrics(NamedTuple):
     handshaking_peers_count: int
     connected_peers_count: int
     known_peers_count: int
-
+    # Counters for the connection pool slots
+    outgoing_connections_count: int
+    incoming_connections_count: int
+    discovered_connections_count: int
+    check_entrypoint_connections_count: int
```
msbrogli, line 77: I guess you haven't really used these counters, so you can delete them.

hathor/p2p/manager.py
```diff
@@ -157,6 +176,30 @@ def __init__(
         # List of peers connected and ready to communicate.
         self.connected_peers = {}
 
+        # List of connected_peers by outgoing connections.
+        self.outgoing_connections = set()
+
+        # Queue for the outgoing connections 
+        self.q_outgoing_connections = deque()
```
msbrogli, line 183: It seems this structure is repeating itself. Maybe create a separate class and use it here multiple times?

hathor/p2p/manager.py
```diff
@@ -417,27 +464,116 @@ def on_connection_failure(self, failure: Failure, peer: Optional[UnverifiedPeer
...
+                    return
+                
+                # If not full, add to the queue and return, waiting for another peer to disconnect.
+                # Note: We return here since the connection is in queue, hence it should not be added
+                # to the connection pool nor broadcasted. It will only be added when other connections are lost.
+                self.q_outgoing_connections.appendleft(protocol)
+                protocol.connection_state = HathorProtocol.ConnectionState.QUEUED_CONNECTING
+                return 
+            # If not full, add to slot
+            self.outgoing_connections.add(protocol)
```
msbrogli, line 496: This same block of code repeats itself for each connection type.

hathor/p2p/manager.py
```diff
@@ -485,9 +621,89 @@ def relay_peer_to_ready_connections(self, peer: PublicPeer) -> None:
...
+                # The protocol to be disconnected may be from the queue.
+                if protocol in self.q_outgoing_connections:
+                    self.q_outgoing_connections.remove(protocol)
+                    return 
+
+                # If protocol in slot but queue not empty, pop from queue and add to slot.
+                dequeued_connection = self.q_outgoing_connections.pop()
+
+                # If the set just discarded a connection, it is guaranteed not to be at full capacity.
+                self.outgoing_connections.add(dequeued_connection)
```
msbrogli, line 651: This same block of code repeats itself for each connection type.

hathor/p2p/manager.py
```diff
@@ -504,17 +720,49 @@ def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
...
+        
     def iter_all_connections(self) -> Iterable[HathorProtocol]:
         """Iterate over all connections."""
         for conn in self.connections:
             yield conn
 
+    def iter_incoming_connections(self) -> Iterable[HathorProtocol]:
+        """Iterate over all of the incoming connections."""
+        for conn in self.incoming_connections:
+            yield conn
```
msbrogli, line 749:
```suggestion
        yield from self.incoming_connections:
```

hathor/p2p/protocol.py
```diff
@@ -104,7 +113,7 @@ def __init__(
         *,
         settings: HathorSettings,
         use_ssl: bool,
-        inbound: bool,
+        inbound: HathorSettings.ConnectionType,
```
msbrogli, line 116: Rename to `connection_type`.

hathor/p2p/protocol.py
```diff
@@ -117,7 +126,15 @@ def __init__(
...
-        self.inbound = inbound
+        # Update: Now, inbound is an enum (0, 1, 2, or 3). 
+        # 0 == Outgoing Connection, 1 == Incoming, 2 == Discovered, 3 == For Checking Entrypoints.
+        self.connection_type = inbound
+
+        # Connection to check a specific entrypoint to be trustworthy or not.
+        #self.check_entrypoint = check_entrypoint
+
+        # Flag of this connection belonging to a discovered peer.
+        # self.discovered = discovered
```
msbrogli, line 137: Commented code. Can we remove them?

msbrogli (changes requested):

hathor/p2p/protocol.py
LFRezende: Removed. No point in leaving this here.

hathor/p2p/manager.py
LFRezende: Cool one-liner, didn't know Python could do that.
Thanks, I did that for every "iteration" function, aside from "iter_not_ready_entrypoints", as it has some extra complexity.

hathor/p2p/manager.py
LFRezende: You're right, too much bloat repetition.

I created now a class "Slot", each with a connections list for its type (outgoing, incoming, ...) and a queue, and the process of adding or removing a connection is now a method of the class.

It made the code far cleaner, thanks for the nudge.

hathor/p2p/manager.py
LFRezende: Create the Slot class, which is instantiated for every type of slot. The slot class itself now handles lists and queues.

hathor/p2p/manager.py
LFRezende: I did code it for "possible future later use", but yes I'm not using it right now - no point in leaving it there. Deleted from the get_peers_count function.

PR #1247, opened 2025-03-19
feat: add tracked object counts metric

glevco:
### Motivation

The Prometheus garbage collector `objects` metric does not currently reflect the tracked objects count, but a different count according to the [docs](https://docs.python.org/3/library/gc.html#gc.get_count):

> Return the current collection counts as a tuple of `(count0, count1, count2)`.

This PR changes it to the tracked objects count. Also, for [reference](https://stackoverflow.com/questions/57462037/what-are-count0-count1-and-count2-values-returned-by-the-python-gc-get-count).

### Acceptance Criteria

- Fix Prometheus garbage collector `objects` metric to get the actual tracked objects count.
- Introduce new Prometheus garbage collector `counts` metric from the previously used `gc.get_counts`.

### Checklist


hathor/prometheus.py
```diff
@@ -212,6 +213,8 @@ def _set_garbage_collection_metrics(self) -> None:
         threshold = gc.get_threshold()
 
         for i in range(3):
+            self.gc_metrics['tracked_objects'].labels(generation=i).set(len(gc.get_objects(i)))
```
glevco, line 216: This is actually kinda slow to run for a large list of objects, which is our case for generation 2. It has millions of objects and takes a couple seconds. So I don't think we should run it every 5 seconds, which is the Prometheus interval. May we should change the Prometheus interval to something like 30 seconds? Or create a separate LoopingCall for this metric?

hathor/prometheus.py
```diff
@@ -212,6 +213,8 @@ def _set_garbage_collection_metrics(self) -> None:
...
+            # This metric is called `objects` for historical reasons, to keep compatibility with existing graphs.
```
luislhl, line 217: I think we can change it and then update the charts. The updated there is trivial.

hathor/prometheus.py
```diff
@@ -63,7 +63,8 @@
 }
 
 GC_METRICS = {
-    'objects': 'Number of objects tracked by the garbage collector',
+    'tracked_objects': 'Number of objects tracked by the garbage collector',
+    'objects': 'Collection counts from the garbage collector',
```
luislhl, line 67: Maybe we should change the description to something like "Counts used by GC to control the moment each generation should be collected"

glevco: Updated to your suggestion in https://github.com/HathorNetwork/hathor-core/pull/1247/commits/4e142bdb2446a1c6d5fd3df9c1de900d73c3b425

luislhl (approved):

jansegre (approved):

PR #1253, opened 2025-04-01
chore: add reorg log

glevco:
### Motivation

### Acceptance Criteria

### Checklist


hathor/consensus/consensus.py
```diff
@@ -135,12 +135,26 @@ def unsafe_update(self, base: BaseTransaction) -> None:
             assert isinstance(old_best_block, Block)
             new_best_block = base.storage.get_transaction(new_best_tip)
             reorg_size = old_best_block.get_height() - context.reorg_common_block.get_height()
+            # TODO: After we remove block ties, should the assert below be true?
+            # assert old_best_block.get_metadata().voided_by
```
msbrogli, line 139: I don't understand why it would not be true even with block ties.

jansegre (approved):

msbrogli (approved):

PR #1256, opened 2025-04-07
Draft - State of Pool Slotting and Check Entrypoint queue

LFRezende:
Draft - Not for merging, only for meeting LUIS/JAN


hathor/p2p/manager.py
```diff
@@ -417,28 +556,65 @@ def on_connection_failure(self, failure: Failure, peer: Optional[UnverifiedPeer
...
+                protocol_connected = self.outgoing_slot.add_connection(protocol)
+
+        if protocol.connection_type == HathorProtocol.ConnectionType.INCOMING:
+            protocol_connected = self.incoming_slot.add_connection(protocol)
+
+        if protocol.connection_type == HathorProtocol.ConnectionType.DISCOVERED:
+            protocol_connected = self.discovered_slot.add_connection(protocol)
+
+        if protocol.connection_type == HathorProtocol.ConnectionType.CHECK_ENTRYPOINTS:
+            protocol_connected = self.check_entrypoints_slot.add_connection(protocol)
```
jansegre, line 586: Could be simplified to:

```python
match protocol.connection_type:
    case HathorProtocol.ConnectionType.OUTGOING:
        ...
    case HathorProtocol.ConnectionType.INCOMING:
        ...
    ...
```

LFRezende: Ok!
I'll leave it like that since we talked about it needing to "come back" from the if clause.

PR #1269, opened 2025-05-08
Whitelist Toggle by CLI

LFRezende:
### Motivation

[1/2] parts of the project of toggling the feature of peer_whitelisting on/off by a full-node runner of Hathor Network.

Hathor Labs envisions further decentralizing its network, allowing full nodes to freely connect to any peer, not only those approved by the company.  This restriction was previously added to protect full-node runners from malicious peers and attacks.

A better method is to provide control to the user. If an attack occurs, node runners should switch to connecting solely to whitelist peers. In Bootstrap, adding the --x-whitelist-only flag forces the node to block all connections coming from peers not in the whitelist.

*Important Note:* This PR only allows whitelist toggle in node bootstrap, not mid-operation. If merged at the current state, node runners should turn off their nodes and reboot with the flag --x-whitelist-only. A later improvement will come at sysctl, in the next PR, to allow changing the flag while running.

### Acceptance Criteria

- Allows full-node runners to bootstrap their nodes listening only to whitelist peers by the usage of --x-whitelist-only
- If the whitelist is empty, any peer will be allowed to connect, regardless of the bool state of the flag.

### Checklist

- This PR is part 1 of the whole whitelist toggling. While production ready, releasing it alone would be incomplete without the next PR on whitelist toggling with sysctl.


LFRezende:
Important Note: The commits of this PR should not be readily merged, since I'm waiting on the review on the Connection Pool Slots. Once done and if approved, I'll merge the commits onto it.

hathor/p2p/states/peer_id.py
```diff
@@ -166,13 +166,15 @@ def _should_block_peer(self, peer_id: PeerId) -> bool:
...
         peer_is_whitelisted = peer_id in self.protocol.node.peers_whitelist
         # never block whitelisted peers
         if peer_is_whitelisted:
             return False
 
-        # when ENABLE_PEER_WHITELIST is set, we check if we're on sync-v1 to block non-whitelisted peers
-        if self._settings.ENABLE_PEER_WHITELIST:
+        # when whitelist_only (old ENABLE_PEER_WHITELIST) is set,
+        # we check if we're on sync-v1 to block non-whitelisted peers
+        if self.enable_whitelist:
```
msbrogli, line 177: You could safely do `self.protocol.is_whltelist_only`.

tests/p2p/test_whitelist.py
```diff
@@ -12,16 +12,23 @@
 from hathor.simulator import FakeConnection
 from tests import unittest
 
+# Re-work the settings in the tests
 
+
+# ALL tests precisam be reworked.
```
msbrogli, line 18: 🇧🇷 🤣

msbrogli (changes requested):

hathor/manager.py
```diff
@@ -220,6 +220,9 @@ def __init__(
         # Thread pool used to resolve pow when sending tokens
         self.pow_thread_pool = ThreadPool(minthreads=0, maxthreads=settings.MAX_POW_THREADS, name='Pow thread pool')
 
+        # Parameter to track whitelist condition of the full-node
+        self.whitelist_only: bool = p2p_manager.whitelist_only
```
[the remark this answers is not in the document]
LFRezende: I copied from the p2p_manager since it was necessary to remove the whitelist_only boolean from the Hathor Settings. This is not important for the Hathor Core to properly function, but it is useful for the unit tests of the whitelist in the testing file.

hathor/p2p/manager.py
```diff
@@ -187,17 +187,17 @@ def __init__(
         self.lc_connect.clock = self.reactor
         self.lc_connect_interval = 0.2  # seconds
 
+        # Parameter to explicitly enable whitelist-only mode, when False it will still check the whitelist for sync-v1
+        self.whitelist_only = whitelist_only
```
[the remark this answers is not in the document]
LFRezende: Good call, it is clearer.

hathor/p2p/states/base.py
```diff
@@ -34,14 +34,15 @@ class BaseState:
...
-    def __init__(self, protocol: 'HathorProtocol', settings: HathorSettings):
+    def __init__(self, protocol: 'HathorProtocol', settings: HathorSettings, enable_whitelist: bool = False):
         self.log = logger.new(**protocol.get_logger_context())
         self._settings = settings
         self.protocol = protocol
         self.cmd_map = {
             ProtocolMessages.ERROR: self.handle_error,
             ProtocolMessages.THROTTLE: self.handle_throttle,
         }
+        self.enable_whitelist = enable_whitelist
```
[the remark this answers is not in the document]
LFRezende: Fair enough, I'll update it.

hathor/builder/cli_builder.py
```diff
@@ -255,13 +255,19 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
...
+            self.whitelist_only = True
+
         p2p_manager = ConnectionsManager(
             settings=settings,
             reactor=reactor,
             my_peer=peer,
             pubsub=pubsub,
             ssl=True,
-            whitelist_only=False,
+            whitelist_only=self.whitelist_only,
```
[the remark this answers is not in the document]
LFRezende: You're right! Way more straight.

hathor/p2p/states/base.py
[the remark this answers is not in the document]
LFRezende: I just checked now and it seems that HathorProtocol does not have any .enable_whitelist parameter, so it would be needed to declare it on the protocol. Doable, however to me it seems to do the same if declared in the baseState.

hathor/p2p/manager.py
```diff
@@ -421,9 +421,19 @@ def on_peer_connect(self, protocol: HathorProtocol) -> None:
...
+
+        # If whitelist only flag ON and whitelist not empty, deny peer not in whitelist.
+        # Important Note: Perhaps no peer ID in the connection...
+        '''
+                if self.whitelist_only and self.manager.peers_whitelist:
+            if protocol.peer.id not in self.manager.peers_whitelist:
+                self.log.warn('Denied: Connection not in whitelist but node is whitelist-only.')
+                protocol.disconnect(force=True)
+                return
+        '''
```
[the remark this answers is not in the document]
LFRezende: Oops, taken care of!

msbrogli (dismissed):

hathor/cli/run_node.py
```diff
@@ -165,6 +165,8 @@ def create_parser(cls) -> ArgumentParser:
                             help='Enables listening on IPv6 interface and connecting to IPv6 peers')
         parser.add_argument('--x-disable-ipv4', action='store_true',
                             help='Disables connecting to IPv4 peers')
+        parser.add_argument("--x-whitelist-only", action="store_true",
```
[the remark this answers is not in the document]
LFRezende: Good suggestions, they'll make the feature more robust. I think having the flag "--p2p-whitelist" and then the input "off" may be a bit verbose, perhaps something like "--use-whitelist" and if so, giving an argument sets the whitelist to be used (perhaps maybe using only suggestions 3 and 4, maybe 2 for clarity)

I make a new branch and a new PR for this.

msbrogli (changes requested):

hathor/builder/builder.py
```diff
@@ -188,6 +189,7 @@ def __init__(self) -> None:
         self._enable_ipv6: bool = False
         self._disable_ipv4: bool = False
 
+        self._peers_whitelist: URLPeersWhitelist | FilePeersWhitelist | None = None
```
msbrogli, line 192: More generically, you can simply use `PeersWhitelist`. This would allow any subclass to be used here.

hathor/builder/builder.py
```diff
@@ -757,6 +759,11 @@ def enable_event_queue(self) -> 'Builder':
         self._enable_event_queue = True
         return self
 
+    def set_whitelist(self, peers_wl: URLPeersWhitelist | FilePeersWhitelist | None) -> 'Builder':
```
msbrogli, line 762: Same here.

hathor/builder/cli_builder.py
```diff
@@ -300,13 +301,34 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
...
+        # We are remodeling the whitelist_only trope.
+        # Now, we must use solely the p2p_whitelist flag, which shall receive
+
+        # Check whitelist pathing. Default values:
+        p2p_wl = self._args.x_p2p_whitelist or 'default'
+        p2p_wl = p2p_wl.strip()
+        p2p_whitelist: URLPeersWhitelist | FilePeersWhitelist | None = None
+
+        # If in testnet, it should always be disabled.
+        testnet = self._args.testnet
```
msbrogli, line 313: Why? The default would be enabled only for mainnet. Any other network would be disabled by default but it could be override by the command line.

hathor/nanocontracts/storage/block_storage.py
```diff
@@ -103,6 +103,7 @@ def _get_trie(self, root_id: Optional[bytes]) -> 'PatriciaTrie':
         return trie
 
     def get_contract_storage(self, contract_id: ContractId) -> NCContractStorage:
+
```
msbrogli, line 106: You can remove this empty line.

hathor/nanocontracts/storage/changes_tracker.py
```diff
@@ -81,6 +81,7 @@ def __init__(self, nc_id: ContractId, storage: NCContractStorage):
         self._balance_diff: dict[BalanceKey, int] = {}
         self._authorities_diff: dict[BalanceKey, _NCAuthorityDiff] = {}
         self._created_tokens: dict[TokenUid, TokenDescription] = {}
+
```
msbrogli, line 84: You can remove this empty line.

hathor/p2p/peers_whitelist.py
```diff
@@ -0,0 +1,171 @@
...
+        """ Changes following_wl to False. Should not be called directly."""
+        self._following_wl = False
+
+    @abstractmethod
+    def _unsafe_update(self) -> Deferred[None]:
+        pass
+
+    @abstractmethod
+    def is_peer_whitelisted(self, peer_id: PeerId) -> bool:
+        raise NotImplementedError
```
msbrogli, line 79: The implementation is the same for all subclasses. So you can move the code here.

hathor/p2p/peers_whitelist.py
```diff
@@ -0,0 +1,171 @@
...
+        return peer_id in self._current
+
+    def _unsafe_update(self) -> Deferred[None]:
+        """
+            Implementation of base class function.
+            Reads the file in the class path.
+        """
+        with open(self._path, 'r', encoding='utf-8') as fp:
+            content = fp.read()
+        new_whitelist = parse_whitelist(content)
```
msbrogli, line 100: I guess it would be nice to log the differences as you do in the `URLPeersWhitelist`. To avoid duplicated code, you can create an internal `_log_diff(self._current, new_whitelist)` that would do the logging for you.

hathor/p2p/peers_whitelist.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def _handle_refresh_err(self, *args: Any, **kwargs: Any) -> None:
+        """This method will be called when an exception happens inside the whitelist update
+           and ends up stopping the looping call.
+           We log the error and start the looping call again.
+        """
+        self.log.error('whitelist refresh had an exception. Start looping call again.', args=args, kwargs=kwargs)
+        self._reactor.callLater(WHITELIST_RETRY_INTERVAL, self._start_lc)
+
+    def update(self) -> Deferred[None]:
```
msbrogli, line 57: To prevent multiple simultaneous executions, you should skip running if `self._is_running` is true.

hathor/p2p/resources/status.py
```diff
@@ -99,6 +99,11 @@ def render_GET(self, request):
         raw_best_blockchain = self.manager.tx_storage.get_n_height_tips(self._settings.DEFAULT_BEST_BLOCKCHAIN_BLOCKS)
         best_blockchain = to_serializable_best_blockchain(raw_best_blockchain)
 
+        whitelist = self.manager.connections.peers_whitelist
+        if whitelist is not None:
+            peer_ids = whitelist._current
```
msbrogli, line 104: You are accessing an internal attribute. It's always better to create a `get_all_peers()` method and simply call it from here.

hathor/p2p/states/peer_id.py
```diff
@@ -161,30 +161,17 @@ async def handle_peer_id(self, payload: str) -> None:
...
-            protocol_is_whitelist_only = self.protocol.connections.whitelist_only
-            if protocol_is_whitelist_only and not peer_is_whitelisted:
-                return True
-
-        # default is not blocking, this will be sync-v2 peers not on whitelist when not on whitelist-only mode
-        return False
+        if peers_whitelist is None:
+            return True
+
+        if not peers_whitelist._current:
```
msbrogli, line 171: Here you should use the `is_peer_whitelisted()` method (instead of accessing an internal attribute of the object).

hathor/p2p/states/peer_id.py
```diff
@@ -161,30 +161,17 @@ async def handle_peer_id(self, payload: str) -> None:
...
+            return True
+
+        if peers_whitelist._following_wl:
+            return peers_whitelist.is_peer_whitelisted(peer_id)
```
msbrogli, line 175: I didn't get it. What is this code for?

hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +286,56 @@ def refresh_auto_hostname(self) -> None:
...
+                if self.connections.peers_whitelist:
+                    self.connections.peers_whitelist.follow_wl()
+                self.connections.whitelist_toggle(True)
+                return
+            else:
+                # Turning the whitelist off will update in the next refresh cycle
+                if self.connections.peers_whitelist:
+                    self.connections.peers_whitelist.unfollow_wl()
+                self.connections.whitelist_toggle(False)
+                return
```
msbrogli, line 332: Here you can use the same factory you used in `cli_builder.py`. It avoids duplicated code and giver you a unified syntax for whitelisting.

hathor/sysctl/p2p/manager.py
```diff
@@ -122,6 +134,11 @@ def __init__(self, connections: ConnectionsManager) -> None:
             None,
             self.reload_entrypoints_and_connections,
         )
+        self.register(
+            'whitelist_on',
```
msbrogli, line 138: What about `whitelist_off`? I guess you could use `whitelist` with getter and setter (using the factory) and `whitelist.status` with a getter that returns information about the number of peers.

hathor/sysctl/p2p/manager.py
```diff
@@ -49,14 +49,26 @@ def parse_sync_version(name: str) -> SyncVersion:
...
+    getMsg += 'FilePeersWhitelist || ' if isinstance(wl_object, FilePeersWhitelist) else ''
+    getMsg += 'URLPeersWhitelist || ' if isinstance(wl_object, URLPeersWhitelist) else ''
+    getMsg += 'Whitelist: '
+    getMsg += 'ON || ' if wl_object._following_wl else 'OFF || '
+    if wl_object._current:
+        getMsg += 'Amount of PeerIds: {len(wl_object._current)}  '
+    else:
+        getMsg += 'Current peerId list EMPTY. '
+
+    return getMsg
```
msbrogli, line 69: In the sysctl, we always try to make the getter output compatible with the setter. This way, we can easily save the current configuration and apply again later. This current output is not compatible with the setter.

hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +286,56 @@ def refresh_auto_hostname(self) -> None:
...
         """Kill all connections and reload entrypoints from the peer config file."""
         self.connections.reload_entrypoints_and_connections()
+
+    def get_whitelist(self) -> str:
+        """Get status of current whitelist."""
+        if self.connections.peers_whitelist:
+            wl_object = self.connections.peers_whitelist
+            return get_whitelist_msg(wl_object)
+
+        return 'No whitelist.'
```
msbrogli, line 296: Maybe simply `'disabled'`?

tests/p2p/test_whitelist.py
```diff
@@ -134,11 +160,21 @@ def test_update_whitelist(self) -> None:
...
             assert isinstance(_update_whitelist_err_mock.call_args.args[0].value, TimeoutError)
+
+    def test_no_whitelist_but_follow(self) -> None:
+        pass
+
+    def test_whitelist_no_follow(self) -> None:
+        pass
+
+    def test_no_whitelist_unfollow(self) -> None:
+        pass
```
msbrogli, line 180: Missing tests.

tests/resources/p2p/test_status.py
```diff
@@ -16,12 +16,21 @@ def setUp(self):
...
+        self.manager.connections.peers_whitelist._current.add(self.get_random_peer_from_pool().id)
+        self.manager.connections.peers_whitelist._current.add(self.get_random_peer_from_pool().id)
 
         self.manager2 = self.create_peer('testnet')
         address2 = IPv4Address('TCP', '192.168.1.1', 54322)
         self.manager2.connections.my_peer.info.entrypoints.add(PeerAddress.from_address(address2))
+
+        # Manager's whitelist is not empty, so its mock whitelist will be followed.
+        # Since manager 2 is a different instance, we need to add it to the whitelist of manager 1
+        self.manager.connections.peers_whitelist._current.add(self.manager2.my_peer.id)
```
msbrogli, line 28: Create a method `add_peer()` to use it here. This avoids accessing internal attributes.

tests/unittest.py
```diff
@@ -203,6 +204,7 @@ def create_peer(  # type: ignore[no-untyped-def]
         enable_event_queue: bool | None = None,
         enable_ipv6: bool = False,
         disable_ipv4: bool = False,
+        mock_peers_whitelist: URLPeersWhitelist | FilePeersWhitelist | None = None,
```
msbrogli, line 207: It doesn't have to be mocked. You can accept any `PeersWhitelist | None` or simply `str` and use the factory.

tests/unittest.py
```diff
@@ -266,6 +268,15 @@ def create_peer(  # type: ignore[no-untyped-def]
...
         manager = self.create_peer_from_builder(builder, start_manager=start_manager)
 
+        if network == 'testnet' and not mock_peers_whitelist:
+            # Mock peers_whitelist must be initiated beforehand
+            # If given None in testing, it defaults to URL.
+            url = "https://something.com"
+            mock_peers_whitelist = URLPeersWhitelist(manager.reactor, url, False)
+            mock_peers_whitelist.follow_wl()
+            mock_peers_whitelist.start(mock_peers_whitelist._on_remove_callback)
+            manager.connections.peers_whitelist = mock_peers_whitelist
```
msbrogli, line 278: You shouldn't do this here since all dependencies should be injected by the builder.

hathor/p2p/manager.py
```diff
@@ -100,7 +98,7 @@ def __init__(
         pubsub: PubSubManager,
         ssl: bool,
         rng: Random,
-        whitelist_only: bool,
+        peers_whitelist: URLPeersWhitelist | FilePeersWhitelist | None,
```
msbrogli, line 101:
```suggestion
        peers_whitelist: PeersWhitelist | None,
```

hathor/p2p/manager.py
```diff
@@ -187,17 +185,16 @@ def __init__(
         self.lc_connect.clock = self.reactor
         self.lc_connect_interval = 0.2  # seconds
 
-        # A timer to try to reconnect to the disconnect known peers.
-        if self._settings.ENABLE_PEER_WHITELIST:
-            self.wl_reconnect = LoopingCall(self.update_whitelist)
-            self.wl_reconnect.clock = self.reactor
+        # Whitelisted peers.
+        self.peers_whitelist: URLPeersWhitelist | FilePeersWhitelist | None = peers_whitelist
```
msbrogli, line 189:
```suggestion
        self.peers_whitelist: PeersWhitelist | None = peers_whitelist
```

hathor/p2p/manager.py
```diff
@@ -187,17 +185,16 @@ def __init__(
...
+
+        # One may chose whether to follow the whitelist or not. Alterable by sysctl.
+        # This is a p2p_manager copy of the wl_object status "_following_wl".
+        self.only_whitelist = bool(self.peers_whitelist)
```
msbrogli, line 193: Why is that? A disabled whitelist means `self.peers_whitelist is None` and an enabled one would have an object. This object supposedly knows whether rules should be applied or not.

hathor/p2p/manager.py
```diff
@@ -421,9 +406,12 @@ def on_peer_connect(self, protocol: HathorProtocol) -> None:
             self.log.warn('reached maximum number of connections', max_connections=self.max_connections)
             protocol.disconnect(force=True)
             return
+
+        # If whitelist only flag ON and whitelist not empty, deny peer not in whitelist.
+        # Important Note: Perhaps no peer ID in the connection...
```
msbrogli, line 411: Old comments?

hathor/p2p/manager.py
```diff
@@ -940,3 +887,44 @@ def reload_entrypoints_and_connections(self) -> None:
         self.log.warn('Killing all connections and resetting entrypoints...')
         self.disconnect_all_peers(force=True)
         self.my_peer.reload_entrypoints_from_source_file()
+
+    def whitelist_swap(self, wl_object: URLPeersWhitelist | FilePeersWhitelist) -> None:
+        """
+            Altering whitelist (URL/PATH) during full-node runtime.
+        """
+
+        if not wl_object:
```
msbrogli, line 896: Should you disable the whitelist in this case?

hathor/p2p/manager.py
```diff
@@ -940,3 +887,44 @@ def reload_entrypoints_and_connections(self) -> None:
...
+            return
+
+        if self.peers_whitelist:
+            self.peers_whitelist.stop()
+
+        # Sysctl may only update to another URL or Path, not None.
+        self.log.warn("Swapping whitelists... ")
+        self.peers_whitelist = wl_object
+
+        self.whitelist_toggle(wl_object._following_wl)
```
msbrogli, line 906: Why is that? I thought you could simply do `wl_object.start()` and that's it.

msbrogli (changes requested):

hathor/sysctl/p2p/manager.py
LFRezende: True, better, done

hathor/p2p/manager.py
LFRezende: I opted not to, since it didn't make much sense to me sending a null whitelist - the effect it will have is to allow any peer to connect, which can be easily done so via toggling "off" the whitelist.  I chose not to allow a whitelist swap to None mid-full node execution, as whitelist_swap only gets called by a sysctl command.

hathor/p2p/manager.py
LFRezende: The self.only_whitelist parameter is just so, if some other project in the future needs to know easily via connections manager whether the whitelist is on or off. 

This passage states: 
Since it is on INIT, it is called when the Connections Manager is instantiated (via Cli_Builder after the cli commands). Thence, if there is a whitelist (not-null list of peers), hence it must have been enabled via the whitelist cli command. If there is none, in the beginning, even if the flag is ON, since there are no peers, it will still not follow whitelist, hence bool FALSE means it is not following only a whitelist. 
However, when altering the WL on the fly with sysctl, this flag is altered accordingly.
I did this only to be more compact. 

But I do see this is less straightforward than passing the peers_whitelist._followingWL parameter, so I created a getter function to pass this argument.

hathor/builder/builder.py
LFRezende: Done in all of the subcases as well.

hathor/p2p/peers_whitelist.py
LFRezende: Added a log and return statement to each implementation of the unsafe_update. If it is running, it returns and skips execution until the next refresh update.

hathor/nanocontracts/storage/changes_tracker.py
LFRezende: Done

hathor/p2p/peers_whitelist.py
LFRezende: Done as well

hathor/builder/cli_builder.py
```diff
@@ -300,13 +301,34 @@ def create_manager(self, reactor: Reactor) -> HathorManager:
...
+
+        if p2p_wl in ('default', 'hathorlabs'):
+            p2p_whitelist = URLPeersWhitelist(reactor, str(settings.WHITELIST_URL), True) if not testnet else None
+        elif p2p_wl.lower() in ('none', 'disabled'):
+            p2p_whitelist = None
+        elif os.path.isfile(p2p_wl):
+            p2p_whitelist = FilePeersWhitelist(reactor, p2p_wl) if not testnet else None
+        else:
+            # URLPeersWhitelist class rejects non-url paths.
+            p2p_whitelist = URLPeersWhitelist(reactor, p2p_wl, True)
```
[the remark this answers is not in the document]
LFRezende: Good Idea, I just did that. I also implemented it in the manager as below cited.

tests/resources/p2p/test_status.py
LFRezende: Done!

tests/p2p/test_whitelist.py
LFRezende: I left these tests more as a placeholder, with significant names so I could recall what to do later. If we should not push to production code with placeholder tests, I can erase them now and put them back later, or I can implement them now.

hathor/sysctl/p2p/manager.py
LFRezende: I must say this one I don't quite get this one. We could elaborate further?

hathor/p2p/peers_whitelist.py
LFRezende: Added a _log_diff method in PeersWhitelist, and in both implementations of the whitelist it  has been called.

hathor/sysctl/p2p/manager.py
LFRezende: Gotta be honest, didn't follow too much this last bit. I can change to whitelist, it becomes clearer, but the whitelist.status bit as a getter in sysctl seems blurry to me. We can talk about it in a meeting and better decide it?

For now, I'll change it to whitelist and leave it as it is.

hathor/sysctl/p2p/manager.py
LFRezende: In complement to the previous highlight, I made a different getter to the whitelist.status. I think that one is more in line with the intended pattern of messages, perhaps we could leave the "get_whitelist" function as a separate kind, which provides a more detailed log of the current whitelist. What do you think?

hathor/manager.py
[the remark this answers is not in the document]

hathor/p2p/manager.py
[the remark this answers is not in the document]
msbrogli: Ok. Thanks!

hathor/p2p/states/base.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/p2p/states/peer_id.py
msbrogli: Done! Thanks!

tests/p2p/test_whitelist.py
msbrogli: It happens! ;) Thanks!

hathor/builder/builder.py
msbrogli: Thanks!

msbrogli (changes requested):

hathor/p2p/peers_whitelist.py
```diff
@@ -0,0 +1,209 @@
...
+    def refresh(self) -> None:
+        self._unsafe_update()
+
+    def _unsafe_update(self) -> Deferred[None]:
+        """
+            Implementation of base class function.
+            Reads the file in the class path.
+        """
+        # Avoiding re-entrancy. If running, should not update once more.
+        if self._is_running:
```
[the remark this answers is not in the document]
LFRezende: Done

hathor/p2p/resources/status.py
```diff
@@ -108,7 +113,7 @@ def render_GET(self, request):
                 'uptime': now - self.manager.start_time,
                 'entrypoints': self.manager.connections.my_peer.info.entrypoints_as_str(),
             },
-            'peers_whitelist': [str(peer_id) for peer_id in self.manager.peers_whitelist],
+            'peers_whitelist': [str(peer_id) for peer_id in peer_ids],
```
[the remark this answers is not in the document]
LFRezende: My POV: The code was designed with the intention of, if needing to disable a whitelist, one can simply toggle it off in sysctl - meaning, instead of passing a different whitelist or a null one, we can pass the "off" option to make the whitelist class stop following its self._current whitelist. In other words, one need not make the whitelist None if one needs it to be disabled - the class instance will be in the self.peers_whitelist parameter, but its internal attribute (self._follow) will be OFF.

In this case, letting it be None allows the return to be None, and any code which access it would need a safeguard condition for method accessing (ex: peers_whitelist.follow_wl() would throw an error).

hathor/p2p/peers_whitelist.py
```diff
@@ -0,0 +1,209 @@
...
+            return
+
+        current_whitelist = set(self._current)
+        self._log_diff(current_whitelist, new_whitelist)
+
+        peers_to_remove = current_whitelist - new_whitelist
+
+        for peer_id in peers_to_remove:
+            if self._on_remove_callback:
+                self._on_remove_callback(peer_id)
```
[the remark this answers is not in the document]
LFRezende: I was doubtful about the necessity of doing so, but it indeed makes sense. 
I thought the callback function was associated to something on the remote fethcing of peers (like the URL fetching), so I didn't put it there.

Looking deeper, the callback is for the removal of peer_ids. We do need to do so for both then.

hathor/p2p/peers_whitelist.py
```diff
@@ -0,0 +1,209 @@
...
+    def _unsafe_update(self) -> Deferred[None]:
+        """
+            Implementation of base class function.
+            Reads the file in the class path.
+        """
+        # Avoiding re-entrancy. If running, should not update once more.
+        if self._is_running:
+            self.log.warning('whitelist update already running, skipping execution.')
+            return Deferred(None)
+        with open(self._path, 'r', encoding='utf-8') as fp:
```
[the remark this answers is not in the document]
LFRezende: I fail to see what would that actually do... Say someone did this on 23/07, and in the same date one wishes to update its whitelist - wouldn't we be blocking the user from doing so?

PR #1270, opened 2025-05-09
Whitelist Toogle via Sysctl [2/2]

LFRezende:

### DIFF
--- a/hathor/sysctl/p2p/manager.py
+++ b/hathor/sysctl/p2p/manager.py
@@ -123,6 +123,12 @@ class ConnectionsManagerSysctl(Sysctl):
             self.reload_entrypoints_and_connections,
         )
 
+        self.register(
+            'toggle_whitelist_on_off',
+            self.get_whitelist_flag,
+            self.set_whitelist_flag,
+        )
+
     def set_force_sync_rotate(self) -> None:
         """Force a sync rotate."""
         self.connections._sync_rotate_if_needed(force=True)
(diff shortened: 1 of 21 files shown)


hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +275,24 @@ def refresh_auto_hostname(self) -> None:
     def reload_entrypoints_and_connections(self) -> None:
         """Kill all connections and reload entrypoints from the peer config file."""
         self.connections.reload_entrypoints_and_connections()
+
+    def get_whitelist_flag(self):
```
msbrogli, line 279: Missing return type.

hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +275,24 @@ def refresh_auto_hostname(self) -> None:
...
+        """Get whether whitelist-only mode is on or off."""
+        return self.connections.whitelist_only
+
+    def set_whitelist_flag(self, on: bool) -> None:
```
msbrogli, line 283: Maybe rename `on: bool` to `enabled: bool`?

hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +275,24 @@ def refresh_auto_hostname(self) -> None:
...
+            self.connections.manager.whitelist_only = on
+
+        # When setting on, all connections that are from peers not in the whitelist must
+        # be discarded, if whitelist not empty.
+        if on:
+            for conn in self.connections.connections:
+                # If there is no whitelist, no point in itering.
+                if conn.get_peer_id():
+                    if conn.get_peer_id() not in conn.node.peers_whitelist:
+                        conn.disconnect(reason='Whitelist turned on', force=True)
```
msbrogli, line 298: Maybe move this logic to `P2PManager.enable_whitelist()` and `P2PManager.disable_whitelist()`.

msbrogli (changes requested):

hathor/sysctl/p2p/manager.py
LFRezende: You're right, it is clearer. Done!

hathor/sysctl/p2p/manager.py
LFRezende: Added return type -> bool.

hathor/sysctl/p2p/manager.py
LFRezende: Becomes more concise, indeed.

I believe only one function to be necessary (enable_whitelist) as disabling whitelist only requires the whitelist flag to be toggled off, and no further logic is needed.

In any case, it was added.

PR #1271, opened 2025-05-09
feat: adds version field to thin_wallet tokens resource

raul-oliveira:
Acceptance criteria:

- Adds the token version in the database (rocksdb) when a tx is created.
- Adds version prop to `thin_wallet/tokens` endpoints

### DIFF
--- a/hathor/indexes/rocksdb_tokens_index.py
+++ b/hathor/indexes/rocksdb_tokens_index.py
@@ -29,6 +29,7 @@ from hathor.indexes.rocksdb_utils import (
 from hathor.indexes.tokens_index import TokenIndexInfo, TokensIndex, TokenUtxoInfo
 from hathor.transaction import BaseTransaction, Transaction
 from hathor.transaction.base_transaction import TxVersion
+from hathor.transaction.token_info import TokenInfoVersion
 from hathor.util import collect_n, json_dumpb, json_loadb
 
 if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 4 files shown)


hathor/indexes/rocksdb_tokens_index.py
```diff
@@ -168,14 +169,16 @@ def _to_value_info(self, info: _InfoDict) -> bytes:
     def _from_value_info(self, value: bytes) -> _InfoDict:
         return cast(_InfoDict, json_loadb(value))
 
-    def _create_token_info(self, token_uid: bytes, name: str, symbol: str, total: int = 0) -> None:
+    def _create_token_info(
+            self, token_uid: bytes, name: str, symbol: str, token_info_version: Optional[int], total: int = 0) -> None:
```
msbrogli, line 173: The format we've been using is one argument per line.

hathor/indexes/rocksdb_tokens_index.py
```diff
@@ -417,6 +421,9 @@ def get_name(self) -> Optional[str]:
     def get_symbol(self) -> Optional[str]:
         return self._info['symbol']
 
+    def get_version(self) -> Optional[int]:
```
msbrogli, line 424: Why not an instance of TokenInfoVersion?

msbrogli (approved):

hathor/indexes/rocksdb_tokens_index.py
raul-oliveira: At first glance, I thought the db class should not have domain references since it's an infrastructure.
I'm adding since we already have domain imports on this layer.

hathor/indexes/rocksdb_tokens_index.py
raul-oliveira: fixed

PR #1308, opened 2025-06-30
chore: add new allowed OCB address

jansegre:
### Motivation

We can do this because `nano-testnet-bravo` is not publicly live yet.

### Acceptance Criteria

- Add `WQFDxic8xWWnMLL4aE5abY2XRKPNvGhtjY` to allowed OCB addresses on `nano-testnet-bravo`'s config

### Checklist


hathor/conf/nano_testnet.py
msbrogli, line 1: I'm fine with it. Any specific reason that we need another address?

msbrogli (approved):

PR #1311, opened 2025-07-04
refactor(nano): improve error handling

glevco:
### Motivation

Current exception handling on nano contract execution has a few problems. It wraps any exception thrown during user code execution (that is, blueprint code) in an `NCFail`, and later any `NCFail` marks a tx execution as failed. This means bugs in our code would become wrapped exceptions and fail tx executions instead of crashing the full node, which would be the expected behavior.

This PR addresses this by recognizing that user-code and internal code should handle errors differently. If both use normal exception handling, it's pretty easy to accidentally include an internal bug in the eternal state of the blockchain.

It introduces a new `Result` type, inspired by Rust, for explicit error handling. All code from the consensus point where we call the runner down to actually executing user code with `exec()` is now handled with this type instead of raising exceptions. Since user code doesn't know about this type, exceptions are used to bridge internal code to user code and vice-versa, via raising/catching only on the boundaries: exceptions are raised on syscalls, and caught in `exec()`.

### Review Notes

- Begin by reviewing the new `hathor/utils/result.py` file, to understand the new type.
- Review the changes in `metered_exec.py` which are the innermost code handling exceptions when user code is executed with `exec()`. There are also comments explaining the new behavior in it.
- Review `runner.py`.
- Review all the rest which is basically adapting code from using exceptions to using Result.

### Acceptance Criteria

- Implement `Result` type for explicit error handling.
- Refactor error handling on contract execution to use the `Result` type.
  - Code that should be updated (mostly Runner and other nano-related code) is changed to explicitly handle Results instead of exceptions.
  - Code that previously used exceptions and now receives results calls the `unwrap_or_raise()` method to bridge from a result to an exception. This includes for examples calls from APIs or test code.
  - Code that receives results that should never fail call `unwrap()` or `expect()`.

### Rationale and Alternatives

Using a custom `Result` type for error handling is something that we've considered for a while, and as it's a natural solution for the problem this PR addresses, I implemented it. I considered two libs:

1. https://github.com/dry-python/returns - this is a pretty popular Python lib that implements not only a `Result` type but also other utility types. I tried it on branch `refactor/nano/error-handling`. It's a good lib and we may even reconsider using it in the future for its other features, but I didn't like the ergonomics of its `Result` type (it's not a sum type).
2. https://github.com/rustedpy/result - an almost direct translation of Rust's Result type to Python. It's also a good lib and I liked its ergonomics, but it's unmaintained.

Considering this, and wanting to prevent new project dependencies, I decided to not use any of them and implement our own type. I did copy and adapt the implementation of the second one though, which is licensed under MIT. It's mostly the same, except for these differences:

1. Removed deprecated properties.
2. Rename `unwrap_or_raise` to `unwrap_or_raise_another`, because I implemented `unwrap_or_raise` with another behavior (analogous to the unmerged https://github.com/rustedpy/result/pull/199).
3. Remove do-notation support because it's too esoteric for Python and we don't need it.
4. Add support for capturing tracebacks on Results when the error type is an exception.
5. Implement `unwrap_or_propagate` which emulates Rust's question mark operator, for better ergonomics (adapted from the unmerged https://github.com/rustedpy/result/pull/197).

#### Incorrectly wrapped exceptions

Before this PR, ALL exceptions raised during contract execution would become part of the blockchain state forever through the tx execution state. This problem has been mitigated, because now only explicitly handled exceptions (with the Result type) become part of the tx failure. Any other exceptions, for example from bugs or asserts, will now crash the full node as expected.

This means we can still incorrectly make a bug part of the blockchain state if we handle an exception as a Result when it should be an assertion instead, for example. The difference is that now it's explicit. The reviewers should pay special attention to identify these cases — whether there's any instance where we handle a Result error but should use `unwrap()` instead.

#### Serialization

The serialization system is too extensive and raises a lot of exceptions internally. Instead of refactoring the whole module to use Results too, I just handled them on the boundary since nano execution only calls it in a single place.

For that, I identified it can raise 3 exception types: `SerializationError, ValueError, TypeError`. This means that any other exceptions thrown in the serialization module can crash the full node (which is way better than becoming part of the blockchain, as it would before). If there are any other know exceptions that can be raised, we should include them. But it's hard to do better than grepping for raises.

It also means that any `ValueError` or `TypeError` raised by Python itself will also become a reason for tx execution failure, even if it's caused by a bug in internal code. Ideally we should refactor the module to use specific exceptions, and even better, to use Results too.

### Checklist

### DIFF
--- a/extras/custom_checks.sh
+++ b/extras/custom_checks.sh
@@ -115,6 +115,28 @@ function check_do_not_import_twisted_reactor_directly() {
     return 0
 }
 
+function check_inherit_from_nc_fail() {
+    PATTERN='class .*\(NCFail\)'
+
+    if grep -R "$PATTERN" "hathor"; then
+        echo 'do not inherit from NCFail, this is only meant to be inherited by blueprint exceptions, created by users.'
+        echo 'we may allow blueprints to catch these exceptions at some point.'
+        return 1
+    fi
+    return 0
+}
+
+function check_raise_nc_fail() {
+    PATTERN='raise NCFail'
+
+    if grep -R "$PATTERN" "hathor"; then
+        echo 'do not raise NCFail, this is only meant to be raised by blueprints, created by users.'
+        echo 'we may allow blueprints to catch these exceptions at some point.'
+        return 1
+    fi
+    return 0
+}
+
 # List of functions to be executed
 checks=(
 	check_version_match
@@ -123,6 +145,8 @@ checks=(
 	check_do_not_import_tests_in_hathor
 	check_do_not_import_from_hathor_in_entrypoints
 	check_do_not_import_twisted_reactor_directly
+	check_inherit_from_nc_fail
+	check_raise_nc_fail
 )
 
 # Initialize a variable to track if any check fails
--- a/hathor/nanocontracts/exception.py
+++ b/hathor/nanocontracts/exception.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 from typing import TypeAlias
 
 from hathor.exception import HathorError
@@ -39,19 +40,15 @@ class NCInvalidPubKey(NCTxValidationError):
     pass
 
 
-class NCFail(NCError):
-    """Raised by Blueprint's methods to fail execution."""
-
-
-class NanoContractDoesNotExist(NCFail):
+class NanoContractDoesNotExist(NCError):
     pass
 
 
-class BlueprintDoesNotExist(NCFail):
+class BlueprintDoesNotExist(NCError):
     pass
 
 
-class NCSerializationError(NCFail):
+class NCSerializationError(NCError):
     pass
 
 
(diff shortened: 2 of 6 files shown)


hathor/nanocontracts/runner/runner.py
```diff
@@ -541,18 +557,20 @@ def _execute_public_method_call(
...
             kwargs = {}
         else:
             if not is_nc_public_method(method):
-                raise NCInvalidMethodCall(f'method `{method_name}` is not a public method')
+                return Err(NCInvalidMethodCall(f'method `{method_name}` is not a public method'))
             match nc_args:
                 case NCRawArgs(args_bytes):
+                    # TODO: Review the `from_callable` call. It contains some `raise TypeError`
+                    #  that may be caused by user input and crash the full node.
                     parser = Method.from_callable(method)
```
glevco, line 572: TODO

hathor/nanocontracts/storage/changes_tracker.py
```diff
@@ -183,7 +192,9 @@ def commit(self) -> None:
             )
 
         for td in self._created_tokens.values():
-            self.storage.create_token(TokenUid(td.token_id), td.token_name, td.token_symbol)
+            # TODO: Check this expect - can I create a situation where it fails?
+            self.storage.create_token(TokenUid(td.token_id), td.token_name, td.token_symbol) \
+                .expect('this err is already handled in the create_token syscall, so it\'s safe to unwrap here')
```
glevco, line 197: TODO

hathor/transaction/storage/transaction_storage.py
```diff
@@ -1210,14 +1226,14 @@ def get_on_chain_blueprint(self, blueprint_id: BlueprintId) -> OnChainBlueprint:
...
             self.log.debug('no transaction with the given id found', blueprint_id=blueprint_id.hex())
-            raise BlueprintDoesNotExist(blueprint_id.hex())
+            return Err(BlueprintDoesNotExist(blueprint_id.hex()))
         if not isinstance(blueprint_tx, OnChainBlueprint):
-            raise OCBInvalidBlueprintVertexType(blueprint_id.hex())
+            raise OCBInvalidBlueprintVertexType(blueprint_id.hex())  # TODO
         tx_meta = blueprint_tx.get_metadata()
         if tx_meta.voided_by or not tx_meta.first_block:
-            raise OCBBlueprintNotConfirmed(blueprint_id.hex())
+            raise OCBBlueprintNotConfirmed(blueprint_id.hex())  # TODO
```
glevco, line 1234: TODO

hathor/nanocontracts/on_chain_blueprint.py
```diff
@@ -254,10 +254,10 @@ def _load_blueprint_code_exec(self) -> tuple[object, dict[str, object]]:
         metered_executor = MeteredExecutor(fuel=fuel, memory_limit=memory_limit)
         try:
             env = metered_executor.exec(self.code.text)
-        except OutOfFuelError as e:
+        except OutOfFuelError as e:  # TODO
             self.log.error('loading blueprint module failed, fuel limit exceeded')
             raise OCBOutOfFuelDuringLoading from e
-        except OutOfMemoryError as e:
+        except OutOfMemoryError as e:  # TODO
```
glevco, line 260: TODO

hathor/nanocontracts/runner/runner.py
```diff
@@ -198,27 +207,30 @@ def has_contract_been_initialized(self, contract_id: ContractId) -> bool:
...
             self._storages[contract_id] = storage
-        return storage
+        return Ok(storage)
 
-    def _create_changes_tracker(self, contract_id: ContractId) -> NCChangesTracker:
+    @propagate_result
+    def _create_changes_tracker(self, contract_id: ContractId) -> Result[NCChangesTracker, NanoContractDoesNotExist]:
         """Return the latest change tracker for a contract."""
-        nc_storage = self.get_current_changes_tracker_or_storage(contract_id)
+        nc_storage = self.get_current_changes_tracker_or_storage(contract_id).unwrap_or_propagate()
```
glevco, line 225: Should this be an `unwrap()` instead?

hathor/nanocontracts/runner/runner.py
```diff
@@ -572,70 +593,65 @@ def _execute_public_method_call(
...
-        except NCFail:
-            raise
-        except Exception as e:
-            # Convert any other exception to NCFail.
-            raise NCFail from e
+        # Although the context is immutable, we're passing a copy to the blueprint method as an added precaution.
+        # This ensures that, even if the blueprint method attempts to exploit or alter the context, it cannot
+        # impact the original context. Since the runner relies on the context for other critical checks, any
+        # unauthorized modification would pose a serious security risk.
+        copied_context = ctx.copy().unwrap_or_propagate()
```
glevco, line 607: Should this be an `unwrap()` instead?

hathor/nanocontracts/runner/runner.py
```diff
@@ -198,27 +207,30 @@ def has_contract_been_initialized(self, contract_id: ContractId) -> bool:
...
-    def get_storage(self, contract_id: ContractId) -> NCContractStorage:
+    @propagate_result
+    def get_storage(self, contract_id: ContractId) -> Result[NCContractStorage, NanoContractDoesNotExist]:
         """Return the storage for a contract.
 
         If no storage has been created, then one will be created."""
         storage = self._storages.get(contract_id)
         if storage is None:
-            storage = self.block_storage.get_contract_storage(contract_id)
+            storage = self.block_storage.get_contract_storage(contract_id).unwrap_or_propagate()
```
glevco, line 217: Should this be an `unwrap()` instead?

PR #1312, opened 2025-07-04
feat(nano): support key mangling for nested container fields

jansegre:
### Motivation

### Acceptance Criteria

### Checklist


hathor/nanocontracts/blueprint_env.py
```diff
@@ -28,6 +28,9 @@
     from hathor.nanocontracts.runner.types import NCArgs
 
 
+Cache: TypeAlias = dict[bytes, Any] | None
```
msbrogli, line 31: Is it used somewhere else? If yes, maybe rename it to `NCAttrCache`?

hathor/nanocontracts/blueprint.py
```diff
@@ -91,6 +110,25 @@ def __new__(cls, name, bases, attrs, **kwargs):
...
                 raise BlueprintSyntaxError(f'fields with default values are currently not supported: `{field_name}`')
 
+        # validation makes sure we already have it
+        original_init_fn = getattr(new_class, NC_INITIALIZE_METHOD)
+        init_containers_fn = _make_initialize_uninitialized_container_fields_fn(container_fields)
+
+        # patch initialize method so it initializes containers fields implicitly
+        def patched_init_fn(self: Blueprint, *args: Any, **kwargs: Any) -> None:
+            original_init_fn(self, *args, **kwargs)
+            init_containers_fn(self)
```
msbrogli, line 120: Why don't we delegate to developers to initialize like they do with regular types?

hathor/nanocontracts/fields/deque_container.py
```diff
@@ -44,88 +45,87 @@ def last_index(self) -> int:
...
+    __slots__ = ('__storage__', '__prefix__', '__value_node', '__metadata_key')
+    __value_node: ContainerNode[T]
     __metadata_key: bytes
 
-    def __init__(self, storage: NCContractStorage, name: str, value: NCType[T]) -> None:
-        self.__storage = storage
-        self.__name = name
-        self.__value = value
-        self.__metadata_key = f'{name}{KEY_SEPARATOR}{_METADATA_KEY}'.encode()
+    def __init__(self, storage: NCContractStorage, prefix: bytes, value_node: ContainerNode[T]) -> None:
```
msbrogli, line 57: `prefix` here is the attribute name, right?

hathor/nanocontracts/fields/deque_container.py
```diff
@@ -44,88 +45,87 @@ def last_index(self) -> int:
...
         return metadata
 
     def __update_metadata(self, new_metadata: _DequeMetadata) -> None:
-        assert new_metadata.length >= 0
-        if new_metadata.length == 0:
-            return self.__storage.del_obj(self.__metadata_key)
-        self.__storage.put_obj(self.__metadata_key, _METADATA_NC_TYPE, new_metadata)
+        # assert new_metadata.length >= 0
+        # if new_metadata.length == 0:
+        #     return self.__storage__.del_obj(self.__metadata_key)
```
msbrogli, line 111: Old code?

hathor/nanocontracts/fields/dict_container.py
```diff
@@ -27,81 +28,87 @@
...
         if not issubclass(origin_type, Mapping):
             raise TypeError('expected Mapping type')
         args = get_args(type_)
         if not args or len(args) != 2:
-            raise TypeError(f'expected {type_.__name__}[<key type>, <value type>]')
+            raise TypeError('expected exactly 2 type arguments')
         key_type, value_type = args
         if not is_origin_hashable(key_type):
             raise TypeError(f'{key_type} is not hashable')
+        # TODO: check value_type?
```
msbrogli, line 68: TODO?

hathor/nanocontracts/fields/dict_container.py
```diff
@@ -111,42 +118,77 @@ def __len__(self) -> int:
     def __getitem__(self, key: K, /) -> V:
         # get the data from the storage
         db_key = self.__to_db_key(key)
-        return self.__storage.get_obj(db_key, self.__value)
+        has_item = self.__value_node.has_value(db_key)
+        item = self.__value_node.get_value(db_key)
```
msbrogli, line 122: Would this traverse the tree twice? If yes, it might be better to adapt `get_value()` to return whether the item was implicitly created or not.

hathor/nanocontracts/fields/dict_container.py
```diff
@@ -159,10 +201,9 @@ def get(self, key: K, default: V | _T | None, /) -> V | _T | None:
     # XXX: `misc` is ignored because mypy thinks this function does not accept all arguments of the second get overload
     def get(self, key: K, default: V | _T | None = None, /) -> V | _T | None:  # type: ignore[misc]
         # return the value for key if key is in the storage, else default
-        try:
+        if key in self:
             return self[key]
```
msbrogli, line 205: Would this traverse the tree twice?

hathor/nanocontracts/fields/container.py
```diff
@@ -0,0 +1,320 @@
...
+            self.cache[prefix] = container
+
+        # XXX: ignore return-value because mypy doesn't know that the built Container is our P
+        return container  # type: ignore[return-value]
+
+    @override
+    def set_value(self, prefix: bytes, value: P) -> None:
+        container = self._build_container(prefix)
+        if isinstance(value, Container):
+            if value == container:
```
msbrogli, line 255: Which case is this?

tests/nanocontracts/fields/test_dict_field.py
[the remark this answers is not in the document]
jansegre: I'll add some tests for this.

hathor/nanocontracts/storage/changes_tracker.py
```diff
@@ -270,6 +273,7 @@ def is_empty(self) -> bool:
         assert not bool(self._authorities_diff)
         assert not bool(self._created_tokens)
         assert not bool(self._blueprint_id)
+        print(self.data)
```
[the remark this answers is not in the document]
jansegre: Yes, I'll remove it.

hathor/nanocontracts/fields/container.py
```diff
@@ -0,0 +1,320 @@
...
+        """Used to initialize the container if it's not already initialized.
+
+        When a Blueprint class is built, the initialize() method is patched by the metaclass to call this method on
+        every container field, so at the end of calling initialize() call, it's guaranteed that every container will
+        have been initialized.
+        """
+        is_init_key = KEY_SEPARATOR.join([self.__prefix__, INIT_KEY])
+        is_init = self.__storage__.get_obj(is_init_key, INIT_NC_TYPE, default=False)
+        if not is_init:
+            self.__init_storage__()
```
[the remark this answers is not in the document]
jansegre: That's exactly the opposite of the desired behavior for `__try_init_storage__`. But it is the behavior of `__init_storage__` (if called twice the second time will raise an exception).

hathor/nanocontracts/blueprint.py
jansegre: We do, developers can initialize the container fields with this implementation. However, if they don't initialize it in the `initialize` method, it will be _implicitly_ initialized.

This was made in order to support the _current_ behavior of not having to explicitly initialize container fields _and_ to support the behavior of preventing double initializations. We can't have both of these without the implicit initialization.

To prevent double initialization we have to record whether the field was previously initialized. And to support being able to use the field without explicitly initializing it (current behavior), we have to implicitly initialize it (which I chose to do as a post-step injected into the `initialize` method).

hathor/nanocontracts/fields/deque_container.py
jansegre: I'll remove it.

hathor/nanocontracts/fields/dict_container.py
jansegre: The call to `get_value` does not traverse the tree, it only walks one step. When a blueprint does `x = self.foo[1][2]` (assuming `foo: dict[int, dict[int, int]]`), it translates roughly to:

```python
field: Field = getattr(self, 'foo')
# internally __get__ will instantiate a `node: ContainerNode` (ContainerProxy in this case) and call `node.get_value('foo')`
attr: DictContainer = field.__get__(self)  # note that __get__'s signature maps to self=field, instance=self
# internally __getitem__ will call `self.__value_node.get_value(1)` which returns another ContainerProxy
item1: DictContainer = attr.__getitem__(1)
# internally __getitem__ will call `self.__value_node.get_value(2) which returns an int
item2: int = item1.__getitem__(2)
x = item2
```

So the "traversal" happens "externally", by each `__getitem__` call.

hathor/nanocontracts/fields/container.py
jansegre: This exists to keep compatibility with the pattern that we used to do before this PR when using nested containers. Namely this from the `Bet` blueprint:

```
# Update dict indexed by address
partial = self.address_details.get(address, {})
partial.update({
    score: self.bets_address[key]
})
self.address_details[address] = partial
```

When doing `self.address_details[address] = partial` it will end up calling `set_value` with the same proxy container returned by `get_value`, so we don't have to do anything _and_ the assignment is correct. If the proxy container was returned by a different attribute or sub-container, this assignment must fail, since we don't support copying from other containers (we currently can't implement this generically without support for iterators on all container types).

hathor/nanocontracts/fields/container.py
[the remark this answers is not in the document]
msbrogli: Got it. Thanks!

hathor/nanocontracts/fields/deque_container.py
msbrogli: I guess I like `attr_path` more. And I don't think we need an extra class here. I mean, I guess a separate class wouldn't make your code simpler or safer.

hathor/nanocontracts/storage/changes_tracker.py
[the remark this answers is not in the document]
jansegre: Removed.

hathor/nanocontracts/fields/dict_container.py
jansegre: Done.

hathor/nanocontracts/storage/changes_tracker.py
```diff
@@ -129,8 +129,11 @@ def get_obj(self, key: bytes, nc_type: NCType[T], *, default: D = _NOT_PROVIDED)
             obj_td = self.storage.get_obj(key, nc_type, default=default)
             obj = obj_td
         if obj is DeletedKey:
+            if default is not _NOT_PROVIDED:
+                return default
             raise KeyError(key)
         assert not isinstance(obj, DeletedKeyType)
+        assert obj is not _NOT_PROVIDED
```
glevco, line 136: Was this a bug?

tests/nanocontracts/fields/test_compound_field.py
```diff
@@ -1,30 +1,45 @@
...
-
 
 class BlueprintWithCompoundField(Blueprint):
     dc: dict[str, list[int]]
 
     @public
     def initialize(self, ctx: Context) -> None:
-        assert self.dc.get('foo', []) == []
-        self.dc['foo'] = [1, 2, 3]
+        # assert self.dc.get('foo', []) == []
```
glevco, line 14: Remove this?

tests/nanocontracts/fields/test_compound_field.py
```diff
@@ -1,30 +1,45 @@
...
+        foo.pop()
+        assert len(foo) == 0
+        assert len(self.dc['foo']) == 0
+        assert 'bar' in self.dc
+        assert len(self.dc) == 2
         del self.dc['foo']
-        try:
-            self.dc['foo']
-        except KeyError as e:
-            assert e.args[0] == b'dc:\x03foo'
```
glevco, line 25: Shouldn't we `assert pytest.raises(KeyError)` for this case?

tests/nanocontracts/fields/test_compound_field.py
```diff
@@ -1,30 +1,45 @@
...
-        assert 'foo' not in self.dc
+        assert len(self.dc) == 1
         assert 'bar' in self.dc
+        assert 'foo' not in self.dc
+        # XXX: implicit creation:
+        assert self.dc['foo'] == []
```
glevco, line 39: This behavior is weird to me, it makes the dict behave like a `defaultdict(list)`, but that's  unexpected for the blueprint dev

tests/nanocontracts/fields/test_compound_field.py
```diff
@@ -60,12 +75,12 @@ def test_dict_field(self) -> None:
...
+        assert b11_storage.get_obj(b'dc:\x03bar:__init__', INIT_NC_TYPE) is True
 
         assert b12.get_metadata().voided_by is None
         b12_storage = self.manager.get_nc_storage(b12, nc1.hash)
 
         with self.assertRaises(KeyError):
-            b12_storage.get_obj(b'dc:\x03foo', INT_VARTUPLE_NC_TYPE)
-        assert b12_storage.get_obj(b'dc:\x03bar', INT_VARTUPLE_NC_TYPE) == (4, 5, 6, 7)
+            b12_storage.get_obj(b'dc:\x03foo:__init__', INIT_NC_TYPE)
+        assert b12_storage.get_obj(b'dc:\x03bar:__init__', INIT_NC_TYPE) is True
```
glevco, line 86: Shouldn't we also assert the actual values, like the old test did?

tests/nanocontracts/test_all_fields.py
```diff
@@ -132,7 +127,7 @@ def initialize(self, ctx: Context) -> None:
         assert cm.exception.args[0] == 'unsupported field type: `invalid_attribute: typing.Union[str, int]`'
         context_exception = cm.exception.__context__
         assert isinstance(context_exception, TypeError)
-        assert context_exception.args[0] == r"type typing.Union[str, int] is not supported by any Field class"
+        assert context_exception.args[0] == r"type must be either `None | T` or `T | None`"
```
glevco, line 130: That error message is misleading now that we support `CallerId: TypeAlias = Address | ContractId`

hathor/nanocontracts/fields/container.py
```diff
@@ -0,0 +1,324 @@
...
+    def build(self, instance: Blueprint) -> ContainerNode:
+        return ContainerNode.from_type(
+            instance.syscall.__storage__,
+            self.type_,
+            type_map=self.type_map,
+            cache=instance.syscall.__cache__,
+        )
+
+
+class ContainerNode(ABC, Generic[T]):
```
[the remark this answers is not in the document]
jansegre: I can add a `__slots__`, but this class is not exposed to blueprints, they are used internally by `Field` and `Container` classes that are exposed.

tests/nanocontracts/fields/test_compound_field.py
jansegre: I kind of agree, however this is required to keep compatibility with current blueprints. I would prefer to not have implicit creations of any kind (dict/list/set, etc), but it seems to align with how we designed the attribute-key system.

hathor/nanocontracts/storage/changes_tracker.py
jansegre: Yes, I believe it was.

tests/nanocontracts/fields/test_dict_field.py
glevco, line 1: This file is the same as `test_deque_field.py`

jansegre: Oh, I actually didn't mean to include this file in this PR. I'll remove it, but still I have to include more tests.

hathor/nanocontracts/fields/deque_container.py
```diff
@@ -286,7 +285,15 @@ def __ge__(self, value: deque[T], /) -> bool:
...
-
-
-DequeField = ContainerField[DequeStorageContainer[T]]
+        if isinstance(value, list):
+            if len(value) != len(self):
+                return False
+            for item in value:
+                if item not in self:
+                    return False
+            return True
```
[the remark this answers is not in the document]
jansegre: Basically it relies on the `item not in self`, which I think we might need an explicit implementation to make sure it behaves correctly. In plain Python, `[1] in [[1]]` is `True` even though the two `[1]` are different objects, and no hash is used because `list` is not hashable.

But ultimately we could use some tests to make sure it behaves like we want. I'm in favor of it behaving as close to plain Python as possible within reasonable performance.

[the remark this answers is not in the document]
jansegre: True, this implementation is incorrect.

hathor/nanocontracts/fields/field.py
```diff
@@ -40,55 +39,49 @@ class Field(Generic[T], ABC):
...
-                raise TypeError(f'type {type_} is not supported by any Field class') from e
-
-    @classmethod
-    @abstractmethod
-    def _from_name_and_type(cls, name: str, type_: type[T], /, *, type_map: TypeMap) -> Self:
-        raise NotImplementedError
-
-    @abstractmethod
+        if not name.isidentifier():
+            raise TypeError('field name must be a valid identifier')
```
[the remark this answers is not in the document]
jansegre: It should probably be an assertion. It would only be possible by manually instantiating a `Field`, which is not possible inside blueprints, so it would be a code error, not an error that could happen normally. But arguably `TypeError`s are on that boundary.

hathor/nanocontracts/blueprint.py
```diff
@@ -91,6 +110,26 @@ def __new__(cls, name, bases, attrs, **kwargs):
...
+
+        # copy important attributes
+        important_attrs = [NC_METHOD_TYPE_ATTR, NC_ALLOWED_ACTIONS_ATTR, '__annotations__']
+        for attr in important_attrs:
+            setattr(patched_init_fn, attr, getattr(original_init_fn, attr))
+        # XXX: this attribute is important for resolving the original method's signature
+        setattr(patched_init_fn, '__wrapped__', original_init_fn)
+
+        # replace the original init method
+        setattr(new_class, NC_INITIALIZE_METHOD, patched_init_fn)
```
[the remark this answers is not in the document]
jansegre: Yes, it makes sense. But also, it depends on how we implement metering. With the AST approach, only the original function is metered, but with tracing the patched function is metered. Either way, moving it to the runner seems like a good idea.

hathor/nanocontracts/fields/dict_container.py
jansegre: What I mean is, this is not turning a `O(N)` into an `O(N^2)`, it's just doing `O(2*N)`.

tests/nanocontracts/fields/test_compound_field.py
jansegre: This doesn't happen because of the implicit creation.

tests/nanocontracts/fields/test_compound_field.py
glevco: I think now is the time to change this if we prefer another way, before mainnet

tests/nanocontracts/fields/test_dict_field.py
jansegre: Removed.

hathor/nanocontracts/fields/deque_container.py
```diff
@@ -286,7 +285,15 @@ def __ge__(self, value: deque[T], /) -> bool:
...
+        elif isinstance(value, DequeContainer):
+            # XXX: only return True if they point to the same data
+            return self.__prefix__ == value.__prefix__
```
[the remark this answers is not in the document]
jansegre: This has been fixed.

tests/nanocontracts/test_all_fields.py
jansegre: This has been fixed.

hathor/nanocontracts/blueprint.py
```diff
@@ -91,6 +111,27 @@ def __new__(cls, name, bases, attrs, **kwargs):
...
+        # validation makes sure we already have it
+        original_init_fn = getattr(new_class, NC_INITIALIZE_METHOD)
+        init_containers_fn = _make_initialize_uninitialized_container_fields_fn(container_fields)
+
+        # patch initialize method so it initializes containers fields implicitly
+        @wraps(original_init_fn)
+        def patched_init_fn(self: Blueprint, *args: Any, **kwargs: Any) -> Any:
+            ret = original_init_fn(self, *args, **kwargs)
+            init_containers_fn(self)
+            return ret
```
msbrogli, line 123: Shouldn't we initialize the containers before calling the `initialize()` method?

msbrogli (approved):

hathor/nanocontracts/storage/changes_tracker.py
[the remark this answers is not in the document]
msbrogli: Thanks!

hathor/nanocontracts/fields/dict_container.py
msbrogli: Thanks!

hathor/nanocontracts/fields/dict_container.py
msbrogli: Got it. We can remove it when we remove the implicit initialization.

tests/nanocontracts/fields/test_compound_field.py
glevco: We should remember to readd this when the implicit creation is removed, then

hathor/nanocontracts/fields/dict_container.py
msbrogli: Ok. Thanks!

hathor/nanocontracts/blueprint.py
msbrogli: As discussed, the implicit initialization will be removed in a separate PR.

hathor/nanocontracts/fields/container.py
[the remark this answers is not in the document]
glevco: I guess it's fine to leave it like this for now.

hathor/nanocontracts/blueprint.py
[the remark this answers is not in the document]
glevco: As agreed this will be removed when we remove implicit creation.

tests/nanocontracts/fields/test_dict_field.py
glevco: Are you still going to include more tests related to this?

jansegre: I'll include more tests in a new PR.

glevco: Ok, thanks!

tests/nanocontracts/fields/test_compound_field.py
jansegre: Done.

hathor/nanocontracts/fields/field.py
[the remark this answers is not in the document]
jansegre: Done, changed to an assert.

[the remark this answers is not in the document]
glevco: Thanks!

tests/nanocontracts/fields/test_compound_field.py
glevco: Thanks!

hathor/nanocontracts/fields/deque_container.py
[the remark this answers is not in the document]
glevco: Thanks!

hathor/nanocontracts/fields/deque_container.py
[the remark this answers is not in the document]
glevco: Resolving as this code is outdated

PR #1316, opened 2025-07-14
fix(nano): use Uint32 for Timestamp instead of Int32

jansegre:
### Motivation

Timestamps don't make use of negative integers, so it's not useful to use a signed integer encoding.

### Acceptance Criteria

- Introduce `Uint32NCType` that is similar to `Int32NCType` but unsigned instead of signed
- Change `Timestamp`'s mapping to `Uint32NCType`

### Checklist

### DIFF
--- a/hathor/nanocontracts/nc_types/__init__.py
+++ b/hathor/nanocontracts/nc_types/__init__.py
@@ -28,7 +28,7 @@ from hathor.nanocontracts.nc_types.nc_type import NCType
 from hathor.nanocontracts.nc_types.null_nc_type import NullNCType
 from hathor.nanocontracts.nc_types.optional_nc_type import OptionalNCType
 from hathor.nanocontracts.nc_types.signed_data_nc_type import SignedDataNCType
-from hathor.nanocontracts.nc_types.sized_int_nc_type import Int32NCType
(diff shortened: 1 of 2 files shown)


glevco (approved):

hathor/nanocontracts/nc_types/__init__.py
```diff
@@ -118,7 +119,7 @@
     Amount: VarUint32NCType,
     BlueprintId: Bytes32NCType,
     ContractId: Bytes32NCType,
-    Timestamp: Int32NCType,
```
msbrogli, line 121: Is this change compatible with the current implementation in the clients? If not, we should add it to the next release notes.

msbrogli (approved):

hathor/nanocontracts/nc_types/__init__.py
jansegre: The only compatibility caveat is that a timestamp higher than `2**31 - 1` (which is `2038-01-19T04:14:07` in ISO format), would previously be deserialized to negative values, which could cause unexpected behavior.

I haven't checked if it affects any actual transaction in the current network, but I think it's highly unlikely that it does.

Clients shouldn't be affected, unless they incorrectly use negative timestamp values, but I'll align this with Pedro anyway.

PR #1320, opened 2025-07-15
feat(nano): restrict dunder names

glevco:
### Motivation

Improve security by adding further syntax restrictions to blueprints, preventing usage of double underscores in name nodes such as `__x__ = 123` and string literals such as `x = '__x__'`.

### Acceptance Criteria

- Restrict double underscores in name nodes and string literals.

### Checklist

### DIFF
--- a/hathor/verification/on_chain_blueprint_verifier.py
+++ b/hathor/verification/on_chain_blueprint_verifier.py
@@ -57,6 +57,8 @@ class _RestrictionsVisitor(ast.NodeVisitor):
     def visit_Name(self, node: ast.Name) -> None:
         if node.id in AST_NAME_BLACKLIST:
             raise SyntaxError(f'Usage or reference to {node.id} is not allowed.')
+        if '__' in node.id:
+            raise SyntaxError('Using dunder names is not allowed.')
         self.generic_visit(node)
 
     def visit_Attribute(self, node: ast.Attribute) -> None:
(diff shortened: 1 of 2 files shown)


hathor/verification/on_chain_blueprint_verifier.py
```diff
@@ -65,6 +67,11 @@ def visit_Attribute(self, node: ast.Attribute) -> None:
                 raise SyntaxError('Access to internal attributes and methods is not allowed.')
         self.generic_visit(node)
 
+    def visit_Constant(self, node: ast.Constant) -> None:
+        if isinstance(node.value, str):
+            if '__' in node.value:
+                raise SyntaxError('Using dunder names is not allowed.')
+
```
jansegre, line 74: I don't really understand this rule, it's entirely bypassable by several different methods: `f'_{'_'}'`, `'_' '_'`, `'_' + '_'`, ..., and those kind of bypasses work basically everywhere (including type annotations when using strings, although we don't allow string annotations in the places where they are processed)

glevco: You're right, I added the rule last minute and didn't think it through.

hathor/verification/on_chain_blueprint_verifier.py
```diff
@@ -57,6 +57,8 @@ def visit_Try(self, node: ast.Try) -> None:
     def visit_Name(self, node: ast.Name) -> None:
         if node.id in AST_NAME_BLACKLIST:
             raise SyntaxError(f'Usage or reference to {node.id} is not allowed.')
+        if '__' in node.id:
```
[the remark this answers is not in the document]
glevco: True.

glevco:
I'm closing this PR because it's not too useful. We could add this restriction on the bytecode-level, though.

PR #1333, opened 2025-07-22
fix(nano): Stop sharing NanoRNG object between contracts

msbrogli:
### Motivation

The Nano runner previously returned the same NanoRNG instance to all contracts, allowing potential exploits where an attacker could override methods in one contract to manipulate random number generation in others. To mitigate this, several layers of protection have been introduced:

- **Class-level Protection**: Method overriding is prevented using `__setattr__()` in a metaclass, ensuring the class definition remains immutable.
- **Instance-level Protection (Setattr)**: Direct overrides via `__setattr__()` are blocked on individual instances.
- **Instance-level Protection (Slots)**: Methods are made read-only through the use of `__slots__`, restricting attribute modifications.
- **Instance Isolation**: Unique NanoRNG instances are now provided to each contract, eliminating sharing.
- **Class Isolation**: Instances are created via a dynamic shell subclass, ensuring that even the NanoRNG class itself is not shared across contracts.

These measures collectively enhance security by isolating RNG behavior and preventing unauthorized modifications. These measures can also be applied to other classes and instances accessible by contracts.

Original discussion: https://github.com/HathorNetwork/nano-hathor-core/pull/168#discussion_r2025167397

### Acceptance Criteria

1. Add a test that exploits the vulnerability.
6. Stop sharing NanoRNG objects between contracts.
7. Each contract will have a single NanoRNG object which will be shared in case of multiple access by that contract.
8. Add slots to NanoRNG.
9. Prevent assigning methods in NanoRNG instances.
10. Prevent assigning methods in NanoRNG classes.

### Checklist

### DIFF
--- a/hathor/nanocontracts/rng.py
+++ b/hathor/nanocontracts/rng.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Sequence, TypeVar
+from __future__ import annotations
+
+from typing import Any, Sequence, TypeVar
 
 from cryptography.hazmat.primitives.ciphers import Cipher, algorithms
 
@@ -21,11 +23,21 @@ from hathor.difficulty import Hash
 T = TypeVar('T')
 
 
-class NanoRNG:
+class NoMethodOverrideMeta(type):
+    __slots__ = ()
+
+    def __setattr__(cls, name: str, value: Any) -> None:
+        raise AttributeError(f'Cannot override method `{name}`')
+
+
+class NanoRNG(metaclass=NoMethodOverrideMeta):
     """Implement a deterministic random number generator that will be used by the sorter.
 
     This implementation uses the ChaCha20 encryption as RNG.
     """
+
+    __slots__ = ('__seed', '__encryptor', '__frozen')
+
     def __init__(self, seed: bytes) -> None:
         self.__seed = Hash(seed)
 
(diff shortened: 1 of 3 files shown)


hathor/nanocontracts/runner/runner.py
```diff
@@ -743,6 +743,7 @@ def syscall_get_rng(self) -> NanoRNG:
         """Return the RNG for the current contract being executed."""
         if self._rng is None:
             raise ValueError('no seed was provided')
+        return NanoRNG(seed=self._rng.randbits(256))
```
msbrogli, line 746: This behavior will break compatibility with the current testnet if any blueprint is using RNG. If not, it's ok.

hathor/nanocontracts/runner/runner.py
```diff
@@ -743,7 +743,7 @@ def syscall_get_rng(self) -> NanoRNG:
         """Return the RNG for the current contract being executed."""
         if self._rng is None:
             raise ValueError('no seed was provided')
-        return self._rng
+        return NanoRNG(seed=self._rng.randbits(256))
```
msbrogli, line 746: This implementation is creating one NanoRNG object per call. Should it be one per contract?

msbrogli: Done! What do you think?

glevco (approved):

hathor/nanocontracts/runner/runner.py
glevco: Using one per contract looks good to me.

hathor/nanocontracts/rng.py
```diff
@@ -36,17 +46,39 @@ def __init__(self, seed: bytes) -> None:
...
+        useful to prevent sharing classes and objects among different contracts.
+        """
+        class ShellNanoRNG(NanoRNG):
+            pass
+        return ShellNanoRNG(seed=seed)
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        if callable(value):
+            raise AttributeError("Cannot assign methods to this object.")
+        super().__setattr__(name, value)
```
glevco, line 63: I think this `if` can be more restrictive, I think `__setattr__` is only called (by python) when we set attributes, which are defined in slots. Then you could raise this exception any time `name not in self.__slots__`. You'll probably also have to account for name mangling.

jansegre (dismissed):

hathor/nanocontracts/rng.py
msbrogli: I guess I don't have to check whether `name in self.__slots__` because Python does it for me. It will only reach this point if it belongs to `__slots__`. So this code is enforcing that no callables can be assigned (which is true for NanoRNG but it might not be for other classes).

glevco: > It will only reach this point if it belongs to `__slots__`.

That's not true, `__setattr__` always runs, even if the attribute is not in `__slots__`. What prevents the assign is not Python implicitly, but actually your call to `super().__setattr__`. If you forget that call, the assignment would be allowed.

I guess adding the check doesn't hurt, it's just another protection layer — in the same vein of adding the shell class even though we know class attributes can't be changed because we also used slots there.

jansegre (dismissed):

hathor/nanocontracts/runner/runner.py
msbrogli: Great! Jan agrees too! Thanks!

jansegre (approved):

PR #1334, opened 2025-07-23
feat(nano): validate types across contracts

glevco:
### Motivation

This PR adds protections for args and return types in calls across contracts. It resolves https://github.com/HathorNetwork/nano-hathor-core/issues/228, read its description for more details.

### Acceptance Criteria

- Refactor `Runner` to add validation of a method's args and return types when making calls across contracts.
- Refactor `method.py`:
  - Add support for serializing both args and kwargs in `Method`, which is necessary for validating calls across contracts.
  - Implement `ReturnOnly` to support validating the fallback method.
  - Unify exception handling.
- Remove `kwargs` from `CallRecord` and `NCCallBeginEntry` since they're now merged with `args`.

### Checklist


hathor/nanocontracts/runner/types.py
```diff
@@ -142,9 +142,8 @@ class CallRecord:
     # The context passed in this call.
     ctx: Context | None
 
-    # The args and kwargs provided to the method.
+    # The args provided to the method.
     args: tuple[Any, ...]
-    kwargs: dict[str, Any]
```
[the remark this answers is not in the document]
glevco: The validation mechanism merges both `args` and `kwargs` into `args`. The behavior is the same, the calls are also the same, the only difference is how the method is called internally.

hathor/nanocontracts/method.py
```diff
@@ -244,25 +277,41 @@ def from_callable(cls, method: Callable) -> Self:
...
+
+        kwargs = kwargs or {}
+        for name, arg in kwargs.items():
+            if name not in self.arg_names:
+                raise NCFail(f"{self.name}() got an unexpected keyword argument '{name}'")
+            if name in merged:
+                raise NCFail(f"{self.name}() got multiple values for argument '{name}'")
+            merged[name] = arg
+
+        return _serialize_map_exception(self.args, tuple(merged.values()))
```
msbrogli, line 309: I guess the safest here would be the following. The main different is that it does not depend on dicts keeping the insertion order (which is something recent in Python).

```suggestion
        return _serialize_map_exception(self.args, tuple(merged[name] for name in self.arg_names))
```

msbrogli (changes requested):

hathor/nanocontracts/runner/runner.py
```diff
@@ -584,7 +577,7 @@ def _execute_public_method_call(
             # This ensures that, even if the blueprint method attempts to exploit or alter the context, it cannot
             # impact the original context. Since the runner relies on the context for other critical checks, any
             # unauthorized modification would pose a serious security risk.
-            ret = self._metered_executor.call(method, ctx.copy(), *args, **kwargs)
+            ret = self._metered_executor.call(method, ctx.copy(), *args)
```
msbrogli, line 580: We should remove `*args` and let just the actual call do it. This prevent some kinds of attack vectors. Should we refactor it here or in another PR?

hathor/nanocontracts/method.py
jansegre: Alternatively we could explicitly use an `OrderedDict`

hathor/nanocontracts/runner/runner.py
jansegre: Oh, I got it after we talked in Slack. Your proposal is to replace `*args` with `args`. I think it makes sense, actually maybe even force it to be kwargonly, so the call would be `metered_executor.call(method, ctx.copy(), args=args)`.

From the POV of the metered executor, I don't think there is any security issues with using `*args` and `**kwargs` to pass the arguments _now_, but if we want to have any argument to configure some aspect of the call then it would be a problem to use `**kwargs` in particular.

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1334/commits/efd70a49b608744115782a44acfca88db5f8f54d

jansegre (approved):

hathor/nanocontracts/runner/runner.py
msbrogli: Thanks!

msbrogli (approved):

tests/nanocontracts/test_types_across_contracts.py
```diff
@@ -0,0 +1,154 @@
...
+    @view
+    def call_view_wrong_kwarg_type(self, other_id: ContractId) -> None:
+        self.syscall.call_view_method(other_id, 'view_method', a='abc')
+
+    @view
+    def call_view_wrong_return_type(self, other_id: ContractId) -> None:
+        self.syscall.call_view_method(other_id, 'view_method_wrong_return_type')
+
+    @fallback
+    def fallback(self, ctx: Context, method_name: str, nc_args: NCArgs) -> int:
```
msbrogli, line 69: I guess you should add a test where Contract A calls a Contract B method passing a list. Then Contract B mutates the list and this mutation should not exist after it gets back to Contract A.

glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1334/commits/d7d3778b7d9ebd6fab9d0ce8223c574919010e40

msbrogli: Great! Thanks!

msbrogli (approved):

jansegre (approved):

jansegre (approved):

PR #1335, opened 2025-07-23
feat(nano): Add reentrancy protection by default

msbrogli:
### Motivation

### Acceptance Criteria

### Checklist


jansegre (approved):

tests/nanocontracts/test_reentrancy.py
glevco, line 1: Missing a test to cover when a method is reentered and it's not allowed (that is, check that the tx fails with the expected exception).

hathor/nanocontracts/runner/runner.py
```diff
@@ -906,6 +910,17 @@ def _validate_context(self, ctx: Context) -> None:
...
+    def _validate_reentrancy(self, contract_id: ContractId, method_name: str, method: Any) -> None:
+        """Check whether a reentrancy is happening and whether it is allowed."""
+        assert self._call_info is not None
+        allow_reentrancy = getattr(method, NC_ALLOW_REENTRANCY, False)
+        if allow_reentrancy:
+            return
+
+        for call_record in self._call_info.stack:
+            if call_record.contract_id == contract_id and call_record.method_name == method_name:
+                raise NCForbiddenReentrancy(f'reentrancy is forbidden on method `{method_name}`')
```
glevco, line 922: Shouldn't we prevent reentrancy in the whole _contract_, instead of just the method? I think the problem could be recreated in a case like this (pseudo-code):

```python
class Fair(Blueprint):
    @public
    def enter():
        call_another_public(attacker, 'attack')
    
    @public
    def reenter():
        enter()

class Attacker(Blueprint):
    @public
    def attack():
        call_another_public(fair, 'reenter')
```

Because the internal call to `enter()` on `Fair.reenter()` does not go through the runner.

glevco (changes requested):

hathor/nanocontracts/runner/runner.py
msbrogli: When coding, I was actually asking myself whether it should be at the method or the blueprint level. I thought that coders would rather have a finer control than simply allowing reentrancy in all methods at once. I agree with the provided example but the `reenter()` method must allow reentrancy so it must add protection against such attacks.

hathor/nanocontracts/runner/runner.py
```diff
@@ -906,6 +910,17 @@ def _validate_context(self, ctx: Context) -> None:
...
                     raise NCInvalidContext('amount must be positive')
 
+    def _validate_reentrancy(self, contract_id: ContractId, method_name: str, method: Any) -> None:
+        """Check whether a reentrancy is happening and whether it is allowed."""
+        assert self._call_info is not None
+        allow_reentrancy = getattr(method, NC_ALLOW_REENTRANCY, False)
+        if allow_reentrancy:
+            return
+
+        for call_record in self._call_info.stack:
```
msbrogli, line 920: This code runs in O(n) where n is the maximum depth allowed which is 100 right now. We can make it run in O(1) by keeping a `dict[tuple[ContractId, method_name], int]`. Is it worth implementing this optimization?

hathor/nanocontracts/runner/runner.py
```diff
@@ -906,6 +910,17 @@ def _validate_context(self, ctx: Context) -> None:
                 if isinstance(action, BaseTokenAction) and action.amount < 0:
                     raise NCInvalidContext('amount must be positive')
 
+    def _validate_reentrancy(self, contract_id: ContractId, method_name: str, method: Any) -> None:
```
msbrogli, line 913: What about `proxy_call_public_method()`? Should it check for reentrancy? The following test failed because it was making a proxy call to a method with the same name but from another blueprint. I guess we should allow it since it's not actually a reentrancy. Do you agree?

```
FAILED tests/nanocontracts/test_contract_upgrade.py::NCDelegateCallTestCase::test_basic - hathor.nanocontracts.exception.NCForbiddenReentrancy: reentrancy is forbidden on method `inc`
```

tests/nanocontracts/test_reentrancy.py
msbrogli: Done!

glevco: Thanks!

hathor/nanocontracts/runner/runner.py
glevco: Considering [my other comment](https://github.com/HathorNetwork/hathor-core/pull/1335#discussion_r2228817838), I think it would naturally work how it should with proxy calls.

The reentrancy flag would be retrieved from the called method, that is, from the proxy blueprint. But the check would be made only on the `contract_id`, preventing reentry independently of the method name.

hathor/nanocontracts/runner/runner.py
msbrogli: It can get up to 1+2+3+...+100 = 100*101/2 = 5050 comparisons, actually. But I guess it's still not much and shouldn't be a problem. We can always optimize it later. Maybe just leave a note about it?

hathor/nanocontracts/runner/runner.py
msbrogli: Good point. I've already accepted your change. Thanks!

hathor/nanocontracts/runner/runner.py
glevco: LGTM

hathor/nanocontracts/runner/runner.py
jansegre: I feel like this can be optimized, but I don't think it should be a blocker. Do we have any idea about the order of magnitude of time it takes when n=100? Being proportional to ~10.000 times a single verification should be *manageable* I think, but certainly not ideal.

PR #1336, opened 2025-07-25
fix(nano): Prevent argument name collision

msbrogli:
### Motivation

### Acceptance Criteria

### Checklist

### DIFF
--- a/hathor/nanocontracts/blueprint_env.py
+++ b/hathor/nanocontracts/blueprint_env.py
@@ -164,26 +164,26 @@ class BlueprintEnvironment:
     @final
     def call_public_method(
         self,
-        nc_id: ContractId,
-        method_name: str,
-        actions: list[NCAction],
+        _nc_id: ContractId,
+        _method_name: str,
+        _actions: list[NCAction],
         *args: Any,
         **kwargs: Any,
     ) -> Any:
         """Call a public method of another contract."""
-        return self.__runner.syscall_call_another_contract_public_method(nc_id, method_name, actions, args, kwargs)
+        return self.__runner.syscall_call_another_contract_public_method(_nc_id, _method_name, _actions, args, kwargs)
 
     @final
     def proxy_call_public_method(
         self,
-        blueprint_id: BlueprintId,
-        method_name: str,
-        actions: list[NCAction],
+        _blueprint_id: BlueprintId,
+        _method_name: str,
+        _actions: list[NCAction],
         *args: Any,
         **kwargs: Any,
     ) -> Any:
         """Execute a proxy call to a public method of another blueprint."""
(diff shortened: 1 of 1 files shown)


hathor/nanocontracts/blueprint_env.py
```diff
@@ -164,26 +164,26 @@ def can_melt(
     @final
     def call_public_method(
         self,
-        nc_id: ContractId,
-        method_name: str,
-        actions: list[NCAction],
+        _nc_id: ContractId,
```
msbrogli, line 167: Should we forbid arguments starting with `_` in Blueprint methods?

hathor/nanocontracts/blueprint_env.py
```diff
@@ -164,26 +164,28 @@ def can_melt(
     @final
     def call_public_method(
         self,
-        nc_id: ContractId,
-        method_name: str,
-        actions: Sequence[NCAction],
+        _nc_id: ContractId,
+        _method_name: str,
+        _actions: Sequence[NCAction],
+        /,
```
glevco, line 170: Why force positional only?

tests/nanocontracts/test_authorities_call_another.py
```diff
@@ -81,7 +81,7 @@ def call_grant_all_to_other_then_revoke(self, ctx: Context, token_uid: TokenUid)
         self.syscall.call_public_method(
             self.other_id,
             'grant_all_to_other',
-            actions=[],
+            [],  # actions
```
glevco, line 84: I don't like this API, why not let the user do `_actions=[]`?

hathor/nanocontracts/blueprint_env.py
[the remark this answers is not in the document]
glevco: Yes, I prefer forcing kw-only than forcing positional-only. It's way clearer and less error-prone, I've suggested it in the past.

Then we wouldn't even need the `_` prefixes, because python wouldn't allow conflicting keywords. The user simply wouldn't be allowed to use the same argument names that we do (this would be changed when we introduce the new `Contract().some_method(...)` API).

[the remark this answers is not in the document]
jansegre: That affects the serialization of arguments. I don't see any foolproof way to serialize keyword arguments without storing the keys, and storing every key for every argument call can add up quickly.

The size of the keys compared to the values obviously depends on the type of the values, but it's possible that the keys end up taking more space than the values, on average, which would mean the serialized args would double in size.

I don't really like that approach too much.

hathor/nanocontracts/blueprint_env.py
jansegre: I think that's fair.

PR #1341, opened 2025-07-30
tests(nano): fix usage of `raises` with match

glevco:
### Motivation

### Acceptance Criteria

### Checklist


glevco:
@jansegre 
> Are these all cases? If they are should we add a custom linter to prevent using `self.assertRaises`?

Those are all cases that were using `assertRaises` + the `match` argument, but not all cases of `assertRaises`. I think that linter would be great, but we would need another PR to remove all usages, and since it would be a big one (even though it would be simple), I decided it was not a priority now.

PR #1385, opened 2025-08-27
fix(p2p): stop accepting parentless vertices (aka genesis vertices)

jansegre:
### Motivation

This fixes an issue where an assertion error happens when calculating the static metadata, because it assumes every vertex has parents.

### Acceptance Criteria

- Add an option on `VertexParser` to toggle whether genesis vertices can be parsed
- Add a `network_vertex_parser` for use in sync agents
- Handle `struct.error` when parsing in sync agents, the handling is to ignore the received vertex, ideally it should be to close the connection, but "ignoring" is the status-quo behavior for other types of invalid vertex and also the node might in fact request the genesis tx or block, and that would be a larger change that we can do in the future, for now ignoring is enough

### Checklist

### DIFF
--- a/hathor/p2p/sync_v2/agent.py
+++ b/hathor/p2p/sync_v2/agent.py
@@ -776,8 +776,13 @@ class NodeBlockSync(SyncAgent):
         assert self.protocol.connections is not None
 
         blk_bytes = base64.b64decode(payload)
-        blk = self.vertex_parser.deserialize(blk_bytes)
+        try:
+            blk = self.vertex_parser.deserialize(blk_bytes)
+        except struct.error:
+            self.log.warn('received invalid block')
+            return
         if not isinstance(blk, Block):
+            self.log.warn('not a block', hash=blk.hash_hex)
             # Not a block. Punish peer?
             return
         blk.storage = self.tx_storage
@@ -1026,7 +1031,11 @@ class NodeBlockSync(SyncAgent):
 
         # tx_bytes = bytes.fromhex(payload)
         tx_bytes = base64.b64decode(payload)
-        tx = self.vertex_parser.deserialize(tx_bytes)
+        try:
+            tx = self.vertex_parser.deserialize(tx_bytes)
+        except struct.error:
+            self.log.warn('received invalid transaction')
+            return
         if not isinstance(tx, Transaction):
             self.log.warn('not a transaction', hash=tx.hash_hex)
             # Not a transaction. Punish peer?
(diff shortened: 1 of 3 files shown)


hathor/indexes/mempool_tips_index.py
```diff
@@ -204,18 +213,20 @@ def update(self, tx: BaseTransaction, *, remove: Optional[bool] = None) -> None:
...
+                continue
             assert isinstance(tx, Transaction)
             if tx.get_metadata().first_block is not None:
                 bfs.skip_neighbors(tx)
             else:
                 yield tx
 
     def get(self) -> set[bytes]:
-        return set(iter(self._index))
+        return set(iter(self._index)) - self._genesis_hashes
```
glevco, line 232: Aren't the genesis transactions considered mempool on an empty network?

hathor/transaction/vertex_parser.py
```diff
@@ -26,10 +26,13 @@
 
 
 class VertexParser:
-    __slots__ = ('_settings',)
+    __slots__ = ('_settings', '_parse_genesis')
 
-    def __init__(self, *, settings: HathorSettings) -> None:
+    def __init__(self, *, settings: HathorSettings, parse_genesis: bool = True) -> None:
+        # the parser can be tuned to not accept parsing them, which is expected to be the case for parsing network
```
glevco, line 32:
```suggestion
        # the parser can be tuned to not accept parsing genesis, which is expected to be the case for parsing network
```

hathor/indexes/mempool_tips_index.py
glevco: Thanks!

glevco (approved):

hathor/p2p/sync_v2/agent.py
```diff
@@ -1144,7 +1153,7 @@ def handle_data(self, payload: str) -> None:
         try:
             tx = self.vertex_parser.deserialize(data)
         except struct.error:
-            # Invalid data for tx decode
+            self.log.warn('received invalid vertex')
```
msbrogli, line 1156: Shouldn't it be a debug message?

hathor/transaction/vertex_parser.py
```diff
@@ -26,10 +26,13 @@
 
 
 class VertexParser:
-    __slots__ = ('_settings',)
+    __slots__ = ('_settings', '_parse_genesis')
 
-    def __init__(self, *, settings: HathorSettings) -> None:
+    def __init__(self, *, settings: HathorSettings, parse_genesis: bool = True) -> None:
```
msbrogli, line 31: I don't like the proposed solution to the problem because it is modifying the parser which, supposedly, has nothing to do with the vertex processing policies. I guess we should either modify the sync-v2 agent or the vertex handler.

msbrogli (changes requested):

PR #1398, opened 2025-09-09
feat(nano): Add salt to tokens created by contracts

msbrogli:
### Motivation

Add a salt to the token creation by contracts. This change allows contracts to create multiple tokens with the same token symbol and was requested by a use case.

### Acceptance Criteria

1. Add `salt: bytes` as an optional attribute when creating a token using `self.syscall.create_token(...)`. Default is no salt.
2. Add `salt: bytes` as an optional attribute in `derive_child_token_id()`. Default is no salt.

### Checklist

### DIFF
--- a/tests/nanocontracts/test_token_creation.py
+++ b/tests/nanocontracts/test_token_creation.py
@@ -34,13 +34,14 @@ class MyBlueprint(Blueprint):
     def create_deposit_token(
         self,
         ctx: Context,
+        salt: bytes,
         token_name: str,
         token_symbol: str,
         amount: int,
         mint_authority: bool,
         melt_authority: bool,
     ) -> None:
-        self.syscall.create_deposit_token(token_name, token_symbol, amount, mint_authority, melt_authority)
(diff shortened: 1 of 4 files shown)


tests/nanocontracts/test_token_creation.py
```diff
@@ -34,13 +34,14 @@ def withdraw(self, ctx: Context) -> None:
...
         ctx: Context,
+        salt: bytes,
         token_name: str,
         token_symbol: str,
         amount: int,
         mint_authority: bool,
         melt_authority: bool,
     ) -> None:
-        self.syscall.create_token(token_name, token_symbol, amount, mint_authority, melt_authority)
+        self.syscall.create_token(salt, token_name, token_symbol, amount, mint_authority, melt_authority)
```
msbrogli, line 44: Add another test to confirm that a contract is allowed to create two tokens with the same token symbol but with different salts.

glevco (dismissed):

glevco (approved):

PR #1415, opened 2025-09-17
feat(dynamic-transaction-model): fee verification using fee header and nc

raul-oliveira:
Depends on [1414](https://github.com/HathorNetwork/hathor-core/pull/1414)
### Motivation

To make fee payment explicit in the transaction this pr adds the `FeeHeader` in the transaction. It carries all the tokens and amounts, indicating they will be used to pay fees.

### Acceptance Criteria
- Should use the `FeeHeader` to manipulate the `token_dict` to make fees act as outputs in the diff.
- The `verify_sum` will have a flag to indicate if it should enforce the HTR amount validation or not. This flag is useful when dealing with nano contracts that might have tokens that aren't created at the verification stage.
- Add a `verify_sum` call when the nano contract executes in the block_consensus
- Adds the `fee_header` verification without storage whithin the `basic` verification
- Specify the block to be used and the source of data in the methods that calls `on_old_new_vertex`
- Change the `VerificationParams` to receive a block or block storage

### Checklist

- [] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

### DIFF
--- a/hathor/cli/mining.py
+++ b/hathor/cli/mining.py
@@ -142,7 +142,7 @@ def execute(args: Namespace) -> None:
             from hathor.verification.vertex_verifiers import VertexVerifiers
             settings = get_global_settings()
             daa = DifficultyAdjustmentAlgorithm(settings=settings)
-            verification_params = VerificationParams.default_for_mempool()
+            verification_params = VerificationParams(nc_block_root_id=None, enable_checkdatasig_count=True)
             verifiers = VertexVerifiers.create_defaults(
                 reactor=Mock(),
                 settings=settings,
--- a/hathor/consensus/block_consensus.py
+++ b/hathor/consensus/block_consensus.py
@@ -25,6 +25,7 @@ from typing_extensions import assert_never
 from hathor.consensus.context import ReorgInfo
 from hathor.feature_activation.feature import Feature
 from hathor.transaction import BaseTransaction, Block, Transaction
+from hathor.transaction.exceptions import TokenNotFound
 from hathor.transaction.nc_execution_state import NCExecutionState
 from hathor.transaction.types import MetaNCCallRecord
 from hathor.util import classproperty
@@ -37,6 +38,7 @@ if TYPE_CHECKING:
     from hathor.nanocontracts.nc_exec_logs import NCLogStorage
     from hathor.nanocontracts.runner import Runner
     from hathor.nanocontracts.runner.runner import RunnerFactory
+    from hathor.nanocontracts.storage import NCBlockStorage
 
 logger = get_logger()
 
@@ -236,8 +238,17 @@ class BlockConsensusAlgorithm:
 
             runner = self._runner_factory.create(block_storage=block_storage, seed=seed_hasher.digest())
             exception_and_tb: tuple[NCFail, str] | None = None
+            token_dict = tx.get_complete_token_info(block_storage)
+            should_verify_sum_after_execution = any(token_info.version is None for token_info in token_dict.values())
+
             try:
                 runner.execute_from_tx(tx)
+
+                # after the execution we have the latest state in the storage
+                # and at this point no tokens pending creation
+                if should_verify_sum_after_execution:
+                    self._verify_sum_after_execution(tx, block_storage)
+
             except NCFail as e:
                 kwargs: dict[str, Any] = {}
                 if tx.name:
@@ -300,6 +311,19 @@ class BlockConsensusAlgorithm:
                     assert NC_EXECUTION_FAIL_ID not in tx_meta.voided_by
                 case _:  # pragma: no cover
                     assert_never(tx_meta.nc_execution)
+
+    def _verify_sum_after_execution(self, tx: Transaction, block_storage: NCBlockStorage) -> None:
+        from hathor import NCFail
+        from hathor.verification.transaction_verifier import TransactionVerifier
+        try:
+            token_dict = tx.get_complete_token_info(block_storage)
+            TransactionVerifier.verify_sum(self._settings, token_dict)
+        except TokenNotFound as e:
+            # At this point, any nonexistent token would have made a prior validation fail. For example, if there
+            # was a withdrawal of a nonexistent token, it would have failed in the balance validation before.
+            raise AssertionError from e
+        except Exception as e:
+            raise NCFail from e
 
     def nc_update_metadata(self, tx: Transaction, runner: 'Runner') -> None:
         from hathor.nanocontracts.runner.types import CallType
--- a/hathor/nanocontracts/balance_rules.py
+++ b/hathor/nanocontracts/balance_rules.py
@@ -30,8 +30,7 @@ from hathor.nanocontracts.types import (
     NCGrantAuthorityAction,
     NCWithdrawalAction,
 )
-from hathor.transaction.token_info import TokenInfoDict
-from hathor.transaction.transaction import TokenInfo
+from hathor.transaction.token_info import TokenInfoDict, TokenVersion
 
 T = TypeVar('T', bound=BaseAction)
 
@@ -101,9 +100,12 @@ class _DepositRules(BalanceRules[NCDepositAction]):
 
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
-        token_info = token_dict.get(self.action.token_uid, TokenInfo.get_default())
+        token_info = token_dict[self.action.token_uid]
         token_info.amount = token_info.amount + self.action.amount
         token_dict[self.action.token_uid] = token_info
+
+        if token_info.version == TokenVersion.FEE:
+            token_info.chargeable_outputs += 1
 
     @override
     def nc_callee_execution_rule(self, callee_changes_tracker: NCChangesTracker) -> None:
@@ -125,9 +127,12 @@ class _WithdrawalRules(BalanceRules[NCWithdrawalAction]):
 
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
-        token_info = token_dict.get(self.action.token_uid, TokenInfo.get_default())
+        token_info = token_dict[self.action.token_uid]
         token_info.amount = token_info.amount - self.action.amount
         token_dict[self.action.token_uid] = token_info
+
+        if token_info.version == TokenVersion.FEE:
+            token_info.chargeable_inputs += 1
 
     @override
     def nc_callee_execution_rule(self, callee_changes_tracker: NCChangesTracker) -> None:
@@ -150,7 +155,7 @@ class _GrantAuthorityRules(BalanceRules[NCGrantAuthorityAction]):
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
         assert self.action.token_uid != HATHOR_TOKEN_UID
-        token_info = token_dict.get(self.action.token_uid, TokenInfo.get_default())
+        token_info = token_dict[self.action.token_uid]
         if self.action.mint and not token_info.can_mint:
             raise NCInvalidAction(
                 f'{self.action.name} token {self.action.token_uid.hex()} requires mint, but no input has it'
@@ -202,7 +207,7 @@ class _AcquireAuthorityRules(BalanceRules[NCAcquireAuthorityAction]):
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
         assert self.action.token_uid != HATHOR_TOKEN_UID
-        token_info = token_dict.get(self.action.token_uid, TokenInfo.get_default())
+        token_info = token_dict[self.action.token_uid]
         token_info.can_mint = token_info.can_mint or self.action.mint
         token_info.can_melt = token_info.can_melt or self.action.melt
         token_dict[self.action.token_uid] = token_info
--- a/hathor/p2p/sync_v2/transaction_streaming_client.py
+++ b/hathor/p2p/sync_v2/transaction_streaming_client.py
@@ -47,11 +47,17 @@ class TransactionStreamingClient:
         self.protocol = self.sync_agent.protocol
         self.tx_storage = self.sync_agent.tx_storage
         self.verification_service = self.protocol.node.verification_service
-        # XXX: since it's not straightforward to get the correct block, it's OK to just disable checkdatasig counting,
-        #      it will be correctly enabled when doing a full validation anyway.
-        self.verification_params = VerificationParams(enable_checkdatasig_count=False)
-        self.reactor = sync_agent.reactor
 
+        # XXX: Since it's not straightforward to get the correct block, it's OK to just disable checkdatasig counting,
+        #      it will be correctly enabled when doing a full validation anyway.
+        #      We can also set the `nc_block_root_id` to `None` because we only call `verify_basic`,
+        #      which doesn't need it.
+        self.verification_params = VerificationParams(
+            enable_checkdatasig_count=False,
+            nc_block_root_id=None,
+        )
+
+        self.reactor = sync_agent.reactor
         self.log = logger.new(peer=self.protocol.get_short_peer_id())
 
         # List of blocks from which we will receive transactions.
--- a/hathor/transaction/resources/create_tx.py
+++ b/hathor/transaction/resources/create_tx.py
@@ -119,7 +119,10 @@ class CreateTxResource(Resource):
         # need to run verify_inputs first to check if all inputs exist
         verifiers.tx.verify_inputs(tx, skip_script=True)
         verifiers.vertex.verify_parents(tx)
-        verifiers.tx.verify_sum(tx.get_complete_token_info())
+
+        best_block = self.manager.tx_storage.get_best_block()
+        block_storage = self.manager.get_nc_block_storage(best_block)
+        verifiers.tx.verify_sum(self.manager._settings, tx.get_complete_token_info(block_storage))
 
 
 CreateTxResource.openapi = {
--- a/hathor/transaction/token_info.py
+++ b/hathor/transaction/token_info.py
@@ -14,12 +14,14 @@
 
 from dataclasses import dataclass
 from enum import IntEnum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from hathor.types import TokenUid
 
 if TYPE_CHECKING:
     from hathor.conf.settings import HathorSettings
+    from hathor.nanocontracts.storage import NCBlockStorage
+    from hathor.transaction.storage import TransactionStorage
 
 
 class TokenVersion(IntEnum):
@@ -31,30 +33,14 @@ class TokenVersion(IntEnum):
 # used when (de)serializing token information
 @dataclass(slots=True, kw_only=True)
 class TokenInfo:
-    amount: int
-    can_mint: bool
-    can_melt: bool
-    version: TokenVersion
+    version: TokenVersion | None
+    amount: int = 0
+    can_mint: bool = False
+    can_melt: bool = False
     # count of non-authority outputs that is used to calculate the fee
     chargeable_outputs: int = 0
     # count of non-authority inputs that is used to calculate the fee
     chargeable_inputs: int = 0
-
-    @classmethod
-    def get_default(cls,
-                    version: TokenVersion = TokenVersion.NATIVE,
-                    can_mint: bool = False,
-                    can_melt: bool = False) -> 'TokenInfo':
-        """
-        Create default deposit token info with zero amount and optional mint/melt permissions.
-        """
-
-        return TokenInfo(
-            amount=0,
-            can_mint=can_mint,
-            can_melt=can_melt,
-            version=version,
-        )
 
     def has_been_melted(self) -> bool:
         """
@@ -78,8 +64,17 @@ class TokenDescription:
     token_symbol: str
     token_version: TokenVersion
 
+    def __post_init__(self) -> None:
+        assert isinstance(self.token_version, TokenVersion)
+
 
 class TokenInfoDict(dict[TokenUid, TokenInfo]):
+    __slots__ = ('fees_from_fee_header',)
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.fees_from_fee_header: int = 0
+
     def calculate_fee(self, settings: 'HathorSettings') -> int:
         """
          Calculate the total fee based on the number of chargeable
@@ -109,3 +104,26 @@ class TokenInfoDict(dict[TokenUid, TokenInfo]):
                 if token_info.chargeable_inputs > 0:
                     fee += settings.FEE_PER_OUTPUT
         return fee
+
+
+def get_token_version(
+    tx_storage: 'TransactionStorage',
+    nc_block_storage: 'NCBlockStorage',
+    token_uid: TokenUid
+) -> TokenVersion | None:
+    """
+    Get the token version for a given token uid.
+    It searches first in the tx storage and then in the block storage.
+    """
+    from hathor.conf.settings import HATHOR_TOKEN_UID
+    if token_uid == HATHOR_TOKEN_UID:
+        return TokenVersion.NATIVE
+    from hathor.transaction.storage.exceptions import TransactionDoesNotExist
+    try:
+        token_creation_tx = tx_storage.get_token_creation_transaction(token_uid)
+        return token_creation_tx.token_version
+    except TransactionDoesNotExist:
+        from hathor.nanocontracts.types import TokenUid
+        if nc_block_storage.has_token(TokenUid(token_uid)):
+            return nc_block_storage.get_token_description(TokenUid(token_uid)).token_version
+    return None
--- a/hathor/transaction/transaction.py
+++ b/hathor/transaction/transaction.py
@@ -29,7 +29,7 @@ from hathor.transaction.exceptions import InvalidToken
 from hathor.transaction.headers import NanoHeader, VertexBaseHeader
 from hathor.transaction.headers.fee_header import FeeHeader
 from hathor.transaction.static_metadata import TransactionStaticMetadata
-from hathor.transaction.token_info import TokenInfo, TokenInfoDict, TokenVersion
+from hathor.transaction.token_info import TokenInfo, TokenInfoDict, TokenVersion, get_token_version
 from hathor.transaction.util import VerboseCallback, unpack, unpack_len
 from hathor.types import TokenUid, VertexId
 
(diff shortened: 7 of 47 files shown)


hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -49,7 +49,12 @@ def __init__(self,
         self.verification_service = self.protocol.node.verification_service
         # XXX: since it's not straightforward to get the correct block, it's OK to just disable checkdatasig counting,
         #      it will be correctly enabled when doing a full validation anyway.
-        self.verification_params = VerificationParams(enable_checkdatasig_count=False)
+        # TODO-RAUL: is correct to grab the best block here?
+        best_block = self.tx_storage.get_best_block()
+        self.verification_params = VerificationParams(
+            enable_checkdatasig_count=False,
+            block_or_block_storage=best_block
+        )
```
glevco, line 57: TODO

hathor/transaction/transaction.py
```diff
@@ -348,21 +352,55 @@ def get_minimum_number_of_inputs(self) -> int:
...
+            return
+
+        fee_header = self.get_fee_header()
+        fees = fee_header.get_fees()
+        # we store the total fee amount from the header to be used in the verify_sum
+        token_dict.fees_from_fee_header = fee_header.total_fee_amount()
+        for fee in fees:
+            token_info = token_dict.get(fee.token_uid)
+            if token_info is None:
+                raise InvalidToken('no inputs for token {}'.format(fee.token_uid.hex()))
```
glevco, line 391: Why do we require an input? Can't the fee be paid with a token from a nano withdrawal?

hathor/transaction/transaction.py
```diff
@@ -348,21 +352,55 @@ def get_minimum_number_of_inputs(self) -> int:
...
+
+            # it should be defined in the inputs
+            assert token_info.version is not TokenVersion.FEE
```
glevco, line 394: I didn't get this comment. Also, where's the validation before the assertion located?

hathor/transaction/transaction.py
```diff
@@ -348,21 +352,55 @@ def get_minimum_number_of_inputs(self) -> int:
...
+
+            # act as a regular output subtracting from the total amount
+            token_info.amount += fee.amount
```
glevco, line 397: Clarify that we "subtract by adding"

hathor/transaction/transaction.py
glevco: Also, it should be `assert ... in (NATIVE, DEPOSIT)`

hathor/nanocontracts/balance_rules.py
```diff
@@ -101,10 +100,13 @@ class _DepositRules(BalanceRules[NCDepositAction]):
 
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
-        token_info = token_dict.get(self.action.token_uid, TokenInfo.get_default())
+        token_info = token_dict[self.action.token_uid]
         token_info.amount = token_info.amount + self.action.amount
         token_dict[self.action.token_uid] = token_info
 
+        if token_info.version == TokenVersion.FEE:
+            token_info.chargeable_outputs += 1
```
glevco, line 108: What about when `token_info is None`, it will postpone the chargeable outputs here, and I couldn't find where we re-run it.

hathor/nanocontracts/balance_rules.py
```diff
@@ -125,10 +127,13 @@ class _WithdrawalRules(BalanceRules[NCWithdrawalAction]):
 
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
-        token_info = token_dict.get(self.action.token_uid, TokenInfo.get_default())
+        token_info = token_dict[self.action.token_uid]
         token_info.amount = token_info.amount - self.action.amount
         token_dict[self.action.token_uid] = token_info
 
+        if token_info.version == TokenVersion.FEE:
+            token_info.chargeable_inputs += 1
```
glevco, line 135: Same here.

glevco (changes requested):

hathor/p2p/sync_v2/transaction_streaming_client.py
glevco: Solved in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/ffa6294f4f822b6913a79c5751aee4aa36b00f7f

hathor/verification/verification_params.py
```diff
@@ -25,14 +28,20 @@ class VerificationParams:
     reject_locked_reward: bool = True
     skip_block_weight_verification: bool = False
     enable_nano: bool = False
+    block_or_block_storage: Block | NCBlockStorage | None = None
```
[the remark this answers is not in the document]
glevco: Changed in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/ffa6294f4f822b6913a79c5751aee4aa36b00f7f

hathor/transaction/token_info.py
```diff
@@ -34,15 +36,15 @@ class TokenInfo:
...
+    version: TokenVersion | None
     # count of non-authority outputs that is used to calculate the fee
     chargeable_outputs: int = 0
     # count of non-authority inputs that is used to calculate the fee
     chargeable_inputs: int = 0
 
     @classmethod
     def get_default(cls,
-                    version: TokenVersion = TokenVersion.NATIVE,
+                    version: TokenVersion | None = TokenVersion.NATIVE,
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/ffa6294f4f822b6913a79c5751aee4aa36b00f7f

hathor/transaction/token_info.py
```diff
@@ -80,6 +82,8 @@ class TokenDescription:
 
 
 class TokenInfoDict(dict[TokenUid, TokenInfo]):
+    fees_from_fee_header: int = 0
+
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/ffa6294f4f822b6913a79c5751aee4aa36b00f7f

hathor/verification/transaction_verifier.py
```diff
@@ -64,18 +67,20 @@
...
         *,
         settings: HathorSettings,
         daa: DifficultyAdjustmentAlgorithm,
         feature_service: FeatureService,
+        nc_storage_factory: NCStorageFactory,
     ) -> None:
         self._settings = settings
         self._daa = daa
         self._feature_service = feature_service
+        self._nc_storage_factory = nc_storage_factory
```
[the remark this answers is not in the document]
glevco: Removed in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/ffa6294f4f822b6913a79c5751aee4aa36b00f7f

hathor/consensus/block_consensus.py
```diff
@@ -237,7 +240,18 @@ def _nc_execute_calls(self, block: Block, *, is_reorg: bool) -> None:
...
+                token_dict = tx.get_complete_token_info(block_storage)
+                for token_uid, token_info in token_dict.items():
+                    if token_info.version is None:
+                        should_verify_after_execution = True
+
                 runner.execute_from_tx(tx)
+                # after the execution we have the latest state in the storage
+                # and at this point no tokens pending creation
+                if should_verify_after_execution:
+                    self.vertex_verifiers.tx.verify_sum(tx.get_complete_token_info(block_storage))
```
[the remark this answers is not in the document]
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/ffa6294f4f822b6913a79c5751aee4aa36b00f7f

hathor/verification/verification_params.py
```diff
@@ -16,11 +16,14 @@
...
 from dataclasses import dataclass
 
+from hathor.transaction import Block
+
 
 @dataclass(slots=True, frozen=True, kw_only=True)
 class VerificationParams:
     """Contains every parameter/setting to run a single verification."""
 
+    best_block: Block | None
```
msbrogli, line 26: This does not seem to be a parameter of the verification. Why are you passing it here if you can get it in the verifier when needed?

hathor/nanocontracts/runner/runner.py
```diff
@@ -187,6 +187,8 @@ def execute_from_tx(self, tx: Transaction) -> None:
         else:
             self.call_public_method_with_nc_args(contract_id, nano_header.nc_method, context, nc_args)
 
+        # TODO-RAUL: add the fee header validation here
+
```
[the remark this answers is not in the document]
glevco: I'm talking to raul, and this can be removed.

hathor/transaction/transaction.py
glevco: This is probably correct because the update token_dict with fees runs after the update with nano. We just need to update this error message. Also, add a test for this case.

hathor/transaction/transaction.py
glevco: I'll change it to a validation instead of assertion.

hathor/verification/transaction_verifier.py
```diff
@@ -387,9 +396,22 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+        It should run only if a fee header isn't present.
+        """
+        assert not tx.has_fees()
+
+        for token_uid in tx.tokens:
+            assert tx.storage is not None
+            token_version = get_token_version(tx.storage, nc_block_storage, token_uid)
+            if token_version == TokenVersion.FEE:
+                raise TxValidationError("All fee tokens should be declared in the fee header.")
+
```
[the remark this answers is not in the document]
glevco: Add test to try to pay fee with a fee token, in these cases:

1. Fee token created by a token creation tx
2. Fee token created by nano on another tx
3. Fee token created by nano on the same tx

hathor/nanocontracts/balance_rules.py
glevco: This will re-run indirectly through the get_complete_token_info that's called on the postponed verification.

hathor/nanocontracts/balance_rules.py
glevco: Same answer.

tests/tx/test_fee_tokens.py
```diff
@@ -420,171 +455,25 @@ def test_fee_token_tx_paid_with_htr_and_deposit(self) -> None:
...
         )
-        fee = tx2.get_complete_token_info().calculate_fee(self.manager._settings)
-        self.assertEqual(fee, 5)
-
-        #  It's the signature of the output of the tx item
-        #  this signature_data allows the tx output to be spent by the tx2 inputs
-        self.sign_inputs(tx2)
-        self.resolve_and_propagate(tx2)
-
-    def test_fee_token_melt_and_deposit_token_to_pay_the_fee_without_melt_authority(self) -> None:
```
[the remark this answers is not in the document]
glevco: This can be removed.

hathor/transaction/token_utils.py
```diff
@@ -0,0 +1,50 @@
...
+
+if TYPE_CHECKING:
+    from hathor.nanocontracts.storage import NCBlockStorage
+    from hathor.nanocontracts.types import TokenUid as NCTokenUid
+    from hathor.transaction.storage import TransactionStorage
+    from hathor.transaction.token_info import TokenVersion
+    from hathor.types import TokenUid as TokenUidType
+
+
+def get_token_version(
```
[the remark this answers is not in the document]
glevco: Removed in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/d3c1f509c1c246143f7630e01c816a89ccdb3790

hathor/verification/transaction_verifier.py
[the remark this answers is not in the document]
glevco: Removed in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/d3c1f509c1c246143f7630e01c816a89ccdb3790

hathor/cli/mining.py
```diff
@@ -142,14 +142,13 @@ def execute(args: Namespace) -> None:
             from hathor.verification.vertex_verifiers import VertexVerifiers
             settings = get_global_settings()
             daa = DifficultyAdjustmentAlgorithm(settings=settings)
-            verification_params = VerificationParams.default_for_mempool()
+            verification_params = VerificationParams(best_block=None, enable_checkdatasig_count=True)
```
raul-oliveira, line 145: Why are you passing this parameter as None if the `VerificationParams.default_for_mempool` alterady satisfies it?

hathor/nanocontracts/balance_rules.py
```diff
@@ -157,8 +155,7 @@ class _GrantAuthorityRules(BalanceRules[NCGrantAuthorityAction]):
     @override
     def verification_rule(self, token_dict: TokenInfoDict) -> None:
         assert self.action.token_uid != HATHOR_TOKEN_UID
-        token_info = token_dict.get(self.action.token_uid)
-        assert token_info is not None
+        token_info = token_dict[self.action.token_uid]
```
raul-oliveira, line 158: Why this change? If you try to access a token that isn't registered it will raise an error, right?

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -49,11 +49,9 @@ def __init__(self,
         self.verification_service = self.protocol.node.verification_service
         # XXX: since it's not straightforward to get the correct block, it's OK to just disable checkdatasig counting,
         #      it will be correctly enabled when doing a full validation anyway.
-        # TODO-RAUL: is correct to grab the best block here?
-        best_block = self.tx_storage.get_best_block()
         self.verification_params = VerificationParams(
```
raul-oliveira, line 52: It is looking like the previous `default_for_mempool`

hathor/transaction/resources/create_tx.py
```diff
@@ -119,10 +119,10 @@ def _verify_unsigned_skip_pow(self, tx: Transaction) -> None:
         # need to run verify_inputs first to check if all inputs exist
         verifiers.tx.verify_inputs(tx, skip_script=True)
         verifiers.vertex.verify_parents(tx)
+
         best_block = self.manager.tx_storage.get_best_block()
-        assert self.manager.verification_service._nc_storage_factory is not None
-        block_storage = self.manager.verification_service._nc_storage_factory.get_block_storage_from_block(best_block)
-        verifiers.tx.verify_sum(tx.get_complete_token_info(block_storage))
+        block_storage = self.manager.get_nc_block_storage(best_block)
```
raul-oliveira, line 124: Could this get_nc_block_storage be null?

hathor/transaction/transaction.py
```diff
@@ -331,15 +331,15 @@ def verify_checkpoint(self, checkpoints: list[Checkpoint]) -> None:
...
         It uses a block storage with the latest token changes in nano contracts
         """
 
-        token_dict = self._get_token_info_from_inputs(block_storage)
-        self._update_token_info_from_nano_actions(token_dict=token_dict, block_storage=block_storage)
-        # This one must be called last so token_dict already contains all tokens in inputs and nano actions.
+        token_dict = self._get_token_info_from_inputs(nc_block_storage)
+        self._update_token_info_from_nano_actions(token_dict=token_dict, nc_block_storage=nc_block_storage)
+        # These must be called last so token_dict already contains all tokens in inputs and nano actions.
         self._update_token_info_from_outputs(token_dict=token_dict)
```
raul-oliveira, line 343: Should we have a test to validate this order?

hathor/verification/transaction_verifier.py
```diff
@@ -240,16 +235,18 @@ def verify_output_token_indexes(self, tx: Transaction) -> None:
...
         token_dict: TokenInfoDict,
-        enforce_htr_check: bool = True
+        allow_nonexistent_tokens: bool = False,
     ) -> None:
         """Verify that the sum of outputs is equal of the sum of inputs, for each token. If sum of inputs
         and outputs is not 0, make sure inputs have mint/melt authority.
 
         In case a token without version is provided, this method will return without raising when the
-        `enforce_htr_check` flag is set to `False`.
+        `allow_nonexistent_tokens` flag is set to `True`.
```
raul-oliveira, line 249: Since you change this variable name, I would add a docstring saying that it won't verify the htr when this flag is set to `True`.

hathor/verification/transaction_verifier.py
```diff
@@ -260,71 +257,66 @@ def verify_sum(
...
                 htr_expected_amount,
             ))
 
-    def validate_token_permissions(self, token_uid: TokenUid, token_info: TokenInfo) -> None:
-        """Verify fee token can be minted/melted based on its authority."""
+    @staticmethod
+    def _check_token_permissions(token_uid: TokenUid, token_info: TokenInfo) -> None:
+        """Verify whether token can be minted/melted based on its authority."""
+        from hathor.conf.settings import HATHOR_TOKEN_UID
+        if token_info.version == TokenVersion.NATIVE:
```
raul-oliveira, line 309: Nice catch adding this assertions!

hathor/wallet/resources/thin_wallet/send_tokens.py
```diff
@@ -217,7 +215,9 @@ def _stratum_deferred_resolve(self, context: _Context) -> None:
     def _stratum_thread_verify(self, context: _Context) -> _Context:
         """ Method to verify the transaction that runs in a separated thread
         """
-        self.manager.verification_service.verify(context.tx, self.params)
+        best_block = self.manager.tx_storage.get_best_block()
```
raul-oliveira, line 218: Just checking. Is this class a singleton? If not, why this change?

tests/nanocontracts/test_fee_tokens.py
```diff
@@ -0,0 +1,302 @@
...
+
+    @public(allow_withdrawal=True)
+    def nop(self, ctx: Context) -> None:
+        pass
+
+
+class FeeTokensTestCase(BlueprintTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        # TODO
```
raul-oliveira, line 58: Comment

tests/tx/test_fee_tokens.py
```diff
@@ -740,3 +741,25 @@ def resolve_and_propagate(self, tx: Transaction) -> None:
         self.manager.cpu_mining_service.resolve(tx)
         self.manager.propagate_tx(tx)
         self.run_to_completion()
+
+    def test_pay_fee_with_fbt(self) -> None:
```
raul-oliveira, line 745: missing docstring about fbt

tests/nanocontracts/test_fee_tokens.py
```diff
@@ -0,0 +1,302 @@
...
+        self.manager = self.create_peer('unittests', nc_indexes=True, nc_log_config=NCLogConfig.FAILED)
+        self.blueprint_id = self.gen_random_blueprint_id()
+        self.manager.tx_storage.nc_catalog.blueprints[self.blueprint_id] = MyBlueprint
+        self.dag_builder = TestDAGBuilder.from_manager(self.manager)
+
+    def test_postponed_verification_success(self) -> None:
+        artifacts = self.dag_builder.build_from_str(f'''
```
raul-oliveira, line 65: Could you add a docstring to this method explaining what "postponed verification" means in this context?

glevco (approved):

hathor/nanocontracts/balance_rules.py
glevco: Yes, accessing the subscript directly will raise an error if not found, but we expect the token to always exist (you had the assertion for that). It was just unnecessary code.

hathor/p2p/sync_v2/transaction_streaming_client.py
glevco: It's not, `default_for_mempool` doesn't accept `None` as the block.

hathor/transaction/resources/create_tx.py
glevco: How? The method doesn't return `None`

hathor/wallet/resources/thin_wallet/send_tokens.py
glevco: Since I'm now using the best block, I have to create the params on every request instead of only once for the class.

hathor/verification/verification_service.py
```diff
@@ -358,3 +377,13 @@ def _verify_without_storage_on_chain_blueprint(
...
+
+    def _get_block_storage(self, params: VerificationParams) -> NCBlockStorage:
+        assert params.best_block is not None
+        assert self._nc_storage_factory is not None
+        meta = params.best_block.get_metadata()
+        if meta.nc_block_root_id is None:
+            # This case only happens for the genesis and during sync of a voided chain.
+            assert params.best_block.is_block or meta.voided_by
+            return self._nc_storage_factory.get_empty_block_storage()
+        return self._nc_storage_factory.get_block_storage_from_block(params.best_block)
```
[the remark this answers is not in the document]
glevco: I still need the factory to call `get_empty_block_storage` on the line above.

glevco (approved):

hathor/verification/transaction_verifier.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1415/commits/3e47585c88ef1ca85cf32c844122b736cb434652

hathor/verification/verification_params.py
jansegre: I think `VerificationParams` should ideally have simple parameters that are more meaningful. My impression is that `best_block` is both a bit vague and a complex object. And ideally they should only change based on the type of verification, instead of on every new block. Although none of this is specified anywhere.

Could we get away with something like a `nc_root_id: NodeId | None` instead? Since it looks like that's the info we need from the `best_block`? Or rather, could we pass the appropriate `NCBlockStorage` directly instead of through `VerificationParams`?

tests/tx/test_fee_tokens.py
raul-oliveira: Ok. I agree

hathor/wallet/resources/thin_wallet/send_tokens.py
raul-oliveira: Got it. tks

hathor/cli/mining.py
raul-oliveira: Got it. tks

hathor/p2p/sync_v2/transaction_streaming_client.py
raul-oliveira: got it. tks

hathor/verification/verification_params.py
glevco: Sure, created here: https://github.com/HathorNetwork/hathor-core/issues/1466

I added it to the spreadsheet under the Consensus Simplify project so we don't loose track of it.

hathor/verification/verification_service.py
```diff
@@ -258,8 +262,15 @@ def _verify_tx(
...
         self.verifiers.tx.verify_sigops_input(tx, params.enable_checkdatasig_count)
         self.verifiers.tx.verify_inputs(tx)  # need to run verify_inputs first to check if all inputs exist
-        self.verifiers.tx.verify_sum(token_dict or tx.get_complete_token_info())
         self.verifiers.tx.verify_version(tx, params)
+
+        block_storage = self._get_block_storage(params)
+        self.verifiers.tx.verify_sum(
+            self._settings,
+            token_dict or tx.get_complete_token_info(block_storage),
+            # if this tx isn't a nano contract we assume we can find all the tokens to validate this tx
```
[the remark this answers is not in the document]
glevco: If it comes from a block, the `block_storage` here would be from the parent block of that block. So if it access a nonexistent token, it would be reject. This means the block itself should also be rejected.

However I don't see how this situation can arise naturally. Other nodes would observe this case and reject both the transaction and the new block, as they would both be invalid.

Do you see any other case? Maybe I could add an explicit test for this.

hathor/transaction/token_info.py
```diff
@@ -109,3 +104,26 @@ def calculate_fee(self, settings: 'HathorSettings') -> int:
...
+    """
+    Get the token version for a given token uid.
+    It searches first in the tx storage and then in the block storage.
+    """
+    from hathor.conf.settings import HATHOR_TOKEN_UID
+    if token_uid == HATHOR_TOKEN_UID:
+        return TokenVersion.NATIVE
+    from hathor.transaction.storage.exceptions import TransactionDoesNotExist
+    try:
+        token_creation_tx = tx_storage.get_token_creation_transaction(token_uid)
```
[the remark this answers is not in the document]
glevco: I believe this example is covered by case 2.ii in my explanation, but let's give names to make the example clearer:

- `tka` = the token creation vertex in the mempool of `node1`;
- `nc1` = the nano tx that has a `WITHDRAWAL` and an output of `tka`, and is both on `node1`'s and `node2`'s mempools;
- `b1` = the block that confirms only `nc1`.

> If a block (`b1`) confirms the nano tx (`nc1`) but not the token creation vertex (`tka`), what would happen?

`nc1` would be executed by `b1`. Without any other tx, `nc1` will always fail in all nodes, because the `WITHDRAWAL` will fail since `nc1` doesn't have any `tka` balance. `nc1` would be accepted as valid, executed, and failed in both `node1` and `node2`. Since `nc1` fails in all nodes, there's no fork.

Now, let's assume `nc1` _does_ have `tka` balance. This means there MUST be a previous nano tx that uses a `DEPOSIT` of `tka`, and that tx MUST be executed before `nc1`, otherwise the balance wouldn't have `tka`. Let's call it `nc0`. Then, ALL those assumptions will hold:

1. `nc0` MUST have `tka` in its inputs for the `DEPOSIT` to be successful. That is, `nc0` spends `tka`.
2. `nc0` MUST be confirmed and executed by `b1` since it has to be executed before `nc1`.
3. Since `tka` is spent by `nc0` and `nc0` is confirmed by `b1`, `tka` is also confirmed by `b1`.

Therefore, the ONLY way `nc1`'s execution can succeed is if `tka` is confirmed by `b1`.

> What if the mempool is different for another node and it doesn't have the token creation vertex? Wouldn't it cause a fork?

The conclusion above means this would be impossible because `tka` is confirmed by `b1`, so `node2` must also have it. `nc1` would succeed in both nodes and no forks would happen.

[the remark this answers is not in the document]
msbrogli: Perfect analysis. I agree. Thanks!

msbrogli (approved):

PR #1431, opened 2025-09-25
feat(nano): implement address balance in global state RFC

msbrogli:
### Motivation

Implements the RFC https://github.com/HathorNetwork/rfcs/pull/108, introducing a protocol-level global address-balance map and a `TransferHeader` that bridges UTXO balances and this shared balance domain. See the linked RFC for the full motivation, design, and rationale.

The implementation is gated behind the `TRANSFER_HEADER` feature activation bit so the new transaction shape and global-state semantics only apply once the feature is active.

### Acceptance Criteria

- A new `TransferHeader` transaction header is parseable, serializable, and included in the transaction sighash, with at most one allowed per transaction.
- `TransferHeader` verification enforces the RFC's structural rules (positive amounts, referenced `InputAddress` entries, unique `(address, token)` pairs on each side, no overlap between debit and credit sides, valid regular addresses, resolved `token_index`, script authorization).
- Transaction-level balance verification extends Hathor's per-token accounting equation to include transfer inputs as sources and transfer outputs as sinks, enabling UTXO<->global transfers, pure global-balance transfers, and caller-funded nano calls.
- Per-address replay protection via monotonic seqnums is implemented, reusing the nano-contract seqnum window constants (`MAX_SEQNUM_JUMP_SIZE` at block execution, `MAX_SEQNUM_DIFF_MEMPOOL` in the mempool).
- Block execution atomically applies transfer-header debits/credits and seqnum updates alongside nano-contract state changes, with rollback on failure; the block executor's topological sorter respects address-seqnum ordering.
- Global address balances and seqnums are persisted in the same patricia-trie structure used for nano-contract state, exposed through `block_storage` APIs (`get_address_balance`, `get_address_seqnum`).
- Nano-contract blueprint-facing syscalls `get_address_balance`, `get_address_balance_before_current_call`, and `transfer_to_address` are available, with `transfer_to_address` rejecting contract IDs and raising on insufficient treasury balance.
- When a transaction carries both `TransferHeader` and `NanoHeader` with non-empty transfer inputs, only the caller address may be debited and all `TxTransferInput`s must reference `address_index = 0`.
- `GET /thin_wallet/address_balance` returns the address's global balances per token and current global seqnum alongside the existing UTXO-derived `tokens_data`.
- The feature is behind the `TRANSFER_HEADER` feature-activation bit.

### Checklist

- [ ] RFC https://github.com/HathorNetwork/rfcs/pull/108 is merged


hathor/nanocontracts/storage/token_proxy.py
```diff
@@ -51,3 +51,6 @@ def create_token(
             token_symbol=token_symbol,
             token_version=token_version
         )
+
+    def add_address_balance(self, address: Address, amount: Amount, token_id: TokenUid) -> None:
```
msbrogli: Missing docstring.

hathor/nanocontracts/blueprint_env.py
```diff
@@ -266,3 +267,7 @@ def setup_new_contract(
             actions=actions,
             fees=fees or (),
         )
+
+    def transfer_to_address(self, address: Address, amount: Amount) -> None:
```
msbrogli: What about the token id?

hathor_tests/tx/test_transfer_header.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_dag_builder_emits_transfer_header_only_when_configured(self) -> None:
+        manager = self.create_peer('unittests')
+        dag_builder = TestDAGBuilder.from_manager(manager)
+        artifacts = dag_builder.build_from_str('''
+            blockchain genesis b[1..5]
+
+            tx1.out[0] = 100 HTR
+            tx2.out[0] = 100 HTR
+            tx2.nc_transfer_input = 10 HTR main
+            tx2.nc_transfer_output = 10 HTR main
```
msbrogli: Should we allow an input and an output of the same token? I guess not.

msbrogli: I just checked the verifier and it is not allowed. How's this test passing? Are they using different addresses?

hathor_tests/tx/test_transfer_header.py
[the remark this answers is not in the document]
jansegre: I've added execution fail tests. Do you mean "the transfer should *not* be persisted as well"? A transfer that tries to move more than the available balance should fail without affecting the balance, no?

jansegre:
Replaced by this sequence:
- #1652
- #1653
- #1654
- #1655 
- #1656

PR #1446, opened 2025-10-06
feat(sync-v2): Orphan blocks handling

msbrogli:
### Motivation

This PR improves the way sync-v2 works with orphan blocks. It also introduces an important change in the consensus: the transactions do NOT affect the block score anymore. In other words, the block score depends exclusively on the blockchain, regardless of how many transactions are confirmed by those blocks.

### Open questions

- This implementation currently discards a chain of orphan blocks that doesn't have enough score. But sync-v2 will keep downloading the same orphan chain over and over. Should we prevent this in this PR?

- Instead of removing transactions in the block score calculation, we could allow the peer to provide an untrusted transactions score but limit it to the score of 10,000 transactions at their peak weight (~35).

### Acceptance Criteria

1. On sync-v2 block streaming client, only add non-orphan blocks to the storage. All orphan blocks are kept in the partial block list.
2. Only add orphan blocks if this orphan chain has at least 95% of the best block score. Otherwise, discard the whole chain of orphan blocks.
3. Remove transactions from the block score calculation. Now, only block weights are used to calculate a score.

### Checklist

### DIFF
--- a/hathor/consensus/block_consensus.py
+++ b/hathor/consensus/block_consensus.py
@@ -369,56 +369,7 @@ class BlockConsensusAlgorithm:
                                                          is_dag_verifications=False)
 
     def update_voided_info(self, block: Block) -> None:
-        """ This method is called only once when a new block arrives.
-
-        The blockchain part of the DAG is a tree with the genesis block as the root.
-        I'll say the a block A is connected to a block B when A verifies B, i.e., B is a parent of A.
-
-        A chain is a sequence of connected blocks starting in a leaf and ending in the root, i.e., any path from a leaf
-        to the root is a chain. Given a chain, its head is a leaf in the tree, and its tail is the sub-chain without
-        the head.
-
-        The best chain is a chain that has the highest score of all chains.
-
-        The score of a block is calculated as the sum of the weights of all transactions and blocks both direcly and
-        indirectly verified by the block. The score of a chain is defined as the score of its head.
-
-        The side chains are the chains whose scores are smaller than the best chain's.
-        The head of the side chains are always voided blocks.
-
-        There are two possible states for the block chain:
-        (i)  It has a single best chain, i.e., one chain has the highest score
-        (ii) It has multiple best chains, i.e., two or more chains have the same score (and this score is the highest
-             among the chains)
-
-        When there are multiple best chains, I'll call them best chain candidates.
-
-        The arrived block can be connected in four possible ways:
-        (i)   To the head of a best chain
-        (ii)  To the tail of the best chain
-        (iii) To the head of a side chain
-        (iv)  To the tail of a side chain
-
-        Thus, there are eight cases to be handled when a new block arrives, which are:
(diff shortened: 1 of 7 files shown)


hathor/p2p/sync_v2/agent.py
```diff
@@ -442,6 +464,13 @@ def run_sync_blocks(self) -> Generator[Any, Any, bool]:
         self._tx_streaming_client = None
         return False
 
+    def get_partial_blocks_score(self, partial_blocks: list[Block]) -> int:
+        # XXX Handle when this block does not exist.
```
glevco: TODO?

tests/consensus/test_soft_voided.py
```diff
@@ -11,7 +11,7 @@
 
 
 class SoftVoidedTestCase(SimulatorTestCase):
-    seed_config = 5988775361793628170
+    seed_config = 2
```
glevco: Why this change?

hathor/p2p/sync_v2/agent.py
glevco: > This implementation currently discards a chain of orphan blocks that doesn't have enough score. But sync-v2 will keep downloading the same orphan chain over and over. Should we prevent this in this PR?

If it's easy enough I think it could be done in this PR. Otherwise, I'm fine doing it on a separate one.

hathor/consensus/block_consensus.py
glevco: > Instead of removing transactions in the block score calculation, we could allow the peer to provide an untrusted transactions score but limit it to the score of 10,000 transactions at their peak weight (~35).

Could you elaborate on this, please?

hathor/consensus/block_consensus.py
```diff
@@ -835,7 +835,7 @@ def _score_block_dfs(self, block: BaseTransaction, used: set[bytes],
                         meta.first_block = block.hash
                         self.context.save(tx)
 
-                    score += weight_to_work(tx.weight)
+                    # score += weight_to_work(tx.weight)
```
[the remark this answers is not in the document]
glevco: Or rather, remove.

hathor/p2p/sync_v2/agent.py
```diff
@@ -424,6 +425,27 @@ def run_sync_blocks(self) -> Generator[Any, Any, bool]:
 
         assert self._blk_streaming_client is not None
         partial_blocks = self._blk_streaming_client._partial_blocks
+        if partial_blocks:
+            score = self.get_partial_blocks_score(partial_blocks)
+            best_block = self.tx_storage.get_best_block()
+            best_block_score = best_block.get_metadata().score
+            threshold = best_block_score * 0.95
```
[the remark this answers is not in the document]
msbrogli: Done! I originally introduced this threshold to factor in the scores of transactions confirmed by orphan chains. However, since those are no longer included in the chain score calculation, we could bump it up to 100%.

PR #1472, opened 2025-10-16
feat(nano): add support for proxy view calls

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1471

### Motivation

Proxy view calls are currently not supported, but there's no reason to not allow them.

### Acceptance Criteria

- Add support for call to view methods via proxies.
  - This means the `get_proxy` syscall is now allowed on view methods.
  - A minor refactor in the Runner was also necessary.

### Checklist


hathor/nanocontracts/proxy_accessor.py
msbrogli: Shouldn't `ProxyAcessor` forbid view methods from calling public methods?

jansegre (approved):

hathor/nanocontracts/proxy_accessor.py
msbrogli: Great! Thanks!

PR #1483, opened 2025-11-03
Feat(dynamic-transaction-model): Bump python-hathorlib, and fix the supported headers list in the vertex parser

raul-oliveira:
### Motivation
The `push_tx` API isn't supporting the fee header due to the absence of FeeHeader in the `supported_headers` list.

### Trace
The vertex parser used by `cls.create_from_struct` (with cls being Tx/TokenCreationTx) when deserializing a transaction that arrived through the `push_tx` API.

### Acceptance Criteria

- Push tx should accept a token creation transaction creating a fee based token `(TokenVersion.FEE = 2)`
- Bump the python-hathorlib version to `0.14.0` that supports the FeeHeader. 
https://github.com/HathorNetwork/python-hathorlib/pull/69

### Checklist

- [X] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged


glevco (dismissed):

hathor/transaction/vertex_parser.py
```diff
@@ -37,6 +37,7 @@ def get_supported_headers(settings: HathorSettings) -> dict[VertexHeaderId, Type
         supported_headers: dict[VertexHeaderId, Type[VertexBaseHeader]] = {}
         if settings.ENABLE_NANO_CONTRACTS:
             supported_headers[VertexHeaderId.NANO_HEADER] = NanoHeader
+            supported_headers[VertexHeaderId.FEE_HEADER] = FeeHeader
```
msbrogli: We can't merge this change otherwise it could cause a hark fork in the network.

hathor/nanocontracts/runner/runner.py
```diff
@@ -1205,6 +1207,10 @@ def syscall_create_child_fee_token(
         fee_payment_token: TokenUid
     ) -> TokenUid:
         """Create a child fee token from a contract."""
+
+        if not self.is_fee_active:
+            raise NCInvalidSyscall('fee feature is not active')
```
glevco: I don't think we need this, it's not currently blocked on mainnet. It's blocked by the fact that we review OCBs. After the fees feature activation is activated, we'll simply start accepting those OCBs.

Then you'll be able to simplify some code because the runner will not need this attribute.

glevco: @msbrogli @jansegre what do you think?

glevco: In fact if we accidentally approve an OCB with this syscall, this change would cause a hard fork.

hathor/transaction/vertex_parser.py
raul-oliveira: Now It's under a conditional. Pls check the updated code :)

hathor/conf/settings.py
```diff
@@ -49,14 +49,14 @@ class NanoContractsSetting(StrEnum):
...
         False
-        >>> bool(NanoContractsSetting.ENABLED)
+        >>> bool(FeatureSetting.ENABLED)
         True
-        >>> bool(NanoContractsSetting.FEATURE_ACTIVATION)
+        >>> bool(FeatureSetting.FEATURE_ACTIVATION)
         True
         """
-        return self in (NanoContractsSetting.ENABLED, NanoContractsSetting.FEATURE_ACTIVATION)
+        return self in (type(self).ENABLED, type(self).FEATURE_ACTIVATION)
```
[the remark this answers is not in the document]
raul-oliveira: Done

hathor/consensus/consensus.py
```diff
@@ -428,8 +428,8 @@ def _unknown_contract_mempool_rule(self, tx: Transaction) -> bool:
 
     def _nano_activation_rule(self, storage: TransactionStorage, tx: Transaction) -> bool:
```
glevco: You need to update this method, it currently assumes fees are activated by the nano feature. You have to remove this logic and create an analogous method for the fee rule.

hathor/transaction/vertex_parser.py
msbrogli: Thanks!

glevco (approved):

PR #1494, opened 2025-11-24
feat(nano): emit a token-creation event including for nano executions

jansegre:
### Motivation

Currently there is no way for a wallet observing events from the reliable integration to know when a token has been created from a nano execution.

~Introduce two new events: `TOKEN_CREATED` and `TOKEN_REMOVED` (for reorgs).~

Introduce a new event: `TOKEN_CREATED`. Removal can be inferred from metadata changes.

### Acceptance Criteria

- Emit `TOKEN_CREATED` for both `TokenCreationTransaction`s and nano-contract executions, including token metadata and optional nc_exec_info.
- ~Emit `TOKEN_REMOVED` only when a token-creation tx is voided, and for nano-created tokens when their execution is rolled back (reorg/un-execution).~
- EventArguments and event data models updated with explicit token fields; tests assert the emitted event payloads match expected metadata.

### Checklist

### DIFF
--- a/hathor/event/event_manager.py
+++ b/hathor/event/event_manager.py
@@ -23,9 +23,13 @@ from hathor.event.model.node_state import NodeState
 from hathor.event.storage import EventStorage
 from hathor.event.websocket import EventWebsocketFactory
 from hathor.execution_manager import ExecutionManager
+from hathor.nanocontracts.runner.index_records import CreateTokenRecord
 from hathor.pubsub import EventArguments, HathorEvents, PubSubManager
 from hathor.reactor import ReactorProtocol as Reactor
-from hathor.transaction import BaseTransaction
+from hathor.transaction import BaseTransaction, Transaction
+from hathor.transaction.nc_execution_state import NCExecutionState
+from hathor.transaction.token_creation_tx import TokenCreationTransaction
+from hathor.transaction.transaction_metadata import TransactionMetadata
 from hathor.util import not_none, progress
 from hathor.utils.iter import batch_iterator
 
@@ -48,6 +52,7 @@ _SUBSCRIBE_EVENTS = [
     HathorEvents.CONSENSUS_TX_UPDATE,
     HathorEvents.CONSENSUS_TX_REMOVED,
     HathorEvents.NC_EVENT,
+    HathorEvents.NC_EXEC_SUCCESS,
 ]
 
 
@@ -170,9 +175,15 @@ class EventManager:
 
     def _handle_hathor_event(self, hathor_event: HathorEvents, event_args: EventArguments) -> None:
         """Handles a PubSub 'HathorEvents' event."""
-        event_type = EventType.from_hathor_event(hathor_event)
 
-        self._handle_event(event_type, event_args)
+        event_type = EventType.from_hathor_event(hathor_event)
+        if event_type is not None:
+            self._handle_event(event_type, event_args)
+
+        if hathor_event == HathorEvents.NETWORK_NEW_TX_ACCEPTED:
+            self._handle_token_creation_events(event_args)
+        elif hathor_event == HathorEvents.NC_EXEC_SUCCESS:
+            self._handle_nc_token_creation_events(event_args)
 
     def _handle_event(self, event_type: EventType, event_args: EventArguments) -> None:
         """Handles an Event Queue feature 'EventType' event."""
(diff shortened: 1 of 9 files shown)


hathor/event/event_manager.py
```diff
@@ -240,6 +251,137 @@ def _create_non_group_edge_event(self, event_type: EventType, event_args: EventA
...
+            meta = tx.get_metadata()
+            if meta.nc_execution == NCExecutionState.SUCCESS:
+                self._emit_nc_token_created_events(tx, meta)
+
+    @staticmethod
+    def _calculate_token_creation_initial_amount(tx: TokenCreationTransaction) -> int:
+        """Calculate the amount minted by a token-creation transaction."""
+        assert tx.hash is not None
+        total = 0
+        for tx_output in tx.outputs:
```
msbrogli: Since it can have a NanoHeader, what about deposits, transferring token authority to the contract and internal minting?

hathor/nanocontracts/runner/index_records.py
```diff
@@ -94,6 +95,47 @@ def from_json(cls, json_dict: dict[str, Any]) -> Self:
         )
 
 
+@dataclass(slots=True, frozen=True, kw_only=True)
+class RemoveTokenRecord:
```
msbrogli: If it's only after a reorg, why does the runner have to do anything?

hathor/event/event_manager.py
```diff
@@ -240,6 +251,137 @@ def _create_non_group_edge_event(self, event_type: EventType, event_args: EventA
...
+        tx = getattr(event_args, 'tx', None)
+        if tx is None:
+            return
+
+        if isinstance(tx, Transaction) and tx.is_nano_contract():
+            meta = tx.get_metadata()
+            self._emit_nc_token_removed_events(tx, meta)
+
+    def _handle_token_metadata_update(self, event_args: EventArguments) -> None:
+        """Emit token create/remove events when nano-contract metadata changes, and removals for TCTs."""
```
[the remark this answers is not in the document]
jansegre: `TokenCreationTransaction`s

hathor/nanocontracts/runner/index_records.py
glevco: I believe a more natural approach, more analogous to what we currently have, would be to remove this `RemoveTokenRecord`, I also thought it was a bit weird.

When a nano in unexecuted and therefore its token creations are removed, the event for the nc tx voidance will be emitted and processed by a client. Then it would be the client responsibility to "void" (remove) all tokens that were created by that voided nc tx.

hathor/event/event_manager.py
jansegre: I assumed they couldn't, and tracking changes to the token is outside of the scope of this event.

hathor_tests/event/test_event_manager.py
```diff
@@ -85,3 +109,107 @@ def test_cannot_finish_group_twice(self) -> None:
...
+        self.assertEqual(token_data.token_symbol, 'DBT')
+        self.assertEqual(token_data.initial_amount, 100)
+        self.assertEqual(token_data.token_version, TokenVersion.DEPOSIT)
+
+        nc_exec_info = not_none(token_data.nc_exec_info)
+        self.assertEqual(nc_exec_info.nc_tx, call_tx.hash_hex)
+        call_meta = call_tx.get_metadata()
+        self.assertIsNotNone(call_meta.first_block)
+        if call_meta.first_block is not None:
+            self.assertEqual(nc_exec_info.nc_block, call_meta.first_block.hex())
```
glevco: In what case would it be `None`?

glevco (changes requested):

hathor/event/event_manager.py
jansegre: > I believe only deposits could, because the initial amount could be partially (or totally) moved to deposits, so the outputs would not reflect the total initial amount.

Wouldn't that be considered 2 different operations in a single transactions, that is:

1. token is created with "initial amount"
2. some of that amount is moved to deposits

Or did I get that wrong? @glevco

glevco: That's correct, but exactly because of that `initial amount = outputs + deposits` , and you're doing `initial amount = outputs`

For example if 100 tokens are created and all 100 tokens are deposited, your initial amount will be 0

hathor_tests/event/test_event_manager.py
jansegre: True, no case, I'll fix the test.

glevco (changes requested):

hathor/event/event_manager.py
```diff
@@ -240,6 +251,74 @@ def _create_non_group_edge_event(self, event_type: EventType, event_args: EventA
...
+            )
+
+            self._handle_event(EventType.TOKEN_CREATED, token_event_args)
+
+    def _handle_nc_token_creation_events(self, event_args: EventArguments) -> None:
+        """Emit NC token-created events when a contract execution succeeds."""
+        tx = getattr(event_args, 'tx', None)
+        if isinstance(tx, Transaction) and tx.is_nano_contract():
+            meta = tx.get_metadata()
+            if meta.nc_execution == NCExecutionState.SUCCESS:
```
glevco: This `if` should be an assertion now, right?

glevco (approved):

hathor/event/event_manager.py
jansegre: Issue created: #1503

PR #1544, opened 2026-01-08
feat(p2p): Add new whitelist policy

msbrogli:
Co-Author: LFRezende <lfsrsprofessional@gmail.com>

### Motivation

The previous peer whitelist implementation was tightly coupled across `HathorManager`, `ConnectionsManager`, and `HathorSettings`, mixing concerns and making it difficult to change the whitelist source (URL, local file) or policy at runtime. The whitelist was controlled by a global `ENABLE_PEER_WHITELIST` setting with special-case logic for sync-v1 vs. sync-v2, and the peer ID list lived on `HathorManager`. This refactor extracts the whitelist into a self-contained, pluggable module (`hathor.p2p.whitelist`) that supports multiple backends (URL-based, file-based), configurable policies (`allow-all` vs. `only-whitelisted-peers`), runtime switching via sysctl, and a grace period for bootstrap peers before the first successful fetch completes. It also removes the now-unused `SyncVersion.V1_1` variant and the `ENABLE_PEER_WHITELIST` setting.

### Acceptance Criteria

- The whitelist is modeled as an abstract `PeersWhitelist` base class with two concrete implementations: `URLPeersWhitelist` (fetches from a remote URL) and `FilePeersWhitelist` (reads from a local file), both supporting periodic refresh with exponential backoff on failure

- Whitelist files support an optional `policy:` directive that controls connection behavior: `allow-all` permits any peer, `only-whitelisted-peers` (default) restricts connections to listed peer IDs only

- A factory function (`create_peers_whitelist`) and CLI argument `--x-p2p-whitelist` allow selecting the whitelist source at startup: `default`/`hathorlabs` (uses settings URL), `none`/`disabled` (no whitelist), a file path, or a custom URL

- Before the first successful whitelist fetch, a grace period allows only bootstrap-discovered peers to connect, preventing connections to arbitrary peers while the whitelist is still loading

- The `ENABLE_PEER_WHITELIST` setting and `SyncVersion.V1_1` are removed; whitelist capability is now always required in the HELLO handshake, and the whitelist is enabled/disabled purely by whether a `PeersWhitelist` instance is provided to `ConnectionsManager`

- Whitelist state is owned by `ConnectionsManager.peers_whitelist` instead of `HathorManager.peers_whitelist`, and `peer_id.py`'s `_is_peer_allowed` simply delegates to the whitelist object

- The sysctl interface exposes `whitelist` (get/set the active whitelist source, or toggle `on`/`off` to suspend/resume) and `whitelist.status` (returns structured state, policy, peer count, and source)

- When a new whitelist is set at runtime, peers not in the updated list are immediately disconnected

### Checklist


hathor/p2p/whitelist/parsing.py
```diff
@@ -0,0 +1,115 @@
...
+2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367
+''')
+    ({'2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367'}, WhitelistPolicy.ONLY_WHITELISTED_PEERS)
+    """
+    lines = parse_file(text, header=header)
+    policy = WhitelistPolicy.ONLY_WHITELISTED_PEERS  # default
+
+    peer_lines: list[str] = []
+    for line in lines:
+        if line.startswith('policy:'):
```
msbrogli: Fix it: Only in the header.

hathor/p2p/manager.py
```diff
@@ -273,7 +265,13 @@ def do_discovery(self) -> None:
         Do a discovery and connect on all discovery strategies.
         """
         for peer_discovery in self.peer_discoveries:
-            coro = peer_discovery.discover_and_connect(self.connect_to_endpoint)
+            # Wrap connect_to_endpoint to register bootstrap peer IDs
+            def connect_with_bootstrap_registration(entrypoint: PeerEndpoint) -> None:
+                if self.peers_whitelist and entrypoint.peer_id is not None:
+                    self.peers_whitelist.add_bootstrap_peer(entrypoint.peer_id)
+                self.connect_to_endpoint(entrypoint)
```
LFRezende: At the beginning of the bootstrap, when peers_whitelist = set(), each entrypoint provided for bootstrapping peers will connect, but it will not add the peer_id of that entrypoint to peers_whitelist.

peers_whitelist AND entrypoint.peer_id is not None --> At the beginning, peers_whitelist == None, hence bootstrap will not add peerId to the set, hence none will be added, just have its entrypoint connected via connect_to_endpoint.

hathor/p2p/manager.py
```diff
@@ -935,3 +880,25 @@ def reload_entrypoints_and_connections(self) -> None:
...
+        """Disconnect all connected peers that are not in the current whitelist."""
+        if not self.peers_whitelist:
+            return
+        self.log.info('Whitelist ON: disconnecting non-whitelisted peers...')
+        connections_snapshot = list(self.connections)
+        for conn in connections_snapshot:
+            peer_id = conn.get_peer_id()
+            if peer_id is None:
+                continue
+            if not self.peers_whitelist.is_peer_whitelisted(peer_id):
```
LFRezende: Small suggestion: perhaps whitelist = self.peers_whitelist would make the function less verbose.

--> if not whitelist;
--> if not whitelist.is_peer_whitelisted(peer_id)

hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +304,66 @@ def refresh_auto_hostname(self) -> None:
...
+            if self._suspended_whitelist is None:
+                return
+            self.connections.set_peers_whitelist(self._suspended_whitelist)
+            self._suspended_whitelist = None
+            return
+        if option == 'off':
+            if self.connections.peers_whitelist is None:
+                return
+            self._suspended_whitelist = self.connections.peers_whitelist
+            self.connections.set_peers_whitelist(None)
```
LFRezende: Small nitpick: perhaps connections = self.connections would reduce verbose, as it is a term used frequently.

LFRezende (commented):
Review - Luis Felipe/LFRezende

**Core:** 

1. Possible oversight which does not add bootstrap peers  at connect_with_bootstrap_registration, lines 268-275
2. Discussion and further recommendation to delete --x-p2p-whitelist-only flag.
3. Discussion on expanding mechanism of suspended_peers.

The rest are smaller refactor suggestions.

hathor/p2p/whitelist/peers_whitelist.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def current_whitelist(self) -> set[PeerId]:
+        """ Returns the current whitelist as a set of PeerId."""
+        return self._current
+
+    def policy(self) -> WhitelistPolicy:
+        """ Returns the current whitelist policy."""
+        return self._policy
+
+    def is_peer_whitelisted(self, peer_id: PeerId) -> bool:
```
jansegre: Nitpick: should we just call this `is_peer_allowed` to mirror `PeerIdState._is_peer_allowed`, since the method is not exactly concerned with the peer actually being on the whitelist, since that's only the case for the `ONLY_WHITELISTED_PEERS` policy?

jansegre (changes requested):
The only relevant points I have are:

- the policy syntax, maybe use `#policy: allow-all` instead
- does the behavior change for testnets/devnets? I think it shouldn't but I couldn't follow the path to reach that conclusion from review alone without having to test this branch.

hathor/p2p/whitelist/peers_whitelist.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def _handle_refresh_err(self, *args: Any, **kwargs: Any) -> None:
+        """This method will be called when an exception happens inside the whitelist update
+           and ends up stopping the looping call.
+           We log the error and start the looping call again with exponential backoff.
+        """
+        self._on_update_failure()
+        retry_interval = self._get_retry_interval()
+        self.log.error(
+            'whitelist refresh had an exception. Start looping call again.',
```
luislhl: question: do we get any detail about why the refresh failed? Or will it just ignore the exception and keep trying?

Having details about it could help in case of recurrent errors.

hathor/p2p/whitelist/peers_whitelist.py
msbrogli: Done! Thanks!

hathor/sysctl/p2p/manager.py
```diff
@@ -269,3 +304,66 @@ def refresh_auto_hostname(self) -> None:
...
+            return
```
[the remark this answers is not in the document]
msbrogli: Blacklisting can help when a specific full node is misbehaving. However, it’s not very effective against targeted attacks. Generating new peer IDs is extremely cheap, so an attacker can simply create a new identity and reconnect to your node almost immediately.

hathor/p2p/whitelist/peers_whitelist.py
```diff
@@ -0,0 +1,177 @@
...
+class PeersWhitelist(ABC):
+    def __init__(self, reactor: Reactor) -> None:
+        self.log = logger.new()
+        self._reactor = reactor
+        self.lc_refresh = LoopingCall(self.update)
+        self.lc_refresh.clock = self._reactor
+        self._current: set[PeerId] = set()
+        self._policy: WhitelistPolicy = WhitelistPolicy.ONLY_WHITELISTED_PEERS
+        self._on_remove_callback: OnRemoveCallbackType = None
+        self._is_running: bool = False
```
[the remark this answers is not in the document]
msbrogli: Done! Thanks!

luislhl (dismissed):

jansegre (dismissed):

hathor/p2p/whitelist/factory.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if spec_lower in (WHITELIST_SPEC_DEFAULT, WHITELIST_SPEC_HATHORLABS):
+        peers_whitelist = URLPeersWhitelist(reactor, str(settings.WHITELIST_URL), True)
+    elif spec_lower in (WHITELIST_SPEC_NONE, WHITELIST_SPEC_DISABLED):
+        peers_whitelist = None
+    elif os.path.isfile(whitelist_spec):
+        peers_whitelist = FilePeersWhitelist(reactor, whitelist_spec)
+    else:
+        # URLPeersWhitelist class rejects non-url paths.
+        peers_whitelist = URLPeersWhitelist(reactor, whitelist_spec, True)
```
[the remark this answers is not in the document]
msbrogli: Done! Can you review it again, please? It's in a separate commit.

luislhl (dismissed):

hathor/p2p/manager.py
LFRezende: Perhaps we could do something like:

```
def connect_with_bootstrap_registration(entrypoint: PeerEndpoint) -> None:

         if not self.peers_whitelist:
              return # Or raise exception

         if entrypoint.peer_id is not None:
                self.peers_whitelist.add_bootstrap_peer(entrypoint.peer_id)
                self.connect_to_endpoint(entrypoint)
```

It ensures that, if no whitelist, there will be no connection, since there are no peers to bootstrap to.
And if there is, check whether they yield an entrypoint peer id. If so, then, and only then shall you add them to the bootstrap peer.

If you want only to connect to entrypoints with a peer_id, then connect to them.

If not, just put `self.connect_to_endpoint(entrypoint)` out of the if clause.

Makes sense?

LFRezende (dismissed):
Last changes on bootstrap peers and their derived tests seem reasonable.

Ran linter and tests locally:
1. All linters work fine 
2. Two tests failed, and the last one seems to be connected to bootstrap (already communicated msbrogli).

It seems to be flaking - I'll approve.

jansegre (dismissed):

hathor/p2p/whitelist/factory.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    peers_whitelist: PeersWhitelist | None = None
+    spec_lower = whitelist_spec.lower().strip()
+
+    if spec_lower in (WHITELIST_SPEC_DEFAULT, WHITELIST_SPEC_HATHORLABS):
+        peers_whitelist = URLPeersWhitelist(reactor, str(settings.WHITELIST_URL), True)
+    elif spec_lower in (WHITELIST_SPEC_NONE, WHITELIST_SPEC_DISABLED):
+        peers_whitelist = None
+    elif _looks_like_url(whitelist_spec):
+        peers_whitelist = URLPeersWhitelist(reactor, whitelist_spec, True)
```
glevco: Why are you passing `mainnet=True` in the constructors? Also it would be better if those args were kw only.

hathor/p2p/whitelist/parsing.py
```diff
@@ -0,0 +1,131 @@
...
+# This whitelist only allows specific peers
+# policy: only-whitelisted-peers
+2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367
+''')
+    ({'2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367'}, WhitelistPolicy.ONLY_WHITELISTED_PEERS)
+
+    parse_whitelist_with_policy('''hathor-whitelist
+2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367
+''')
+    ({'2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367'}, WhitelistPolicy.ONLY_WHITELISTED_PEERS)
```
glevco: Use doctests.

hathor/p2p/whitelist/parsing.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    if header is None:
+        header = 'hathor-whitelist'
```
glevco: Move to a constant.

hathor/p2p/whitelist/url_whitelist.py
```diff
@@ -0,0 +1,123 @@
...
+            self._url = None
+            return
+
+        result = urlparse(self._url)
+        if mainnet:
+            if result.scheme != 'https':
+                raise ValueError(f'invalid scheme: {self._url}')
+
+            if not result.netloc:
+                raise ValueError(f'invalid url: {self._url}')
```
glevco: This should be checked even for non-mainnet, right?

glevco (changes requested):

hathor/p2p/resources/status.py
```diff
@@ -99,7 +99,7 @@ def render_GET(self, request):
         if whitelist is not None:
             peer_ids = whitelist.current_whitelist()
         else:
-            peer_ids = set()
+            peer_ids = frozenset()
```
LFRezende: If we freeze the set, how can we update it in sysctl?

LFRezende (dismissed):

.gitignore
```diff
@@ -30,3 +30,6 @@ keys.json
 
 # Pycharm
 .idea
+
```
jansegre, line 33: Suggestion: use `.git/info/exclude` for personal tools. Not a blocker, since we do have other setup specific ignores.

jansegre (dismissed):

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,467 @@
...
+
+### CLI
+
+The whitelist is configured via a single CLI argument:
+
+```
+--p2p-whitelist-source <spec>
+```
+
+**RULE-01**: When `--p2p-whitelist-source` is omitted (or its value is falsy),
```
msbrogli: I don't like to state where in the code it is going to happen. I guess here it would be something like "When it is omitted (or falsy), it should be treat as `default`".

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,467 @@
...
+the builder MUST substitute the literal string `default`
+(`hathor_cli/builder.py`).
+
+**RULE-02**: The `<spec>` value MUST be resolved case-insensitively by
+`create_peers_whitelist` (`hathor/p2p/whitelist/factory.py`) in the following
```
msbrogli: Same here. Do not force `create_peers_whitelist` to do it. State how it should behave.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,467 @@
...
+order:
+
+| `<spec>` value (lowercased) | Resolution |
+|---|---|
+| `default` or `hathorlabs` | If `settings.WHITELIST_URL` is `None`, the whitelist MUST be disabled (return `None`). Otherwise construct `URLPeersWhitelist(reactor, settings.WHITELIST_URL, mainnet=True)`. |
+| `none` or `disabled` | Whitelist MUST be disabled — all peers allowed. |
+| A value for which `os.path.isfile(<spec>)` is `True` | Construct `FilePeersWhitelist(reactor, <spec>)`. |
+| Anything else | Construct `URLPeersWhitelist(reactor, <spec>, mainnet=True)`. |
+
+**RULE-03**: `URLPeersWhitelist` with `mainnet=True` MUST reject the URL with
```
msbrogli: These rules are too close to the code. They should be general rules that are enforced "regardless of how it is implemented (which will be covered by the RFCs)". Some architectural decisions might be in the spec, though.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,467 @@
...
+Note: the P2P handshake capability `CAPABILITY_WHITELIST` (`'whitelist'`,
+advertised in HELLO) is a separate mechanism and is not part of this
+feature's enforcement path. It is flagged here only because the naming
+overlaps; see Open Questions.
+
+### Constants
+
+| Constant | Value | Defined in | Purpose |
+|---|---|---|---|
+| `WHITELIST_REFRESH_INTERVAL` | 30 s | `peers_whitelist.py` | Period of the refresh `LoopingCall`. |
```
msbrogli: What's the LoopingCall? The intent matters more than the detail. LoopingCall is probably refreshing the whitelist. So the purpose should be "Refresh period for the whitelist".

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Data format
+
+**RULE-05**: Whitelist content (whether served over HTTP or read from a file)
+MUST conform to:
+
+```
+hathor-whitelist
+# Optional comment
+policy: <policy-type>
```
msbrogli: That's wrong: `# policy: <policy-type>` (for compatibility)

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,467 @@
...
+<peer-id-hex>
+<peer-id-hex>
+...
+```
+
+**RULE-06**: The first line MUST be the literal string `hathor-whitelist`. If
+the header does not match, `parse_file` MUST raise `ValueError('invalid header')`.
```
msbrogli: You don't have to say which exception will be raised unless it is essential.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+the node MUST behave as if `default` was passed.
+
+**RULE-02**: The `<spec>` value MUST be resolved case-insensitively, in the
+following order:
+
+| `<spec>` value (lowercased) | Resolution |
+|---|---|
+| `default` or `hathorlabs` | Use the URL from `settings.WHITELIST_URL`. If `WHITELIST_URL` is unset, the whitelist MUST be disabled (all peers allowed). |
+| `none` or `disabled` | Whitelist MUST be disabled. |
+| A value that resolves to an existing local file | Use a file-based whitelist reading from that path. |
```
msbrogli, line 43: Actually, let's use URI. Examples: `file:///absolute/path/to/file` for files, `http://url/` or `https://url/` for URLs. Any other format generates an error and stops the full node initialization.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+| Anything else | Treat the value as a URL. |
+
+**RULE-03**: On mainnet, URL sources MUST use the `https` scheme and MUST have
```
msbrogli, line 46: Let's remove this rule.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+been accumulated MUST cause the content to be rejected.
+
+**RULE-09**: Valid policy values (case-insensitive) are:
+
+| Value | Behavior |
+|---|---|
+| `only-whitelisted-peers` | Default when no `policy:` line is present. Only peers in the list are accepted. |
+| `allow-all` | All peers are accepted regardless of list membership. |
+
+Any other policy value MUST cause the content to be rejected.
```
msbrogli, line 107: In this case, what happens with the node? Does it allow all connections? Or ignore the new whitelist and keeps the current old? What if this happens on the first load?

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+### State transitions
+
+| From | Event | To | Side effects |
+|---|---|---|---|
+| GRACE_PERIOD | Fetch succeeds | ENFORCING | Apply whitelist; drop bootstrap-peer state; reset backoff. |
+| GRACE_PERIOD | Fetch fails | GRACE_PERIOD | Increment failure counter. Keep bootstrap-only admission. |
+| ENFORCING | Fetch succeeds | ENFORCING | Apply new whitelist (diff, disconnect removed peers); reset backoff. |
+| ENFORCING | Fetch fails | ENFORCING | Preserve the active whitelist and policy. Increment failure counter. |
+| ENFORCING / GRACE_PERIOD | Refresh mechanism crashes | same (restarted after backoff) | Restart scheduled after `min(BASE * 2^failures, MAX)` (see RULE-14). |
+| ENFORCING | Sysctl `off` | SUSPENDED | Whitelist retained for later restore; enforcement paused. |
```
msbrogli, line 134: Should this whitelist keep refreshing? I guess not but we need good rules when it is activated again.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+| SUSPENDED | Sysctl `on` | ENFORCING | Enforcement resumes; currently-connected non-whitelisted peers MUST be disconnected immediately (see DEC-07). |
```
msbrogli, line 135: What if the whitelist is outdated? Should it fetch successfully at least once before applying? Or should sysadmins choose when to apply the whitelist?

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+| Any | Sysctl set URL/path | GRACE_PERIOD → ENFORCING | Old whitelist stopped; any suspended whitelist discarded; new whitelist begins refreshing. |
+
+### Startup
+
+**RULE-11**: When the node starts and a whitelist is configured, the whitelist
+MUST begin refreshing periodically and MUST be registered with a
+disconnect-on-removal callback that drops any peer removed from the list.
```
msbrogli, line 142: Why does it have to be a callback? Isn't it better to just state that, after a refresh, peers removed from the list MUST be disconnected?

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+
+**RULE-12**: During peer discovery, every entrypoint whose peer ID is known
+MUST be registered as a bootstrap peer so the node can connect to it during
+the grace period.
+
+### Periodic refresh
+
+**RULE-13**: The whitelist MUST be refreshed every `WHITELIST_REFRESH_INTERVAL`.
+Refreshes MUST NOT overlap: if a refresh is triggered while another is still
+in progress, the new invocation MUST be skipped (with a warning log) and MUST
```
msbrogli, line 152: Do we need a warning? I guess not.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+| 2 | 120 s |
+| 3 | 240 s |
+| 4+ | 300 s (cap) |
+
+**RULE-15**: A successful refresh MUST reset the consecutive-failure counter
+to 0.
+
+### Shutdown
+
+**RULE-16**: When the node shuts down, any pending retry MUST be cancelled and
```
msbrogli, line 172: Should the node persist the latest whitelist and use it on the next start?

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+1. Issue an HTTP GET with header `User-Agent: hathor-core`.
+2. Reject any HTTP status outside the 2xx range.
+3. Apply `WHITELIST_REQUEST_TIMEOUT` to the entire round trip.
+4. Decode the body as UTF-8.
+5. Parse the body per the data-format rules above.
+6. On success, apply the parsed whitelist per RULE-19.
+7. On any failure, leave the active whitelist unchanged per RULE-20.
+
+**RULE-22**: A fetch timeout SHOULD be logged at a lower severity (e.g.
+`warning`) than other fetch failures (e.g. `error`), since timeouts are
```
msbrogli, line 223: No need to warn for a single failure. It should only log a warn if multiple failures happen in a row.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+is incremented, and the event is logged. A missing file MUST NOT cause the
+node to lock out all peers.
+
+## Integration points
+
+### Runtime replacement
+
+**RULE-25**: Replacing the active whitelist at runtime MUST:
+
+1. Stop the previous whitelist (if any).
```
msbrogli, line 241: But I guess the "stop action" should not close any connections.

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+2. Install the new whitelist and begin refreshing it.
```
msbrogli, line 242: Should it have at least one successful fetching before replacing? I mean, I don't want the full node to close all connections but the ones with bootstrap peers because the URL was changed, do it?

docs/p2p/peer-whitelist.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Sysctl
+
+**RULE-28**: The sysctl key `p2p.connections.whitelist` MUST support:
+
+- **Get**: returns the source string (URL or file path) of the active
+  whitelist, or the literal `"none"` if no whitelist is currently active
+  (covers DISABLED, SUSPENDED, and a URL source whose URL is unset).
+- **Set `"on"`**: resume enforcement of a previously suspended whitelist. If
+  nothing is suspended, this is a silent no-op. (See DEC-08.)
```
msbrogli, line 264: Should we return an error message to the sysadmin?

hathor/p2p/whitelist/parsing.py
```diff
@@ -0,0 +1,131 @@
...
+
+# node3
+G2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367
+2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367
+''')
+    {'2ffdfbbfd6d869a0742cff2b054af1cf364ae4298660c0e42fa8b00a66a30367'}
+
+    """
+    lines = parse_file(text, header=header)
+    return {PeerId(line.split()[0]) for line in lines}
```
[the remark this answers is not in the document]
glevco: for inline comments

PR #1575, opened 2026-01-24
feat(nano): Add HTTP API and CLI to re-run blocks and nano transactions

msbrogli:
### Motivation

Add HTTP API and CLI support for dry-running nano contract block execution without modifying state. This allows developers to inspect what would happen during NC execution for debugging and verification purposes.

### Acceptance Criteria

1. Add `/nano_contract/dry_run` HTTP endpoint that accepts either `block_hash` or `tx_hash` and returns detailed execution results
2. Add `x-nc-dry-run` CLI command with `--block-hash` or `--tx-hash` options, supporting JSON and text output formats
3. Add `NCDryRunBlockExecutor` that wraps `NCBlockExecutor` for read-only execution with in-memory voided state tracking
4. Refactor `NCBlockExecutor.execute_block()` to accept a `should_skip` predicate so consensus and dry-run can supply their own skip semantics

### Additional changes in this PR (second commit)

The follow-up commit (`refactor(nano): Inject NCBlockExecutor and split dry-run models`) contains a dependency-injection cleanup that the dry-run feature benefits from:

- **Breaking change to `ConsensusAlgorithm.__init__`**: no longer accepts `runner_factory` + `nc_calls_sorter`. Instead accepts `block_executor: NCBlockExecutor`, which is now built by `Builder` / `CliBuilder` and injected. Direct callers of `ConsensusAlgorithm(...)` must update their construction sites.
- **`ConsensusAlgorithm._block_executor` is now the public attribute `block_executor`** (one test updated accordingly).
- **`NCDryRunResource(manager, block_executor)`**: the dry-run resource receives the executor via constructor injection instead of reaching into `consensus_algorithm._block_executor`.
- **New module `hathor/nanocontracts/execution/dry_run_models.py`**: `ExecutionStatus`, `DryRunCallRecord`, `DryRunTxResult`, and `DryRunResult` moved out of `dry_run_block_executor.py` to keep pydantic models separate from the executor logic.
- **Events are now recorded on dry-run failure**, not only on success.
- **New invariant**: `assert block_storage.get_root_id() == parent_root_id` at the top of `execute_block`.
- **Added `test_dry_run_skip_cascade`** + `build_cascade_dry_run_dag` fixture: verifies a tx spending an output of a failing NC tx is SKIPPED via the dependency cascade in the same block.

This refactor commit can be split into a separate PR on request; the dry-run feature commit stands on its own.

### Checklist

### DIFF
--- a/hathor/nanocontracts/execution/block_executor.py
+++ b/hathor/nanocontracts/execution/block_executor.py
@@ -19,16 +19,18 @@ from __future__ import annotations
 import hashlib
 import traceback
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Iterator
+from typing import TYPE_CHECKING, Callable, Iterator, Mapping, TypeAlias, cast
 
 from hathor.feature_activation.utils import Features
 from hathor.nanocontracts.exception import NCFail
 from hathor.nanocontracts.nano_runtime_version import NanoRuntimeVersion
 from hathor.transaction import Block, Transaction
-from hathor.transaction.exceptions import TokenNotFound
-from hathor.transaction.nc_execution_state import NCExecutionState
+from hathorlib.nanocontracts.runner.call_info import CallInfo
 from hathorlib.nanocontracts.runner.runner import MAX_SEQNUM_JUMP_SIZE
 from hathorlib.nanocontracts.types import Address, BlueprintId, ContractId, NCRawArgs, VertexId
+
+# Type alias for the skip predicate
+ShouldSkipPredicate = Callable[[Transaction], bool]
 
 if TYPE_CHECKING:
     from hathor.conf.settings import HathorSettings
@@ -37,6 +39,8 @@ if TYPE_CHECKING:
     from hathor.nanocontracts.runner.runner import RunnerFactory
     from hathor.nanocontracts.sorter.types import NCSorterCallable
     from hathor.nanocontracts.storage import NCBlockStorage, NCStorageFactory
+    from hathor.nanocontracts.types import TokenUid
+    from hathor.transaction.token_info import TokenDescription
 
 
 # Transaction execution result types (also used as block execution effects)
@@ -49,9 +53,12 @@ class NCTxExecutionSuccess:
 
 @dataclass(slots=True, frozen=True)
 class NCTxExecutionFailure:
-    """Result type for failed NC execution."""
+    """Result type for failed NC execution.
+
+    `call_info` is None for cyclic-dependency failures, which never run a Runner.
+    """
     tx: Transaction
-    runner: 'Runner'
+    call_info: CallInfo | None
     exception: NCFail
     traceback: str
 
@@ -62,7 +69,7 @@ class NCTxExecutionSkipped:
     tx: Transaction
 
 
-NCTxExecutionResult = NCTxExecutionSuccess | NCTxExecutionFailure | NCTxExecutionSkipped
+NCTxExecutionResult: TypeAlias = NCTxExecutionSuccess | NCTxExecutionFailure | NCTxExecutionSkipped
 
 
 # Block execution lifecycle effect types for generator-based execution
@@ -72,7 +79,8 @@ class NCBeginBlock:
     block: Block
     parent_root_id: bytes
     block_storage: 'NCBlockStorage'
-    nc_sorted_calls: list[Transaction]
+    nc_sorted_calls: tuple[Transaction, ...]
+    nc_cyclic_fails: tuple[Transaction, ...]
 
 
 @dataclass(slots=True, frozen=True)
@@ -96,10 +104,34 @@ class NCEndBlock:
     final_root_id: bytes
 
 
-NCBlockEffect = (
+NCBlockEffect: TypeAlias = (
     NCBeginBlock | NCBeginTransaction | NCTxExecutionResult |
     NCEndTransaction | NCEndBlock
 )
+
+
+class _UncommittedRestrictedBlockProxy:
+    """Read-only token overlay for pre-commit validation."""
+
+    def __init__(
+        self,
+        block_storage: 'NCBlockStorage',
+        *,
+        created_tokens: Mapping['TokenUid', 'TokenDescription'],
+    ) -> None:
+        self._block_storage = block_storage
+        self._created_tokens = created_tokens
+
+    def has_token(self, token_id: 'TokenUid') -> bool:
+        if token_id in self._created_tokens:
+            return True
+        return self._block_storage.has_token(token_id)
+
+    def get_token_description(self, token_id: 'TokenUid') -> 'TokenDescription':
+        token_description = self._created_tokens.get(token_id)
+        if token_description is not None:
+            return token_description
+        return self._block_storage.get_token_description(token_id)
 
 
 class NCBlockExecutor:
@@ -135,7 +167,12 @@ class NCBlockExecutor:
         self._nc_calls_sorter = nc_calls_sorter
         self._feature_service = feature_service
 
-    def execute_block(self, block: Block) -> Iterator[NCBlockEffect]:
+    def execute_block(
+        self,
+        block: Block,
+        *,
+        should_skip: ShouldSkipPredicate,
+    ) -> Iterator[NCBlockEffect]:
         """Execute block as generator, yielding effects without applying them.
 
         This is the pure execution method that yields lifecycle events as it processes
@@ -143,18 +180,17 @@ class NCBlockExecutor:
 
         Args:
             block: The block to execute.
+            should_skip: A predicate function that determines if a transaction should be skipped.
+                This allows the caller to provide context-specific voided state tracking.
 
         Yields:
             NCBlockEffect instances representing each step of execution.
         """
-        from hathor.nanocontracts import NC_EXECUTION_FAIL_ID
-
         assert self._settings.ENABLE_NANO_CONTRACTS
         assert not block.is_genesis, "Genesis blocks should be handled separately"
 
         meta = block.get_metadata()
         assert not meta.voided_by
-        assert meta.nc_block_root_id is None
 
         parent = block.get_block_parent()
         parent_meta = parent.get_metadata()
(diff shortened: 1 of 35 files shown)


hathor/nanocontracts/execution/dry_run_block_executor.py
```diff
@@ -0,0 +1,381 @@
...
+            target_tx_hash: Optional TX hash when looking up via transaction.
+
+        Returns:
+            DryRunResult containing all execution information.
+        """
+        from hathor.nanocontracts import NC_EXECUTION_FAIL_ID
+        from hathor.transaction import Transaction
+
+        block_meta = block.get_metadata()
+        expected_root_id = block_meta.nc_block_root_id.hex() if block_meta.nc_block_root_id else ''
```
msbrogli: Use bytes instead.

hathor/nanocontracts/execution/dry_run_block_executor.py
```diff
@@ -0,0 +1,383 @@
...
+        events: list[dict[str, Any]] = []
+        for event in call_info.nc_logger.__events__:
+            events.append({
+                'nc_id': event.nc_id.hex(),
+                'data': event.data.hex(),
+            })
+        return events
+
+
+def format_dry_run_text(result: DryRunResult, verbose: bool = False) -> str:
```
msbrogli: Move elsewhere.

jansegre (dismissed):

hathor/nanocontracts/resources/dry_run.py
```diff
@@ -0,0 +1,170 @@
...
+        self.manager = manager
+
+    @api_endpoint(
+        path='/nano_contract/dry_run',
+        method='GET',
+        operation_id='nano_contracts_dry_run',
+        summary='Dry-run NC block execution',
+        description='Dry-run nano contract execution for a block without modifying state.',
+        tags=['nano_contracts'],
+        visibility='public',
```
glevco: This could be expensive, shouldn't it be private by default?

glevco (changes requested):

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -132,6 +132,8 @@ def __init__(
     def execute_block(
         self,
         block: Block,
+        *,
+        dry_run: bool = False,
```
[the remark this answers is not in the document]
msbrogli: Done!

hathor/nanocontracts/execution/dry_run_utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    _validate_block(block)
+    return DryRunTarget(block=block, target_tx_hash=tx_hash_bytes)
+
+
+def _resolve_via_block(tx_storage: 'TransactionStorage', block_hash_hex: str) -> DryRunTarget:
+    """Resolve block directly by hash."""
+    if not _HEX_HASH_RE.match(block_hash_hex):
+        raise DryRunValidationError('Invalid hash format')
+    block_hash_bytes = bytes.fromhex(block_hash_hex)
```
glevco: Isn't it cheaper and simpler to just do `bytes.fromhex` and catch the exception?

glevco (dismissed):

jansegre (dismissed):

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -135,26 +137,30 @@ def __init__(
...
             NCBlockEffect instances representing each step of execution.
         """
-        from hathor.nanocontracts import NC_EXECUTION_FAIL_ID
-
         assert self._settings.ENABLE_NANO_CONTRACTS
         assert not block.is_genesis, "Genesis blocks should be handled separately"
 
         meta = block.get_metadata()
         assert not meta.voided_by
-        assert meta.nc_block_root_id is None
```
msbrogli, line 157: When we are doing a dry-run, this assert fails. So it was moved to the `consensus_block_executor.py`.

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -165,14 +171,11 @@ def execute_block(self, block: Block) -> Iterator[NCBlockEffect]:
         for tx in block.iter_transactions_in_this_block():
             if not tx.is_nano_contract():
                 continue
-            tx_meta = tx.get_metadata()
-            assert tx_meta.nc_execution in {None, NCExecutionState.PENDING}
```
msbrogli, line 169: Same here: this assert fails when doing a dry-run.

hathor_cli/nc_dry_run.py
```diff
@@ -0,0 +1,228 @@
...
+                tx_hash=tx_hash_arg,
+            )
+        except (DryRunValidationError, DryRunConflictError, DryRunNotFoundError) as e:
+            self.log.error(str(e))
+            sys.exit(1)
+
+        # Execute dry run using shared executor
+        try:
+            dry_run_executor = NCDryRunBlockExecutor(
+                self.manager.consensus_algorithm.block_executor
```
msbrogli: Why don't we inject `block_executor` in this resource?

hathor/nanocontracts/resources/dry_run.py
```diff
@@ -0,0 +1,120 @@
...
+        response_model=Union[DryRunResult, ErrorResponse, NotFoundResponse, ConflictResponse],
+    )
+    def render_GET(self, request: 'Request', *, params: NCDryRunParams) -> Union[ResponseModel, Deferred[Any]]:
+        request.setHeader(b'cache-control', b'no-store')
+
+        logger.info('nc_dry_run.start', block_hash=params.block_hash, tx_hash=params.tx_hash)
+
+        try:
+            target = resolve_block_for_dry_run(
+                self.manager.tx_storage,
```
msbrogli: Inject `tx_storage` (instead of `manager`).

hathor/nanocontracts/execution/dry_run_block_executor.py
```diff
@@ -0,0 +1,274 @@
...
+        rng_seed: Optional[bytes],
+        runner: Any,
+        include_changes: bool,
+    ) -> DryRunTxResult:
+        """Build result for successful execution."""
+        call_info = runner.get_last_call_info()
+        call_records = self._build_call_records(call_info, include_changes)
+        events = self._build_events(call_info)
+
+        return DryRunTxResult(
```
msbrogli, line 204: Why not simply use `NCTxExecutionSuccess`?

hathor/nanocontracts/execution/dry_run_block_executor.py
```diff
@@ -0,0 +1,274 @@
...
+        exception: Exception,
+        tb: str,
+        include_changes: bool,
+    ) -> DryRunTxResult:
+        """Build result for failed execution."""
+        call_info = runner.get_last_call_info()
+        call_records = self._build_call_records(call_info, include_changes)
+        events = self._build_events(call_info)
+
+        return DryRunTxResult(
```
msbrogli, line 228: Why not use `NCTxExecutionFailure` (instead of `DryRunTxResult`)?

hathor/nanocontracts/execution/dry_run_utils.py
```diff
@@ -0,0 +1,130 @@
...
+from hathor.transaction.storage.exceptions import TransactionDoesNotExist, TransactionIsNotABlock
+
+_HEX_HASH_RE = re.compile(r'^[0-9a-fA-F]{64}$')
+
+if TYPE_CHECKING:
+    from hathor.transaction import Block
+    from hathor.transaction.storage import TransactionStorage
+
+
+class DryRunValidationError(Exception):
```
msbrogli, line 30: Move to exceptions file?

hathor/nanocontracts/execution/dry_run_utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Returns:
+        DryRunTarget with the resolved block and optional target tx hash.
+
+    Raises:
+        DryRunValidationError: For invalid input (bad hash, genesis, voided, not NC).
+        DryRunNotFoundError: When the referenced block or transaction is not found.
+    """
+    if tx_hash:
+        return _resolve_via_tx(tx_storage, tx_hash)
```
msbrogli, line 73: If a block was provided, it must match the block found by `_resolve_via_tx()`, right?

hathor/nanocontracts/execution/dry_run_utils.py
msbrogli: Good call. I replaced the regex with a reusable `parse_hash32()` helper in `hathor/utils/hex.py` that does `bytes.fromhex` and also enforces the 64-char length (`bytes.fromhex('ab')` would otherwise succeed). The regex is gone from `dry_run_utils.py`.

hathor/nanocontracts/execution/dry_run_block_executor.py
msbrogli: These internal types carry live objects (`runner: Runner`, `tx: Transaction`) because `NCConsensusBlockExecutor` needs to call the runner back to apply its accumulated state. A pydantic model would need `arbitrary_types_allowed` just to hold them, and even then they would not serialize to JSON. `DryRunTxResult` also aggregates info that does not live in `NCTxExecutionSuccess` alone: `rng_seed` comes from the preceding `NCBeginTransaction` effect, and `call_records`/`events` are extracted from `runner.get_last_call_info()`. Merging would force every caller (including consensus) to pay the extraction cost eagerly. Keeping the split: internal types = execution lifecycle effects, `DryRunTxResult` = HTTP/CLI serialization boundary.

hathor/nanocontracts/execution/dry_run_block_executor.py
msbrogli: Same reasoning as the success case: `NCTxExecutionFailure` carries live `runner`, `tx`, and `exception: NCFail` — consensus needs to re-raise the exception and apply runner state; pydantic would not serialize any of those cleanly. `DryRunTxResult` is the serialization boundary: it stringifies the exception into `exception_type` / `exception_message` / `traceback` so the result is JSON-safe.

glevco (dismissed):

jansegre (dismissed):

hathor/nanocontracts/execution/dry_run_block_executor.py
```diff
@@ -0,0 +1,303 @@
...
+            # voids, or being voided by another tx) means consensus skipped it
+            # and dry-run must skip it too.
+            meta = tx.get_metadata()
+            if meta.voided_by:
+                only_nc_fail = (
+                    NC_EXECUTION_FAIL_ID in meta.voided_by
+                    and not (meta.voided_by - {NC_EXECUTION_FAIL_ID, tx.hash})
+                )
+                if not only_nc_fail:
+                    return True
```
glevco: Why not `if meta.voided_by != {NC_EXECUTION_FAIL_ID, tx.hash}: return True`?

glevco (dismissed):

hathor/nanocontracts/execution/dry_run_block_executor.py
msbrogli: Done! Thanks!

glevco (dismissed):

jansegre (dismissed):

PR #1603, opened 2026-02-24
feat(tx): Add shielded outputs for amount and token privacy

msbrogli:
### Motivation

### Acceptance Criteria

### Checklist


hathor/indexes/utxo_index.py
```diff
@@ -143,6 +143,9 @@ def _update_executed(self, tx: BaseTransaction) -> None:
         # remove all inputs
         for tx_input in tx.inputs:
             spent_tx = tx.get_spent_tx(tx_input)
+            # CONS-018: skip shielded outputs — they are never in the UTXO index
```
msbrogli: This change is wrong. It should actually include the utxo to the index regardless of the output type. But it might need some change when querying and filtering by value/token.

hathor/nanocontracts/vertex_data.py
```diff
@@ -40,7 +40,9 @@ def _get_txin_output(vertex: BaseTransaction, txin: TxInput) -> TxOutput | None:
     except TransactionDoesNotExist:
         assert False, f'missing dependency: {txin.tx_id.hex()}'
 
-    assert len(vertex2.outputs) > txin.index, 'invalid output index'
+    # CONS-024: return None for shielded output indices instead of crashing
+    if txin.index >= len(vertex2.outputs):
+        return None
```
msbrogli: Why not return the shielded output?

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -61,6 +61,7 @@ def __init__(self,
                 nanocontracts=False,
                 fee_tokens=False,
                 opcodes_version=OpcodesVersion.V1,
+                shielded_transactions=True,
```
msbrogli: Should it be false?

hathor/transaction/scripts/execute.py
```diff
@@ -115,9 +115,20 @@ def script_eval(tx: Transaction, txin: TxInput, spent_tx: BaseTransaction, versi
...
     """
+    # VULN-002: Handle shielded output references
+    # CONS-007: Bounds-check shielded index to raise InvalidScriptError, not IndexError
+    if txin.index < len(spent_tx.outputs):
+        output_script = spent_tx.outputs[txin.index].script
+    elif spent_tx.shielded_outputs:
+        shielded_idx = txin.index - len(spent_tx.outputs)
+        if shielded_idx >= len(spent_tx.shielded_outputs):
+            raise InvalidScriptError(f'input index {txin.index} out of range')
+        output_script = spent_tx.shielded_outputs[shielded_idx].script
```
msbrogli: Refactor to spent_tx.get_output(n) (or similar if you know a better name). Actually, maybe tx.outputs should be refactores to tx._outputs but that's a big task for this PR.

msbrogli: Why not use resolve_spent_output()?

hathor/transaction/exceptions.py
```diff
@@ -278,3 +278,31 @@ class InvalidFeeAmount(InvalidFeeHeader):
...
+class InvalidSurjectionProofError(TxValidationError):
+    """Surjection proof is invalid."""
+
+
+class ShieldedBalanceMismatchError(TxValidationError):
+    """Shielded balance equation does not hold."""
+
+
+class TrivialCommitmentError(TxValidationError):
+    """Rule 4: All transparent inputs require >= 2 shielded outputs."""
```
msbrogli: Should we allow 1:1? Such transactions wouldn't actually shield the value.

hathor/transaction/shielded_tx_output.py
```diff
@@ -0,0 +1,231 @@
...
+        raise ValueError(f'Unknown shielded output mode: {mode_byte}')
+
+    return output, bytes(view[offset:])
+
+
+def get_sighash_bytes(output: ShieldedOutput) -> bytes:
+    """Return sighash bytes for a shielded output.
+
+    Includes commitment + mode + token_data/asset_commitment + script.
+    Does NOT include proofs (range_proof, surjection_proof).
```
msbrogli: Why not?

hathor/verification/verification_service.py
```diff
@@ -139,6 +140,21 @@ def verify_basic(
...
+            if not params.features.shielded_transactions:
+                from hathor.transaction.exceptions import InvalidShieldedOutputError
+                raise InvalidShieldedOutputError('shielded transactions are not enabled')
+            self._verify_basic_shielded_header(vertex)
+
+    def _verify_basic_shielded_header(self, tx: Transaction) -> None:
+        """Shielded verifications that don't need storage."""
+        try:
+            self.verifiers.shielded_tx.verify_shielded_outputs(tx)
+        except Exception:
```
msbrogli: Why Exception?

hathor/verification/vertex_verifier.py
```diff
@@ -225,15 +229,39 @@ def get_allowed_headers(self, vertex: BaseTransaction, params: VerificationParam
...
             case TxVersion.ON_CHAIN_BLUEPRINT:
                 pass
-            case TxVersion.REGULAR_TRANSACTION | TxVersion.TOKEN_CREATION_TRANSACTION:
+            case TxVersion.TOKEN_CREATION_TRANSACTION:
                 if params.features.nanocontracts:
                     allowed_headers.add(NanoHeader)
                 if params.features.fee_tokens:
                     allowed_headers.add(FeeHeader)
+                # CONS-006: Token creation txs must NOT allow shielded outputs,
+                # as it bypasses minting verification when combined with verify_sum skip.
```
msbrogli: Remove "as it bypasses minting verification when combined with verify_sum skip".

hathor/verification/vertex_verifiers.py
```diff
@@ -101,4 +104,5 @@ def create(
             token_creation_tx=token_creation_tx_verifier,
             nano_header=nano_header_verifier,
             on_chain_blueprint=on_chain_blueprint_verifier,
+            shielded_tx=shielded_tx_verifier,
```
msbrogli: There's no need for a shielded transaction verifier. Split the verifiers according to the vertex type, instead.

hathor/wallet/resources/thin_wallet/address_search.py
```diff
@@ -46,6 +46,9 @@ def has_token_and_address(self, tx: 'BaseTransaction', address: str, token: byte
         """
         for tx_input in tx.inputs:
             spent_tx = tx.get_spent_tx(tx_input)
+            # CONS-022: skip shielded outputs
+            if spent_tx.is_shielded_output(tx_input.index):
```
msbrogli: Same here. Should we add a marker to the response?

hathor/wallet/resources/send_tokens.py
```diff
@@ -134,7 +135,12 @@ def _render_POST_thread(self, values: dict[str, Any], request: Request) -> Union
         self.manager.cpu_mining_service.resolve(tx)
         tx.init_static_metadata_from_storage(self._settings, self.manager.tx_storage)
         best_block = self.manager.tx_storage.get_best_block()
-        params = VerificationParams.default_for_mempool(best_block=best_block)
+        features = Features.from_vertex(
+            settings=self._settings,
+            feature_service=self.manager.feature_service,
+            vertex=best_block,
+        )
+        params = VerificationParams.default_for_mempool(best_block=best_block, features=features)
```
msbrogli: Why this change? Doesn't VerificationParams already do it if features is not provided?

hathor/wallet/base_wallet.py
```diff
@@ -599,6 +611,123 @@ def on_new_tx(self, tx: BaseTransaction) -> None:
...
+        """Verify that a recovered token UID and asset blinding factor match the asset_commitment.
+
+        AUDIT-C015: Prevents social engineering attacks where a malicious sender
+        embeds a wrong token UID in the range proof message.
+
+        Raises ValueError if the token UID is inconsistent.
+        """
+        from hathor.crypto.shielded.asset_tag import create_asset_commitment, derive_tag
+        expected_tag = derive_tag(token_id)
+        expected_commitment = create_asset_commitment(expected_tag, asset_bf)
```
msbrogli: Shouldn't it be done by `derive_tag()` itself?

hathor-ct-crypto/src/types.rs
```diff
@@ -0,0 +1,11 @@
+/// A 32-byte token UID.
+pub type TokenUid = [u8; 32];
+
+/// Zero token UID representing HTR.
+pub const HTR_TOKEN_UID: TokenUid = [0u8; 32];
```
msbrogli: Can we keep it with one byte?

hathor-ct-crypto/src/generators.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// Derive an unblinded asset generator for a given token UID.
+///
+/// Returns `Generator::new_unblinded(SECP256K1, tag)` where tag is derived from the token UID.
+pub fn derive_asset_tag(token_uid: &TokenUid) -> Result<Generator> {
+    let tag = derive_tag(token_uid)?;
+    let generator = Generator::new_unblinded(SECP256K1, tag);
+    Ok(generator)
+}
+
```
LFRezende: **[Feature Tweak]**

As stated in [comment](https://github.com/HathorNetwork/hathor-core/pull/1603/changes/03f932b9cb5ee598bccb481c0b6a1a557f041c3e#r2879838795), if we wish to maintain the core design of the NUMS specified in [internal-rfc: NUMS](https://github.com/HathorNetwork/internal-rfcs/blob/design/confidential-transactions/projects/confidential-transactions/03-amount-privacy/020-SHIELDED-TRANSACTIONS-DESIGN.md) we'd need to keep the structure:

``` asset_tag = SHA256(SHA256(token_uid) || counter) ```

But counter isn't used here in a loop.  The following line:

`let generator  = Generator::new_unblinded(SECP256K1, tag)`

Calls (eventually): `secp256k1_generator_generate_blinded(ctx, key32=tag_bytes, blind32=zeros)`

Which generates two different points with the tag provided and sums them up, returning the result (the sum will be on the curve, so we do return a valid asset tag).

**Not a bug** - just a deviation from the design detailed in the internal-rfcs that I'm pointing out, in case we're willing to stick to it.

File: hathor/crypto/shielded/asset_tag.py [hunk 15]
```diff
@@ -0,0 +1,54 @@
...
+
+def derive_asset_tag(token_uid: bytes) -> bytes:
+    """Derive a deterministic NUMS generator (33 bytes) for a token UID.
+
+    Accepts both 1-byte (HTR) and 32-byte token UIDs.
+    """
+    if _lib is None:
+        raise RuntimeError('hathor_ct_crypto native library is not available')
+    return _lib.derive_asset_tag(_normalize_token_uid(token_uid))
+
```
LFRezende: **[Feature Tweak]**

If to match NUMS, as said in [comment](https://github.com/HathorNetwork/hathor-core/pull/1603/changes/03f932b9cb5ee598bccb481c0b6a1a557f041c3e#r2879838795), instead of changing `derive_tag`, as it is used also for raw tags, we should pass the normalized_token_uid hashed.

hathor-ct-crypto/src/surjection.rs
```diff
@@ -0,0 +1,180 @@
...
+    codomain_tag: &Tag,
+    codomain_blinding_factor: &Tweak,
+    domain: &[(Generator, Tag, Tweak)],
+) -> Result<SurjectionProof> {
+    if domain.is_empty() {
+        return Err(HathorCtError::SurjectionProofError(
+            "domain must not be empty".into(),
+        ));
+    }
+
```
LFRezende: Perhaps a more complete error check:

```
if domain.is_empty() {
    return Err(...);
}

if domain.len() > SOME_MAX_VALUE {
    return Err(...);
}
```

I did some research and it seems a high domain could lead to some DOS-verification attacks, if left allowed to be too large.

hathor/crypto/shielded/range_proof.py
```diff
@@ -0,0 +1,44 @@
...
+    generator: bytes,
+) -> tuple[int, bytes, bytes]:
+    """Rewind a Bulletproof range proof to recover committed value, blinding factor, and message.
+
+    Requires the same nonce key that was used when creating the proof.
+    Returns (value, blinding_factor, message).
+    """
+    if _lib is None:
+        raise RuntimeError('hathor_ct_crypto native library is not available')
+    return _lib.rewind_range_proof(proof, commitment, nonce, generator)
```
LFRezende: **[OK]** - just lends the logic to their rust counterparts, no reason of concern.

hathor/dag_builder/vertex_exporter.py
```diff
@@ -505,21 +505,145 @@ def _add_or_augment_shielded_fee(self, node: DAGNode, vertex: BaseTransaction) -
...
+                asset_comm = create_asset_commitment(raw_tag, asset_blinding)
+                commitment = create_commitment(amount, blinding, asset_comm)
+
+                # Embed token_uid(32B) + asset_blinding(32B) in range proof message
+                message = token_uid + asset_blinding
+                range_proof = create_range_proof(
+                    amount, blinding, commitment, asset_comm,
+                    message=message, nonce=nonce,
+                )
+
```
LFRezende: Why do we send `tokenUid || r_asset` here as a message?

hathor/transaction/base_transaction.py
```diff
@@ -309,6 +358,29 @@ def hash_hex(self) -> str:
...
+        """
+        if index < len(self.outputs):
+            return self.outputs[index]
+        shielded_idx = index - len(self.outputs)
+        shielded = self.shielded_outputs
+        if shielded_idx < len(shielded):
+            return shielded[shielded_idx]
+        raise IndexError(f'output index {index} out of range (transparent={len(self.outputs)}, '
+                         f'shielded={len(shielded)})')
+
```
LFRezende: I didn't get that much the idea of this function - its inner purpose. What is "resolve" the spent output?

PR #1613, opened 2026-03-10
feat(p2p): SlotsManager class for managing connection pool slots

LFRezende:
### Motivation

RFCs available: https://github.com/HathorNetwork/internal-rfcs/tree/design/connection-pool-slots/projects/p2p-projects/connection-pool-slots/

The original Pull Request for slotting the connection pool (#1266 ) was proposed for subdividing the connections into four main groups, allowing for a more solid infrastructure against attacks onto hathor-core. 

The original PR has been subdivided into two smaller ones - this focusing on the development of a SlotsManager class to handle the distribution and coordination of protocols between the ConnectionsSlot classes, which for this pull request will be restricted to three:

- OUTGOING: Outbound connections started by the node to other peers, after all bootstrap set-ups have been finalized.
- INCOMING: Inbound connections, started by other peers towards our node. 
- BOOTSTRAP: Outbound connections right at the beginning of the full node boot up, trying to fetch hardcoded honest peers to connect to.

The second PR will deal with the fourth slot class - CHECK_ENTRYPOINTS - which deserves its separate pipeline due to the inner intrincacies of the checking mechanism.

When starting the ConnectionsManager class, the SlotsManager will be booted up, which will bring up by itself the slots.

The SlotsManager will be called whenever some manipulation of the protocol flux towards or outwards the slots needs to be performed, particularly when connecting/disconnecting a HathorProtocol instance.

**Notes**
1. This PR has flaking issues and may have compatibility constraints, as the idea for now is to discuss the ideas and implementations described within its code body;
2. Many of the new features regarding the SlotsManager still need to be properly tested, and the old tests need to be updated as well.
3. Check Entrypoints has been purposefully extracted out of the code, so to test and discuss specifically the slotting of the connection pool first, rather than both it and the check entrypoints mechanism in the same breath.

### Acceptance Criteria

- Connection establishment must be capped by the slot limitations
- Connections must be subdivided into the proper slots by their connection type
- When a protocol is disconnected, it must be removed from the slot
- When a protocol is connected, it must be added to the proper slot
- No protocol of type A should be appended to slot of type B !=A
- Total active connections == Connections(slotA) + Connections(slotB) + Connectioons(slotC): in other words, the slots must be able to limit the connection flow to their inner circles.

### Checklist

- Connection establishment must be capped by the slot limitations - OK
- Connections must be subdivided into the proper slots by their connection type - OK
- When a protocol is disconnected, it must be removed from the slot - OK
- When a protocol is connected, it must be added to the proper slot - OK
- No protocol of type A should be appended to slot of type B !=A - OK
- Total active connections == Connections(slotA) + Connections(slotB) + Connectioons(slotC): in other words, the slots must be able to limit the connection flow to their inner circles. - OK

### DIFF
--- a/hathor/p2p/connect_classes.py
+++ b/hathor/p2p/connect_classes.py
@@ -1,0 +1,48 @@
+# Copyright 2021 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from dataclasses import dataclass
+from enum import Enum
+
+
+class ConnectionType(Enum):
+    """ Types of Connection as inputs for an instance of the Hathor Protocol. """
+    OUTGOING = 0
+    INCOMING = 1
+    BOOTSTRAP = 2
+
+    def is_outbound(self) -> bool:
+        """ If value is 1, then the connection is inbound. If not, outbound."""
+        return self in (ConnectionType.OUTGOING, ConnectionType.BOOTSTRAP)
+
+
+@dataclass(slots=True, frozen=True)
+class ConnectionAllowed:
+    confirmation: str
+
+
+@dataclass(slots=True, frozen=True)
+class ConnectionRejected:
+    reason: str
+
+
+@dataclass(slots=True, frozen=True)
+class ConnectionRemoved:
+    reason: str
+
+
+@dataclass(slots=True, frozen=True)
+class ConnectionNotRemoved:
+    reason: str
--- a/hathor/p2p/connection_slot.py
+++ b/hathor/p2p/connection_slot.py
@@ -1,0 +1,140 @@
+# Copyright 2021 Hathor Labs
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from dataclasses import dataclass
+
+from typing_extensions import assert_never
+
+from hathor.p2p.connect_classes import (
+    ConnectionAllowed,
+    ConnectionNotRemoved,
+    ConnectionRejected,
+    ConnectionRemoved,
+    ConnectionType,
+)
+from hathor.p2p.protocol import HathorProtocol
+
+AddToSlotResult = ConnectionAllowed | ConnectionRejected
+RemoveFromSlotResult = ConnectionRemoved | ConnectionNotRemoved
+
+
+class ConnectionSlots:
+    """ Class of a connection pool slot - outgoing, incoming, bootstrap connections. """
+    connection_slot: set[HathorProtocol]
+    max_slot_connections: int
+
+    def __init__(self, type: ConnectionType, max_connections: int) -> None:
+
+        assert max_connections > 0, 'Slot max number must allow at least one connection'
+
+        self.connection_slot = set()
+        self.max_slot_connections = max_connections
+
+    def __contains__(self, protocol: HathorProtocol) -> bool:
+        return protocol in self.connection_slot
+
+    def add_connection(self, protocol: HathorProtocol) -> AddToSlotResult:
+        """ Adds connection protocol to the slot. Checks whether the slot is full or not. If full,
+            disconnects the protocol. If the type is 'check_entrypoints', the returns peers of it
+            may go to a queue."""
+
+        if protocol in self.connection_slot:
+            return ConnectionRejected("Protocol already in Slot.")
+
+        if self.is_full():
+            return ConnectionRejected(f"Slot {protocol.connection_type} is full")
+
+        self.connection_slot.add(protocol)
+
+        return ConnectionAllowed(f"Added to slot {protocol.connection_type}.")
+
+    def remove_connection(self, protocol: HathorProtocol) -> ConnectionRemoved:
+        """ Removes from given instance the protocol passed. Returns protocol from queue
+            when disconnection leads to free space in slot."""
+
+        # Discard does nothing if protocol not in connection_slot.
+        self.connection_slot.discard(protocol)
+        return ConnectionRemoved('Connection successfully removed from slot.')
+
+    def is_full(self) -> bool:
+        """ Checks if connection slot has reached the limit of allowed connections."""
+        return len(self.connection_slot) >= self.max_slot_connections
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class SlotsManagerSettings:
+    max_outgoing: int
+    max_incoming: int
+    max_bootstrap: int
+
+
+class SlotsManager:
+    """Manager of slot connections - selects the slot to which must we send the
+     arriving protocol. Three protocol slots: OUTGOING, INCOMING, BOOTSTRAP.
+    """
+    outgoing_slot: ConnectionSlots
+    incoming_slot: ConnectionSlots
+    bootstrap_slot: ConnectionSlots
+
+    def __init__(self, settings: SlotsManagerSettings) -> None:
+        self.outgoing_slot = ConnectionSlots(ConnectionType.OUTGOING, settings.max_outgoing)
+        self.incoming_slot = ConnectionSlots(ConnectionType.INCOMING, settings.max_incoming)
+        self.bootstrap_slot = ConnectionSlots(ConnectionType.BOOTSTRAP, settings.max_bootstrap)
+
+    def add_to_slot(self, protocol: HathorProtocol) -> AddToSlotResult:
+        """Add received protocol to one of the slots.
+        If slot is full, protocol is not added. """
+
+        conn_type = protocol.connection_type
+        slot: ConnectionSlots
+        match conn_type:
+            case ConnectionType.OUTGOING:
+                slot = self.outgoing_slot
+            case ConnectionType.INCOMING:
+                slot = self.incoming_slot
+            case ConnectionType.BOOTSTRAP:
+                slot = self.bootstrap_slot
+            case _:
+                assert_never(conn_type)
+
+        status = slot.add_connection(protocol)
+
+        return status
+
+    def remove_from_slot(self, protocol: HathorProtocol) -> None:
+        """ Removes protocol from slot of same type.
+            If OUTGOING, INCOMING or BOOTSTRAP, simply remove from slot and disconnect.
+            Should be called by manager when disconnecting a protocol.
+            Wraps _remove_from_slot_result and reduces API exposure."""
+
+        self._remove_from_slot_result(protocol)
+
+    def _remove_from_slot_result(self, protocol: HathorProtocol) -> RemoveFromSlotResult:
+        """ Removes protocol from slot of same type.
+            If OUTGOING, INCOMING or BOOTSTRAP, simply remove from slot and disconnect.
+            Returns type for result, used in tests."""
+
+        conn_type = protocol.connection_type
+        slot: ConnectionSlots
+        match conn_type:
+            case ConnectionType.OUTGOING:
+                slot = self.outgoing_slot
+            case ConnectionType.INCOMING:
+                slot = self.incoming_slot
+            case ConnectionType.BOOTSTRAP:
+                slot = self.bootstrap_slot
+            case _:
+                assert_never(conn_type)
+
+        return slot.remove_connection(protocol)
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -23,8 +23,11 @@ from twisted.internet.interfaces import IListeningPort, IProtocol, IProtocolFact
 from twisted.internet.task import LoopingCall
 from twisted.protocols.tls import TLSMemoryBIOFactory, TLSMemoryBIOProtocol
 from twisted.python.failure import Failure
+from typing_extensions import assert_never
 
 from hathor.conf.settings import HathorSettings
+from hathor.p2p.connect_classes import ConnectionAllowed, ConnectionRejected
+from hathor.p2p.connection_slot import SlotsManager
 from hathor.p2p.netfilter.factory import NetfilterFactory
 from hathor.p2p.peer import PrivatePeer, PublicPeer, UnverifiedPeer
 from hathor.p2p.peer_discovery import PeerDiscovery
@@ -98,6 +101,7 @@ class ConnectionsManager:
         rng: Random,
         enable_ipv6: bool,
         disable_ipv4: bool,
+        slots_manager: SlotsManager
     ) -> None:
         self.log = logger.new()
         self._settings = settings
@@ -110,7 +114,6 @@ class ConnectionsManager:
 
         self.reactor = reactor
         self.my_peer = my_peer
-
         # List of address descriptions to listen for new connections (eg: [tcp:8000])
         self.listen_address_descriptions: list[str] = []
 
@@ -124,12 +127,15 @@ class ConnectionsManager:
         self.localhost_only = False
 
         # Factories.
-        from hathor.p2p.factory import HathorClientFactory, HathorServerFactory
+        from hathor.p2p.factory import HathorBootstrapFactory, HathorClientFactory, HathorServerFactory
         self.use_ssl = ssl
         self.server_factory = HathorServerFactory(
             self.my_peer, p2p_manager=self, use_ssl=self.use_ssl, settings=self._settings
         )
         self.client_factory = HathorClientFactory(
+            self.my_peer, p2p_manager=self, use_ssl=self.use_ssl, settings=self._settings
+        )
+        self.bootstrap_factory = HathorBootstrapFactory(
             self.my_peer, p2p_manager=self, use_ssl=self.use_ssl, settings=self._settings
         )
 
@@ -151,6 +157,8 @@ class ConnectionsManager:
 
         # List of peers connected and ready to communicate.
         self.connected_peers = {}
+
+        self.slots_manager = slots_manager
 
         # Queue of ready peer-id's used by connect_to_peer_from_connection_queue to choose the next peer to pull a
         # random new connection from
@@ -259,12 +267,16 @@ class ConnectionsManager:
         """Add a peer discovery method."""
         self.peer_discoveries.append(peer_discovery)
 
+    def connect_to_bootstrap(self, entrypoint: PeerEndpoint) -> None:
+        """Connect to a bootstrap entrypoint discovered via a PeerDiscovery strategy."""
+        self.connect_to_endpoint(entrypoint, discovery_call=True)
+
     def do_discovery(self) -> None:
         """
         Do a discovery and connect on all discovery strategies.
         """
         for peer_discovery in self.peer_discoveries:
-            coro = peer_discovery.discover_and_connect(self.connect_to_endpoint)
+            coro = peer_discovery.discover_and_connect(self.connect_to_bootstrap)
             Deferred.fromCoroutine(coro)
 
     def disable_rate_limiter(self) -> None:
@@ -331,7 +343,7 @@ class ConnectionsManager:
             len(self.connecting_peers),
             len(self.handshaking_peers),
             len(self.connected_peers),
-            len(self.verified_peer_storage)
+            len(self.verified_peer_storage),
         )
 
     def get_sync_factory(self, sync_version: SyncVersion) -> SyncAgentFactory:
@@ -387,10 +399,26 @@ class ConnectionsManager:
 
     def on_peer_connect(self, protocol: HathorProtocol) -> None:
         """Called when a new connection is established."""
+
+        # Checks whether connections in the network are at limit.
         if len(self.connections) >= self.max_connections:
             self.log.warn('reached maximum number of connections', max_connections=self.max_connections)
             protocol.disconnect(force=True)
             return
+
+        connection_status = self.slots_manager.add_to_slot(protocol)
+
+        match connection_status:
+            case ConnectionRejected():
+                self.log.warn('Connection Rejected')
+                protocol.disconnect(force=True)
+                return
+            case ConnectionAllowed():
+                pass
+            case _:
+                assert_never(connection_status)
+
+
         self.connections.add(protocol)
         self.handshaking_peers.add(protocol)
 
@@ -404,8 +432,8 @@ class ConnectionsManager:
         """Called when a peer is ready."""
         assert protocol.peer is not None
         self.verified_peer_storage.add_or_replace(protocol.peer)
-
         self.handshaking_peers.remove(protocol)
+
         for conn in self.iter_all_connections():
             conn.unverified_peer_storage.remove(protocol.peer)
 
@@ -455,9 +483,16 @@ class ConnectionsManager:
 
     def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
         """Called when a peer disconnect."""
+
+        # Discard handles case when not in connections.
         self.connections.discard(protocol)
+
+        # Each conn is from a slot - discard from it as well.
+        self.slots_manager.remove_from_slot(protocol)
+
         if protocol in self.handshaking_peers:
             self.handshaking_peers.remove(protocol)
+
         if protocol._peer is not None:
             peer_id = protocol.peer.id
             existing_protocol = self.connected_peers.pop(peer_id, None)
@@ -474,6 +509,7 @@ class ConnectionsManager:
             elif peer_id in self.new_connection_from_queue:
                 # now we're sure it can be removed from new_connection_from_queue
                 self.new_connection_from_queue.remove(peer_id)
+
         self.pubsub.publish(
             HathorEvents.NETWORK_PEER_DISCONNECTED,
             protocol=protocol,
@@ -482,13 +518,11 @@ class ConnectionsManager:
 
     def iter_all_connections(self) -> Iterable[HathorProtocol]:
         """Iterate over all connections."""
-        for conn in self.connections:
-            yield conn
+        yield from self.connections
 
     def iter_ready_connections(self) -> Iterable[HathorProtocol]:
         """Iterate over ready connections."""
-        for conn in self.connected_peers.values():
-            yield conn
+        yield from self.connected_peers.values()
 
     def iter_not_ready_endpoints(self) -> Iterable[PeerEndpoint]:
         """Iterate over not-ready connections."""
@@ -624,6 +658,7 @@ class ConnectionsManager:
         entrypoint: PeerEndpoint,
         peer: UnverifiedPeer | PublicPeer | None = None,
         use_ssl: bool | None = None,
+        discovery_call: bool = False
     ) -> None:
         """ Attempt to connect directly to an endpoint, prefer calling `connect_to_peer` when possible.
 
@@ -633,6 +668,7 @@ class ConnectionsManager:
 
         If `use_ssl` is True, then the connection will be wraped by a TLS.
         """
+
         if entrypoint.peer_id is not None and peer is not None and entrypoint.peer_id != peer.id:
             self.log.debug('skipping because the entrypoint peer_id does not match the actual peer_id',
                            entrypoint=str(entrypoint))
@@ -663,11 +699,10 @@ class ConnectionsManager:
 
         endpoint = entrypoint.addr.to_client_endpoint(self.reactor)
 
-        factory: IProtocolFactory
+        factory: IProtocolFactory = self.bootstrap_factory if discovery_call else self.client_factory
+
         if use_ssl:
-            factory = TLSMemoryBIOFactory(self.my_peer.certificate_options, True, self.client_factory)
-        else:
-            factory = self.client_factory
+            factory = TLSMemoryBIOFactory(self.my_peer.certificate_options, True, factory)
 
         if peer is not None:
             now = int(self.reactor.seconds())
@@ -753,9 +788,10 @@ class ConnectionsManager:
         assert protocol.peer.id is not None
         assert protocol.my_peer.id is not None
         other_connection = self.connected_peers[protocol.peer.id]
+        is_outbound = protocol.connection_type.is_outbound()
         if bytes(protocol.my_peer.id) > bytes(protocol.peer.id):
             # connection started by me is kept
-            if not protocol.inbound:
+            if is_outbound:
                 # other connection is dropped
                 return other_connection
             else:
@@ -763,7 +799,7 @@ class ConnectionsManager:
                 return protocol
         else:
             # connection started by peer is kept
-            if not protocol.inbound:
+            if is_outbound:
                 return protocol
             else:
                 return other_connection
--- a/hathor/p2p/protocol.py
+++ b/hathor/p2p/protocol.py
@@ -25,6 +25,7 @@ from twisted.protocols.basic import LineReceiver
 from twisted.python.failure import Failure
 
 from hathor.conf.settings import HathorSettings
+from hathor.p2p.connect_classes import ConnectionType
 from hathor.p2p.messages import ProtocolMessages
 from hathor.p2p.peer import PrivatePeer, PublicPeer, UnverifiedPeer
 from hathor.p2p.peer_endpoint import PeerEndpoint
@@ -108,7 +109,7 @@ class HathorProtocol:
         *,
         settings: HathorSettings,
         use_ssl: bool,
-        inbound: bool,
+        connection_type: ConnectionType,
     ) -> None:
         self._settings = settings
         self.my_peer = my_peer
@@ -120,8 +121,9 @@ class HathorProtocol:
         assert self.connections.reactor is not None
         self.reactor = self.connections.reactor
 
-        # Indicate whether it is an inbound connection (true) or an outbound connection (false).
-        self.inbound = inbound
+        # Type of Connection
+        # 0 == Outgoing, 1 == Incoming, 2 == Bootstrap
+        self.connection_type = connection_type
 
         # Maximum period without receiving any messages.
         self.idle_timeout = self._settings.PEER_IDLE_TIMEOUT
--- a/hathor/simulator/fake_connection.py
+++ b/hathor/simulator/fake_connection.py
@@ -21,6 +21,7 @@ from OpenSSL.crypto import X509
 from structlog import get_logger
 from twisted.internet.address import IPv4Address, IPv6Address
 from twisted.internet.testing import StringTransport
+from typing_extensions import assert_never
 
 from hathor.p2p.peer import PrivatePeer
 from hathor.p2p.peer_endpoint import PeerAddress, PeerEndpoint
@@ -273,7 +274,15 @@ class FakeConnection:
         self._buf2.clear()
 
         self._proto1 = self.manager1.connections.server_factory.buildProtocol(self.addr2)
-        self._proto2 = self.manager2.connections.client_factory.buildProtocol(self.addr1)
+
+        # If in bootstrap mode, now we instantiate the Bootstrap Factory.
+        match self._fake_bootstrap_id:
+            case False:
+                self._proto2 = self.manager2.connections.client_factory.buildProtocol(self.addr1)
+            case None | PeerId():
+                self._proto2 = self.manager2.connections.bootstrap_factory.buildProtocol(self.addr1)
+            case _:
+                assert_never(self._fake_bootstrap_id)
 
         # When _fake_bootstrap_id is set we don't pass the peer because that's how bootstrap calls
         # connect_to_endpoint()
--- a/hathor_tests/p2p/test_connection_slots.py
+++ b/hathor_tests/p2p/test_connection_slots.py
@@ -1,0 +1,243 @@
+from twisted.python.failure import Failure
+
+from hathor.p2p.connection_slot import SlotsManager, SlotsManagerSettings
+from hathor.simulator import FakeConnection
+from hathor_tests.simulation.base import SimulatorTestCase
+
+
+class ConnectionSlotsTestCase(SimulatorTestCase):
+    def test_slot_limit(self) -> None:
+        """ Tests whether the slots and the pool stop increasing connections after cap is reached.
+
+            Important note: create_peer caps the peer pool at 100. Hence, if more than 100 connections
+            are opened (if settings.P2P_..._OUTGOING + settings.P2P_...INCOMING _+ ... exceed 100)
+            then the tests will fail."""
+
+        # Full-Node: May receive incoming connections, deliver outgoing connections, etc.
+        # If the total amount of connections exceeds 100, create_peer will not yield more than 100 peers as well.
+        full_node = self.create_peer()
+
+        # Set the limits for each slot (sum MUST NOT exceed 100).
+        # We set the limits here since grabbing the values from the settings exceeds limits in simulator.
+        max_outgoing = 20
+        max_incoming = 15
+        max_bootstrap = 10
+        max_connections = max_bootstrap + max_incoming + max_outgoing
+
+        # Set SlotsManager Settings:
+        slots_manager_settings = SlotsManagerSettings(
+            max_outgoing=max_outgoing,
+            max_incoming=max_incoming,
+            max_bootstrap=max_bootstrap,
+        )
+
+        # Set SlotsManager:
+        slots_manager = SlotsManager(slots_manager_settings)
+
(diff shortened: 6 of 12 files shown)


hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,164 @@
...
+    def is_in_slot(self, protocol: HathorSettings) -> bool:
+        return protocol in self.connection_slot
+
+
+@dataclass
+class SlotsManagerSettings:
+    max_outgoing: int
+    max_incoming: int
+    max_bootstrap: int
+
```
LFRezende: Check if sum is lower or equal than max conn

hathor/p2p/manager.py
```diff
@@ -480,9 +515,16 @@ def relay_peer_to_ready_connections(self, peer: PublicPeer) -> None:
 
     def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
         """Called when a peer disconnect."""
+
+        # Discard handles case when not in connections.
         self.connections.discard(protocol)
+
```
LFRezende: Check if we should remove connecitons

hathor/p2p/manager.py
```diff
@@ -675,13 +720,18 @@ def _connect_to_callback(
         peer: UnverifiedPeer | PublicPeer | None,
         endpoint: IStreamClientEndpoint,
         entrypoint: PeerEndpoint,
+        discovery_call: bool = False
     ) -> None:
         """Called when we successfully connect to a peer."""
         if isinstance(protocol, HathorProtocol):
+            if discovery_call:
+                protocol.connection_type = HathorProtocol.ConnectionType.BOOTSTRAP
             protocol.on_outbound_connect(entrypoint, peer)
```
LFRezende: Delete this, as the factory sets it already.

hathor/p2p/manager.py
```diff
@@ -730,10 +782,16 @@ def connect_to_endpoint(
...
+            if use_ssl:
+                factory = TLSMemoryBIOFactory(self.my_peer.certificate_options, True, self.discovered_factory)
+            else:
+                factory = self.discovered_factory
         else:
-            factory = self.client_factory
+            if use_ssl:
+                factory = TLSMemoryBIOFactory(self.my_peer.certificate_options, True, self.client_factory)
+            else:
+                factory = self.client_factory
```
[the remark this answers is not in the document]
LFRezende: 1. Take 1614, all that is not check ep in 1613, and updated
2. Put the comments of the discussion
3. Refactor with Gabriel
4. Verify if acceptance criteria is updated

hathor/p2p/manager.py
LFRezende: Done

hathor/p2p/manager.py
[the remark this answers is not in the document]
LFRezende: Done, but could not figure a simple way of setting the boolean without passing it as an argument to connect_to_endpoint, so copied the function of connect_to_endpoint and changed some of its inner structure. I'll definitely revisit this function.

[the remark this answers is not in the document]
LFRezende: Done above.

hathorlib/hathorlib/conf/settings.py
```diff
@@ -368,7 +368,12 @@ def _validate_token_deposit_percentage(cls, token_deposit_percentage: float) ->
     PEER_CONNECTION_RETRY_MAX_RETRY_INTERVAL: int = 300
 
     # Number max of connections in the p2p network
-    PEER_MAX_CONNECTIONS: int = 125
+    PEER_MAX_CONNECTIONS: int = 85
+
+    # Max Number of each connection slot (int):
+    P2P_PEER_MAX_INCOMING_CONNECTIONS: int = 50
+    P2P_PEER_MAX_OUTGOING_CONNECTIONS: int = 60
+    P2P_PEER_MAX_DISCOVERED_PEERS_CONNECTIONS: int = 15
```
glevco: Why the reduction from 125 to 85? We should add a validation on settings to check that `PEER_MAX_CONNECTIONS == P2P_PEER_MAX_INCOMING_CONNECTIONS + P2P_PEER_MAX_OUTGOING_CONNECTIONS + P2P_PEER_MAX_DISCOVERED_PEERS_CONNECTIONS`

hathor/p2p/connect_classes.py
```diff
@@ -0,0 +1,55 @@
...
+
+class ConnectionType(Enum):
+    """ Types of Connection as inputs for an instance of the Hathor Protocol. """
+    OUTGOING = 0
+    INCOMING = 1
+    BOOTSTRAP = 2
+
+    def is_outbound(self) -> bool:
+        """ If value is 1, then the connection is inbound. If not, outbound."""
+        return self.value != self.INCOMING
```
glevco, line 28: Do this instead, so this method keeps working if new types are added in the future:

```suggestion
    def is_outbound(self) -> bool:
        """ If value is 1, then the connection is inbound. If not, outbound."""
        return self.value in (ConnectionType.OUTGOING, ConnectionType.BOOTSTRAP)
```

hathor/p2p/connect_classes.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@dataclass
+class ConnectionRemoved:
+    reason: str
+
+
+@dataclass
+class ConnectionNotRemoved:
+    reason: str
```
glevco, line 55: Missing slots and frozen

hathor/p2p/protocol.py
```diff
@@ -120,8 +122,12 @@ def __init__(
...
         self.reactor = self.connections.reactor
 
-        # Indicate whether it is an inbound connection (true) or an outbound connection (false).
-        self.inbound = inbound
+        # Type of Connection
+        # 0 == Outgoing, 1 == Incoming, 2 == Discovered
+        self.connection_type = connection_type
+
+        # Connection State
+        self.connection_state = ConnectionState.CREATED
```
glevco, line 130: Why not just reuse `self.state` instead of this new attribute and type?

glevco (changes requested):

hathor/simulator/fake_connection.py
```diff
@@ -273,7 +273,12 @@ def reconnect(self) -> None:
         self._buf2.clear()
 
         self._proto1 = self.manager1.connections.server_factory.buildProtocol(self.addr2)
-        self._proto2 = self.manager2.connections.client_factory.buildProtocol(self.addr1)
+
+        # If in bootstrap mode, now we instantiate the Discovery Factory.
+        if self._fake_bootstrap_id is False:
+            self._proto2 = self.manager2.connections.client_factory.buildProtocol(self.addr1)
+        elif self._fake_bootstrap_id or self._fake_bootstrap_id is None:
+            self._proto2 = self.manager2.connections.discovered_factory.buildProtocol(self.addr1)
```
glevco: Use a match with `assert_never` instead

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+RemoveFromSlotResult = ConnectionRemoved | ConnectionNotRemoved
+
+
+class ConnectionSlots:
+    """ Class of a connection pool slot - outgoing, incoming, discovered connections. """
+    connection_slot: set[HathorProtocol]
+    type: ConnectionType
+    max_slot_connections: int
+
+    def __init__(self, type: ConnectionType, max_connections: int):
```
glevco: Missing `-> None` annotation (this prevents mypy from type checking the method)

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if max_connections <= 0:
+            raise ValueError("Slot max number must allow at least one connection")
```
glevco: Should be as assertion as this comes from settings

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+        return ConnectionAllowed(f"Added to slot {self.type}.")
+
+    def remove_connection(self, protocol: HathorProtocol) -> ConnectionRemoved:
+        """ Removes from given instance the protocol passed. Returns protocol from queue
+            when disconnection leads to free space in slot."""
+
+        # Discard does nothing if protocol not in connection_slot.
+        self.connection_slot.discard(protocol)
+        return ConnectionRemoved('Connection successfully removed.')
```
glevco: Missing `assert self.type == protocol.connection_type`

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def is_full(self) -> bool:
+        return len(self.connection_slot) >= self.max_slot_connections
+
+    def is_in_slot(self, protocol: HathorSettings) -> bool:
+        return protocol in self.connection_slot
+
+
+@dataclass
```
glevco: Missing slots, frozen, kw_only

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    outgoing_slot: ConnectionSlots
+    incoming_slot: ConnectionSlots
+    bootstrap_slot: ConnectionSlots
+
+    types_allowed: list[ConnectionType] = [
+        ConnectionType.OUTGOING,
+        ConnectionType.INCOMING,
+        ConnectionType.BOOTSTRAP,
+    ]
```
glevco: Use a `frozenset` instead

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+        self.bootstrap_slot = ConnectionSlots(ConnectionType.BOOTSTRAP, settings.max_bootstrap)
+
+    def add_to_slot(self, protocol: HathorProtocol) -> AddToSlotResult:
+        """Add received protocol to one of the slots.
+        If slot is full, protocol is not added. """
+
+        conn_type = protocol.connection_type
+        types = self.types_allowed
+
+        assert conn_type in types
```
glevco: `protocol.connection_type` is typed as `ConnectionType`, and `self.types_allowed` contains all possible values of `ConnectionType`. So type checking already guarantees that this assert is always true, you may remove that assert and the `types_allowed` entirely. Also, if a new type is added, the `assert_never` case on the `match` below would catch it.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+        slot: ConnectionSlots | None = None
```
glevco: It should never be `None`

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+        return status
+
+    def remove_from_slot(self, protocol: HathorProtocol) -> RemoveFromSlotResult:
+        """ Removes protocol from slot of same type.
+            If OUTGOING, INCOMING or BOOTSTRAP, simply remove from slot and disconnect.
+            Should be called by manager when disconnecting a protocol."""
+
+        conn_type = protocol.connection_type
+        assert conn_type in self.types_allowed
```
glevco: Same as above

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+        slot: ConnectionSlots | None = None
```
glevco: Same as above

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+            case ConnectionType.OUTGOING:
+                slot = self.outgoing_slot
+            case ConnectionType.INCOMING:
+                slot = self.incoming_slot
+            case ConnectionType.BOOTSTRAP:
+                slot = self.bootstrap_slot
+            case _:
+                assert_never(conn_type)
+
+        assert protocol in slot.connection_slot, 'Protocol not in slot.'
```
glevco: Move this to `slot.remove_connection`

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+        slot.remove_connection(protocol)
+        return ConnectionRemoved(reason=f'Connection on slot {slot.type} removed.')
```
glevco: Shouldn't you `return slot.remove_connection(...)`?

glevco (changes requested):

hathor/p2p/manager.py
```diff
@@ -129,14 +130,17 @@ def __init__(
...
         self.use_ssl = ssl
         self.server_factory = HathorServerFactory(
             self.my_peer, p2p_manager=self, use_ssl=self.use_ssl, settings=self._settings
         )
         self.client_factory = HathorClientFactory(
             self.my_peer, p2p_manager=self, use_ssl=self.use_ssl, settings=self._settings
         )
+        self.discovered_factory = HathorDiscoveredFactory(
+            self.my_peer, p2p_manager=self, use_ssl=self.use_ssl, settings=self._settings
+        )
```
glevco: The factory is called "discovered" but the `ConnectionType` is called `BOOTSTRAP`. Please unify those namings.

hathor/p2p/manager.py
```diff
@@ -272,8 +286,13 @@ def do_discovery(self) -> None:
...
         Do a discovery and connect on all discovery strategies.
         """
+        def connect_to_bootstrap(entrypoint: PeerEndpoint, peer: UnverifiedPeer | PublicPeer | None = None,
+                                 use_ssl: bool | None = None) -> None:
+
+            self.connect_to_endpoint(entrypoint, peer, use_ssl=use_ssl, discovery_call=True)
+
         for peer_discovery in self.peer_discoveries:
-            coro = peer_discovery.discover_and_connect(self.connect_to_endpoint)
+            coro = peer_discovery.discover_and_connect(connect_to_bootstrap)
```
glevco: I just noticed the `discover_and_connect` method takes a `Callable[[PeerEndpoint], None]`, so we can simply remove the `peer` and `use_ssl` args from `connect_to_bootstrap`

hathor/p2p/manager.py
```diff
@@ -412,13 +431,28 @@ def on_connection_failure(self, failure: Failure, peer: Optional[UnverifiedPeer
...
     def on_peer_connect(self, protocol: HathorProtocol) -> None:
         """Called when a new connection is established."""
+
+        # Checks whether connections in the network are at limit.
         if len(self.connections) >= self.max_connections:
             self.log.warn('reached maximum number of connections', max_connections=self.max_connections)
             protocol.disconnect(force=True)
             return
+
+        connection_status: ConnectionAllowed | ConnectionRejected
```
glevco: Unnecessary line

hathor/p2p/manager.py
```diff
@@ -412,13 +431,28 @@ def on_connection_failure(self, failure: Failure, peer: Optional[UnverifiedPeer
...
+
+        connection_status = self.slots_manager.add_to_slot(protocol)
+
+        if isinstance(connection_status, ConnectionRejected):
+            self.log.warn('Connection Rejected')
+            protocol.disconnect(force=True)
+            return
```
glevco, line 444: Use an exhaustive `match` instead

hathor/p2p/manager.py
```diff
@@ -480,9 +516,16 @@ def relay_peer_to_ready_connections(self, peer: PublicPeer) -> None:
 
     def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
         """Called when a peer disconnect."""
+
+        # Discard handles case when not in connections.
         self.connections.discard(protocol)
+
+        # Each conn is from a slot - discard from it as well.
+        self.slots_manager.remove_from_slot(protocol)
```
glevco, line 520: We have other places that call `protocol.disconnect` too, for example in `disconnect_all_peers`. Shouldn't they also call `remove_from_slot`? It also looks like this method doesn't remove them from `self.connections`, could you please check that?

hathor/p2p/manager.py
```diff
@@ -499,21 +542,24 @@ def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
...
+
         self.pubsub.publish(
             HathorEvents.NETWORK_PEER_DISCONNECTED,
             protocol=protocol,
             peers_count=self._get_peers_count()
         )
 
+        # SUGGESTION: To make a NETWORK_PEER_DEQUEUED. It would be clearer, since in the case of a dequeue,
+        # the order of events would be "NETWORK_PEER_READY" and then "NETWORK_PEER_CONNECTED", which is
+        # the opposite.
```
glevco: Remove?

hathor/p2p/manager.py
```diff
@@ -675,6 +721,7 @@ def _connect_to_callback(
         peer: UnverifiedPeer | PublicPeer | None,
         endpoint: IStreamClientEndpoint,
         entrypoint: PeerEndpoint,
+        discovery_call: bool = False
```
glevco, line 716: Remove

hathor/p2p/manager.py
```diff
@@ -819,17 +867,22 @@ def get_connection_to_drop(self, protocol: HathorProtocol) -> HathorProtocol:
         assert protocol.peer.id is not None
         assert protocol.my_peer.id is not None
         other_connection = self.connected_peers[protocol.peer.id]
+        _outbound_types = (
+            ConnectionType.OUTGOING,
+            ConnectionType.BOOTSTRAP,
+        )
+        is_outbound = protocol.connection_type in _outbound_types
```
glevco: Use `protocol.is_outbound()`

glevco (changes requested):

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def slot_number(self, slot: ConnectionSlots) -> int:
+        return len(slot.connection_slot)
+
+    def slot_size(self, slot: ConnectionSlots) -> int:
+        return slot.max_slot_connections
```
glevco: These are unused, should we remove them?

hathor/p2p/manager.py
glevco, line 520: We don't use the returned value from `remove_from_slot`, should we? Otherwise can we remove it?

hathor_tests/p2p/test_connection_slots.py
```diff
@@ -0,0 +1,196 @@
...
+            in_peerList.append(self.create_peer())
+
+        # Generate incoming connections - full_node is the target.
+        in_connList = []
+        for i in range(0, max_incoming + 1):
+            in_connList.append(FakeConnection(full_node, in_peerList[i]))
+
+        # Add connections to simulator.
+        for i in range(len(in_connList)):
+            self.simulator.add_connection(in_connList[i])
```
glevco, line 47: Just do it in a single loop

hathor_tests/p2p/test_connection_slots.py
```diff
@@ -0,0 +1,196 @@
...
+        out_peerList = []
+        for _ in range(max_outgoing + 1):
+            out_peerList.append(self.create_peer())
+
+        out_connList = []
+        for i in range(0, max_outgoing + 1):
+            out_connList.append(FakeConnection(out_peerList[i], full_node))
+
+        for i in range(len(out_connList)):
+            self.simulator.add_connection(out_connList[i])
```
glevco: Same

hathor_tests/p2p/test_connection_slots.py
```diff
@@ -0,0 +1,196 @@
...
+        bootstrap_peerList = []
+        for _ in range(max_bootstrap + 1):
+            bootstrap_peerList.append(self.create_peer())
+
+        bootstrap_connList = []
+        for i in range(0, max_bootstrap + 1):
+            bootstrap_connList.append(FakeConnection(bootstrap_peerList[i], full_node, fake_bootstrap_id=None))
+
+        for i in range(len(bootstrap_connList)):
+            self.simulator.add_connection(bootstrap_connList[i])
```
glevco, line 80: Same

hathor_tests/p2p/test_connection_slots.py
```diff
@@ -0,0 +1,196 @@
...
+        in_conn_list = []
+        for _, peer in enumerate(in_peer_list):
+            in_conn_list.append(FakeConnection(full_node, peer))
+            self.simulator.add_connection(in_conn_list[-1])
+
+        # Bootstrap:
+        boot_conn_list = []
+        for _, peer in enumerate(boot_peer_list):
+            boot_conn_list.append(FakeConnection(peer, full_node, fake_bootstrap_id=None))
+            self.simulator.add_connection(boot_conn_list[-1])
```
glevco, line 182: Same

hathor_tests/p2p/test_connection_slots.py
glevco, line 1: Missing tests for peer removal after disconnection

glevco (changes requested):

hathorlib/hathorlib/conf/settings.py
LFRezende: The limits has been set back to 125. When running tests with the values on settings, there is a limit on the tests executed in the connection pool. It caps the number of fake protocols open to 100, so that's why I did so earlier.

About the method, just added one.

hathor/p2p/connect_classes.py
LFRezende: Better, just did that, thanks.

hathor/p2p/connect_classes.py
LFRezende: Indeed, just added.

hathor/p2p/connection_slot.py
LFRezende: Done

hathor/p2p/connection_slot.py
LFRezende: Done

hathor/p2p/connection_slot.py
LFRezende: This will have to be changed when we go to CheckEntrypoints in PR 1614, but we can do it for now. Done.

hathor/p2p/connection_slot.py
LFRezende: I'll keep the frozen and slots, but kw_only I'll skip since it gets too verbose and the linter will flag it.

hathor/p2p/connection_slot.py
LFRezende: Good idea. Just did so.

hathor/p2p/connection_slot.py
LFRezende: Fair enough. I'll delete the types allowed parameter from the class.

hathor/p2p/connection_slot.py
LFRezende: Removed.

hathor/p2p/connection_slot.py
LFRezende: Removed.

hathor/p2p/connection_slot.py
LFRezende: Done

hathor/p2p/connection_slot.py
LFRezende: Damn, you're right. No idea what I was thinking haha.

Done.

hathor/p2p/manager.py
LFRezende: Just did. All instances of Discovered turned into Bootstrap.

hathor/p2p/manager.py
LFRezende: Fair enough, just did so.

hathor/p2p/manager.py
LFRezende: By placing these on `on_peer_disconnect`, which is called always when a `protocol.disconnect` happens, we ensure this disconnected protocol is discarded both from the ` self.connections ` and from its respective slot - I'm not sure if I misunderstood you, but I see no point in adding `remove_from_slot` to other disconnect functions that already call `protocol.disconnect`, as all of them will then eventually reach `on_peer_disconnect`.

Additionally, you are correct in saying that there is no removal from `self.connections` in the `remove_from_slot` method. But we leave this to the connections manager, not the slots manager. `on_peer_disconnect` discards the protocol from the connections set of protocols of the manager.

hathor/p2p/manager.py
LFRezende: Done.

hathor/p2p/manager.py
LFRezende: But how will then the `connect_to_bootstrap` function call the `self.bootstrap_factory`, if there is no way for the `connect_to_endpoint` function know we're calling it from `connect_to_bootstrap `function?



```python
def do_discovery(self) -> None:
        """
        Do a discovery and connect on all discovery strategies.
        """
        def connect_to_bootstrap(entrypoint: PeerEndpoint) -> None:
            self.connect_to_endpoint(entrypoint, discovery_call=True)

        for peer_discovery in self.peer_discoveries:
            coro = peer_discovery.discover_and_connect(connect_to_bootstrap)
            Deferred.fromCoroutine(coro)
```

hathor/p2p/manager.py
LFRezende: Good call. It'd actually be `protocol.connection_type.is_outbound()`, fixed and took away the outbound_types.

hathor/p2p/connection_slot.py
LFRezende: Yeah, we'll change them for other getter functions in the future.

Just took them out.

hathor/p2p/manager.py
LFRezende: I intended to leave there for future tests. Also, in the future, with check_entrypoints, we'll need to return the statuses. I'll leave it like that for now.

hathor_tests/p2p/test_connection_slots.py
LFRezende: Indeed. Done.

hathor_tests/p2p/test_connection_slots.py
LFRezende: Done.

hathor_tests/p2p/test_connection_slots.py
LFRezende: Done.

hathor_tests/p2p/test_connection_slots.py
LFRezende: Here is actually not possible since the indexes do not match the same value.

hathor_tests/p2p/test_connection_slots.py
LFRezende: Brought one from the 1614 to this one. Passes.

hathor/p2p/protocol.py
LFRezende: At the time I wanted a separate Enum to differentiate between ping-pong/handshaking and the lifecycle of protocols in the slots context. I guess we could unify them - would be eventuall the same work and we'd delete unnecessary code.

Since it changes much of the code and tests, I'll wait for other comments from Marcelo and Jan before I batch all these fixes in a next commit. For now I'll leave it like that, but we can change it, I believe.

hathor/simulator/fake_connection.py
LFRezende: Match feels finicky here, as the id can be False, None or 'Something'. But if it is passed "True" and not a PeerId, it should not pass, for example. So some sort of is_instance would be needed.

I instead added an `else ValueError`.

hathor/p2p/manager.py
LFRezende: Taken out.

hathor/p2p/manager.py
LFRezende: What would be the advantage in this case? 
It is only one case for matching, and there is no assert_never gains.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self.connection_slot.add(protocol)
+
+        return ConnectionAllowed(f"Added to slot {self.type}.")
+
+    def remove_connection(self, protocol: HathorProtocol) -> ConnectionRemoved:
+        """ Removes from given instance the protocol passed. Returns protocol from queue
+            when disconnection leads to free space in slot."""
+
+        assert self.type == protocol.connection_type
```
msbrogli: Be careful here: When a protocol is transitioning to another type, it can try to remove itself from the previous slot and trigger an assertion error.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,139 @@
...
+        assert protocol in self.connection_slot, 'Protocol not in slot.'
+
+        # Discard does nothing if protocol not in connection_slot.
+        self.connection_slot.discard(protocol)
+        return ConnectionRemoved('Connection successfully removed from slot.')
+
+    def is_full(self) -> bool:
```
msbrogli, line 70: Missing docstring.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,139 @@
...
+        return len(self.connection_slot) >= self.max_slot_connections
+
+    def is_in_slot(self, protocol: HathorSettings) -> bool:
```
msbrogli: You could have used `__in__` instead.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,139 @@
...
+        self.connection_slot.add(protocol)
+
+        return ConnectionAllowed(f"Added to slot {self.type}.")
+
+    def remove_connection(self, protocol: HathorProtocol) -> ConnectionRemoved:
+        """ Removes from given instance the protocol passed. Returns protocol from queue
+            when disconnection leads to free space in slot."""
+
+        assert self.type == protocol.connection_type
+        assert protocol in self.connection_slot, 'Protocol not in slot.'
```
msbrogli: You can have a remove and a discard method, the only difference is that discard fails silently.

hathor/p2p/manager.py
```diff
@@ -157,6 +161,16 @@ def __init__(
         # List of peers connected and ready to communicate.
         self.connected_peers = {}
 
+        # Instances of Connection Slots
+        max_outgoing: int = settings.P2P_PEER_MAX_OUTGOING_CONNECTIONS
+        max_incoming: int = settings.P2P_PEER_MAX_INCOMING_CONNECTIONS
+        max_bootstrap: int = settings.P2P_PEER_MAX_BOOTSTRAP_PEERS_CONNECTIONS
+
+        slots_manager_settings = SlotsManagerSettings(max_outgoing, max_incoming, max_bootstrap)
```
msbrogli: I guess it would be even cleaner to do it in the builder and inject `SlotsManager` directly.

hathor/p2p/manager.py
```diff
@@ -272,8 +286,11 @@ def do_discovery(self) -> None:
         """
         Do a discovery and connect on all discovery strategies.
         """
+        def connect_to_bootstrap(entrypoint: PeerEndpoint) -> None:
```
msbrogli: Why are you using a nested function (instead of the previous method)?

hathor/p2p/manager.py
```diff
@@ -480,9 +512,16 @@ def relay_peer_to_ready_connections(self, peer: PublicPeer) -> None:
 
     def on_peer_disconnect(self, protocol: HathorProtocol) -> None:
         """Called when a peer disconnect."""
+
+        # Discard handles case when not in connections.
         self.connections.discard(protocol)
+
+        # Each conn is from a slot - discard from it as well.
+        self.slots_manager.remove_from_slot(protocol)
```
msbrogli, line 491: This call might raise assertion failures and other errors. Should we protect against these?

msbrogli (dismissed):

hathor/p2p/connection_slot.py
LFRezende: Good catch, but don't worry - in 1613, no protocol  shifts types. In 1614, the mechanism deals with it. Nevertheless I believe it is worth a change before pushing to prod. No protocols in 1613 will shift but we can standardize.

hathor/p2p/connection_slot.py
LFRezende: Removed is_in_slot and added __in__ dunder method.

hathor/p2p/connection_slot.py
LFRezende: Added.

hathor/p2p/connection_slot.py
LFRezende: I disliked the idea of having a silent failing system, so I left only the remove method. if you think i'd be best having both, let me know.

hathor/p2p/manager.py
LFRezende: Done. I created some getter methods for the Builder and also added directly these lines to the CLiBuilder and ran the tests in the p2p folder - seem to work fine.

hathor/p2p/manager.py
LFRezende: So we can set the connect_to_endpoint function with a discovery_call=True flag, and change the factory used when instantiating a new protocol. The protocol needs to be of type BOOTSTRAP, which is emitted solely by bootstrap_factories. When creating the protocol, this is set with the discovery_call flag set to true, but for this to happen we wrap it into a nested function.

In connect_to_endpoint:

```python

factory: IProtocolFactory = self.bootstrap_factory if discovery_call else self.client_factory
```

But this is done so only when called by do_discovery - which needs to build bootstrap-typed protocols, thence calling `connect_to_endpoint ` with `discovery_call=True`

hathor/p2p/manager.py
LFRezende: I took away the assert when protocol is not present, but left the one with different type. I believe that if types are mixing (without CheckEntrypoints mechanism present), this should be flagged. But the removal of a non-existent protocol does not pose a problem.

hathorlib/hathorlib/conf/settings.py
glevco: Thanks!

hathor/p2p/manager.py
LFRezende: I did self instead of self.value.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,138 @@
...
+    """ Class of a connection pool slot - outgoing, incoming, bootstrap connections. """
+    connection_slot: set[HathorProtocol]
+    type: ConnectionType
+    max_slot_connections: int
+
+    def __init__(self, type: ConnectionType, max_connections: int) -> None:
+
+        assert max_connections > 0, 'Slot max number must allow at least one connection'
+
+        self.type = type
```
jansegre: Maybe we can remove this field. It's only used for asserts (which are double checks), for the strings we can just use the `protocol_connection_type` directly.

hathor/p2p/manager.py
jansegre: The alternative would be a dedicated `connect_to_bootstrap` dedicated method instead of a "loose function". I think I'd prefer the dedicated method as well. I get the `discovery_call=True` argument, but a method can do exactly that as well, and it's more "testable" and not a new function on every `do_discovery` call.

hathor/p2p/manager.py
jansegre: In the cases where the return value is useful for for a test, but not in general, I would rather have something like a `_remove_from_slot_result` with the return value that the test calls, and a `remove_from_slot` that calls the previous function and ignores the result, that way the exposed API stays "sane".

jansegre (dismissed):

hathor/p2p/connection_slot.py
LFRezende: Done.

hathor/simulator/fake_connection.py
glevco: That's not a problem, `isinstance` is not necessary:

```python
        match self._fake_bootstrap_id:
            case False:
                self._proto2 = self.manager2.connections.client_factory.buildProtocol(self.addr1)
            case None | PeerId():
                self._proto2 = self.manager2.connections.bootstrap_factory.buildProtocol(self.addr1)
            case _:
                assert_never(self._fake_bootstrap_id)
```

hathor/p2p/connection_slot.py
glevco: Still missing.

hathor/p2p/connection_slot.py
glevco: I think `kw_only` is specially important in this case of class because all its 3 fields are of the same type `int`. It's easy to accidentally mess up the order and the linter would pass and code would work, but the limits would be swapped. I would say it's the most important of the 3 dataclass configs.

What do you mean by "the linter will flag it"?

hathor/p2p/manager.py
glevco: Yes there is, the `add_to_slot` method returns `AddToSlotResult = ConnectionAllowed | ConnectionRejected`. So you can do the disconnect and return for `ConnectionRejected`, `pass` for `ConnectionAllowed`, and `assert_never` to future-proof against any updates in that code.

hathor/p2p/manager.py
glevco: This parameter is not used in this method.

hathor/p2p/manager.py
glevco: You're right! Sorry.

hathor/p2p/connection_slot.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, type: ConnectionType, max_connections: int) -> None:
+
+        assert max_connections > 0, 'Slot max number must allow at least one connection'
+
+        self.connection_slot = set()
+        self.max_slot_connections = max_connections
+
+    def __in__(self, protocol: HathorProtocol) -> bool:
+        return protocol in self.connection_slot
```
glevco: This doesn't work. Did you mean to define `__contains__` instead? But I believe it can just be removed since it's unused.

LFRezende: I updated it to `__contains__`, thanks. I'll leave it like this for testing and 1614 I've previously updated this same dunder method.

hathor/p2p/manager.py
LFRezende: I followed @jansegre's suggestion and wrapped it up with a `_remove_from_slot_result`.

hathor/p2p/connection_slot.py
LFRezende: I removed it entirely, following @msbrogli 's suggestion. We'll let removal of non-present protocols to silently fail in discard.

hathor/p2p/connection_slot.py
LFRezende: I followed @jansegre 's suggestion and took self.type entirely for now, and removed the type assertions. @msbrogli 's also shown concern on type assertion, so it made sense to me to take it out entirely.

hathor/simulator/fake_connection.py
LFRezende: Great suggestion, added it.

hathor/p2p/manager.py
LFRezende: I used the alterate method Jan suggested.

hathor/p2p/connection_slot.py
LFRezende: I took the protocol in slot safeguard out in removal - we may discard it solely, letting it to silently fail if not there.

hathor/p2p/manager.py
LFRezende: Also, took away type asserts.

hathor/p2p/manager.py
LFRezende: It is used in connect_to_endpoint, tot select the factory:

```python

factory: IProtocolFactory = self.bootstrap_factory if discovery_call else self.client_factory
```

Which is used by `connect_to_bootstrap` to configure `connect_to_endpoint`:

```python

def connect_to_bootstrap(self, entrypoint: PeerEndpoint) -> None:
        """Connect to a bootstrap entrypoint discovered via a PeerDiscovery strategy."""
        self.connect_to_endpoint(entrypoint, discovery_call=True)
```

hathor/p2p/manager.py
LFRezende: Done!

hathor/p2p/connection_slot.py
LFRezende: Nah nevermind, I was being silly. Added, and refactored.

hathor/p2p/protocol.py
LFRezende: It is funny - it is not used in this PR, only in CheckEntrypoints in 1614. I'll delete it from this one, since we do not read in any place the actual connection state for now.

hathor/p2p/manager.py
LFRezende: Line code matched precisely with connect_to_endpoint's discovery_call boolean - misread.

Removed.

glevco (dismissed):

jansegre (dismissed):

hathor/p2p/connection_slot.py
LFRezende: Update: I took away self.type, it makes no sense having it for this PR when it will only be truly useful in 1614.

msbrogli: Ok. Thanks!

hathor/p2p/connection_slot.py
msbrogli: Thanks!

hathor/p2p/connection_slot.py
msbrogli: Thanks!

hathor/p2p/connection_slot.py
msbrogli: Ok. Thanks!

hathor/p2p/manager.py

hathor/p2p/manager.py
msbrogli: Great. Thanks!

hathor/p2p/manager.py
msbrogli: Ok. Thanks!

msbrogli (dismissed):

jansegre (approved):

PR #1625, opened 2026-03-18
feat(nano): implement Blueprint V2 feature

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1624

### Motivation

Add general support for updating the Blueprint API via Feature Activation (changing existing methods), and specific support for Blueprint V2, with updates on the `Context` API (details on the acceptance criteria below). The goal of the `Context` API change is to simplify the handling of actions for users. The current API is error-inducing, and could use better names.

The mechanism for updating user-facing APIs in Blueprints allows the coexistence of multiple versions of Blueprints. OCBs deployed before the feature activation continue to use the old API without any compatibility issues. OCBs deployed after the feature activation have to necessarily use the new API. This is done by extracting the feature state from the OCB's first_block's parent, NOT from the Nano tx's first_block's parent (in contrast to the `NanoRuntimeVersion`).

### Acceptance Criteria

- Add new feature `BLUEPRINT_V2` and settings `ENABLE_BLUEPRINT_V2`, currently disabled in all networks.
- Refactor some `BlueprintService` methods and add retrieval of the Blueprint's version using the state of the OCB's first block's parent.
- Refactor `NCBlueprintCatalog` so it also holds blueprint versions.
- Update and introduce new `Context` API:
  - New `Context.blueprint_version` property, which returns the Blueprint's version. Only available in V2.
  - Deprecate `Context.actions` (only available in V1). Replaced by `Context.actions_by_token` in V2.
  - Deprecate `Context.actions_list` (only available in V1). Replaced by `Context.all_actions` in V2.
  - Change behavior of `Context.get_single_action()`. In V1, it allowed N actions as long as they're for different tokens. In V2, only 1 action is allowed in the whole context.
  - New `Context.get_token_single_action()`, which mimics the behavior of `get_single_action()` from V1, that is, it returns the single action of a token, but allows other actions for other tokens. Only available in V2.
- Rename `Context.copy` to `__prepare_for_new_runner_call__ `. We must reject any OCBs that try to use it.
- Extra: change behavior of the `NANO_RUNTIME_V2` feature, so it uses the state of the first block's parent instead of the first block, for consistency between all 3 features (`NANO_CONTRACTS`, `NANO_RUNTIME_V2`, and `BLUEPRINT_V2`).

### Checklist


jansegre (dismissed):

jansegre (dismissed):

jansegre (approved):

hathor/feature_activation/utils.py
```diff
@@ -61,13 +63,15 @@ def from_vertex(*, settings: HathorSettings, feature_service: FeatureService, ve
         nano_runtime_version = (
             NanoRuntimeVersion.V2 if feature_is_active[Feature.NANO_RUNTIME_V2] else NanoRuntimeVersion.V1
         )
+        blueprint_version = BlueprintVersion.V2 if feature_is_active[Feature.BLUEPRINT_V2] else BlueprintVersion.V1
```
msbrogli, line 66: I guess the OCB vertex should explicitly choose a version. We can remove versions as part of a feature activation if we don't want it to be used anymore.

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -173,7 +173,7 @@ def execute_block(self, block: Block) -> Iterator[NCBlockEffect]:
 
         nc_sorted_calls = self._nc_calls_sorter(block, nc_calls) if nc_calls else []
         block_storage = self._nc_storage_factory.get_block_storage(parent_root_id)
-        features = Features.from_vertex(settings=self._settings, feature_service=self._feature_service, vertex=block)
+        features = Features.from_vertex(settings=self._settings, feature_service=self._feature_service, vertex=parent)
```
msbrogli, line 176: Why this change? Which feature activation is used in this part of the code? Couldn't it cause inconsistencies with the mainnet?

hathor/nanocontracts/blueprint_service.py
```diff
@@ -77,44 +80,73 @@ def get_blueprint_class(self, blueprint_id: BlueprintId) -> type[Blueprint]:
...
+        # nanos that depend on this Blueprint anyway.
+        # Considering BlueprintVersion.V2 for example, this means that OCBs deployed before its activation date can be
+        # confirmed by either V1 or V2 blocks, but OCBs deployed after the activation date are guaranteed to be V2.
+        features = Features.from_vertex(
+            settings=self.settings,
+            feature_service=self.feature_service,
+            vertex=first_block_parent,
         )
-        return self.get_on_chain_blueprint(blueprint_id)
+        return ocb.get_blueprint_class(), features.blueprint_version
```
msbrogli, line 128: I don't like implying the version from the feature activation. I guess the feature activation should add/remove versions and the OCBs must explicitly choose a version (for compatibility, if a version is not provided, we can consider it to be V1).

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -63,6 +63,7 @@ def __init__(self,
                 fee_tokens=False,
                 opcodes_version=OpcodesVersion.V1,
                 nano_runtime_version=NanoRuntimeVersion.V1,
+                blueprint_version=BlueprintVersion.V1,
```
msbrogli, line 66: What if a block confirms a deprecated blueprint version?

hathor/transaction/headers/nano_header.py
```diff
@@ -223,11 +224,12 @@ def get_actions(self) -> list[NCAction]:
         """Get a list of NCActions from the header actions."""
         return [header_action.to_nc_action(self.tx) for header_action in self.nc_actions]
 
-    def get_context(self) -> Context:
+    def get_context(self, blueprint_version: BlueprintVersion | None = None) -> Context:
```
msbrogli, line 227: We won't need it anymore.

hathorlib/hathorlib/nanocontracts/runner/runner.py
```diff
@@ -648,7 +649,8 @@ def _execute_public_method_call(
         # This ensures that, even if the blueprint method attempts to exploit or alter the context, it cannot
         # impact the original context. Since the runner relies on the context for other critical checks, any
         # unauthorized modification would pose a serious security risk.
-        ret = self._metered_executor.call(method, args=(ctx.copy(), *args))
+        ctx = ctx.__prepare_for_new_runner_call__(blueprint_version)
```
msbrogli, line 652: It won't be needed anymore.

hathorlib/hathorlib/nanocontracts/context.py
```diff
@@ -104,30 +140,103 @@ def get_caller_contract_id(self) -> ContractId | None:
...
-        if actions is None or len(actions) != 1:
+        """
+        Utility method to get the single action for the provided token,
+        and that is the only action in the whole Context.
+
+        - If there are no actions, this method will fail.
+        - If there are any other actions for the provided token, this method will fail.
+        - If there are any other actions for any other tokens, this method will fail.
+        """
+        match self.__blueprint_version:
```
msbrogli, line 197: Like BlueprintV1 and BlueprintV2, should we have UtilsV1 and Utilsv2 to avoid this checks?

msbrogli (changes requested):

hathor/feature_activation/utils.py
glevco: What would be the advantage, in this case? The new API is supposed to be better for security, so there's no reason to keep V1. If we're just going to introduce V2 and deprecate V1 in another Feature Activation, it's analogous to what's implemented in this PR, with the extra step of having to introduce a new VertexHeader to configure the version in the OCB.

hathor/nanocontracts/execution/block_executor.py
glevco: This is the implementation of this point from the Acceptance Criteria:

> - Extra: change behavior of the `NANO_RUNTIME_V2` feature, so it uses the state of the first block's parent instead of the first block, for consistency between all 3 features (`NANO_CONTRACTS`, `NANO_RUNTIME_V2`, and `BLUEPRINT_V2`).

It doesn't affect mainnet because the `features` variable here is only used below, on:

```python
result = self.execute_transaction(
    runtime_version=features.nano_runtime_version,
    tx=tx,
    block_storage=block_storage,
    rng_seed=rng_seed,
)
```

That is, it only changes the behavior of the nano runtime version, which is not activated yet, and therefore it's safe to change. The nano runtime version test is also updated to reflect this.

hathor/nanocontracts/blueprint_service.py
glevco: This is the same behavior for Opcodes V2 and Nano Runtime V2. Together with Blueprint V2, all 3 features work the same: when a new version is activated it becomes the current version, and the only one that can be used. Also, there's a fourth feature that will share this logic, DAA V2 in your PR.

PR #1628, opened 2026-03-19
feat(nano): implement action authorization on Blueprint V2

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1625

### Motivation

Harden Nano security by implementing mandatory action authorization in Blueprints V2. This means any action in the context that is not explicitly authorized by calling `ctx.authorize(action)` will make the Nano tx fail execution.

### Acceptance Criteria

- Add mandatory action authorization to the `BlueprintVersion.V2` feature, through new methods `Context.authorize(action)` and `Context.is_authorized(action)`, only available in V2.
- Rename `Context.create_from_vertex` to `__create_from_vertex__`. We must reject any OCBs that try to use it.

### Checklist


jansegre (approved):
Please add a comment to https://github.com/HathorNetwork/ops-tools/pull/1109 with the impact to the OCB review. I'll update the guide accordingly.

hathorlib/hathorlib/nanocontracts/context.py
```diff
@@ -227,17 +231,56 @@ def __legacy_get_single_action(self, token_uid: TokenUid) -> NCAction:
...
+                        raise NCFail(f'unauthorized action: {action}')
+                assert len(self.__all_actions__) == len(self.__authorized_actions)
+            case _:  # pragma: no cover
+                assert_never(self.__blueprint_version)
+
+    def authorize(self, action: NCAction) -> None:
+        """Authorize an action."""
+        match self.__blueprint_version:
+            case BlueprintVersion.V1:
+                raise NCFail('`Context.authorize` is not supported yet.')
```
msbrogli, line 266: I'm ok failing but it could be a no-op as well.

hathorlib/hathorlib/nanocontracts/context.py
```diff
@@ -227,17 +231,56 @@ def __legacy_get_single_action(self, token_uid: TokenUid) -> NCAction:
...
+            case BlueprintVersion.V2:
+                if action in self.__authorized_actions:
+                    raise NCFail(f'action is already authorized: {action}')
```
msbrogli, line 269: I'm not sure what's a better dev ux here. No-op vs Failing?

msbrogli (approved):

hathorlib/hathorlib/nanocontracts/context.py
glevco: I believe this may prevent mistakes. There's no reason to authorize an action twice, it's likely the dev meant to authorize another action. And he can check with `is_authorized` to prevent this error.

hathorlib/hathorlib/nanocontracts/context.py
glevco: The method doesn't exist currently, so if blueprints call it, they'll fail. This replicates that behavior, and also prevents mistakes: we shouldn't allow them to call authorize without actually running any authorization mechanism.

PR #1629, opened 2026-03-20
chore(linter): replace flake8 and isort with ruff

jansegre:
### Motivation

We want to consolidate Python linting and import sorting under Ruff so the project relies on a single, faster toolchain instead of separate `flake8`, `isort`, and several grep-based custom checks. This keeps existing developer and CI entry points intact while moving project-specific lint policies into centrally managed Ruff configuration.

### Acceptance Criteria

- `make check` and `make dcheck` should validate Python code with Ruff instead of `flake8` and `isort`, while the legacy `make flake8` and `make isort-check` targets should continue to work as compatibility aliases.
- some specific lints that were previously implemented in `extras/custom_checks.sh` should run as dedicated Ruff checks with the same scope and allowlists preserved, while the shell custom checks should continue covering the remaining rules for version consistency and enum comparisons.
- The repository should declare `ruff` as the development linting dependency, remove `flake8` and `isort` from project configuration and the lockfile.

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,13 +76,16 @@ mypy:
 dmypy:
 	dmypy run --timeout 86400 -- -p hathor -p hathor_tests -p extras.custom_tests
 
+.PHONY: ruff
+ruff:
+	ruff check $(py_sources)
+
 .PHONY: flake8
-flake8:
-	flake8 $(py_sources)
+flake8: ruff
 
 .PHONY: isort-check
 isort-check:
-	isort --ac --check-only $(py_sources)
+	ruff check $(py_sources) --select I
 
 .PHONY: yamllint
 yamllint:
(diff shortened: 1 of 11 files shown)


msbrogli (dismissed):

Makefile
glevco, line 1: The description says `make flake8` and `make isort-check` are kept as alias, but they're not. But `make isort` is. I think we should unify either keeping all or removing all (I'm fine with this).

pyproject.toml
```diff
@@ -90,13 +89,26 @@ packaging = "=26.0"
...
-known_first_party = "hathor,hathor_tests"
-line_length = 119
-multi_line_output = 3
+[tool.ruff]
+line-length = 119
+target-version = "py311"
+
+[tool.ruff.lint]
+select = ["E", "F", "W", "I"]
+ignore = ["E731"]
```
glevco, line 100: Add comments explaining each?

pyproject.toml
```diff
@@ -90,13 +89,26 @@ packaging = "=26.0"
...
+known-first-party = ["hathor", "hathor_tests", "hathorlib"]
+
+[tool.ruff.lint.per-file-ignores]
+"hathor/__init__.py" = ["F401"]
+"hathor/_openapi/register.py" = ["F401"]
+"hathor/event/websocket/protocol.py" = ["E701"]
+"hathor/nanocontracts/exception.py" = ["I001"]
+"hathor/nanocontracts/storage/__init__.py" = ["I001"]
+"hathor/nanocontracts/storage/backends.py" = ["I001"]
+"hathor/nanocontracts/storage/factory.py" = ["I001"]
```
glevco, line 117: Add comments explaining each?

Makefile
jansegre: I've added those aliases.

pyproject.toml
jansegre: Done.

pyproject.toml
jansegre: Done.

glevco (approved):

PR #1636, opened 2026-03-25
feat: add REDUCE_DAA_TARGET feature activation for 4x faster blocks

msbrogli:
### Motivation

Implements the design in https://github.com/HathorNetwork/rfcs/pull/99

Reduce the DAA target block time from 30s to 7.5s (4x faster blocks) via the feature activation system. This enables higher throughput while maintaining the same inflation rate by proportionally reducing block rewards.

### Acceptance Criteria

- Add `REDUCE_DAA_TARGET` to the `Feature` enum with consensus handling (no-op for tx verification, only affects DAA parameters)
- DAA receives `FeatureService` via dependency injection; checks feature state internally
- Configurable target via `REDUCED_AVG_TIME_BETWEEN_BLOCKS_10X` setting (default 75 = 7.5s)
- Reward reduction factor: `(AVG_TIME * 10) // REDUCED_10X` (e.g. 300 // 75 = 4x)
- Two-context pattern: `is_feature_active(block)` for verification, `is_feature_active_for_next_block(parent_block)` for template creation (handles LOCKED_IN → ACTIVE boundary)
- DAA reduction takes effect at the activation block itself
- 21 unit tests covering reward reduction, boundary detection, effective avg time, and consensus rules
- No mainnet Criteria added in this PR — only code + Feature enum entry

### Analysis

The simulation runs a mainnet-like scenario with ~8.93 EH/s hashpower (weight ~67.86), where the `REDUCE_DAA_TARGET` feature activates at height 600. 1600 total blocks are mined with a DAA window of 134 blocks.

**Expected weight drop:**

The DAA weight formula includes a `log₂(T)` term where T is the target block time:
- Before activation: `W = log₂(H × 30)`
- After activation: `W = log₂(H × 7.5)`
- Expected drop: `log₂(30) − log₂(7.5) = log₂(30/7.5) = log₂(4) = 2.0`

This 2.0 drop is independent of hashpower — it's purely the ratio of target times. The **instantaneous** weight drop at the activation block (h=599→600) is exactly 2.0 (67.79 → 65.79), matching the theoretical prediction. The **steady-state** average weight drop shown in the dashboard summary is 1.81, slightly less than 2.0 due to sampling variance over finite windows.

**DAA behavior:**

The DAA correctly estimated the hashrate and adjusted mining difficulty right at activation. No overshoot, no prolonged oscillation. Solvetimes converge to the new 7.5s target within ~20 blocks, as visible in the solvetime chart (avg solvetime drops from 25.09s to 7.37s, max transition solvetime 50s).

**No decay triggered:**

The weight decay mechanism (triggered when solvetime ≥ 3600s) was never needed. The DAA handled the transition smoothly on its own.

**Raw data:**

The per-block simulation data points (height, weight, solvetime, hashpower, feature state) are available in [`tools/daa-reduction/simulator/daa_runs/hp8928412586733189120_s0.json`](tools/daa-reduction/simulator/daa_runs/hp8928412586733189120_s0.json).

image

### Checklist


hathor/verification/block_verifier.py
```diff
@@ -69,7 +69,7 @@ def verify_weight(self, block: Block) -> None:
     def verify_reward(self, block: Block) -> None:
         """Validate reward amount."""
         parent_block = block.get_block_parent()
-        tokens_issued_per_block = self._daa.get_tokens_issued_per_block(parent_block.get_height() + 1)
+        tokens_issued_per_block = self._daa.get_tokens_issued_per_block(parent_block.get_height() + 1, block=block)
```
msbrogli, line 72: Should this method receive only the block? It can get the parent block height from the block itself if needed.

hathor/feature_activation/feature_service.py
```diff
@@ -57,6 +57,33 @@ def is_feature_active(self, *, vertex: Vertex, feature: Feature) -> bool:
         state = self.get_state(block=block, feature=feature)
         return state.is_active()
 
+    def is_feature_active_for_next_block(self, *, parent_block: 'Block', feature: Feature) -> bool:
```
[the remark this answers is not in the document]
glevco: I would prefer starting it at the very next block, so we don't need this extra and specific method in feature service. The practical effect is the same.

tools/daa-reduction/simulator/daa_runs/hp8928412586733189120_s0.json
glevco, line 1: Remove?

tools/daa-reduction/simulator/daa_live_dashboard.html
glevco, line 1: Remove?

hathor/daa.py
glevco, line 1: This implementation is pretty specific and ad-hoc for the reduction feature. You could instead use the `Features.from_vertex` structure and make a `BlockTimeVersion` analogous to `OpcodesVersion`, `NanoRuntimeVersion`, and the unmerged `BlueprintVersion`.

By doing that, the structure would be ready to make another change if we decide to do it in the future (which is possible after we observe the real network behavior after the update). With the current solution, it would get unnecessarily complex to add another conditional on top of the implemented ones.

tools/daa-reduction/simulator/daa_runs/hp8928412586733189120_s0.json
msbrogli: This was the result of the simulation. I thought on adding the raw values so we can compare the actual activation with the expected activation.

tools/daa-reduction/simulator/daa_live_dashboard.html
msbrogli: Same for the tool. I thought on adding it to a separate repository but it made sense to have a collection of tools for future use.

hathor/feature_activation/utils.py
```diff
@@ -114,6 +122,7 @@ def all_enabled() -> Features:
             fee_tokens=True,
             opcodes_version=OpcodesVersion.V2,
             nano_runtime_version=NanoRuntimeVersion.V2,
+            block_time_version=BlockTimeVersion.V2,
```
msbrogli: Rename to `DAAVersion`.

hathor/daa/daa.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_tokens_issued_per_block(self, height: int, *, block: Block | None = None) -> int:
+        """Return the number of tokens issued per block of a given height.
+
+        When a block is provided, checks feature state and reduces the reward proportionally
+        if REDUCE_DAA_TARGET is active.
+        """
+        if block is not None:
+            return self._select(block).get_tokens_issued_per_block(height)
+        return self._v1.get_tokens_issued_per_block(height)
```
msbrogli: I don't like this default. Should we require a block? Or the version to use?

hathor/daa/daa.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_reward_for_next_block(self, parent_block: Block) -> int:
+        """Return the reward for the next block after parent_block."""
+        height = parent_block.get_height() + 1
+        return self._select(parent_block).get_tokens_issued_per_block(height)
+
+    def minimum_tx_weight(self, tx: Transaction) -> float:
+        """Returns the minimum weight for the param tx. Version-independent."""
+        return self._v1.minimum_tx_weight(tx)
```
msbrogli, line 100: It should require the block and select as well.

hathor/daa/daa.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_mined_tokens(self, height: int) -> int:
+        """Return the number of tokens mined in total at height. Version-independent."""
+        return self._v1.get_mined_tokens(height)
```
msbrogli, line 109: It should require the block and select as well.

hathor/daa/daa.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_weight_decay_amount(self, distance: int) -> float:
+        """Return the amount to be reduced in the weight of the block. Version-independent."""
+        return self._v1.get_weight_decay_amount(distance)
```
msbrogli, line 118: It should require the block and select as well.

hathor/daa/daa.py
msbrogli: Or, if a block is not provided, use the best block?

hathor/daa/v2.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+    def calculate_next_weight(
+        self,
+        parent_block: Block,
+        timestamp: int,
+        parent_block_getter: Callable[[Block], Block],
+    ) -> float:
+        """Public method for template creation."""
+        if self.TEST_MODE & TestMode.TEST_BLOCK_WEIGHT:
```
msbrogli, line 82: Should we remove `TEST_MODE` for once and for all?

hathor_cli/mining.py
```diff
@@ -145,12 +145,14 @@ def execute(args: Namespace) -> None:
...
             settings = get_global_settings()
             daa = DifficultyAdjustmentAlgorithm(settings=settings)
+            from hathor.daa import DAAVersion
             verification_params = VerificationParams(nc_block_root_id=None, features=Features(
                 count_checkdatasig_op=True,
                 nanocontracts=False,
                 fee_tokens=False,
                 opcodes_version=OpcodesVersion.V2,
                 nano_runtime_version=NanoRuntimeVersion.V2,
+                block_time_version=DAAVersion.V1,
```
msbrogli: `block_time_version`? It should be `daa_version`.

hathor/feature_activation/utils.py
```diff
@@ -61,13 +64,17 @@ def from_vertex(*, settings: HathorSettings, feature_service: FeatureService, ve
         nano_runtime_version = (
             NanoRuntimeVersion.V2 if feature_is_active[Feature.NANO_RUNTIME_V2] else NanoRuntimeVersion.V1
         )
+        block_time_version = (
```
msbrogli: `daa_version`

hathor/p2p/sync_v2/transaction_streaming_client.py
```diff
@@ -63,6 +64,7 @@ def __init__(self,
                 fee_tokens=False,
                 opcodes_version=OpcodesVersion.V1,
                 nano_runtime_version=NanoRuntimeVersion.V1,
+                block_time_version=DAAVersion.V1,
```
msbrogli: `daa_version`

hathor/daa/common.py
```diff
@@ -0,0 +1,227 @@
...
+    min_block_weight: float,
+    test_mode: TestMode,
+) -> float:
+    """Calculate the next block weight, aka DAA/difficulty adjustment algorithm.
+
+    The algorithm used is described in `RFC 22`_.
+
+    The weight must not be less than ``min_block_weight``.
+
+    .. _RFC 22: https://gitlab.com/HathorNetwork/rfcs/merge_requests/22
```
glevco, line 66: What is this syntax?

hathor/daa/common.py
```diff
@@ -0,0 +1,227 @@
...
+
+def _get_mined_tokens(settings: HathorSettings, height: int) -> int:
+    """Return the number of tokens mined in total at height."""
+    assert settings.BLOCKS_PER_HALVING is not None
+    number_of_halvings = (height - 1) // settings.BLOCKS_PER_HALVING
+    number_of_halvings = max(0, number_of_halvings)
+
+    blocks_in_this_halving = height - number_of_halvings * settings.BLOCKS_PER_HALVING
+
+    tokens_per_block = settings.INITIAL_TOKENS_PER_BLOCK
```
glevco, line 198: This has to be updated, right?

hathor/daa/daa.py
```diff
@@ -0,0 +1,126 @@
...
+        self._v1 = DifficultyAdjustmentAlgorithmV1(settings=settings, test_mode=test_mode)
+        self._v2 = DifficultyAdjustmentAlgorithmV2(settings=settings, test_mode=test_mode)
+        self.TEST_MODE = test_mode
+        self.MIN_BLOCK_WEIGHT = settings.MIN_BLOCK_WEIGHT
+        DifficultyAdjustmentAlgorithm.singleton = self
+
+    def _select(self, block: Block) -> DifficultyAdjustmentAlgorithmV1 | DifficultyAdjustmentAlgorithmV2:
+        """Select V1 or V2 based on feature activation state for the given block."""
+        if self._feature_service is None:
+            return self._v1
```
glevco, line 61: We should add an assert somewhere to make sure this is set in non-test environments

hathor/manager.py
```diff
@@ -796,8 +796,11 @@ def generate_mining_block(self, timestamp: Optional[int] = None,
         return block
 
     def get_tokens_issued_per_block(self, height: int) -> int:
-        """Return the number of tokens issued (aka reward) per block of a given height."""
-        return self.daa.get_tokens_issued_per_block(height)
+        """Return the number of tokens issued (aka reward) per block of a given height.
+
+        This is an info-only method (used by APIs) with no block context, so it always uses V1.
+        """
+        return self.daa._v1.get_tokens_issued_per_block(height)
```
glevco, line 803: It should return the correct value based on the height.

hathor/daa/daa.py
```diff
@@ -0,0 +1,126 @@
...
+            return self._select(block).get_weight_decay_amount(distance)
+        return self._v1.get_weight_decay_amount(distance)
+
+    def get_block_dependencies(
+        self,
+        block: Block,
+        parent_block_getter: Callable[[Block], Block],
+    ) -> list[VertexId]:
+        """Return the ids of the required blocks to call `calculate_block_difficulty`."""
+        return self._v1.get_block_dependencies(block, parent_block_getter)
```
glevco, line 126: Call the standalone `_get_block_dependencies` function instead , and remove it from v1 and v2

hathor/daa/v2.py
```diff
@@ -0,0 +1,114 @@
...
+            return 1.0
+        if block.is_genesis:
+            return self.MIN_BLOCK_WEIGHT
+        parent_block = parent_block_getter(block)
+        return _calculate_next_weight(
+            self._settings, parent_block, block.timestamp, parent_block_getter,
+            avg_time=self.avg_time_between_blocks, min_block_weight=self.MIN_BLOCK_WEIGHT,
+            test_mode=self.TEST_MODE,
+        )
```
glevco, line 73: This method is repeated in both v1 and v2, with the only difference being what arguments they pass to `_calculate_next_weight`. The same for other methods in this file. It seems to me the common could just be an abstract base class and the versions would override the missing methods, without duplication.

Even better, I would model this abstraction in inversion with values and composition instead: a single DAA class that gets a `DAASettings` class that provides values. Each one of v1 and v2 would correspond to an instance of `DAASettings`, with different values.

glevco (changes requested):

hathor/dag_builder/default_filler.py
```diff
@@ -210,8 +210,8 @@ def run(self) -> None:
                     assert set(balance.keys()).issubset({'HTR'})
                     diff = balance.get('HTR', 0)
 
-                    # TODO Use the actual height.
-                    target = self._daa_factory.get_tokens_issued_per_block(1)
+                    # TODO Use the actual height. DAG construction has no chain context, so V1.
+                    target = self._daa_factory.create_v1().get_tokens_issued_per_block(1)
```
msbrogli, line 214: Should we keep the tests on v1 or move to v2?

hathor/daa/factory.py
```diff
@@ -68,45 +68,88 @@ def __init__(
...
+        config: DAAConfig,
+        *,
+        v2_start_height: int | None = None,
+    ) -> DifficultyAdjustmentAlgorithm:
         return DifficultyAdjustmentAlgorithm(
             settings=self._settings,
             config=config,
             test_mode=self.TEST_MODE,
             min_block_weight=self.MIN_BLOCK_WEIGHT,
+            v2_start_height=v2_start_height,
```
msbrogli: Move this to DAAConfig.

hathor/feature_activation/feature_service.py
```diff
@@ -224,6 +224,33 @@ def get_feature_states(self, *, vertex: Vertex) -> dict[Feature, FeatureState]:
             for feature, info in feature_infos.items()
         }
 
+    def get_activation_height(self, *, block: 'Block', feature: Feature) -> int | None:
```
msbrogli, line 227: I don't like this implementation. Should we keep this registered in the FeatureState when it transitions to ACTIVE? @glevco

hathor/dag_builder/default_filler.py
glevco: I guess we can keep V1 for now, and think about updating them after the feature is activated.

hathor/feature_activation/feature_service.py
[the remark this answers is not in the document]
jansegre: I agree.

hathor/daa/common.py
```diff
@@ -200,31 +219,63 @@ def _get_base_tokens_issued_per_block(settings: HathorSettings, height: int) ->
...
-    # Sum the blocks in the current halving
-    mined_tokens += blocks_in_this_halving * tokens_per_block
 
-    return mined_tokens
+def _get_mined_tokens(
+    settings: HathorSettings,
+    height: int,
+    *,
+    reward_reduction_factor: int = 1,
+    v2_start_height: int | None = None,
```
glevco: I would rather remove the defaults to avoid mistakenly calling this with default values when they should be set.

hathor/daa/common.py
```diff
@@ -200,31 +219,63 @@ def _get_base_tokens_issued_per_block(settings: HathorSettings, height: int) ->
     return max(amount, settings.MINIMUM_TOKENS_PER_BLOCK)
 
 
-def _get_mined_tokens(settings: HathorSettings, height: int) -> int:
-    """Return the number of tokens mined in total at height.
-
-    The base reward at each halving is ``INITIAL_TOKENS_PER_BLOCK // 2**halving`` (capped at
-    ``MINIMUM_TOKENS_PER_BLOCK``).
+def _sum_block_rewards_in_range(
```
glevco, line 225: It seems the logic in this method could reuse `_get_base_tokens_issued_per_block`.

hathor/daa/factory.py
```diff
@@ -68,45 +68,88 @@ def __init__(
...
     def create_v1(self) -> DifficultyAdjustmentAlgorithm:
         """Build a DAA with the V1 config."""
         return self._build(DAAConfig.for_v1(self._settings))
 
+    def create_v2(self, *, v2_start_height: int | None = None) -> DifficultyAdjustmentAlgorithm:
+        """Build a DAA with the V2 config.
+
+        ``v2_start_height`` — the height of the first V2 block — is only needed to compute
+        cumulative mined tokens correctly. When omitted, ``get_mined_tokens`` will treat
+        the entire chain as V2.
```
glevco: > When omitted, ``get_mined_tokens`` will treat the entire chain as V2.

Why would this be used?

hathor/daa/factory.py
```diff
@@ -68,45 +68,88 @@ def __init__(
...
+        if self._tx_storage is None or self._feature_service is None:
             return self.create_v1()
         return self.create_from_parent(self._tx_storage.get_best_block())
 
     def _select_config(self, parent_block: Block) -> DAAConfig:
-        # Currently V1-only. Future feature-gated DAA versions plug in here.
+        if self._feature_service is None:
+            return DAAConfig.for_v1(self._settings)
+        if parent_block._static_metadata is None:
+            return DAAConfig.for_v1(self._settings)
```
glevco: This is all dangerous because we cannot be sure the production path won't reach this. We should never fallback to v1 automatically. We should use asserts for production, and separate methods for tests/APIs if necessary.

glevco (changes requested):

hathor/feature_activation/feature_service.py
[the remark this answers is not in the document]
msbrogli: 🤖 Done — Shape B is in `DAAFactory._select_config`: a block is V2 iff its parent is `REDUCE_DAA_TARGET ACTIVE`. The dedicated `is_feature_active_for_next_block` helper was dropped.

hathorlib/hathorlib/conf/settings.py
```diff
@@ -157,6 +157,9 @@ def MAXIMUM_NUMBER_OF_HALVINGS(self) -> int:
     # Average time between blocks.
     AVG_TIME_BETWEEN_BLOCKS: int = 30  # in seconds
 
+    # Average time between blocks after REDUCE_DAA_TARGET feature activation.
+    REDUCED_AVG_TIME_BETWEEN_BLOCKS: int = 30  # in seconds, networks configure the post-activation target
```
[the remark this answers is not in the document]
msbrogli: 🤖 Leaving at 7.5s for now (`REDUCED_AVG_TIME_BETWEEN_BLOCKS_10X = 75`, giving a 4× speedup and `reward_reduction_factor=300//75=4`). Happy to switch if there is a strong preference for 6s or 10s — pinging here so it does not get lost.

hathor/verification/block_verifier.py
msbrogli: 🤖 Done — `verify_reward(self, block)` derives `parent_block` internally.

hathor/feature_activation/utils.py
msbrogli: 🤖 Done — enum is `DAAVersion` in `hathor/daa/common.py`.

hathor/daa/daa.py
msbrogli: 🤖 Addressed by the composition refactor. `DifficultyAdjustmentAlgorithm` no longer carries a default version — it takes a `DAAConfig`, and version selection happens at the factory boundary via `DAAFactory.create_from_block(block)` / `create_from_parent(parent)`. For code paths without a feature_service (CLI tools, dag_builder synthetic blocks) the explicit `create_v1()` is used.

hathor/daa/daa.py
msbrogli: 🤖 Same as the thread above — methods live on a per-block `DifficultyAdjustmentAlgorithm` instance built by `DAAFactory.create_from_*(block)`, so the block is required and selection happens at construction.

hathor/daa/daa.py
msbrogli: 🤖 Same as the thread above — methods live on a per-block `DifficultyAdjustmentAlgorithm` built by `DAAFactory.create_from_*(block)`.

hathor/daa/daa.py
msbrogli: 🤖 Same as the thread above — methods live on a per-block `DifficultyAdjustmentAlgorithm` built by `DAAFactory.create_from_*(block)`.

hathor/daa/v2.py
msbrogli: 🤖 Partially addressed: `DAAFactory.__init__` now asserts `test_mode == DISABLED` unless `NETWORK_NAME in {unittests, privatenet}`, so it cannot be silently enabled on mainnet/testnet. Full removal needs a settings-level rework that touches ~30 test files — tracking it as a follow-up PR.

hathor_cli/mining.py
msbrogli: 🤖 Done — `block_time_version` is gone from the codebase; everything uses `daa_version: DAAVersion`.

hathor/feature_activation/utils.py
msbrogli: 🤖 Done — `daa_version: DAAVersion`.

hathor/p2p/sync_v2/transaction_streaming_client.py
msbrogli: 🤖 Done — `daa_version=DAAVersion.V1`.

hathor/daa/common.py
msbrogli: 🤖 Line context shifted after the refactor — the original referent is gone. If the underlying concern still applies, can you point me at the current location?

hathor/daa/daa.py
msbrogli: 🤖 Done — `DAAFactory.__init__` asserts `test_mode == TestMode.DISABLED` unless `settings.NETWORK_NAME in {unittests, privatenet}`.

hathor/manager.py
msbrogli: 🤖 Done — `manager.get_tokens_issued_per_block(height)` now goes through `daa_factory.create_from_parent(best_block).get_tokens_issued_per_block(height)`, so the right version is selected based on the best blocks feature state.

hathor/daa/daa.py
msbrogli: 🤖 Done — `DifficultyAdjustmentAlgorithm.get_block_dependencies` calls `_get_block_dependencies` from `common.py`. The v1/v2 files are gone.

hathor/daa/v2.py
msbrogli: 🤖 Done — implemented exactly that pattern. `DifficultyAdjustmentAlgorithm` is the single class; `DAAConfig` is the values object built via `DAAConfig.for_v1(settings)` / `DAAConfig.for_v2(settings, v2_start_height=…)`. v1.py/v2.py are gone.

hathor/dag_builder/default_filler.py
msbrogli: 🤖 Going with your call: keeping V1 for now, revisit after activation.

hathor/daa/factory.py
msbrogli: 🤖 Done — `v2_start_height` is now a field on `DAAConfig` (`DAAConfig.for_v2(settings, v2_start_height=…)`). The DAA constructor no longer takes it; the factory just builds the config and hands it over.

hathor/feature_activation/feature_service.py
msbrogli: 🤖 Leaving as-is for now and treating this as a follow-up. If you (@glevco) prefer caching the activation height on the FeatureState transition to ACTIVE, happy to do that in a separate PR — would touch FeatureService internals more than this PRs scope.

hathor/daa/common.py
msbrogli: 🤖 Done — `_get_mined_tokens` no longer has defaults; `reward_reduction_factor` and `v2_start_height` are required keyword args.

hathor/daa/common.py
msbrogli: 🤖 Agreed in principle — deferring to a follow-up to keep this PR scoped to the activation + architecture changes.

hathor/daa/factory.py
msbrogli: 🤖 The docstring no longer says that. `create_v2(v2_start_height=None)` is allowed for callers that only use per-block methods; `get_mined_tokens` asserts when called on a V2 instance without `v2_start_height`.

hathor/daa/factory.py
msbrogli: 🤖 Done — `_select_config` now asserts `feature_service is not None` and `parent._static_metadata is not None`. CLI/synthetic-block paths (dag_builder, mining CLI) call `create_v1()` directly. The test/simulator fixtures auto-wire feature_service via the Builder so they go through the same production path instead of a silent fallback.

hathor/feature_activation/feature_service.py
glevco: I agree we can leave it like this for now and persist it in a separate PR. But instead of registering it in the per-block feature state, I would just store the activation height of each feature in the `FeatureActivationStorage`, which already exists for Feature Activation metadata.

hathor/daa/common.py
```diff
@@ -195,31 +222,63 @@ def _get_base_tokens_issued_per_block(settings: HathorSettings, height: int) ->
...
+    v2_start_height: int | None,
+) -> int:
+    """Return the number of tokens mined in total at height.
 
-    return mined_tokens
+    Heights below ``v2_start_height`` are treated as V1 (no reduction). Heights at or above
+    ``v2_start_height`` get their per-block reward divided by ``reward_reduction_factor``.
+    When ``v2_start_height`` is ``None`` or beyond ``height``, the entire range is V1.
+    """
+    if v2_start_height is None or v2_start_height > height:
```
glevco, line 269: Shouldn't `v2_start_height` be asserted here?

glevco (dismissed):

jansegre (dismissed):

jansegre (dismissed):

PR #1643, opened 2026-04-07
chore: reset feature states

glevco:
### Motivation

Due to misconfig, `FEE_TOKENS` was not activated on mainnet and `OPCODES_V2` was not activated on neither testnet nor mainnet. This PR renames them to `FAILED_*` by performing a migration, which allows us to reuse those feature names.

### Acceptance Criteria

- Create new features `FAILED_FEE_TOKENS` and `FAILED_OPCODES_V2`. They're NOT used to change any behavior in code.
- Update settings:
  - On mainnet, rename feature `FEE_TOKENS` to `FAILED_FEE_TOKENS`, and `OPCODES_V2` to `FAILED_OPCODES_V2`.
  - On testnet, rename feature `OPCODES_V2` to `FAILED_OPCODES_V2`.
- Create migration to clear `metadata.feature_states` from all blocks, so we can perform the feature renaming. This metadata is just a cache that stores pre-calculated feature states, they'll be naturally recalculated with the new feature names.

### Checklist


jansegre (approved):
Should we reset every other feature that was activated for a wider clean up?

hathor/transaction/storage/migrations/reset_feature_state_cache.py
```diff
@@ -0,0 +1,56 @@
...
+        assert isinstance(storage, TransactionRocksDBStorage)
+        feature_activation_storage = FeatureActivationStorage(
+            settings=storage._settings,
+            rocksdb_storage=storage._rocksdb_storage,
+        )
+
+        log.info('resetting existing feature activation settings...')
+        feature_activation_storage.reset_settings()
+
+        log.info('cleaning up feature states cache...')
```
msbrogli, line 49: Just to be sure: This will trigger a feature recalculation, right? I'm not sure we should do it or simply force a sync from scratch.

hathor/conf/mainnet.py
msbrogli, line 1: Why rename instead of just remove them?

msbrogli (approved):

hathor/transaction/storage/migrations/reset_feature_state_cache.py
glevco: It will, but as we have 6,495,587 blocks in mainnet and only calculate new states every 20,160 blocks, and on top of that we've only been using feature activation in the last few years, this calculation will run for less than 300 blocks. I guess it's safe and way faster than requiring a sync from scratch.

hathor/conf/mainnet.py
glevco: I guess we could remove them and that would be ideal, however I determined that a rename would be safer, to prevent edge cases and conflicts by reusing the names. For example, depending on whether the BigSignalingService runs before or after the migration, its logic could be impacted.

I would rather just keep rename in this PR, and then we can do a separate PR for a general feature cleanup, including old features that were already activated, as suggested by @jansegre in https://github.com/HathorNetwork/hathor-core/pull/1643#pullrequestreview-4069768021. So we can just remove them, without name reuse.

hathor/transaction/storage/migrations/reset_feature_state_cache.py
msbrogli: Got it. Ok!

hathor/conf/mainnet.py
msbrogli: Indeed! Their presence protects us from misconfiguring something. Thanks!

PR #1639, opened 2026-03-27
feat: hathorlib move runner

r4mmer:
### Motivation

To be able to implement the blueprint simulator on hathorlib there is a significant amount of code that needs to be moved from `hathor/` to `hathorlib/`.
Since moving this amount of code can cause instabilities we are breaking down these changes into more manageable and reviewable PRs.

This is the final PR.

### Acceptance Criteria

- Move runner and other core requirements to run NCs to hathorlib

### Changes of note

- Moved more hathor exceptions to hathorlib
- `BlueprintServiceProtocol` to decouple the blueprint service from the runner
- New hathorlib tests
- Decreased coverage requirements from 85% to 84%

### Checklist

- [ ] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

PR #1648, opened 2026-04-09
Feat/hathorlib move runner 2

r4mmer:
### Motivation

To be able to implement the blueprint simulator on hathorlib there is a significant amount of code that needs to be moved from `hathor/` to `hathorlib/`.
Since moving this amount of code can cause instabilities we are breaking down these changes into more manageable and reviewable PRs.

This is the final PR.

### Acceptance Criteria

- Move runner and other core requirements to run NCs to hathorlib

### Changes of note

- Moved more hathor exceptions to hathorlib
- `BlueprintServiceProtocol` to decouple the blueprint service from the runner
- Decreased coverage requirements from 85% to 84%

### Checklist


hathorlib/hathorlib/nanocontracts/runner/runner.py
```diff
@@ -0,0 +1,1449 @@
...
+
+        try:
+            token_description = self.tx_storage.get_token_description(token_uid)
+        except Exception:
+            token_description = None
+        if token_description is None:
+            raise NCInvalidSyscall(
+                f'contract {call_record.contract_id.hex()} could not find {token_uid.hex()} token'
+            )
+        return token_description
```
glevco: Before the move to hathorlib, it was:

- Get token creation tx from storage and raise `NCInvalidSyscall` on `TransactionDoesNotExist`
- Raise `NCInvalidSyscall` on `first_block is None`
- Return `TokenDescription`

Then, currently on `master`, it was changed to:

- Get token description from storage 
  - Get token creation tx from storage
  - Return `None` on `first_block is None`
  - Return `TokenDescription`
- Raise `NCInvalidSyscall` on `token_description is None`
- Raise `NCInvalidSyscall` on  `TransactionDoesNotExist`

That seems like the exact same behavior to me. However, in this PR, it's changed to:

- Get token description from storage 
  - Get token creation tx from storage
  - Return `None` on `first_block is None`
  - Return `TokenDescription`
- Raise `NCInvalidSyscall` on `token_description is None`
- **Raise `NCInvalidSyscall` on  `Exception`**

So there was a change in the last line which _could_ be a behavior change. Knowing Nano internals, I know both would cause the tx to fail. However, at the very least the exception message would be misleading to users: before it was "could not find token" only for `TransactionDoesNotExist`, and now this same message is applied to any exception.

glevco (changes requested):

hathorlib/hathorlib/nanocontracts/runner/runner.py
glevco: Moved from https://github.com/HathorNetwork/hathor-core/pull/1639#discussion_r3054179146

r4mmer: Sorry, this specific change is because `TransactionDoesNotExist` is a hathor exception, we would need to move this exception to hathorlib as well.

If you want we can make `get_token_description` change the exception so the runner can react to a specific exception.

r4mmer: Done

glevco: Thanks!

glevco (approved):

msbrogli (approved):

PR #1649, opened 2026-04-09
feat: blueprint simulator

r4mmer:
### Motivation

Blueprint developers currently need a running Hathor full node to test their NanoContract logic, which makes the development loop slow and heavyweight. This PR introduces a self-contained, in-memory simulator that enables testing
blueprints using the production `Runner` without any network or node dependencies. It lives entirely within `hathorlib`, so it can be used as a standalone library.

### Acceptance Criteria

- The simulator mirrors the production `NCBlockExecutor` pattern: per-call Runner creation, commit on success, discard on failure, block storage commit on `new_block()`
- Blueprints can be registered from a file (`register_blueprint(path)`, the preferred flow — mirrors the on-chain `OnChainBlueprint` import surface) or from a Python class (`register_blueprint_class(cls)`, for quick tests)
- Contracts are created using `BlueprintId` returned by the registration calls
- `create_contract(bid, …)` is the primary entry point and returns a `ContractProxy` that exposes blueprint methods as regular Python methods; `create_contract_raw(bid, …)` is the low-level variant that returns a `TxResult`
- `get_contract(contract_id)` returns a `ContractProxy` for an already-deployed contract
- A low-level string-based API (`call_public`, `call_view`) is available for generic tooling and dynamic calls, with the same semantics as production
- Token deposits, withdrawals, and authority grants/acquires are supported via action helpers (`NanoSimulator.deposit()`, `NanoSimulator.withdrawal()`, etc.)
- Block lifecycle supports both automatic (one block per call) and manual (`new_block()`) modes, switchable at runtime via `auto_new_block`
- Simulated clock can be controlled via `advance_time()` / `set_time()`, with lazy block data creation so time changes take effect on the next call
- Events emitted by blueprints are captured in `TxResult` and queryable via `get_events(tx_hash=..., block_hash=...)`
- Execution logs are captured (including for failed calls) and queryable via `get_logs(tx_hash=..., block_hash=...)`
- Snapshot/restore enables capturing and rolling back simulation state for test isolation
- Failed calls raise `NCFail` and do not commit state changes
- Custom tokens can be created and used in contract interactions
- Deterministic ID generation for addresses, contracts, blueprints, and tokens
- Simulated `checksig` backend with sentinel values (`CHECKSIG_VALID` / `CHECKSIG_INVALID`) allows testing `SignedData` verification without real cryptography, enabled by default and scoped to each execution call

### Limitations

Since this is a simulator, not all features can be exactly the same as production. Some of the more notable differences are:

- Token authorities are not enforced
- Transactions and balances are not validated
- A view method called before a `new_block()` will get the current state of the storage; in production the state available to view methods is only from validated and confirmed blocks
- There is no mempool or enforceable transaction ordering
- The simulated `checksig` backend does not perform real cryptographic verification — use `CHECKSIG_VALID` / `CHECKSIG_INVALID` sentinel values as `script_input`, or provide a custom backend via
`NanoSimulatorBuilder().with_checksig(my_backend)`
- In production a block comes after the nano contract call txs and defines which txs are executed in the block; in the simulation we create a block and all subsequent calls are part of the block until a new block appears

### Architecture

The simulator is a thin orchestration layer on top of the production `Runner`. It does **not** reimplement execution logic — it wires together in-memory implementations of the infrastructure the Runner depends on.

```
NanoSimulatorBuilder (configuration & wiring)
    └── NanoSimulator (orchestration)
            ├── Runner / RunnerFactory (production execution engine, unmodified)
            ├── InMemoryNCStorageFactory (trie-backed contract storage)
            ├── InMemoryBlueprintService (blueprint registry)
            ├── InMemoryTxStorage (token registry)
            ├── SimulatorClock (controllable time)
            ├── ContextFactory (synthetic Context/VertexData/BlockData)
            ├── EventStore (event & log capture)
            ├── IdGenerator (deterministic IDs)
            ├── SimulatorSnapshot (state capture/restore)
            └── ContractProxy (OO wrapper for deployed contracts)
```

### Files

**Source** (`hathorlib/hathorlib/nanocontracts/simulator/`):

| File                    | Purpose                                                                        |
|-------------------------|--------------------------------------------------------------------------------|
| `builder.py`            | `NanoSimulatorBuilder` — configuration API                                     |
| `simulator.py`          | `NanoSimulator` — core orchestration, block lifecycle, method calls            |
| `proxy.py`              | `ContractProxy` — object-oriented wrapper exposing blueprint methods           |
| `checksig.py`           | Simulated checksig backend and `CHECKSIG_VALID` / `CHECKSIG_INVALID` sentinels |
| `context_factory.py`    | Synthetic `Context`, `VertexData`, `BlockData` creation                        |
| `id_generator.py`       | Deterministic ID generation                                                    |
| `in_memory_services.py` | `SimulatorClock`, `InMemoryBlueprintService`, `InMemoryTxStorage`              |
| `in_memory_storage.py`  | `InMemoryNCStorageFactory` wrapping `InMemoryNodeTrieStore`                    |
| `event_store.py`        | Event and execution log capture, indexed by tx/block hash                      |
| `result.py`             | `TxResult` and `BlockResult` dataclasses                                       |
| `snapshot.py`           | `SimulatorSnapshot` — state capture and restore via deep copy                  |

### Usage Example

```python
from hathorlib.nanocontracts.simulator import NanoSimulatorBuilder, CHECKSIG_VALID
from hathorlib.nanocontracts.types import NC_HTR_TOKEN_UID

sim = NanoSimulatorBuilder().build()

# Register a blueprint from a file (preferred — mirrors the on-chain flow)
bid = sim.register_blueprint('./blueprints/my_blueprint.py')
alice = sim.create_address('alice')

# Deploy a contract — returns a ContractProxy
contract = sim.create_contract(bid, caller=alice)

# Call methods directly on the proxy
contract.do_something(caller=alice, actions=[sim.deposit(NC_HTR_TOKEN_UID, 1000)])
value = contract.get_value()

# Retrieve an existing contract later
contract = sim.get_contract(contract_id)

# SignedData with simulated checksig
signed = SignedData[str]('oracle result', CHECKSIG_VALID)
contract.set_result(signed, caller=oracle)

# Time control
sim.advance_time(3600)

# Snapshot/restore
snap = sim.snapshot()
contract.mutate(caller=alice)
sim.restore(snap)  # roll back

# --- Low-level API (for generic tooling or dynamic calls) ---

# Register from an already-imported class
bid = sim.register_blueprint_class(MyBlueprint)

# Create a contract and get the raw TxResult instead of a proxy
tx_result = sim.create_contract_raw(bid, caller=alice)

# String-based public/view calls
sim.call_public(tx_result.contract_id, 'do_something', caller=alice)
value = sim.call_view(tx_result.contract_id, 'get_value')
```

### Coverage Analysis

pytest-cov works seamlessly without any extra configuration when using the simulator to test blueprint files. The example below was made with a simple test on the bet blueprint, loaded with `sim.register_blueprint("./blueprints/bet.py")`
using the same `bet.py` file as the one uploaded to mainnet. The example project only had hathorlib, pytest and pytest-cov as dependencies.

{A3B9A29B-A006-49B2-8EE8-AAA7CD253770}

### Checklist


hathorlib/hathorlib/nanocontracts/types.py
```diff
@@ -149,6 +150,26 @@ def set_checksig_backend(fn: ChecksigBackend) -> None:
     _checksig_backend = fn
 
 
+@contextmanager
+def checksig_backend(fn: ChecksigBackend) -> Iterator[None]:
```
msbrogli, line 154: Why this?

hathorlib/hathorlib/nanocontracts/simulator/__init__.py
msbrogli, line 1: I don't like `hathorlib.simulator`. Maybe `hathorlib.nanocontracts.simulator`? Or `hathorlib.simulator.nanocontracts`?

hathorlib/hathorlib/nanocontracts/simulator/checksig.py
```diff
@@ -0,0 +1,46 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Sentinel values for use as script_input when constructing SignedData.
+# Use CHECKSIG_VALID to make checksig() return True, CHECKSIG_INVALID to make it return False.
+CHECKSIG_VALID: bytes = b'__checksig_valid__'
+CHECKSIG_INVALID: bytes = b'__checksig_invalid__'
+
+
+def simulated_checksig_backend(sighash_all_data: bytes, script_input: bytes, script: bytes) -> bool:
```
msbrogli, line 29: What's this?

hathorlib/hathorlib/simulator/README.md
```diff
@@ -0,0 +1,403 @@
...
+sim.call_public(cid, 'do_something', caller=alice)
+value = sim.call_view(cid, 'get_value')
+```
+
+## SimulatorBuilder
+
+The builder configures and creates `Simulator` instances with a fluent API:
+
+```python
+sim = (SimulatorBuilder()
```
msbrogli: Maybe `NanoSimulatorBuilder()`?

hathorlib/hathorlib/simulator/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+### Creating contracts
+
+```python
+# Basic creation
+result = sim.create_contract(bid, caller=alice)
+cid = result.contract_id
+
+# With constructor arguments
+result = sim.create_contract(bid, caller=alice, args=(oracle_script, token_uid, Timestamp(1000)))
```
msbrogli: Does it return a ContractProxy?

hathorlib/hathorlib/simulator/README.md
```diff
@@ -0,0 +1,403 @@
...
+result = sim.create_contract(
+    bid,
+    caller=alice,
+    actions=[sim.deposit(NC_HTR_TOKEN_UID, 1000)],
+)
+```
+
+### ContractProxy (object-oriented API)
+
+`create_instance` returns a `ContractProxy` that exposes blueprint methods as regular Python methods:
```
msbrogli: Maybe `get_contract()`?

hathorlib/hathorlib/nanocontracts/simulator/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+```python
+# Option 1: Disable at build time
+sim = SimulatorBuilder().with_auto_new_block(False).build()
+
+# Option 2: Toggle at runtime
+sim.auto_new_block = False
+
+# Multiple calls in the same block
+sim.call_public(cid, 'increment', caller=alice)
```
msbrogli, line 240: Why `call_public()` if we can use the proxy? I guess we should promote the proxy.

hathorlib/hathorlib/nanocontracts/simulator/README.md
```diff
@@ -0,0 +1,403 @@
...
+# Execution logs (same filtering options)
+logs = sim.get_logs(tx_hash=tx_result.tx_hash)
+all_logs = sim.get_logs()
+```
+
+Each event has a `.data` attribute (bytes). Execution logs (`NCExecEntry`) include an `error_traceback` field that is `None` on success and contains the traceback string on failure.
+
+## Checksig (Simulated Signatures)
+
+The simulator provides a simulated checksig backend so you can test `SignedData` without real cryptography:
```
msbrogli, line 278: Why? Signing is so easy.

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli, line 1: We don't it now, but maybe we should add a FakeRNG that can be fully controlled for testing. So users can set a seed or simply provide a list of values to be returned. So they can test multiple random scenarios.

hathorlib/hathorlib/nanocontracts/simulator/result.py
```diff
@@ -0,0 +1,38 @@
...
+
+from dataclasses import dataclass
+
+from hathorlib.nanocontracts.nc_exec_logs import NCEvent, NCExecEntry
+from hathorlib.nanocontracts.types import ContractId, VertexId
+
+
+@dataclass(frozen=True, slots=True)
+class TxResult:
+    """Result of a single call_public or create_contract call."""
```
msbrogli, line 25: Why would they return a TxResult? I got lost here.

hathorlib/hathorlib/nanocontracts/types.py
r4mmer: The hathorlib have the utilities to verify valid scripts or generating scripts from addresses but it cannot execute a script.
We provide a good default replacement that will work when passing the sentinel values but the user may want to use in another way, with this context manager he can just do:

```python
with checksig_backend(my_backend):
    val = SignedData[str]('data', input_data)
    contract.do_something(val, caller=alice)
```

and the backend will simulate whatever condition the user wants.

msbrogli (changes requested):

hathorlib/hathorlib/nanocontracts/simulator/checksig.py
r4mmer: This is how `SignedData` validation works during a simulation since we dont have the capability to run the oracle script to verify the signature.

hathorlib/hathorlib/simulator/README.md
r4mmer: No, it returns a `TxResult` which is (in the simulated space) the information on the tx this call belongs to.
So `result.tx_hash` or `result.contract_id` would be the contract id that may be used in other calls.

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: I think we should have both, while the proxy is the promoted and intended way to use we may still want to have a "low level" api.
The proxy has a few constraints, for instance there are reserved words for arguments (caller and actions).

Another reason is that the proxy is an instance that does not indicate which simulator instance it belongs to, making it harder to track where the call is taking effect when using multiple simulators.
An example of this scenario is a side-dag interacting with mainnet.

hathorlib/hathorlib/nanocontracts/simulator/result.py
r4mmer: A call to a contract is a Tx, but we are calling a function instead.
Since we don't have a transaction storage and managing utxos seems unnecessary for the purpuses of this simulator I created this `TxResult`, which has informations on the Tx of the simulated call.
This includes tx_hash, which contract is being called, from which block this was executed and the events and logs of the call.

Personally, I like the idea of calling a method and having the events and logs returned as a result.
The tx, block and contract are basically metadata for the call.

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: I like this idea, we would have to change a little the interface of the runner but makes a lot of sense.

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: Running a script is not, and we currently intend the oracle to be a P2PKH output script but it could be any kind of script, even one without signing.
Since we cannot run script from hathorlib and we cannot guarantee that the script will be a P2PKH (if we could I could just write a function that simulates the P2PKH execution) I figured the user could indicate the intended result of the validation

hathorlib/hathorlib/simulator/README.md
r4mmer: Done, changed the api so the proxy seems like the intended way to use the simulator

hathorlib/hathorlib/simulator/README.md
r4mmer: Done

hathorlib/hathorlib/nanocontracts/simulator/__init__.py
r4mmer: Done

hathorlib/hathorlib/nanocontracts/simulator/README.md
jansegre, line 1: We don't have any other case with a `README.md` in a module like this. I think it might be better if this was in the docstring of `__init__.py`, that way the examples could be testable too.

hathorlib/hathorlib/nanocontracts/simulator/result.py
jansegre: Why not `NcCallResult` instead?

hathorlib/hathorlib/nanocontracts/simulator/result.py
```diff
@@ -0,0 +1,38 @@
...
+    block_hash: VertexId
+    contract_id: ContractId
+    events: list[NCEvent]
+    exec_entry: NCExecEntry | None
+
+
+@dataclass(frozen=True, slots=True)
+class BlockResult:
+    """Result of new_block(), summarizing all transactions in the block."""
+    block_hash: VertexId
```
jansegre, line 36: Similarly this could be a `NcExecResult`, since the execution happens at the block level and executes each call in order.

hathorlib/hathorlib/nanocontracts/types.py
```diff
@@ -35,9 +37,6 @@
 from hathorlib.utils.address import decode_address, get_address_b58_from_bytes
 from hathorlib.utils.typing import InnerTypeMixin
 
-# Well-known constant: HTR token UID is always 0x00
-HATHOR_TOKEN_UID: bytes = b'\x00'
-
 # XXX: mypy gives the following errors on all subclasses of `bytes` that use FauxImmutableMeta:
```
jansegre, line 40: We might have to keep this. Please check if there's any blueprint published in `mainnet` or `testnet` that imports this constant.

.gitignore
```diff
@@ -30,3 +30,5 @@ keys.json
 
 # Pycharm
 .idea
+
+docs/plans/
```
jansegre: We should try to use `.git/info/exclude` unless the ignored path is really meant to be shared between devs.

jansegre (dismissed):

hathorlib/hathorlib/nanocontracts/types.py
msbrogli: But why would we skip signature verification if that's cheap to run? I mean, I understand skipping mining but not signing and checking signatures.

hathorlib/hathorlib/nanocontracts/simulator/__init__.py
msbrogli: Thanks!

hathorlib/hathorlib/nanocontracts/simulator/checksig.py
msbrogli: Why don't we have the capability to run the oracle script to verify the signature?

hathorlib/hathorlib/simulator/README.md
msbrogli: Thanks!

hathorlib/hathorlib/simulator/README.md
msbrogli: Thanks!

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: Good point. Should we introduce a SimulatorProxy that knows about the simulator being used?

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: Oh! Can't we run scripts from hathorlib? Is it a pending task to migrate script execution to hathorlib? If yes, I finally understood other parts of this PR.

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: Should we prepare for it now or simply refactor later?

hathorlib/hathorlib/nanocontracts/types.py
r4mmer: But we can't run the script to check the signature.

hathorlib/hathorlib/nanocontracts/simulator/checksig.py
r4mmer: The hathorlib does not have the capability to run scripts (as of today)
We can move the scripts infrastructure to hathorlib to mitigate this

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: We currently do not have a task to migrate the script to hathorlib, we can add this task if you think its best.

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: Since this changes the Runner, I think this should be made in its own PR, what do you think?
Should I open an issue to track this?

hathorlib/hathorlib/nanocontracts/types.py
r4mmer: I think it does not matter since the blueprint import comes from [here](https://github.com/HathorNetwork/hathor-core/blob/feat/blueprint-simulator/hathorlib/hathorlib/nanocontracts/allowed_imports.py#L68)

But correct me if I'm wrong on this

.gitignore
r4mmer: Sorry, added by mistake

hathorlib/hathorlib/nanocontracts/simulator/result.py
r4mmer: Done

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: Removed the simulator readme, and we dont run docstring tests on the hathorlib.
I will add the readme as a comment here for documentation since it may be used later

hathorlib/hathorlib/nanocontracts/simulator/result.py
r4mmer: Done

r4mmer:
# NanoContracts Blueprint Simulator

An in-memory simulator for developing and testing Hathor NanoContract blueprints without running a full node. It provides a developer-friendly API that mirrors the on-chain execution model while giving you full control over time, blocks, tokens, and state.

## Quick Start

```python
from hathorlib.nanocontracts.simulator import NanoSimulatorBuilder

# Create a simulator
sim = NanoSimulatorBuilder().build()

# Register a blueprint from a file (mirrors the on-chain flow)
bid = sim.register_blueprint('./blueprints/my_blueprint.py')  # str or Path
alice = sim.create_address('alice')

# Deploy a contract and interact with it via a ContractProxy
contract = sim.create_contract(bid, caller=alice)
contract.do_something(caller=alice)
value = contract.get_value()
```

## NanoSimulatorBuilder

The builder configures and creates `NanoSimulator` instances with a fluent API:

```python
sim = (NanoSimulatorBuilder()
    .with_seed(b'my_test_seed')        # Deterministic RNG seed
    .with_initial_time(1700000000)     # Starting clock timestamp
    .with_auto_new_block(False)        # Manual block control
    .with_unlimited_fuel()             # Disable fuel metering
    .with_checksig(my_backend)         # Custom signature verification
    .with_settings(custom_settings)    # Custom HathorSettings
    .build())
```

| Method                      | Default                     | Description                                                                     |
|-----------------------------|-----------------------------|---------------------------------------------------------------------------------|
| `with_seed(seed)`           | `b'simulator_default_seed'` | RNG seed for deterministic ID generation                                        |
| `with_initial_time(ts)`     | Current time                | Initial simulated clock timestamp                                               |
| `with_auto_new_block(bool)` | `True`                      | Auto-advance blocks after each successful call                                  |
| `with_unlimited_fuel()`     | Off                         | Disable fuel/memory limits for rapid prototyping                                |
| `with_checksig(backend)`    | Simulated backend           | Signature verification backend (see [Checksig](#checksig-simulated-signatures)) |
| `with_settings(settings)`   | Mainnet-like                | Custom `HathorSettings`                                                         |
| `with_runtime_version(v)`   | `V2`                        | Nano runtime version                                                            |

## Addresses and Tokens

```python
# Create deterministic test addresses
alice = sim.create_address('alice')
bob = sim.create_address('bob')

# Get the output script for an address (needed by some blueprints)
oracle_script = sim.get_output_script(alice)

# Create custom tokens
from hathorlib.nanocontracts.types import NC_HTR_TOKEN_UID
my_token = sim.create_token('MyToken', 'MTK')
```

The built-in HTR token (`NC_HTR_TOKEN_UID`) is always available.

## Blueprint Registration

The preferred way to register a blueprint is to load it from a file. This mirrors the
on-chain `OnChainBlueprint` flow, uses the same `from hathor import ...` import surface
blueprints run against in production, and keeps your simulator tests honest about the
blueprint's actual source layout.

```python
bid = sim.register_blueprint('./blueprints/bet.py')
```

The file must contain a class decorated with `@export` that subclasses `Blueprint`. The
simulator resolves `from hathor import ...` imports automatically via a shim module.

For special cases (not entirely bound to the on-chain constraints) the developer can define a
blueprint class and register it into the simulator with `register_blueprint_class`:

```python
class MyBlueprint(Blueprint):
    def initialize(self, ctx: Context):
        self.value = 0

    @public
    def increase(self, ctx: Context, value: int):
        self.value += value

    @value
    def get_value(self, ctx: Context):
        return self.value

bid = sim.register_blueprint_class(MyBlueprint)
```

Both forms are idempotent: calling them again with the same file or class returns the
same `BlueprintId`.

## Contract Lifecycle

### Creating contracts

`create_contract` returns a `ContractProxy` that exposes the blueprint's public and view
methods as regular Python methods.

```python
# Basic creation
contract = sim.create_contract(bid, caller=alice)

# With constructor arguments
contract = sim.create_contract(bid, caller=alice, args=(oracle_script, token_uid, Timestamp(1000)))

# With actions (e.g. initial deposit)
contract = sim.create_contract(
    bid,
    caller=alice,
    actions=[sim.deposit(NC_HTR_TOKEN_UID, 1000)],
)
```

The proxy also exposes the underlying `NcCallResult` via `contract.tx_result` and the
contract's id via `contract.contract_id`.

### Calling methods via the proxy

Public methods require `caller` (and optionally `actions`) as keyword arguments and
return a `NcCallResult`. View methods take positional arguments and return the view's value
directly.

```python
contract.bet(alice, '1x0', caller=alice, actions=[sim.deposit(NC_HTR_TOKEN_UID, 1000)])
result = contract.has_result()
amount = contract.get_max_withdrawal(bob)
```

### Retrieving an existing contract

```python
contract = sim.get_contract(contract_id)
```

### Low-level API

For generic tooling, dynamic calls, or tests that want to pin down the low-level
surface, the simulator exposes string-based primitives:

```python
# Raw creation returns NcCallResult instead of a proxy
tx_result = sim.create_contract_raw(
    bid,
    caller=alice,
    args=(...),
    actions=[sim.deposit(NC_HTR_TOKEN_UID, 500)],
)

# Public method calls by name
tx_result = sim.call_public(
    contract_id, 'method_name',
    caller=alice,
    args=(arg1, arg2),
    actions=[sim.deposit(NC_HTR_TOKEN_UID, 500)],
)

# View method calls by name
value = sim.call_view(contract_id, 'get_something', arg1, arg2)
```

Prefer the proxy-based API above for day-to-day blueprint testing.

## Actions (Deposits, Withdrawals, Authority)

Actions represent token operations attached to method calls:

```python
# Deposit tokens into a contract
sim.deposit(token_uid, amount)

# Withdraw tokens from a contract
sim.withdrawal(token_uid, amount)

# Grant token authority to a contract
sim.grant_authority(token_uid, mint=True)
sim.grant_authority(token_uid, melt=True)

# Acquire token authority from a contract
sim.acquire_authority(token_uid, mint=True, melt=True)
```

Actions are passed as a list to `actions=` in `call_public`, `create_contract`, or proxy public methods.

## NcCallResult

Every successful `call_public` or `create_contract_raw` returns a `NcCallResult`:

```python
result = sim.call_public(cid, 'emit_one', caller=alice)

result.tx_hash        # Transaction hash (VertexId)
result.block_hash     # Block hash this tx belongs to (VertexId)
result.contract_id    # Contract ID
result.events         # List of NCEvent emitted during execution
result.exec_entry     # Execution log entry (NCExecEntry)
```

Failed calls raise `NCFail` (or a subclass). Changes are **not** committed on failure.

## Time Control

The simulator has a virtual clock that you control:

```python
sim = NanoSimulatorBuilder().with_initial_time(100).build()

# Advance time by a number of seconds
sim.advance_time(1000)

# Set time to an absolute timestamp
sim.set_time(5000)

# Read the current time
print(sim.clock_time)
```

Time is used by the block's `timestamp` field, which blueprints access via `ctx.block.timestamp`.

## Block Management

By default (`auto_new_block=True`), each successful call creates its own block. For fine-grained control:

```python
# Option 1: Disable at build time
sim = NanoSimulatorBuilder().with_auto_new_block(False).build()

# Option 2: Toggle at runtime
sim.auto_new_block = False

# Multiple calls in the same block
sim.call_public(cid, 'increment', caller=alice)
sim.call_public(cid, 'increment', caller=alice)
sim.call_public(cid, 'increment', caller=alice)

# Commit the block manually
block_result = sim.new_block()
print(block_result.block_height)
print(len(block_result.tx_results))  # 3
```

`NcExecResult` contains:
- `block_hash` -- the hash of the committed block
- `block_height` -- the block height
- `tx_results` -- list of `NcCallResult` for all transactions in the block

## Events and Logs

Blueprints can emit events via `self.syscall.emit_event(data)`. The simulator captures these along with execution logs.

```python
# Events from a specific transaction
events = sim.get_events(tx_hash=tx_result.tx_hash)

# Events from a specific block
events = sim.get_events(block_hash=block_result.block_hash)

# All events across the simulation
all_events = sim.get_events()

# Execution logs (same filtering options)
logs = sim.get_logs(tx_hash=tx_result.tx_hash)
all_logs = sim.get_logs()
```

Each event has a `.data` attribute (bytes). Execution logs (`NCExecEntry`) include an `error_traceback` field that is `None` on success and contains the traceback string on failure.

## Checksig (Simulated Signatures)

The simulator provides a simulated checksig backend so you can test `SignedData` without real cryptography:

```python
from hathorlib.nanocontracts.simulator import CHECKSIG_VALID, CHECKSIG_INVALID
from hathorlib.nanocontracts.types import SignedData

# Create signed data that will pass checksig
valid_data = SignedData[str]('my_value', CHECKSIG_VALID)

# Create signed data that will fail checksig
invalid_data = SignedData[str]('my_value', CHECKSIG_INVALID)
```

Use `CHECKSIG_VALID` / `CHECKSIG_INVALID` as the `script_input` parameter when constructing `SignedData`. The simulated backend recognizes these sentinel values and returns `True` / `False` accordingly.

To disable checksig or provide a custom backend:

```python
# Disable (checksig raises NotImplementedError)
sim = NanoSimulatorBuilder().with_checksig(None).build()

# Custom backend
def my_backend(sighash_all_data: bytes, script_input: bytes, script: bytes) -> bool:
    return verify_signature(script_input, script)

sim = NanoSimulatorBuilder().with_checksig(my_backend).build()
```

## Snapshot and Restore

Save and restore the full simulation state for branching test scenarios:

```python
sim.call_public(cid, 'increment', caller=alice)
sim.call_public(cid, 'increment', caller=alice)
assert sim.call_view(cid, 'get_count') == 2

# Take a snapshot
snap = sim.snapshot()

# Continue modifying state
sim.call_public(cid, 'increment', caller=alice)
sim.call_public(cid, 'increment', caller=alice)
assert sim.call_view(cid, 'get_count') == 4

# Restore to the snapshot
sim.restore(snap)
assert sim.call_view(cid, 'get_count') == 2
```

Snapshots capture contract storage, token registry, clock time, block height, ID counters, and events/logs. You can take multiple snapshots and restore to any of them.

## State Inspection

```python
# Check contract balance
balance = sim.get_balance(contract_id, token_uid)
print(balance.available)

# Check if a contract exists
sim.has_contract(contract_id)

# Current block height
sim.block_height

# Current clock time
sim.clock_time
```

## Error Handling

Failed calls raise `NCFail` (or subclasses). State is **not** modified on failure:

```python
from hathorlib.nanocontracts import NCFail

try:
    contract.bet(bob, '1x1', caller=bob, actions=[sim.deposit(NC_HTR_TOKEN_UID, 5000)])
except NCFail as err:
    print(f"Call failed: {err}")
    # Contract state is unchanged
```

Failed calls still record execution logs (with traceback) for debugging, queryable via `sim.get_logs()`.

## Writing Blueprints for the Simulator

Blueprint files use `from hathor import ...` for all imports. The simulator provides a shim that maps these to the correct `hathorlib` types.

```python
from hathor import (
    Address, Blueprint, Context, NCFail, SignedData,
    Timestamp, TokenUid, TxOutputScript,
    export, public, view,
)

@export
class MyBlueprint(Blueprint):
    value: int

    @public
    def initialize(self, ctx: Context) -> None:
        self.value = 0

    @public(allow_deposit=True)
    def deposit(self, ctx: Context) -> None:
        action = ctx.get_single_action(self.token_uid)
        self.value += action.amount

    @view
    def get_value(self) -> int:
        return self.value
```

Key decorators:
- `@export` -- marks the class as the blueprint to be loaded by `register_blueprint()`
- `@public` -- marks a method as callable via transactions
- `@public(allow_deposit=True)` / `@public(allow_withdrawal=True)` -- allows deposit/withdrawal actions
- `@view` -- marks a read-only method (no `ctx` parameter, no state changes)

## Using with pytest

```python
import pytest
from hathorlib.nanocontracts.types import NC_HTR_TOKEN_UID
from hathorlib.nanocontracts.simulator import NanoSimulatorBuilder

@pytest.fixture
def sim():
    return NanoSimulatorBuilder().with_initial_time(100).build()

@pytest.fixture
def alice(sim):
    return sim.create_address('alice')

@pytest.fixture
def contract(sim, alice):
    bid = sim.register_blueprint('./blueprints/my_blueprint.py')
    return sim.create_contract(bid, caller=alice)

class TestMyBlueprint:
    def test_initial_state(self, contract):
        assert contract.get_value() == 0

    def test_deposit(self, sim, contract, alice):
        contract.deposit(caller=alice, actions=[sim.deposit(NC_HTR_TOKEN_UID, 100)])
        assert contract.get_value() == 100
```

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: Got it. We can plan it separately. Let's discuss it offline. Thanks!

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: Let's open an issue to track this and also add as a pending task to the project. Thanks!

hathorlib/hathorlib/simulator/README.md
msbrogli: Got it. Thanks!

hathorlib/hathorlib/nanocontracts/simulator/checksig.py
msbrogli: Got it. Thanks for clarifying!

hathorlib/hathorlib/nanocontracts/types.py
msbrogli: Got it. Thanks!

msbrogli (dismissed):

hathorlib/hathorlib/nanocontracts/types.py
jansegre: That link is only for the allow-rule, not the source. The actual source is the `hathor` module (`hathor/__init__.py`), but it imports `HATHOR_TOKEN_UID` from `hathor.nanocontracts` (`hathor/nanocontracts/__init__.py`) which in turn defines it directly (and as a `TokenUid`, not `bytes`), so in the end this removal is correct..

hathorlib/hathorlib/nanocontracts/simulator/README.md
jansegre: Even if we don't currently run docstrings I think it shouldn't be an argument *not* to do this. We can just turn on docstring testing in the future (like we do with `hathor`). And the text is a good guide, if it only lives in this PR as a comment it will for sure be forgotten.

hathorlib/hathorlib/nanocontracts/simulator/simulator.py
```diff
@@ -0,0 +1,628 @@
...
+
+    Example:
+        from hathorlib.nanocontracts.simulator import NanoSimulator, NanoSimulatorBuilder
+
+        sim = NanoSimulatorBuilder().build()
+        bid = sim.register_blueprint('./blueprints/my_blueprint.py')
+        alice = sim.create_address('alice')
+        contract = sim.create_contract(bid, caller=alice)
+        contract.do_something(caller=alice)
+        value = contract.get_value()
```
jansegre, line 87: Why don't we format the example with `>>>` so it's properly formatted for docstring testing, even if we don't have it enabled yet.

jansegre (dismissed):

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: Great! I will create a new PR with tests in the examples

hathorlib/hathorlib/nanocontracts/simulator/simulator.py
r4mmer: Will do so in a separate PR, along with allowing tests on the hathorlib docstrings.

hathorlib/hathorlib/nanocontracts/simulator/README.md
r4mmer: Created https://github.com/HathorNetwork/hathor-core/issues/1681

hathorlib/hathorlib/nanocontracts/simulator/simulator.py
r4mmer: Created thie [issue](https://github.com/HathorNetwork/hathor-core/issues/1682) to track this

jansegre (approved):

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: We can do it separately.

hathorlib/hathorlib/nanocontracts/simulator/README.md
msbrogli: Thanks!

hathorlib/hathorlib/nanocontracts/simulator/result.py
msbrogli: Thanks!

msbrogli (approved):

PR #1658, opened 2026-04-20
refactor: update code according to VM requirements

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1666

### Motivation

The new Nano runtime will start being tested soon. In preparation for it, in order to keep all hathor-core tests passing, we must update test Blueprints to remove prohibited items. This PR also includes some other related small refactors/improvements.

### Acceptance Criteria

- Fix `--nc-exec-fail-trace` (it wasn't working after the block executor refactor).
- Add `ENABLE_HTR_VM` flag, which is used to skip some implementation-specific tests.
- Update Blueprint tests to remove usage of prohibited types/functions/syntax/etc.
- Call `setup_logging` for tests.
- Update `NC_FIELDS_ATTR` from `__fields` to `__blueprint_fields`.

### Checklist

### DIFF
--- a/hathor/nanocontracts/execution/consensus_block_executor.py
+++ b/hathor/nanocontracts/execution/consensus_block_executor.py
@@ -303,7 +303,7 @@ class NCConsensusBlockExecutor:
                 if tx.name:
                     kwargs['__name'] = tx.name
                 if self._nc_exec_fail_trace:
-                    kwargs['exc_info'] = True
+                    kwargs['traceback'] = tb
                 self.log.info(
                     'nc execution failed',
                     tx=tx.hash.hex(),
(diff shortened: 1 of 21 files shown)


hathor/nanocontracts/execution/consensus_block_executor.py
```diff
@@ -286,7 +286,7 @@ def _apply_effect(
                 if tx.name:
                     kwargs['__name'] = tx.name
                 if self._nc_exec_fail_trace:
-                    kwargs['exc_info'] = True
+                    kwargs['traceback'] = tb
```
glevco, line 289: This was a bug after the block executor refactor, because the `exc_info` was only available inside a `try-except` block. We fix it by just printing the traceback as a log attribute, instead.

PR #1661, opened 2026-04-23
feat(verification): VerificationContext completeness mechanism

msbrogli:
### Motivation

Consensus verification has a silent-failure class of bug: when two dispatch gates disagree, a check can be skipped and nothing in the system notices. We hit exactly this recently on the shielded-outputs work — a tx with shielded inputs → transparent outputs was having its balance check skipped by both the transparent path (gated on `is_shielded()`) and the shielded path (gated on `has_shielded_outputs()`). Neither fired. No exception, no log, just an inflation hole.

This PR adds a defense-in-depth mechanism that makes this class of bug loud. Each verify_* method records a flag when its invariant has been enforced; at the end of each verification stage, the dispatcher asserts the recorded flag set covers the required set for the vertex's shape. Any skip — dispatcher-level OR internal to a verifier — raises `VerificationChecksMissingError`.

The mechanism is self-enforcing: a cross-check test enumerates (vertex type × params) combinations and asserts the required-flag table stays 1:1 with what the dispatcher actually records. That test already earned its keep — it caught two real drifts in the required-checks table on first run, which are fixed in this PR.

### Acceptance Criteria

- `VerificationCheck` IntFlag enum covers every verify_* invariant across all verifier classes (42 flags)
- `VerificationContext` threads through `verify_basic` and `verify`; `ctx` is a **required** keyword argument on every verifier method — no silent no-op if a caller forgets it
- Verifiers self-record at their success tail, so stubbing/early-returning/deleting a check's work will fail the completeness assertion
- Nano-contract postponement: `verify_sum` records `BALANCE_POSTPONED` when a nonexistent token defers the check; `NCBlockExecutor._verify_sum_after_execution` records `BALANCE` after re-verification; a post-nano completeness check voids the tx (via `NCFail`) if `BALANCE` wasn't promoted
- `verification_checks` persisted in `TransactionMetadata` via `to_storage_json` (internal bookkeeping, deliberately **not** exposed in public `to_json`)
- Cross-check test (`test_required_checks_match_dispatcher`) enumerates vertex shape × params combinations and asserts `ctx.checks_run` equals `required_checks_for(...)` bidirectionally. Future drift fails loudly
- Regression coverage: `test_completeness` simulates a broken verifier (stub returns no-op) and asserts `VerificationChecksMissingError` fires. `test_post_nano_voiding` simulates a bypassed post-execution re-check and asserts `NCFail` is raised
- Full test suite clean: 2530 passed, lint/mypy clean across the whole tree

### Scope note

This lands on `master`, which does not yet carry the shielded-outputs feature set. The shielded flag bundle (`SHIELDED_COMMITMENTS`, `SHIELDED_BALANCE`, `SHIELDED_SURJECTION`, etc.) is intentionally deferred to the follow-up port on `feat/ct-amount-token-privacy` — that branch has the verifiers to instrument. The framework here is additive and ready to accept them.

### Checklist

- [x] If you are requesting a merge into \`master\`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

### DIFF
--- a/hathor/nanocontracts/execution/block_executor.py
+++ b/hathor/nanocontracts/execution/block_executor.py
@@ -289,6 +289,11 @@ class NCBlockExecutor:
             # after the execution we have the latest state in the storage
             # and at this point no tokens pending creation
             self._verify_sum_after_execution(tx, block_storage)
+
+            # Defense-in-depth: the flag-set must now be complete.
+            # If _verify_sum_after_execution is ever removed or bypassed, the tx
+            # will arrive here with BALANCE_POSTPONED but no BALANCE → void it.
+            self._assert_verification_checks_complete_after_execution(tx)
         except NCFail as e:
             return NCTxExecutionFailure(
                 tx=tx,
@@ -300,14 +305,37 @@ class NCBlockExecutor:
         return NCTxExecutionSuccess(tx=tx, runner=runner)
 
     def _verify_sum_after_execution(self, tx: Transaction, block_storage: 'NCBlockStorage') -> None:
-        """Verify token sums after execution for dynamically created tokens."""
+        """Verify token sums after execution for dynamically created tokens.
+
+        Passes a VerificationContext so verify_sum records BALANCE into tx
+        metadata on success — letting the post-execution completeness check
+        confirm re-verification ran."""
         from hathor.verification.transaction_verifier import TransactionVerifier
+        from hathor.verification.verification_context import VerificationContext
         try:
             token_dict = tx.get_complete_token_info(block_storage)
-            TransactionVerifier.verify_sum(self._settings, tx, token_dict)
+            ctx = VerificationContext(vertex_hash=tx.hash or b'')
+            TransactionVerifier.verify_sum(self._settings, tx, token_dict, ctx=ctx)
+            tx.get_metadata().verification_checks |= ctx.checks_run
         except TokenNotFound as e:
             # At this point, any nonexistent token would have made a prior validation fail. For example, if there
             # was a withdrawal of a nonexistent token, it would have failed in the balance validation before.
             raise AssertionError from e
         except Exception as e:
             raise NCFail from e
+
+    def _assert_verification_checks_complete_after_execution(self, tx: Transaction) -> None:
+        """After nano execution, every required VerificationCheck flag must be
+        recorded. Missing flags mean a check was silently skipped — void the
+        tx by raising NCFail, which the caller converts to NCTxExecutionFailure."""
+        from hathor.transaction.exceptions import VerificationChecksMissingError
+        from hathor.verification.required_checks import Stage, required_post_nano_checks
+        from hathor.verification.verification_context import VerificationContext
+
+        meta = tx.get_metadata()
+        ctx = VerificationContext(vertex_hash=tx.hash or b'', checks_run=meta.verification_checks)
+        required = required_post_nano_checks(tx, self._settings)
+        try:
+            ctx.check(required, stage=Stage.POST_NANO_EXECUTION.value)
+        except VerificationChecksMissingError as e:
+            raise NCFail(str(e)) from e
--- a/hathor/stratum/stratum.py
+++ b/hathor/stratum/stratum.py
@@ -43,6 +43,7 @@ from hathor.reactor import ReactorProtocol as Reactor
 from hathor.transaction import BaseTransaction, BitcoinAuxPow, Block, MergeMinedBlock, Transaction, sum_weights
 from hathor.transaction.exceptions import PowError, ScriptError, TxValidationError
 from hathor.util import json_dumpb, json_loadb
+from hathor.verification.verification_context import VerificationContext
 from hathor.verification.vertex_verifier import VertexVerifier
 from hathor.wallet.exceptions import InvalidAddress
 
@@ -537,7 +538,7 @@ class StratumProtocol(JSONRPC):
         )
 
         try:
-            verifier.verify_pow(tx, override_weight=job.weight)
+            verifier.verify_pow(tx, override_weight=job.weight, ctx=VerificationContext(vertex_hash=tx.hash or b''))
         except PowError:
             self.log.error('bad share, discard', job_weight=job.weight, tx=tx)
             return self.send_error(INVALID_SOLUTION, msgid, {
@@ -553,7 +554,7 @@ class StratumProtocol(JSONRPC):
         self.manager.reactor.callLater(0, self.job_request)
 
         try:
-            verifier.verify_pow(tx)
+            verifier.verify_pow(tx, ctx=VerificationContext(vertex_hash=tx.hash or b''))
         except PowError:
             # Transaction pow was not enough, but the share was succesfully submited
             self.log.info('high hash, keep mining', tx=tx)
--- a/hathor/transaction/transaction_metadata.py
+++ b/hathor/transaction/transaction_metadata.py
@@ -25,6 +25,7 @@ from hathor.transaction.types import MetaNCCallRecord
 from hathor.transaction.validation_state import ValidationState
 from hathor.util import collect_n, json_dumpb, json_loadb, practically_equal
 from hathor.utils.weight import work_to_weight
+from hathor.verification.verification_check import VerificationCheck
 
 if TYPE_CHECKING:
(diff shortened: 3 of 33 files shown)


hathor/transaction/transaction_metadata.py
```diff
@@ -244,6 +253,7 @@ def to_storage_json(self) -> dict[str, Any]:
         else:
             data['first_block'] = None
         data['validation'] = self.validation.name.lower()
+        data['verification_checks'] = int(self.verification_checks)
```
msbrogli, line 256: Should we enforce a maximum number of flags to prevent this number from getting too big?

hathor/stratum/stratum.py
```diff
@@ -553,7 +554,7 @@ def handle_submit(self, params: dict, msgid: Optional[str]) -> None:
         self.manager.reactor.callLater(0, self.job_request)
 
         try:
-            verifier.verify_pow(tx)
+            verifier.verify_pow(tx, ctx=VerificationContext(vertex_hash=tx.hash or b''))
```
msbrogli, line 557: Create a class method `VerificationContext.create_empty()`.

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -300,14 +305,37 @@ def execute_transaction(
...
+        metadata on success — letting the post-execution completeness check
+        confirm re-verification ran."""
         from hathor.verification.transaction_verifier import TransactionVerifier
+        from hathor.verification.verification_context import VerificationContext
         try:
             token_dict = tx.get_complete_token_info(block_storage)
-            TransactionVerifier.verify_sum(self._settings, tx, token_dict)
+            ctx = VerificationContext(vertex_hash=tx.hash or b'')
+            TransactionVerifier.verify_sum(self._settings, tx, token_dict, ctx=ctx)
+            tx.get_metadata().verification_checks |= ctx.checks_run
```
msbrogli, line 319: Should it be a separate method `tx.update_verification_checks(ctx.checks_run)`?

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -300,14 +305,37 @@ def execute_transaction(
...
+        tx by raising NCFail, which the caller converts to NCTxExecutionFailure."""
+        from hathor.transaction.exceptions import VerificationChecksMissingError
+        from hathor.verification.required_checks import Stage, required_post_nano_checks
+        from hathor.verification.verification_context import VerificationContext
+
+        meta = tx.get_metadata()
+        ctx = VerificationContext(vertex_hash=tx.hash or b'', checks_run=meta.verification_checks)
+        required = required_post_nano_checks(tx, self._settings)
+        try:
+            ctx.check(required, stage=Stage.POST_NANO_EXECUTION.value)
```
msbrogli, line 339: After execution, required must always be true.

hathor/verification/nano_header_verifier.py
```diff
@@ -157,72 +163,74 @@ def verify_action_list(actions: Sequence[NCAction]) -> None:
...
+            for action in nano_header.nc_actions:
+                if action.type not in allowed_actions:
+                    exception = NCForbiddenAction(f'action {action.type} is forbidden on method `{method_name}`')
+                    raise NCTxValidationError from exception
+        ctx.record(VerificationCheck.NANO_METHOD_CALL)
+
+    def verify_seqnum(
+        self, tx: BaseTransaction, params: VerificationParams, *, ctx: VerificationContext
+    ) -> None:
+        if params.harden_nano_restrictions:
```
msbrogli, line 225: That's bad. If a check is skipped for any reason, its related flag should be unset.

hathor/verification/nano_header_verifier.py
```diff
@@ -157,72 +163,74 @@ def verify_action_list(actions: Sequence[NCAction]) -> None:
...
-            try:
-                contract_storage = block_storage.get_contract_storage(ContractId(nano_header.nc_id))
-            except NanoContractDoesNotExist as e:
-                raise NCTxValidationError from e
-            blueprint_id = contract_storage.get_blueprint_id()
-            allow_fallback = True
+    def verify_method_call(
+        self, tx: BaseTransaction, params: VerificationParams, *, ctx: VerificationContext
+    ) -> None:
+        if params.harden_nano_restrictions:
```
msbrogli, line 169: That's bad. If a check is skipped for any reason, its related flag should be unset.

hathor/verification/on_chain_blueprint_verifier.py
```diff
@@ -141,14 +143,15 @@ class OnChainBlueprintVerifier:
...
         self._settings = settings
 
-    def verify_pubkey_is_allowed(self, tx: OnChainBlueprint) -> None:
+    def verify_pubkey_is_allowed(self, tx: OnChainBlueprint, *, ctx: VerificationContext) -> None:
         """Verify if the on-chain blueprint's pubkey is allowed."""
         if self._settings.NC_ON_CHAIN_BLUEPRINT_RESTRICTED:
             address = get_address_b58_from_public_key_bytes(tx.nc_pubkey)
             if address not in self._settings.NC_ON_CHAIN_BLUEPRINT_ALLOWED_ADDRESSES:
                 raise OCBPubKeyNotAllowed(f'nc_pubkey with address {address} is not allowed')
+        ctx.record(VerificationCheck.OCB_PUBKEY_ALLOWED)
```
msbrogli, line 152: That's bad. If a check is skipped for any reason, its related flag should be unset.

hathor/verification/required_checks.py
```diff
@@ -0,0 +1,193 @@
...
+
+def _is_genesis_or_skipped(vertex: BaseTransaction, settings: HathorSettings) -> bool:
+    if vertex.hash in settings.SKIP_VERIFICATION:
+        return True
+    if getattr(vertex, 'is_genesis', False):
+        return True
+    return False
+
+
+def required_checks_for(
```
msbrogli, line 93: Let's have a list of required verifications for each vertex type and ingestion type (mempool, block sync, etc).

hathor/verification/transaction_verifier.py
```diff
@@ -362,66 +395,72 @@ def verify_version(self, tx: Transaction, params: VerificationParams) -> None:
...
-        if sorted(seen_token_indexes) != list(range(1, len(tx.tokens) + 1)):
-            raise UnusedTokensError('unused tokens are not allowed')
-
-    def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
+    def verify_tokens(
+        self, tx: Transaction, params: VerificationParams, *, ctx: VerificationContext
+    ) -> None:
+        """Verify that all tokens are used and unique. Records TOKENS
+        unconditionally — the invariant is "this check ran for these params"."""
+        if params.harden_token_restrictions:
```
msbrogli, line 405: That's bad. If a check is skipped for any reason, its related flag should be unset.

hathor/verification/transaction_verifier.py
```diff
@@ -362,66 +395,72 @@ def verify_version(self, tx: Transaction, params: VerificationParams) -> None:
...
-                # outputs cannot be spent.
-                raise InputVoidedAndConfirmed(spent_tx.hash.hex())
-            if txin.index not in spent_tx_meta.spent_outputs:
-                continue
-            spent_by_list = spent_tx_meta.spent_outputs[txin.index]
-            within_counter = 0
-            for h in spent_by_list:
-                if h == tx.hash:
-                    # Skip tx itself.
+        if params.reject_conflicts_with_confirmed_txs:
```
msbrogli, line 432: That's bad. If a check is skipped for any reason, its related flag should be unset.

hathor/verification/verification_check.py
```diff
@@ -0,0 +1,82 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from enum import IntFlag, auto
+
+
+class VerificationCheck(IntFlag):
```
msbrogli, line 18: Add a prefix with the vertex type (ex: VERTEX_VERSION_BASIC, BLOCK_REWARD). Also add a comment explain what each of them mean.

PR #1664, opened 2026-04-27
feat(shielded): mint/melt headers (per RFC) + canonical invariants

msbrogli:
## Summary

Lifts the parent branch's Rule 8 (\"no mint/melt in shielded txs\") by introducing two new headers — **MintHeader (0x14)** and **MeltHeader (0x15)** — that publicly declare per-token supply changes inside an otherwise-shielded transaction. The declared scalars enter the augmented homomorphic balance equation as unblinded terms, so the no-inflation guarantee still holds while the recipient set stays hidden.

Token Creation Transactions can now carry shielded outputs of the new token. Both headers are gated by the parent shielded transactions feature flag (\`ENABLE_SHIELDED_TRANSACTIONS\`); there is no separate gate.

## Why no range proof on the headers

Each \`MintMeltEntry\` enforces \`1 <= amount < 2**64\` at construction and at deserialize time. That bound is what protects the augmented balance equation from negative-value attacks via Pedersen commitments: by construction the declared scalar cannot encode a negative or zero supply change, so it cannot be used to sneak inflation in through the unblinded term. Range proofs are only needed where the value is hidden in a commitment; here it is publicly declared and bounded.

## DEPOSIT vs. FEE — fee policy on shielded mint/melt

The augmented balance equation handles both token-version fee models inline:

- **DEPOSIT-version tokens** pay the existing 1% HTR rule. Mint adds \`(deposit, htr_uid)\` to the output side; melt adds \`(withdraw, htr_uid)\` to the input side. Rounding follows the existing helpers (ceil for deposit, floor for withdraw).
- **FEE-version tokens** pay one \`FEE_PER_OUTPUT\` in HTR per \`MintHeader\` / \`MeltHeader\` entry, on the output side, regardless of mint vs. melt. The charge is per header entry — not per shielded recipient — so it matches the spirit of the transparent per-output rule without leaking how many recipients an entry was split across. \`AmountShieldedOutput\`s of FEE tokens still pay the generic \`FEE_PER_AMOUNT_SHIELDED_OUTPUT\` on top.

Both charges are folded directly into the balance equation, NOT into \`FeeHeader\`. \`FeeHeader\` continues to cover only chargeable transparent outputs and shielded outputs.

The matching RFC update (rfcs.git \`75fd4030\`) picks up the \`fee_token_charge\` term in Rule M4 and clarifies Rule M6's FeeHeader scope.

## Commits

- \`b3355b65\` docs(shielded): add PR split plan for shielded transactions
- \`1346edab\` feat(shielded): implement mint/melt headers per RFC
- \`7633c843\` fix(shielded): tighten mint/melt invariants from review feedback
- \`23fa5e3c\` refactor(shielded): fold mint/melt gate into ENABLE_SHIELDED_TRANSACTIONS
- \`d2e0042e\` refactor(shielded): move mint/melt wire format to hathorlib
- \`3112e133\` fix(shielded): tokens-index reorg ordering + double-count cancellation
- \`d7bc82b0\` test(shielded): close mint/melt review gaps + maintainability cleanups
- \`45471358\` test(shielded): pin FEE-version mint/melt augmented-balance behavior
- \`040674bc\` feat(shielded): charge FEE_PER_OUTPUT per Mint/Melt entry on FEE tokens

## Scope

- New: \`hathorlib/hathorlib/headers/mint_melt_header.py\` (canonical wire format + entry validation) and \`hathor/transaction/headers/mint_melt_header.py\` (consensus-typed wrapper that imports the wire-format helpers from hathorlib).
- Header IDs: \`MINT_HEADER = 0x14\`, \`MELT_HEADER = 0x15\` in \`headers/types.py\` (both packages).
- Parser: registered in \`vertex_parser/_headers.py\` + \`_vertex_parser.py\` and in \`hathorlib/vertex_parser.py\`, gated by \`ENABLE_SHIELDED_TRANSACTIONS\`.
- Verification: new methods in \`verification/transaction_verifier.py\` for mint/melt invariants (Rules M1–M4), authority checks, fee accounting, and shielded-balance integration. \`_fold_mint_melt_entry\` folds each entry into the augmented balance equation (DEPOSIT 1% on mint/melt, FEE_PER_OUTPUT per entry for FEE-version tokens); \`_resolve_token_version_for_mint_melt\` resolves the token version (TCT self-reference, tx-storage fallback, nano-block-storage path) and raises \`TokenNotFound\` on resolution failure rather than returning a sentinel.
- Settings: only \`ENABLE_SHIELDED_TRANSACTIONS\` controls everything; \`FEE_PER_AMOUNT_SHIELDED_OUTPUT\` / \`FEE_PER_FULL_SHIELDED_OUTPUT\` / \`FEE_PER_OUTPUT\` / \`TOKEN_DEPOSIT_PERCENTAGE\` unchanged.
- Indexes: \`rocksdb_tokens_index\` updates supply on mint/melt headers. The header amount is treated as the authoritative supply delta for the tokens it covers — the per-utxo loops' visible contribution for those tokens is cancelled to avoid double-counting in mixed transparent + shielded mint/melt. \`destroy_token\` runs last in \`remove_tx\` so reorg of a shielded TCT does not orphan a row with a negative total.
- Token Creation: \`token_creation_transaction_verifier\` accepts shielded outputs of the new token.
- Tests: \`hathor_tests/tx/test_shielded_mint_melt.py\` (~1200 lines) covers wire format, Rules M1–M4, authority checks (positive + negative paths), the full augmented balance equation (mint/melt sides + DEPOSIT-version deposit/withdraw with boundary rounding + FEE-version per-entry charge with the per-entry-not-per-amount property), every branch of \`_resolve_token_version_for_mint_melt\`, indexer net-cancellation correctness, and the reorg-ordering invariant.
- Client guide: \`hathor-ct-crypto/SHIELDED-OUTPUTS-CLIENT-GUIDE.md\` §3.4 gains a \"Balance with Mint/Melt headers\" subsection documenting how clients fold mint/melt entries into \`verify_balance\` for both DEPOSIT and FEE versions.

## Base

Targets \`feat/ct-amount-token-privacy\`, not \`master\`. Stacked on top of the parent shielded-transactions branch and follows the PR-split plan in \`docs/plans/shielded-pr-split.md\` — see the new \"Post-plan extensions\" section at the bottom of that file.

## Test plan

- [ ] \`poetry run pytest hathor_tests/tx/test_shielded_mint_melt.py -v\`
- [ ] \`poetry run pytest hathor_tests/tx/test_shielded_cons_fixes.py -v\`
- [ ] \`poetry run pytest hathorlib/tests -q\`
- [ ] \`poetry run make check\` (flake8 + isort + mypy + yamllint)
- [ ] Verify \`ENABLE_SHIELDED_TRANSACTIONS=DISABLED\` keeps existing tx behavior unchanged
- [ ] Verify mint/melt headers are rejected when shielded transactions are not active
- [ ] Verify reorg of a shielded TCT does not produce orphaned negative-total rows in the tokens index
- [ ] Verify a shielded mint of a FEE-version token requires \`FEE_PER_OUTPUT\` HTR per MintHeader entry on top of FeeHeader fees


hathor/transaction/headers/mint_melt_header.py
```diff
@@ -0,0 +1,203 @@
...
+    token_index: int
+    amount: int
+
+    def __post_init__(self) -> None:
+        if not 1 <= self.token_index <= MAX_MINT_MELT_ENTRIES:
+            raise ValueError(
+                f'token_index must be in [1, {MAX_MINT_MELT_ENTRIES}]; got {self.token_index}'
+            )
+        if not 1 <= self.amount < 2 ** 64:
+            raise ValueError(f'amount must be in [1, 2**64); got {self.amount}')
```
msbrogli: Explain that this limit already protects from "negative mint/melt values in Pedersen commitments so we don't need any range proof in these headers". This should be highlighted in the PR description and in the RFC too.

hathor/transaction/headers/mint_melt_header.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def _serialize_entries(entries: list[MintMeltEntry]) -> bytes:
```
msbrogli: Move serialization/deserialization to the hathorlib.

hathor/transaction/headers/mint_melt_header.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Subclasses set `_HEADER_ID` and `_HEADER_NAME`. The wire format and entry
+    constraints are identical (RFC §4.1).
+    """
+
+    _HEADER_ID: ClassVar[bytes] = b''
+    _HEADER_NAME: ClassVar[str] = ''
+
+    tx: Transaction
+    entries: list[MintMeltEntry] = field(default_factory=list)
```
msbrogli, line 75: Why a list? It's it easier to let multiple headers be present when needed?

hathor/transaction/headers/mint_melt_header.py
msbrogli: Done in d2e0042e. Added a docstring on `MintMeltEntry` explaining that the `1 <= amount < 2**64` bound is what lets us skip a range proof on these headers (the declared scalar can never encode a negative or zero supply change, so it cannot be used to sneak inflation into the Pedersen-augmented balance equation). Also called this out in the PR description under "Why no range proof on the headers". RFC update is out of this repo's scope but flagged for follow-up.

hathor/transaction/headers/mint_melt_header.py
msbrogli: Done in d2e0042e. Created `hathorlib/hathorlib/headers/mint_melt_header.py` as the canonical wire-format source (`serialize_entries` / `deserialize_entries` / `MintMeltEntry` constraints) plus the `MintHeader` / `MeltHeader` base classes. Registered in `hathorlib.headers.__init__`, `hathorlib.headers.types` (`MINT_HEADER = 0x14`, `MELT_HEADER = 0x15`), and `hathorlib.vertex_parser`. hathor-core's `mint_melt_header.py` now imports the wire-format helpers from hathorlib and only adds the consensus-typed `InvalidMintMeltHeaderError` translation, mirroring the pattern `shielded_outputs_header` already follows.

hathor/transaction/headers/mint_melt_header.py
msbrogli: Investigated this — the codebase enforces two invariants in `vertex_verifier.verify_headers` that make multi-instance Mint/Melt headers expensive: (a) "only one instance of {type} is allowed" — fires on duplicate type, and (b) "Headers must be in canonical order (ascending VertexHeaderId)" — strict `<`, so a run of equal IDs is rejected. Allowing multiple MintHeader/MeltHeader instances would require relaxing both rules globally (with knock-on effects on canonical-order determinism and malleability surface) and growing `get_maximum_number_of_headers` to ~37 (5 base + 16 mint + 16 melt).

Bundling entries inside a single header keeps both invariants intact and keeps the per-entry uniqueness check local to the header. Documented this rationale in d2e0042e on `_MintMeltHeaderBase`'s docstring so future readers don't have to rediscover it. Happy to revisit if the multi-instance approach has benefits I'm not seeing.

PR #1674, opened 2026-04-28
feat(p2p): introduce WhitelistManager and `# policy:` directive

msbrogli:
### Motivation

The peer-whitelist mechanism on `master` mixes operational state (peer list, fetch loop) with admission policy and pulls those concerns across `ConnectionsManager`, `HathorManager`, and `PeerIdState`. There is no way for the *whitelist content itself* to declare a policy, so operators cannot opt the network into an `allow-all` mode at runtime without a code/config change.

This PR introduces:

1. A `# policy: <value>` directive in the whitelist content. It is written as a comment so older parsers ignore it. Accepted values: `only-whitelisted-peers` (default; preserves current behavior) and `allow-all` (admits peers regardless of list membership). The directive must appear before any peer ID; duplicates and unknown values are rejected (case-insensitive).
2. A new `WhitelistManager` (`hathor/p2p/whitelist_manager.py`) that owns the enabled flag, the active policy, the peer list, the periodic refresh loop, and the connection-admission decision. It is loose-coupled to the connections layer via a `drop_connection: Callable[[PeerId], None]` predicate received at construction.

The admission rule is reduced to: peer in the list → allow; `allow-all` policy → allow; otherwise `enabled or whitelist_only` → block. The sync-v1 special case and the `USE_PEER_WHITELIST_ON_SYNC_V2` setting (also removed from `hathorlib/conf/settings.py`) are gone.

`ConnectionsManager` loses the old fields/methods (`whitelist_only`, `whitelist_policy`, `update_whitelist`, `_update_whitelist_*`, `wl_reconnect`, the whitelist HTTP agent). `HathorManager` loses `peers_whitelist`, `add_peer_to_whitelist`, and `remove_peer_from_whitelist_and_disconnect`. The status resource and tests read the peer list off `manager.connections.whitelist`.

### Acceptance Criteria

- [x] Whitelist parser recognizes the optional `# policy: <value>` directive and validates it (precedence, no duplicates, known values, case-insensitive).
- [x] `allow-all` admits peers that are not on the list; `only-whitelisted-peers` preserves the existing behavior.
- [x] `WhitelistManager` owns the enabled flag, active policy, peer list, refresh loop, and admission decision; `ConnectionsManager` and `HathorManager` no longer host whitelist state or methods.
- [x] `PeerIdState._should_block_peer` is a thin delegation to `whitelist.is_connection_allowed(peer_id)`.
- [x] `USE_PEER_WHITELIST_ON_SYNC_V2` and the sync-v1 branch are removed.
- [x] Unit tests cover parser cases (default, allow-all, only-whitelisted, case-insensitive, policy-after-peer rejection, duplicate rejection, invalid value, invalid header, unrelated comments) and policy enforcement (`allow-all` admits unlisted peers; `_update_cb` updates the active policy).
- [x] `pytest hathor_tests/{p2p,resources/p2p,others}` passes (205/205).
- [x] `mypy` and `ruff check` clean on changed files.
- [ ] CI green on the PR.

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged.

### DIFF
--- a/hathor/p2p/manager.py
+++ b/hathor/p2p/manager.py
@@ -23,7 +23,6 @@ from twisted.internet.interfaces import IListeningPort, IProtocol, IProtocolFact
 from twisted.internet.task import LoopingCall
 from twisted.protocols.tls import TLSMemoryBIOFactory, TLSMemoryBIOProtocol
 from twisted.python.failure import Failure
-from twisted.web.client import Agent
 
 from hathor.conf.settings import HathorSettings
 from hathor.p2p.netfilter.factory import NetfilterFactory
@@ -37,7 +36,7 @@ from hathor.p2p.rate_limiter import RateLimiter
 from hathor.p2p.states.ready import ReadyState
 from hathor.p2p.sync_factory import SyncAgentFactory
 from hathor.p2p.sync_version import SyncVersion
-from hathor.p2p.utils import parse_whitelist
+from hathor.p2p.whitelist_manager import WhitelistManager
 from hathor.pubsub import HathorEvents, PubSubManager
 from hathor.reactor import ReactorProtocol as Reactor
 from hathor.transaction import BaseTransaction
@@ -47,9 +46,6 @@ if TYPE_CHECKING:
     from hathor.manager import HathorManager
 
 logger = get_logger()
-
-# The timeout in seconds for the whitelist GET request
-WHITELIST_REQUEST_TIMEOUT = 45
 
 
 class _SyncRotateInfo(NamedTuple):
@@ -85,7 +81,7 @@ class ConnectionsManager:
     new_connection_from_queue: deque[PeerId]
     connecting_peers: dict[IStreamClientEndpoint, _ConnectingPeer]
     handshaking_peers: set[HathorProtocol]
-    whitelist_only: bool
+    whitelist: WhitelistManager
     verified_peer_storage: VerifiedPeerStorage
     _sync_factories: dict[SyncVersion, SyncAgentFactory]
     _enabled_sync_versions: set[SyncVersion]
@@ -100,7 +96,6 @@ class ConnectionsManager:
         pubsub: PubSubManager,
         ssl: bool,
         rng: Random,
-        whitelist_only: bool,
         enable_ipv6: bool,
         disable_ipv4: bool,
     ) -> None:
@@ -187,16 +182,15 @@ class ConnectionsManager:
         self.lc_connect.clock = self.reactor
         self.lc_connect_interval = 0.2  # seconds
 
-        # A timer to try to reconnect to the disconnect known peers.
-        if self._settings.ENABLE_PEER_WHITELIST:
-            self.wl_reconnect = LoopingCall(self.update_whitelist)
-            self.wl_reconnect.clock = self.reactor
-
         # Pubsub object to publish events
         self.pubsub = pubsub
 
-        # Parameter to explicitly enable whitelist-only mode, when False it will still check the whitelist for sync-v1
-        self.whitelist_only = whitelist_only
+        # Whitelist manager: owns enabled flag, active policy, peer list, refresh loop and gating decision.
+        self.whitelist = WhitelistManager(
+            settings=self._settings,
+            reactor=self.reactor,
+            drop_connection=self.drop_connection_by_peer_id,
+        )
 
         # Parameter to enable IPv6 connections
         self.enable_ipv6 = enable_ipv6
@@ -210,9 +204,6 @@ class ConnectionsManager:
         # sync-manager factories
         self._sync_factories = {}
         self._enabled_sync_versions = set()
-
-        # agent to perform HTTP requests
-        self._http_agent = Agent(self.reactor)
 
     def add_sync_factory(self, sync_version: SyncVersion, sync_factory: SyncAgentFactory) -> None:
         """Add factory for the given sync version, must use a sync version that does not already exist."""
@@ -298,28 +289,12 @@ class ConnectionsManager:
         self.lc_reconnect.start(5, now=False)
         self.lc_sync_update.start(self.lc_sync_update_interval, now=False)
 
-        if self._settings.ENABLE_PEER_WHITELIST:
-            self._start_whitelist_reconnect()
+        self.whitelist.start()
 
         for description in self.listen_address_descriptions:
             self.listen(description)
 
         self.do_discovery()
-
-    def _start_whitelist_reconnect(self) -> None:
-        # The deferred returned by the LoopingCall start method
-        # executes when the looping call stops running
-        # https://docs.twistedmatrix.com/en/stable/api/twisted.internet.task.LoopingCall.html
-        d = self.wl_reconnect.start(30)
-        d.addErrback(self._handle_whitelist_reconnect_err)
-
-    def _handle_whitelist_reconnect_err(self, *args: Any, **kwargs: Any) -> None:
-        """ This method will be called when an exception happens inside the whitelist update
-            and ends up stopping the looping call.
-            We log the error and start the looping call again.
-        """
-        self.log.error('whitelist reconnect had an exception. Start looping call again.', args=args, kwargs=kwargs)
-        self.reactor.callLater(30, self._start_whitelist_reconnect)
 
     def _start_peer_connect_loop(self) -> None:
         # The deferred returned by the LoopingCall start method
@@ -596,47 +571,6 @@ class ConnectionsManager:
         # when the peer is disconnected and without entrypoint
         for peer in list(self.verified_peer_storage.values()):
             self.connect_to_peer(peer, int(now))
-
-    def update_whitelist(self) -> Deferred[None]:
-        from twisted.web.client import readBody
-        from twisted.web.http_headers import Headers
-        assert self._settings.WHITELIST_URL is not None
-        self.log.info('update whitelist')
-        d = self._http_agent.request(
-            b'GET',
-            self._settings.WHITELIST_URL.encode(),
-            Headers({'User-Agent': ['hathor-core']}),
-            None)
-        d.addCallback(readBody)
-        d.addTimeout(WHITELIST_REQUEST_TIMEOUT, self.reactor)
-        d.addCallback(self._update_whitelist_cb)
-        d.addErrback(self._update_whitelist_err)
-
-        return d
-
-    def _update_whitelist_err(self, *args: Any, **kwargs: Any) -> None:
-        self.log.error('update whitelist failed', args=args, kwargs=kwargs)
-
-    def _update_whitelist_cb(self, body: bytes) -> None:
-        assert self.manager is not None
-        self.log.info('update whitelist got response')
-        try:
-            text = body.decode()
-            new_whitelist = parse_whitelist(text)
-        except Exception:
-            self.log.exception('failed to parse whitelist')
-            return
-        current_whitelist = set(self.manager.peers_whitelist)
-        peers_to_add = new_whitelist - current_whitelist
-        if peers_to_add:
-            self.log.info('add new peers to whitelist', peers=peers_to_add)
-        peers_to_remove = current_whitelist - new_whitelist
-        if peers_to_remove:
-            self.log.info('remove peers peers from whitelist', peers=peers_to_remove)
-        for peer_id in peers_to_add:
-            self.manager.add_peer_to_whitelist(peer_id)
-        for peer_id in peers_to_remove:
-            self.manager.remove_peer_from_whitelist_and_disconnect(peer_id)
 
     def connect_to_peer(self, peer: UnverifiedPeer | PublicPeer, now: int) -> None:
         """ Attempts to connect if it is not connected to the peer.
--- a/hathor/p2p/states/peer_id.py
+++ b/hathor/p2p/states/peer_id.py
@@ -162,29 +162,10 @@ class PeerIdState(BaseState):
         self.send_ready()
 
     def _should_block_peer(self, peer_id: PeerId) -> bool:
-        """ Determine if peer should not be allowed to connect.
+        """Determine whether the peer should be denied admission.
 
-        Currently this is only because the peer is not in a whitelist and whitelist blocking is active.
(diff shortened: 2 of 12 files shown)


hathor/p2p/manager.py
```diff
@@ -198,6 +199,10 @@ def __init__(
         # Parameter to explicitly enable whitelist-only mode, when False it will still check the whitelist for sync-v1
         self.whitelist_only = whitelist_only
```
msbrogli: Let's unify this in a WhitelistManager that will have an enabled flag, the current policy and will check if a connection is allowed. This WhitelistManager can update itself, moving the code away from P2PManager.

msbrogli: Done in 3d081203 — moved enabled/policy/peer list/loop/admission into a new `WhitelistManager` (hathor/p2p/whitelist_manager.py). `_should_block_peer` now delegates to `whitelist.is_connection_allowed()`. Removed the corresponding fields and methods from ConnectionsManager and HathorManager.

hathor/p2p/whitelist_manager.py
```diff
@@ -0,0 +1,165 @@
...
+        self.peers: list[PeerId] = []
+
+        self._connections: Optional['ConnectionsManager'] = None
+        self._http_agent: Optional[Agent] = None
+        self._loop: Optional[LoopingCall] = None
+        if self.enabled:
+            self._loop = LoopingCall(self.update)
+            self._loop.clock = reactor
+
+    def set_connections(self, connections: 'ConnectionsManager') -> None:
```
msbrogli: Why not a predicate that can be called to drop connections?

hathor/p2p/whitelist_manager.py
```diff
@@ -0,0 +1,165 @@
...
+            self._connections.drop_connection_by_peer_id(peer_id)
+
+    def is_connection_allowed(self, peer_id: PeerId, sync_version: SyncVersion) -> bool:
+        """Return True if the peer is allowed to connect under the current policy."""
+        if peer_id in self.peers:
+            return True
+        if self.policy is WhitelistPolicy.ALLOW_ALL:
+            return True
+        if self.enabled:
+            if sync_version.is_v1():
```
msbrogli: Let's remove sync version v1 and also USE_PEER_WHITELIST_ON_SYNC_V2.

hathor/p2p/whitelist_manager.py
msbrogli: Done in 102ab07e — `WhitelistManager` now accepts a `drop_connection: Callable[[PeerId], None]` predicate at construction; `set_connections` is gone.

hathor/p2p/whitelist_manager.py
msbrogli: Done in 102ab07e — dropped the sync-v1 branch and removed `USE_PEER_WHITELIST_ON_SYNC_V2` (also from `hathorlib/hathorlib/conf/settings.py`). `is_connection_allowed` no longer takes `sync_version`; gating is now: peer in list → allow; ALLOW_ALL → allow; otherwise `enabled or whitelist_only` → block.

jansegre (dismissed):

hathor/p2p/utils.py
```diff
@@ -130,6 +131,15 @@ def generate_certificate(private_key: RSAPrivateKey, ca_file: str, ca_pkey_file:
     return certificate
 
 
+class WhitelistPolicy(Enum):
```
glevco: Create a `StrEnum`?

hathor/p2p/utils.py
```diff
@@ -143,8 +153,15 @@ def parse_file(text: str, *, header: Optional[str] = None) -> list[str]:
     return list(nonblank_lines)
 
 
-def parse_whitelist(text: str, *, header: Optional[str] = None) -> set[PeerId]:
-    """ Parses the list of whitelist peer ids
+def parse_whitelist(
+    text: str, *, header: Optional[str] = None
```
glevco: The `header` argument looks unused — every caller of `parse_whitelist` only passes `text`. Should we remove it?

hathor/p2p/utils.py
```diff
@@ -143,8 +153,15 @@ def parse_file(text: str, *, header: Optional[str] = None) -> list[str]:
...
+) -> tuple[set[PeerId], WhitelistPolicy]:
+    """Parses the whitelist content and returns peers and active policy.
+
+    The optional `# policy: <value>` directive MUST appear before any peer-ID
+    line. It is written as a commented line so older parsers (which strip
+    `#` lines) remain compatible. If absent, the policy defaults to
+    ONLY_WHITELISTED_PEERS.
```
glevco: Use doctests.

hathor/p2p/utils.py
```diff
@@ -158,11 +175,46 @@ def parse_whitelist(text: str, *, header: Optional[str] = None) -> set[PeerId]:
...
+    (set(), WhitelistPolicy.ALLOW_ALL)
     """
-    lines = parse_file(text, header=header)
-    return {PeerId(line.split()[0]) for line in lines}
+    if header is None:
+        header = 'hathor-whitelist'
+    lines = text.splitlines()
+    _header = lines.pop(0)
+    if _header != header:
+        raise ValueError('invalid header')
```
glevco: Why duplicate the header parsing from `parse_file`?

hathor/p2p/utils.py
```diff
@@ -158,11 +175,46 @@ def parse_whitelist(text: str, *, header: Optional[str] = None) -> set[PeerId]:
...
+                if policy_seen:
+                    raise ValueError('duplicate policy directive')
+                value = comment.split(':', 1)[1].strip().lower()
+                try:
+                    policy = WhitelistPolicy(value)
+                except ValueError:
+                    raise ValueError(f'invalid whitelist policy: {value}')
+                policy_seen = True
+            continue
+        peers.add(PeerId(line.split()[0]))
```
glevco: A single malformed peer ID will raise and abort the entire whitelist refresh. Should we warn and skip instead?

hathor/p2p/utils.py
```diff
@@ -158,11 +175,46 @@ def parse_whitelist(text: str, *, header: Optional[str] = None) -> set[PeerId]:
...
+
+    return peers, policy
```
glevco, line 223: Should we check that the set of peers is empty when the policy is `ALLOW_ALL`?

hathor/p2p/utils.py
msbrogli: 🤖 Done in f8573cd1 — `WhitelistPolicy` now extends `StrEnum`.

hathor/p2p/utils.py
msbrogli: 🤖 Done in f8573cd1 — dropped the `header` kwarg from `parse_whitelist`. Kept it on `parse_file` since the netfilter caller passes a different header.

hathor/p2p/utils.py
msbrogli: 🤖 Done in f8573cd1 — converted the docstring examples to runnable `>>>` doctests. Verified with `python -m doctest`.

hathor/p2p/utils.py
msbrogli: 🤖 Done in f8573cd1 — extracted `_pop_header(text, expected)` and routed both `parse_file` and `parse_whitelist` through it. (Could not have `parse_whitelist` simply call `parse_file` because it needs to see the comment lines to find the `# policy:` directive.)

hathor/p2p/utils.py
msbrogli: 🤖 Done in f8573cd1 — malformed peer IDs are now caught per-line (`ValueError`/`TypeError`), logged via structlog with the offending token, and skipped. The refresh continues with the rest of the file. Added a regression test (`test_parse_whitelist_skips_malformed_peer_ids`).

hathor/p2p/utils.py
msbrogli: 🤖 Done in f8573cd1 — `parse_whitelist` now raises `ValueError("peer list must be empty when policy is allow-all")` when both are present. Added a regression test (`test_parse_whitelist_allow_all_with_peers_is_rejected`).

jansegre (dismissed):

hathor/p2p/utils.py
glevco: Thanks!

hathor/p2p/utils.py
glevco: Thanks!

hathor/p2p/utils.py
glevco: Thanks!

hathor/p2p/utils.py
glevco: Thanks!

hathor/p2p/utils.py
glevco: Thanks!

hathor/p2p/utils.py
glevco: Thanks!

hathor/p2p/states/peer_id.py
```diff
@@ -162,29 +162,11 @@ async def handle_peer_id(self, payload: str) -> None:
...
+        """Determine whether the peer should be denied admission.
 
-        Currently this is only because the peer is not in a whitelist and whitelist blocking is active.
+        The decision is delegated to the connection's WhitelistManager, which knows the
+        active policy, the whitelist contents, and the local enforcement flags.
         """
-        peer_is_whitelisted = peer_id in self.protocol.node.peers_whitelist
-        # never block whitelisted peers
-        if peer_is_whitelisted:
+        if self.protocol.connections is None:
```
glevco: This should be an assertion, right?

hathor/p2p/whitelist_manager.py
```diff
@@ -0,0 +1,152 @@
...
+    the periodic refresh loop. Centralizes the connection-admission decision
+    so that other modules only consult `is_connection_allowed`.
+    """
+
+    def __init__(
+        self,
+        *,
+        settings: HathorSettings,
+        reactor: Reactor,
+        whitelist_only: bool,
```
glevco: It looks like this is not set to `True` by any code path, so we could just remove it.

hathor/p2p/whitelist_manager.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _handle_loop_err(self, *args: Any, **kwargs: Any) -> None:
+        """Called when the LoopingCall stops due to an exception. Restart it."""
+        self.log.error('whitelist reconnect had an exception. Start looping call again.',
+                       args=args, kwargs=kwargs)
+        self._reactor.callLater(30, self._start_loop)
+
+    def add_peer(self, peer_id: PeerId) -> None:
+        if not self.enabled:
+            return
```
glevco: Should this be an assertion?

hathor/p2p/whitelist_manager.py
```diff
@@ -0,0 +1,152 @@
...
+
+        # When True, peers absent from the list are blocked.
+        self.whitelist_only = whitelist_only
+
+        # When True, the protocol-level whitelist mechanism is active and the manager
+        # periodically fetches and applies updates from WHITELIST_URL.
+        self.enabled = settings.ENABLE_PEER_WHITELIST
+
+        self.policy: WhitelistPolicy = WhitelistPolicy.ONLY_WHITELISTED_PEERS
+        self.peers: list[PeerId] = []
```
glevco: Why not a `set`?

hathor/p2p/whitelist_manager.py
```diff
@@ -0,0 +1,152 @@
...
+        if not self.enabled:
+            return
+        if peer_id in self.peers:
+            self.log.info('peer already in whitelist', peer_id=peer_id)
+            return
+        self.peers.append(peer_id)
+
+    def remove_peer_and_disconnect(self, peer_id: PeerId) -> None:
+        if not self.enabled:
+            return
```
glevco: Should this be an assertion?

glevco (dismissed):

hathor/p2p/states/peer_id.py
msbrogli: 🤖 Done in ec831f5d — replaced with `assert self.protocol.connections is not None`. The type is non-Optional `ConnectionsManager` (protocol.py:82) and protocol.py:290 already uses the assertion form, so this matches the existing convention.

hathor/p2p/whitelist_manager.py
msbrogli: 🤖 Done in ec831f5d — removed the `whitelist_only` parameter from `WhitelistManager`, `ConnectionsManager`, and both builders (it was hardcoded to `False` everywhere). `is_connection_allowed` is now `peer in list → allow; ALLOW_ALL → allow; enabled → block; else allow`.

hathor/p2p/whitelist_manager.py
msbrogli: 🤖 Done in ec831f5d — replaced the `if not self.enabled: return` guard with `assert self.enabled`. `add_peer` is only called from `_update_cb`, which only runs from the LoopingCall that is created and started exclusively when `enabled=True`.

hathor/p2p/whitelist_manager.py
msbrogli: 🤖 Done in ec831f5d — switched `self.peers` from `list[PeerId]` to `set[PeerId]`. All access patterns are membership tests, dedup, and set diffs in `_update_cb`, which now use `self.peers` directly (the temporary `current = set(self.peers)` is gone).

hathor/p2p/whitelist_manager.py
msbrogli: 🤖 Done in ec831f5d — same change as `add_peer`: replaced `if not self.enabled: return` with `assert self.enabled`. `remove_peer_and_disconnect` is also only reachable from `_update_cb`, which only fires when the LoopingCall is running (i.e. `enabled=True`).

glevco (dismissed):

jansegre (dismissed):

jansegre (approved):

glevco (approved):

PR #1677, opened 2026-04-28
refactor(daa): split into pure algorithm + DAAFactory

msbrogli:
### Motivation

The DAA module was a monolithic `DifficultyAdjustmentAlgorithm` class that mixed three concerns: the algorithm itself, version-specific values, and the dispatch logic that picks which version applies to a given block. As we prepare to add feature-gated DAA versions (e.g. `REDUCE_DAA_TARGET` for faster blocks), this entanglement makes each new version a sprawling change that touches both the algorithm and every holder.

This PR separates the three roles so future feature-gated DAA versions can plug in cleanly, without changing the algorithm or the call sites:

- **`DifficultyAdjustmentAlgorithm`** — pure algorithm, takes `(settings, config)` and runs.
- **`DAAConfig`** — frozen value object describing a version (currently just `avg_time_between_blocks`); `DAAConfig.for_v1(settings)` builds the V1 instance.
- **`DAAFactory`** — carries dispatch context (`settings`, `tx_storage`, `feature_service`) plus mutable `TEST_MODE` / `MIN_BLOCK_WEIGHT`, and exposes `create_v1`, `create_from_parent`, `create_from_block`, `create_from_best_block`.

Holders (manager, verifiers, dag_builder, simulator, builder, peer, resources) now receive `daa_factory: DAAFactory` and build the right DAA per call. `feature_service` is wired into the factory but unused in this PR — the seam is in `_select_config`, which currently always returns V1. The follow-up feature PR will only need to extend `DAAConfig`, add `for_v2` / `create_v2`, and fill in `_select_config` — no holder changes.

### Acceptance Criteria

- [x] `hathor/daa.py` is split into a `hathor/daa/` package: `common.py` (utilities, types, `DAAConfig`), `daa.py` (`DifficultyAdjustmentAlgorithm`), `factory.py` (`DAAFactory`).
- [x] `DifficultyAdjustmentAlgorithm` is parameterized by `DAAConfig`; no version-specific logic lives in the class.
- [x] `DAAFactory` exposes `create_v1`, `create_from_parent`, `create_from_block`, `create_from_best_block`, and the `_select_config` seam.
- [x] All holders take `daa_factory: DAAFactory` and build per-call DAAs via the factory methods.
- [x] Zero behavior change vs `master`: V1 mining / verification / template generation are unchanged.
- [x] `poetry run pytest hathor_tests/` — full suite passes (2502 tests, no regressions).
- [x] `poetry run ruff check` and `poetry run mypy` — no new issues vs `master` baseline.

### Checklist


hathor/verification/transaction_verifier.py
```diff
@@ -94,7 +94,7 @@ def verify_parents_basic(self, tx: Transaction) -> None:
     def verify_weight(self, tx: Transaction) -> None:
         """Validate minimum tx difficulty."""
         assert self._settings.CONSENSUS_ALGORITHM.is_pow()
-        min_tx_weight = self._daa.minimum_tx_weight(tx)
+        min_tx_weight = self._daa_factory.create_from_best_block().minimum_tx_weight(tx)
```
msbrogli: DAAFactory should have a tool that calculate `minimum_tx_weight(tx)` directly because it is independent of the block.

hathor/daa/common.py
```diff
@@ -0,0 +1,244 @@
...
+    min_block_weight: float,
+    test_mode: TestMode,
+) -> float:
+    """Calculate the next block weight, aka DAA/difficulty adjustment algorithm.
+
+    The algorithm used is described in `RFC 22`_.
+
+    The weight must not be less than ``min_block_weight``.
+
+    .. _RFC 22: https://gitlab.com/HathorNetwork/rfcs/merge_requests/22
```
glevco: What's this syntax?

hathor/verification/transaction_verifier.py
msbrogli: Done in 0741faef. Moved both `minimum_tx_weight` and `get_weight_decay_amount` to `DAAFactory` and removed them from `DifficultyAdjustmentAlgorithm` — they only need `settings` + `TEST_MODE`, never a `DAAConfig`, so this puts them where they belong and gives a single obvious call site. All production call sites are now `factory.minimum_tx_weight(tx)` directly (no more `create_from_best_block()` indirection); the two tests that called them on a directly-constructed DAA were migrated to the factory.

hathor/manager.py
```diff
@@ -801,7 +802,7 @@ def generate_mining_block(self, timestamp: Optional[int] = None,
 
     def get_tokens_issued_per_block(self, height: int) -> int:
         """Return the number of tokens issued (aka reward) per block of a given height."""
-        return self.daa.get_tokens_issued_per_block(height)
+        return self.daa_factory.create_from_best_block().get_tokens_issued_per_block(height)
```
msbrogli: As it is per height, should this method be moved to the factory? It seems to be independent of the best block.

msbrogli: Done in fb089208. Same principle as the previous review: methods that don't depend on a `DAAConfig` belong on the factory. Moved `get_tokens_issued_per_block`, `get_mined_tokens`, and `get_reward_for_next_block` to `DAAFactory` and removed them from `DifficultyAdjustmentAlgorithm`. Updated all call sites — the `create_from_best_block()` / `create_from_parent()` / `create_v1()` indirection in `manager.get_tokens_issued_per_block`, `BlockVerifier.verify_reward`, `mining_info`, and `default_filler` is gone; they call the factory method directly. `DifficultyAdjustmentAlgorithm` now only holds methods that actually use `_config` (`calculate_block_difficulty`, `calculate_next_weight`) plus `get_block_dependencies`.

jansegre (dismissed):

hathor/daa/daa.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @property
+    def avg_time_between_blocks(self) -> float:
+        return self._config.avg_time_between_blocks
+
+    @cpu.profiler(key=lambda _, block: 'calculate_block_difficulty!{}'.format(block.hash.hex()))
+    def calculate_block_difficulty(self, block: Block, parent_block_getter: Callable[[Block], Block]) -> float:
+        """Calculate block weight according to the ascendants of `block`."""
+        if self.TEST_MODE & TestMode.TEST_BLOCK_WEIGHT:
+            return 1.0
```
glevco, line 68: This is already in `_calculate_next_weight`

hathor/daa/daa.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def calculate_next_weight(
+        self,
+        parent_block: Block,
+        timestamp: int,
+        parent_block_getter: Callable[[Block], Block],
+    ) -> float:
+        """Public method for template creation."""
+        if self.TEST_MODE & TestMode.TEST_BLOCK_WEIGHT:
+            return 1.0
```
glevco, line 85: This is already in `_calculate_next_weight`

hathor/daa/factory.py
```diff
@@ -0,0 +1,131 @@
...
+            return self.create_v1()
+        return self.create_from_parent(block.get_block_parent())
+
+    def create_from_best_block(self) -> DifficultyAdjustmentAlgorithm:
+        """Build a DAA treating the current best block as the parent of a hypothetical next block.
+
+        Falls back to V1 when no tx_storage is wired (e.g. CLI tools).
+        """
+        if self._tx_storage is None:
+            return self.create_v1()
```
glevco: I would rather make this behavior explicit so we don't accidentally use it in a place where it's supposed to retrieving feature state.

glevco (changes requested):

hathor/daa/common.py
msbrogli: 🤖 Standard reStructuredText hyperlink target syntax — pairs with `` `RFC 22`_ `` above to render as a link in Sphinx-rendered docs. Replaced with a plain inline URL in e0711073 since the rest of the codebase doesn't use rST link references.

hathor/daa/daa.py
msbrogli: 🤖 Addressed in e0711073 by deduplicating the other direction: removed the `test_mode` parameter and short-circuit from the inner `_calculate_next_weight` helper, kept it here. The outer check is needed to win over the `block.is_genesis` branch below — without it, `test_genesis_weight` regresses (genesis returns `MIN_BLOCK_WEIGHT=2` instead of `1.0` under `TEST_BLOCK_WEIGHT`). The private helper is now agnostic; the test-mode policy lives only at the public API.

hathor/daa/daa.py
msbrogli: 🤖 Same fix as the previous thread (e0711073): the duplication is gone, but the direction reversed — the inner `_calculate_next_weight` no longer takes `test_mode` and no longer short-circuits. The check stays in the public `calculate_next_weight` and `calculate_block_difficulty`.

hathor/daa/factory.py
msbrogli: 🤖 Made it structurally impossible in e0711073: `create_from_best_block` had zero callers after the helper moves in 0741faef / fb089208 (everything that used to need it now goes through factory methods or factory-level helpers directly). Deleted the method entirely, and since `_tx_storage` was only used here, removed it from `DAAFactory.__init__` too. Updated `hathor/builder/builder.py` and `hathor_cli/builder.py` to stop passing `tx_storage=`. No fallback to accidentally misuse.

glevco (approved):

PR #1688, opened 2026-05-07
feat(indexes): fast mempool-tips initialization

jansegre:
### Motivation

The mempool tips index is kept in memory because the original RocksDB-backed implementation was too slow, but the in-memory variant has to be rebuilt at every startup by iterating over all transactions. In practice the mempool is empty most of the time (notably right after a clean shutdown when sync has caught up), so that rebuild is pure wasted work.

This PR adds a lightweight persisted "empty" marker for the mempool tips index so that, when it applies, initialization can be skipped entirely. To support this it also introduces a generic mechanism on `BaseIndex` that lets any index opt out of rebuild based on persisted metadata, keeping the change reusable for future indexes.

### Acceptance Criteria

- `BaseIndex` exposes two new optional hooks — `still_needs_initialization(tx_storage)` and `init_finish(tx_storage)` — defaulting to "always rebuild" and a no-op respectively, so existing indexes are unaffected.
- `IndexesManager` consults `still_needs_initialization` after `init_start` to drop indexes from the rebuild set, logs which ones were skipped, and calls `init_finish` on every index that was initially selected for initialization.
- A new `SimpleRocksDBMempoolTipsIndex` replaces `MemoryMempoolTipsIndex` in `RocksDBIndexesManager`: it stays memory-backed for reads/writes but persists an "empty" marker in a dedicated column family whenever the in-memory set transitions to/from empty.
- On startup, if the empty marker is present, the mempool tips index skips the full transaction scan; if the marker is absent, the normal rebuild still runs and the marker is re-synced at the end of initialization.
- Tests cover the three relevant cases: marker is written and lets a subsequent startup skip iteration entirely, a missing marker correctly rebuilds a non-empty mempool, and the marker is kept in sync as transactions enter and leave the mempool at runtime.

### Checklist

### DIFF
--- a/hathor/indexes/manager.py
+++ b/hathor/indexes/manager.py
@@ -142,14 +142,16 @@ class IndexesManager(ABC):
             if db_last_started_at != index_last_started_at:
                 indexes_to_init.append(index)
 
-        if indexes_to_init:
-            indexes_names = [type(index).__name__ for index in indexes_to_init]
+        initial_indexes_to_init = indexes_to_init.copy()
+
+        if initial_indexes_to_init:
+            indexes_names = [type(index).__name__ for index in initial_indexes_to_init]
             self.log.info('there are indexes that need initialization', indexes_to_init=indexes_names)
         else:
             self.log.info('there are no indexes that need initialization')
 
         # make sure that all the indexes that we're rebuilding are cleared
-        for index in indexes_to_init:
+        for index in initial_indexes_to_init:
             index_db_name = index.get_db_name()
             if index_db_name:
                 tx_storage.set_index_last_started_at(index_db_name, NULL_INDEX_LAST_STARTED_AT)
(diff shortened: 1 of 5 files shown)


hathor/indexes/manager.py
```diff
@@ -187,6 +195,9 @@ def _manually_initialize(self, tx_storage: 'TransactionStorage') -> None:
                 if index.get_scope().matches(tx):
                     index.init_loop_step(tx)
 
+        for index in initial_indexes_to_init:
+            index.init_finish(tx_storage)
```
glevco: Shouldn't we only call it for `indexes_to_init`?

glevco (dismissed):

hathor/indexes/manager.py
jansegre: Yeah, you're right. Fixed.

glevco: Thanks!

glevco (approved):

pedroferreira1 (approved):

PR #1670, opened 2026-04-28
feat(shielded): add ShieldedOutputs/UnshieldBalance/Mint/Melt headers + accessors

msbrogli:
### Motivation

Round 2 of this PR introduces the four new wire-format headers needed by the upcoming shielded transactions feature, plus everything else needed to consume them from Python code: the `ENABLE_SHIELDED_TRANSACTIONS` feature flag, the parser registration that gates on it, and the `Transaction` / `GenericVertex` accessors that surface header presence and shielded outputs to callers. Verification semantics, the Rust crypto wrappers, and wallet/index/sync integration still ship in follow-up PRs.

The structural shape changed in response to PR review feedback:
- The four header classes live entirely in **hathorlib**; hathor-core re-exports them from `hathor.transaction.headers` so consensus code keeps a single import path while there is only one canonical implementation.
- Wire-format work moved out of the header classes into per-header modules under a new `hathorlib/vertex_parser/` package — mirroring `hathor/transaction/vertex_parser/_<name>_header.py` (the layout hathor-core already uses for `FeeHeader` / `NanoHeader`). Each module exposes Serializer/Deserializer-receiving free functions, and the header classes become thin wrappers (~6 lines each) that build a `BytesSerializer` / `BytesDeserializer` and delegate. This is the inversion-of-control reviewers asked for: a future whole-vertex orchestrator can call the helpers directly with one Serializer pass instead of going through each class's bytes-in/bytes-out wrapper.

#### Headers (`serialize` / `deserialize` / `get_sighash_bytes`)

| Header | ID | Wire format |
|---|---|---|
| `ShieldedOutputsHeader` | `0x12` | `header_id(1) | num_outputs(1) | <output blobs>` |
| `UnshieldBalanceHeader` | `0x13` | `header_id(1) | excess_blinding_factor(32)` |
| `MintHeader` | `0x14` | `header_id(1) | num_entries(1) | (token_index(1) | amount(8 BE))*N` |
| `MeltHeader` | `0x15` | same shape as `MintHeader` |

- New IDs registered in `headers/types.py` for both packages; new exports in `headers/__init__.py` for both packages (and `serialize_entries` / `deserialize_entries` are NO LONGER public — they're private to `vertex_parser/_mint_melt_header.py`).
[...]

PR #1702, opened 2026-05-18
feat(tx): add shielded output headers

LFRezende:
### Summary

  This is **PR 4 of 8** in the staged rollout of confidential ("shielded") transactions — the reviewable split of the squashed work in https://github.com/HathorNetwork/hathor-core/blob/feat/shielded-outputs/docs/plans/shielded-pr-split.md 
  (confidential transactions, modeled on Liquid Confidential Assets: Pedersen commitments, Borromean range proofs, asset surjection proofs, ECDH rewind).

  > [!IMPORTANT]
  > **This is an intermediate, foundational step only.** It introduces the *data model*, the *wire-format headers*, and the *feature flag* — and nothing
  else. There is **no verification logic, no wallet support, no indexing/sync/API, and no native crypto** in this PR. The feature ships **disabled by 
  default** and is **non-functional** on its own; it exists so later PRs have a stable base to build on.

  The full sequence:

  | PR | Scope | Status |
  |----|-------|--------|
  | 1 | `chore`: rename `verify_sum → verify_transparent_balance` | ✅ merged |
  | 2 | `refactor`: `resolve_spent_output` plumbing | ✅ merged |
  | 3 | ~~canonical header ordering~~ | retracted (no ordering enforced) |
  | **4** | **`feat`: data model + headers + feature flag (gated OFF)** | **← this PR** |
  | 5 | `feat`: **verification** (balance dispatch, range/surjection proofs) — security-critical | upcoming |
  | 6 | `feat`: wallet (ECDH UTXO discovery) | upcoming |
  | 7 | `feat`: indexes, sync, API, events, DAG builder | upcoming |
  | 8 | `feat`: Rust crypto lib + build/CI wiring (**ships last**) | upcoming |

  ### What's in this PR

  **Data model**
  - `hathor/transaction/shielded_tx_output.py` — shielded output types (`AmountShieldedOutput`, `FullShieldedOutput`), their `OutputMode`, plus
  (de)serialization and sighash bytes.
  - `hathor/crypto/shielded/` — the Python surface for the future native crate (`hathor_ct_crypto`). It **degrades gracefully**: if the native lib is absent
  (which it is, until PR 8), `SHIELDED_CRYPTO_AVAILABLE` is `False` and importing the package never fails.

  **Headers (wire format)**
  - `ShieldedOutputsHeader` (type id `0x12`) and `UnshieldBalanceHeader` (type id `0x13`) — definitions + (de)serialization, registered in the vertex parser.
  - _MintHeader (`0x14`) / MeltHeader (`0x15`) are a separate post-plan extension and are intentionally **not** here._

  **Feature flag + plumbing**
  - `ENABLE_SHIELDED_TRANSACTIONS: FeatureSetting = DISABLED` in settings.
  - `Feature.SHIELDED_TRANSACTIONS` + `Features.shielded_transactions`, threaded through `Features.{from_vertex, for_mempool, all_enabled}`.
  - Read-only `Transaction` accessors (`has_shielded_outputs`, `is_shielded`, `shielded_outputs`, `has_unshield_balance_header`, `excess_blinding_factor`,
  …).
  - A node-startup guard (`validate_shielded_crypto_available`) that fails fast **only** if the flag is enabled without the native lib.
  - Data-model-level typed exceptions and fee-rate settings constants.

  **Paired `hathorlib` changes** — the matching header/serialization additions land in `hathorlib` (its own version bump).

  | Deferred | Lands in |
  |----------|----------|
  | Verification semantics (balance, range/surjection proofs, commitment validity) | PR 5 |
  | Wallet (ECDH UTXO discovery + decryption) | PR 6 |
  | Indexes, sync, API, events, DAG builder | PR 7 |
  | Native Rust crate (`hathor_ct_crypto`) + build/CI wiring | PR 8 |
  | Mint/melt headers (`0x14`/`0x15`) | post-plan extension |
  | Canonical header ordering / `get_header_id()` | retracted — not present |

  ### 🔒 Safety: fully gated by `ENABLE_SHIELDED_TRANSACTIONS` (default `DISABLED`)

  With the flag disabled (the default — and the only state this PR ships), **no shielded behavior is reachable**:

  - **The parser is the airtight gate.** The vertex parser only registers the shielded header IDs when the flag is enabled. A flag-off node **cannot 
  deserialize** a transaction carrying header `0x12`/`0x13` — it raises `"header type not supported"` before any shielded code runs.
  - **Defense-in-depth at verification.** `get_allowed_headers` admits the shielded headers only when `features.shielded_transactions` is true (and only for
  `REGULAR_TRANSACTION`), which is `False` when the flag is off.
  - **The startup guard is a no-op** while disabled; it only fires if someone enables the flag without the native lib.
  - **Everything else is inert or dead code in this PR**: the consensus feature-activation hook is a documented no-op `pass`; the new `Transaction` accessors
   have **no production callers** yet (their consumers arrive in PRs 5–7) and would return empty/`False` regardless, since no shielded data can exist when
  the flag is off.

  Net effect: a flag-off node behaves exactly as before this PR, and enabling the flag without the native lib fails fast at startup rather than silently
  misbehaving.

  ### Testing

  - `make check` (custom checks + yamllint + ruff + mypy) passes.
  - Existing test suite is green. The data-model / header-round-trip / verification tests that exercise this code require the native crate and therefore land
   alongside it in PR 8 (or behind a skip-guard once it's a dev dependency).


hathor/consensus/consensus.py
```diff
@@ -460,6 +460,12 @@ def _feature_activation_rules(self, tx: Transaction, new_best_block: Block) -> b
                 case Feature.REDUCE_DAA_TARGET:
                     # This feature does not affect transaction verification, only DAA parameters and the Nano runtime.
                     pass
+                case Feature.SHIELDED_TRANSACTIONS:
+                    # Verification semantics for shielded transactions — including the
+                    # reorg activation rule — are deferred to a later PR (PR 5). The
+                    # feature ships gated OFF, so no shielded tx can be accepted yet and
+                    # there is nothing to revalidate on reorg here.
+                    pass
```
glevco: You should leave a `raise NotImplementedError` instead of `pass` in this case

hathor/transaction/vertex_parser/_vertex_parser.py
```diff
@@ -41,6 +49,9 @@ def get_supported_headers(settings: HathorSettings) -> dict[VertexHeaderId, Type
             supported_headers[VertexHeaderId.NANO_HEADER] = NanoHeader
         if settings.ENABLE_FEE_BASED_TOKENS:
             supported_headers[VertexHeaderId.FEE_HEADER] = FeeHeader
+        if settings.ENABLE_SHIELDED_TRANSACTIONS != FeatureSetting.DISABLED:
```
glevco: You can just do this, like the others:

```suggestion
        if settings.ENABLE_SHIELDED_TRANSACTIONS:
```

Because `FeatureSetting` implements `__bool__`.

hathor/transaction/shielded_tx_output.py
glevco, line 1: You can remove this file and just use `hathorlib` directly.

hathor/transaction/headers/shielded_outputs_header.py
glevco, line 1: Why not remove this and just use hathorlib's ?

hathorlib/hathorlib/transaction/__init__.py
glevco, line 1: Why this file rename?

hathor/transaction/headers/shielded_outputs_header.py
jansegre: I agree. We can also do this with other headers too later.

hathor/transaction/vertex_parser/_vertex_parser.py
LFRezende: Simpler, I'll do it.

hathor/consensus/consensus.py
LFRezende: Fair enough, I'll include it.

hathor/transaction/shielded_tx_output.py
LFRezende: In this case (the following not so much) I do agree - this is more of a leftover. When I deleted files from future PRs, this one was left due to Mirroring hathorliib, but since `recover_shielded_secrets` is deleted, no need to keep it. I'll remove this and stick to hathorlib's version.

hathor/transaction/headers/shielded_outputs_header.py
LFRezende: I thought it'd be interesting to keep the structure of the other transaction types. After all, ` NanoHeader ` and `FeeHeader ` also have header mirrors in hathor-core, not just hathorlib, so it just made sense to keep this homogeinity. I guess we could, but then we'd have to do this for all header types in my opinion.

I get it is a smell having such duplicates (and Marcelo even started to address this in #1670 ), but as it is, we'd better do this to all headers in the future.

hathorlib/hathorlib/transaction/__init__.py
LFRezende: The previous was a module/file `transaction.py` -> We're turning it into a package `transaction/__init__.py` so we can include different files. We have the shielded_tx_output.py and we're gearing towards matching the structure of #1670 , which implements it as is.

hathor/consensus/consensus.py
glevco: Thanks!

hathor/transaction/vertex_parser/_vertex_parser.py
glevco: Thanks!

hathor/transaction/shielded_tx_output.py
glevco: Thanks!

hathor/transaction/headers/shielded_outputs_header.py
glevco: The duplication exists for nano and fee headers because they were introduced before the hathorlib was unified into this repo. Now that it's here, we should make an effort to put new code there directly, without duplication. It's less refactor work for the future. Indeed, we must move all other types too, but that's for a separate task — let's just do it for new types in this PR.

hathorlib/hathorlib/transaction/__init__.py
glevco: You could just have added a new, empty `__init__.py` file, but I guess this is fine.

hathor/transaction/headers/shielded_outputs_header.py
LFRezende: So, "break the structure" - we'll not mirror here, but leave the rest mirrored, right?

I guess we can do it, by this perspective.

glevco: Yes, for now

LFRezende: One small hiccup that I saw before: the deserializations do not match. We have two `is_instance(tx, Transaction)` guarding each:

- Hathorlib: hathorlib/hathorlib/headers/shielded_outputs_header.py

```[python]
from hathorlib.transaction import Transaction       
  if not isinstance(tx, Transaction):
      raise ValueError
```

- Hathor-Core: hathor/transaction/headers/shielded_outputs_header.py

```[python]
 from hathor.transaction.transaction import Transaction
  if not isinstance(tx, Transaction):
      raise InvalidShieldedOutputError(...)
```
But we pass a hathor-core tx to be deserialized in hathor/transaction/vertex_parser/_headers.py:67

```[python]
shielded_header, leftover = ShieldedOutputsHeader.deserialize(vertex, remaining_bytes)
```

What I mean is, if we get ShieldedOutputsHeader solely by importing from hathorlib, it would call the hathorlib deserialize, guarded by this `is_instance(tx, Transaction)`, but `tx` is hathor-core transaction, not hathorlib. It would  fail at runtime, as these are different.

Does that make sense?

glevco: It does make sense, but this `VertexHeader.deserialize` is an outdated code path. You can remove it and just raise `NotImplementedError` there. The actual deserializer path goes through the standalone functions on `hathor/transaction/vertex_parser/_headers.py`

LFRezende: Fair enough, thanks! 
I'll do it.

hathorlib/hathorlib/base_transaction.py
```diff
@@ -500,6 +500,29 @@ def is_standard(
 
                 return False
 
+        # Check shielded output scripts (same rules as transparent)
+        if hasattr(self, 'has_shielded_outputs') and self.has_shielded_outputs():
```
jansegre: Why do we need a `hasattr` check here? In general I think `hasattr` should be avoided. If there are cases where `self` does not have the `has_shielded_outputs` method, those cases should be more explicitly checked directly instead of using `hasattr`.

hathorlib/hathorlib/base_transaction.py
```diff
@@ -218,7 +218,7 @@ def get_header_from_bytes(self, buf: bytes) -> bytes:
 
     def get_maximum_number_of_headers(self) -> int:
         """Return the maximum number of headers for this vertex."""
-        return 2
+        return 3
```
jansegre, line 221: Maybe this increase needs to be more careful? (i.e. rely on the feature activation)

If you find that it's safe (compatibility-wise) to increase this value without a feature-activation check, could you explain it here? (We don't need to leave a comment on code for this)

hathor/transaction/vertex_parser/_shielded_outputs_header.py
```diff
@@ -0,0 +1,77 @@
...
+    deserialize_shielded_output,
+)
+
+
+def deserialize_shielded_outputs_header(
+    buf: bytes,
+    *,
+    verbose: VerboseCallback = None,
+) -> tuple[list[ShieldedOutput], bytes]:
+    """Parse `header_id(1) | num_outputs(1) | outputs…` and return (outputs, leftover)."""
```
jansegre: I noticed that this function is basically used with the serialization system (the classes in `hathorlib/hathorlib/serialization`) in `deserialize_headers` (`hathor/transaction/vertex_parser/_headers.py`) and its implementation even required adding `replace_remaining` which isn't ideal, instead the implementation of `deserialize_shielded_outputs_header` would be smoother if it was made using the serialization system directly, and it wouldn't require adding `replace_remaining` (which seems very specifically needed for this case). Please consider this refactor.

hathor/transaction/vertex_parser/_unshield_balance_header.py
```diff
@@ -0,0 +1,57 @@
...
+from hathor.transaction.util import VerboseCallback
+from hathorlib.headers.unshield_balance_header import EXCESS_BLINDING_FACTOR_SIZE
+
+
+def deserialize_unshield_balance_header(
+    buf: bytes,
+    *,
+    verbose: VerboseCallback = None,
+) -> tuple[bytes, bytes]:
+    """Parse `header_id(1) | excess_blinding_factor(32)` and return (excess_bf, leftover)."""
```
jansegre: [This comment](https://github.com/HathorNetwork/hathor-core/pull/1702/changes#r3300022570) also applies here.

hathor/transaction/vertex_parser/_shielded_outputs_header.py
jansegre: Oh, and for symmetry the serializer could be implemented using the serialization system as well. Initially when I looked at it I thought it was fine to leave it using `bytes` directly, but since this implementation is a bit tangled we might as well port both serialization and deserialization.

hathorlib/hathorlib/serialization/deserializer.py
```diff
@@ -89,6 +89,14 @@ def read_struct(self, format: str) -> tuple[Any, ...]:
...
         return struct.unpack_from(format, data)
 
+    def replace_remaining(self, data: Buffer) -> None:
+        """Replace the remaining bytes in the deserializer with new data.
+
+        This is used when delegating to a raw-bytes deserializer and then pushing
+        the unconsumed leftover back into the Deserializer.
+        """
+        raise TypeError('this deserializer does not support replace_remaining')
+
```
jansegre: Instead of adding this method we should re-write any raw-bytes deserializer to either use the deserializer system directly, which helps avoid over-consuming bytes, (which I'd say is the better option since we eventually want all serializer/deserializer using this serialization system, _or_ implement it more carefully so it doesn't over-consume bytes and we end up needing to stuff them back into the deserializer.

hathorlib/tests/test_basic.py
jansegre, line 1: Please add tests that check that when `ENABLE_SHIELDED_TRANSACTIONS` is `disabled` it will reject parsing shielded output headers.

jansegre (changes requested):

hathorlib/hathorlib/base_transaction.py
LFRezende: Since we'll have an extra header type tied to a Tx, we'll need to increase it: having NanoHeader, FeeHeader and ShieldedOutputsHeader (or UnshieldedBalanceHeader, but never both at the same time). So it needs to go to three. When we implement mint/melt headers, then at least four.


Now, the explanation on why we need not worry about this, in my opinion:

1. Increasing it to 3 does not invalidate any of the previous vertex, since no tx has previously had more than 2 headers. We ran a scan previously and (up until the time of the scan, a week ago), there were only ~75 multi-header transactions, and they all fit a canonical ordering (no duplicates or invalid parsing). 
2. In the definition of `verify_headers` (hathor/verification/vertex_verifier.py), we avoid header duping and unallowed headers. So if any other header appears in the slot, it should be rejected. The only way to fill in a third-header would be a ShieldedOutputsHeader/UnshieldedBalanceHeader, and these are gated by `ENABLE_SHIELDED_TRANSACTIONS` (off).

hathorlib/hathorlib/base_transaction.py
LFRezende: I changed it to an authoral method and dropped the `hasattr` implementation. See if it fits.

hathor/transaction/vertex_parser/_shielded_outputs_header.py
LFRezende: I dropped the `replace_remaining` approach and  used the serialization system while preserving symmetry between the serializer and deserializer. See if it is decent.

hathor/transaction/vertex_parser/_unshield_balance_header.py
LFRezende: Done as well, see if it is satisfactory as well.

hathorlib/hathorlib/serialization/deserializer.py
LFRezende: Removed it and implemented the serialization system directly.

hathorlib/tests/test_basic.py
LFRezende: Done, check to see if they are enough.

jansegre (dismissed):

hathorlib/hathorlib/transaction/shielded_tx_output.py
```diff
@@ -0,0 +1,188 @@
...
+
+@dataclass(slots=True, frozen=True)
+class FullShieldedOutput:
+    """Both amount and token type hidden. Surjection proof required."""
+    commitment: bytes           # 33B Pedersen commitment
+    range_proof: bytes          # ~3213B Borromean (40-bit)
+    script: bytes               # Locking script
+    asset_commitment: bytes     # 33B blinded asset tag (A = H_token + r_asset*G)
+    surjection_proof: bytes     # Variable, asset surjection proof
+    ephemeral_pubkey: bytes = b''  # 33B compressed secp256k1 pubkey for ECDH recovery
```
glevco: Why not `None` for not present instead of `b''`?

hathor/transaction/vertex_parser/_headers.py
```diff
@@ -57,14 +57,43 @@ def deserialize_headers(
...
+                    deserialize_shielded_outputs_header,
+                )
+                assert isinstance(vertex, Transaction)
+                # Deserialization goes through the standalone free function (framework-based),
+                # not the header class's (outdated) deserialize classmethod. It consumes exactly
+                # the header's bytes from the deserializer, like the Nano/Fee cases above.
+                shielded_outputs = deserialize_shielded_outputs_header(deserializer)
+                # hathorlib's header types `tx` as a hathorlib Transaction; we store the
+                # hathor-core vertex (cross-lib by design — the header never uses `self.tx`).
+                header = ShieldedOutputsHeader(tx=vertex, shielded_outputs=shielded_outputs)  # type: ignore[arg-type]
```
glevco: "the header never uses `self.tx`" - then let's remove it

hathor/transaction/vertex_parser/_headers.py
```diff
@@ -57,14 +57,43 @@ def deserialize_headers(
...
+            case VertexHeaderId.UNSHIELD_BALANCE_HEADER:
+                from hathor.transaction import Transaction
+                from hathor.transaction.headers import UnshieldBalanceHeader
+                from hathor.transaction.vertex_parser._unshield_balance_header import (
+                    deserialize_unshield_balance_header,
+                )
+                assert isinstance(vertex, Transaction)
+                excess_bf = deserialize_unshield_balance_header(deserializer)
+                # See note above: cross-lib `tx` ref, unused by the hathorlib header.
+                header = UnshieldBalanceHeader(tx=vertex, excess_blinding_factor=excess_bf)  # type: ignore[arg-type]
```
glevco: Same here

hathor/verification/vertex_verifier.py
```diff
@@ -208,11 +214,19 @@ def get_allowed_headers(self, vertex: BaseTransaction, params: VerificationParam
...
+                if params.features.fee_tokens:
+                    allowed_headers.add(FeeHeader)
+            case TxVersion.REGULAR_TRANSACTION:
                 if params.features.nanocontracts:
                     allowed_headers.add(NanoHeader)
                 if params.features.fee_tokens:
                     allowed_headers.add(FeeHeader)
+                if params.features.shielded_transactions:
+                    allowed_headers.add(ShieldedOutputsHeader)
+                    allowed_headers.add(UnshieldBalanceHeader)
```
glevco, line 229: Why TokenCreationTx will not support shielded outputs?

LFRezende: We're excluding mint/melt operations for now in the context of shielded outputs, as part of the pr-split plan from marcelo. Check https://github.com/HathorNetwork/hathor-core/blob/feat/shielded-outputs/docs/plans/shielded-pr-split.md, last session (Post 1-8 steps). So if we are able to create tokens in the context of shielded outputs, we mint a shielded token, which we are post-poning th e implementation for the next PRs.

hathor/transaction/headers/shielded_outputs_header.py
glevco: There's still some duplication (for example deserialization of the shielded outputs header), but I guess we can improve it with the other headers.

hathor/verification/vertex_verifier.py
LFRezende: Just so I am clearer: a transaction which carries shielded outputs AND is creating tokens needs to declare the initial supply in its MintHeader (new Header designed for it) which will be done in newer RFCs, not in this 8 steps plan.

hathorlib/hathorlib/transaction/shielded_tx_output.py
LFRezende: Better, indeed. Done.

hathor/transaction/vertex_parser/_headers.py
LFRezende: Removed it from both headers.

hathor/transaction/vertex_parser/_headers.py
LFRezende: Removed.

hathor/transaction/headers/shielded_outputs_header.py
LFRezende: Okay, so we leave it like this?

glevco (approved):

jansegre (approved):

PR #1709, opened 2026-05-25
feat(shielded): Shielded Outputs creation/proof implementations.

LFRezende:
### Motivation

Following the guidelines of the drafted project plan (https://github.com/HathorNetwork/hathor-core/blob/feat/shielded-outputs/docs/plans/shielded-pr-split.md), we are now in conditions to **introduce the layout for the verification of shielded outputs** into hathor-core.

The goal of this PR is to implement the python abstractions of the verification process of shielded outputs. 

While it already advances some of the rust implementation (mainly lib and the creation of an empty crate to store function bindings for type checks), we are postponing rust implementation of the libs to the eighth an final step of the PR-split plan. 

***Mint/Melt headers have also been postponed to beyond the PR-split plan, as the own plan suggests.*

### Acceptance Criteria

* Add `transaction_verifier.py`: Implement verification checks on transactions for Pedersen Commitments, Range Proofs and Asset Surjection Proofs;

* Implement the Verification Service (`hathor/verification/verification_service.py` ) to check shielded output related headers (if so, gate them off), balances (transparent inputs/outputs) and the rules of the token if not shielded.;

* Have `shielded_activation_rule` for the consensus on handling shielded transactions.

* Token update info in transactions via input read and output read (add `hathor/transaction/transaction.py`).

* Add `hathor/crypto/shielded/`:
  * Include `asset_tag.py` - Wrapper to normalize tokenUid
  * Include `balance.py` - Wrapper to verify balance and safeguard `excess_blinding_factor` scenarios.
  * Include `_bindings.py` - Current implementation imports them from rust crate.

* All shielded-themed coded must be GATED OFF
* No type check - bindings must cover all functions.
* No Mint/Melt properties for shielded transactions must be active (if shielded is detected, is must raise).

_____________________________________________________

### Thorough Criteria of PR-5 from Original Plan:
For each item: ✅ implemented · ⚠️ implemented in a different form / partially · ❌ not present (with the reason).

---

## `hathor/verification/transaction_verifier.py` — 15/16 as specified, 1 substituted

| Requirement | Status | Location / note |
|---|---|---|
| Extract `_check_token_permissions_and_deposits` (with `is_shielded`) | ✅ Implemented | line 283 |
| Add `is_shielded` param to `_check_token_permissions` | ✅ Implemented | line 418 — `is_shielded: bool = False` (byte-identical signature to cleanUp) |
| `verify_transparent_balance(..., shielded_fee=0)` | ✅ Implemented | line ~210 — signature wraps; `*, shielded_fee: int = 0` |
| `verify_token_rules` | ✅ Implemented | line 393 |
| `verify_shielded_outputs` (basic dispatcher) | ✅ Implemented | line 561 |
| `verify_shielded_outputs_with_storage` | ✅ Implemented | line 569 |
| `verify_commitments_valid` | ✅ Implemented | line 590 |
| `verify_authority_restriction` | ✅ Implemented | line 760 |
| `verify_range_proofs` | ✅ Implemented | line 641 |
| `verify_surjection_proofs` | ✅ Implemented | line 685 (de-augmented — mint-domain block dropped) |
| `verify_shielded_balance` | ✅ Implemented | line 803 (de-augmented — `_fold_mint_melt_entry` block + `nc_block_storage` param dropped) |
| `verify_trivial_commitment_protection` | ✅ Implemented | line 775 |
| `verify_shielded_fee` | ✅ Implemented | line 548 |
| `calculate_shielded_fee` | ✅ Implemented | line 537 |
| `_normalize_token_uid` | ✅ Implemented | line 528 |
| `_get_or_derive_asset_tag` | ✅ Implemented | line 671 |
| **`verify_no_undeclared_mint_melt`** | ⚠️ **Substituted** as `verify_no_mint_melt` (forbid-all) at line 785 | The plan lists this as the renamed-by-mint/melt-extension version. We shipped the **pre-mint/melt forbid-all variant** because mint/melt is a post-plan extension. **Why:** the declared-exception machinery (the part `verify_no_undeclared_mint_melt` adds) depends on `MintHeader`/`MeltHeader`, which don't exist in PR 5. *(Separate concern, flagged earlier: the on-disk body is a verbatim copy from `shielded-outputs-1`, awaiting remediation choice — see "Open caveats" at the bottom.)* |

## `hathor/verification/verification_service.py` — 4/5 explicit, 1 covered indirectly

| Requirement | Status | Location / note |
|---|---|---|
| Early feature-gate in `verify_basic` | ✅ Implemented | line 117 — `if not params.features.shielded_transactions: raise InvalidShieldedOutputError` |
| `_verify_basic_shielded_header` (no storage) | ✅ Implemented | line 156 |
| `_verify_shielded_header` (storage-bound) | ✅ Implemented | line 239 |
| **Reject shielded outputs in `TokenCreationTransaction`** | ❌ **Not at this layer** | No explicit verification-layer reject. **Why:** the parser-level `get_allowed_headers` (`vertex_verifier.py:241`) only admits `ShieldedOutputsHeader` for `TxVersion.REGULAR_TRANSACTION` — so a TCT with shielded outputs is rejected *upstream* at parser admission, not at the dispatch layer the plan asked for. Decision recorded in memory: *"shielded-TCT is mint/melt-coupled (requires `MintHeader` per RFC §4.4) → deferred to mint/melt extension."* That decision conflated the *flat* reject (could ship now) with the *mint/melt-coupled admission* (post-plan). |
| `_verify_tx` balance dispatch split | ✅ Implemented | line 308 — `verify_token_rules` + `verify_shielded_balance` on the shielded branch; `verify_transparent_balance` on the transparent branch |

## `hathor/transaction/transaction.py` — 2/3

| Requirement | Status | Location / note |
|---|---|---|
| `_get_token_info_from_inputs` skip shielded inputs | ✅ Implemented | line 379 — uses `mode()` / `OutputMode` check, wrapped in `try/except IndexError` |
| `_update_token_info_from_outputs` seed `TokenInfo` for output-only tokens | ✅ Implemented | line 425 |
| **`to_json_extended` include `tokens` array** | ❌ **Deferred to PR 7** | The one-line `json_extended['tokens'] = [h.hex() for h in self.tokens]` is missing on step5. **Why:** the fetch checklist (Group 2) records *"DEFER → PR 7: `get_related_addresses` shielded iteration + `to_json_extended` shielded block"* — bundled with PR 7's *"indexes / sync / API / events / DAG-builder"* scope because it's a JSON-API surface change, not a verification rule. |

## `hathor/transaction/token_info.py` — 1/1

| Requirement | Status | Location / note |
|---|---|---|
| `calculate_fee(settings, *, shielded_fee=0)` | ✅ Implemented | line 61  |

## `hathor/consensus/consensus.py` — 2/2

| Requirement | Status | Location / note |
|---|---|---|
| `_shielded_activation_rule` reorg handling for `SHIELDED_TRANSACTIONS` | ✅ Implemented | line 464 (definition); the `case Feature.SHIELDED_TRANSACTIONS:` reorg case at line 463 (call site at line 517). De-augmented — only `has_shielded_outputs()` / `has_unshield_balance_header()` clauses; the `has_mint_header` / `has_melt_header` clauses are dropped (post-plan). |
| Add `FAILED_FEE_TOKENS` / `FAILED_OPCODES_V2` to the no-op feature list | ✅ Implemented | lines 468–469 |

## Tests — 0/6 (all deferred to PR 8)

| File | Status | Why |
|---|---|---|
| `hathor_tests/tx/test_shielded_verification.py` (660 lines) | ❌ Deferred to PR 8 | module-level `import hathor_ct_crypto` |
| `hathor_tests/tx/test_shielded_security.py` (638 lines) | ❌ Deferred to PR 8 | module-level `import hathor_ct_crypto` |
| `hathor_tests/tx/test_shielded_audit_fixes.py` (874 lines) | ❌ Deferred to PR 8 | module-level `import hathor_ct_crypto` |
| `hathor_tests/tx/test_shielded_cons_fixes.py` (1,737 lines) | ❌ Deferred to PR 8 | module-level `import hathor_ct_crypto` |
| `hathor_tests/tx/test_shielded_post_audit_fixes.py` (413 lines) | ❌ Deferred to PR 8 | module-level `import hathor_ct_crypto` |
| `hathor_tests/tx/test_shielded_v3_audit_fixes.py` (139 lines) | ❌ Deferred to PR 8 | No module-level Rust import, but bundled for consistency — body never audited for transitive Rust dependence |
_____________________________________________________

### Summary

  This is **PR 5 of 8** in the staged rollout of confidential ("shielded") transactions — the reviewable split of the squashed work in
  https://github.com/HathorNetwork/hathor-core/blob/feat/shielded-outputs/docs/plans/shielded-pr-split.md
  (confidential transactions, modeled on Liquid Confidential Assets: Pedersen commitments, Borromean range proofs, asset surjection proofs, ECDH rewind).

  > [!IMPORTANT]
  > **This is the security-critical heart of the feature — and it ships disabled.** It adds the *verification semantics* (balance dispatch,
  range/surjection/commitment proof checks, shielded-aware token rules) plus the *Python crypto surface* those checks call into. There is still **no wallet, 
  no indexing/sync/API, and no native crypto** here. The feature ships **disabled by default** and is **non-functional** on its own: the crypto layer
  degrades gracefully (`SHIELDED_CRYPTO_AVAILABLE` is `False` until the native crate lands in PR 8), so the proof checks are dormant and unreachable while
  the flag is off.

  > [!NOTE]
  > **Stacked on #1702 (PR 4).** Until that merges and this branch is rebased, the diff against `master` also shows PR 4's changes. Review/merge after #1702.

  The full sequence:

  | PR | Scope | Status |
  |----|-------|--------|
  | 1 | `chore`: rename `verify_sum → verify_transparent_balance` | ✅ merged |
  | 2 | `refactor`: `resolve_spent_output` plumbing | ✅ merged |
  | 3 | ~~canonical header ordering~~ | retracted (no ordering enforced) |
  | 4 | `feat`: data model + headers + feature flag (gated OFF) |✅ merged |
  | **5** | **`feat`: verification (balance dispatch, range/surjection proofs) — security-critical** | **← this PR** |
  | 6 | `feat`: wallet (ECDH UTXO discovery) | upcoming |
  | 7 | `feat`: indexes, sync, API, events, DAG builder | upcoming |
  | 8 | `feat`: Rust crypto lib + build/CI wiring (**ships last**) | upcoming |

  ### What's in this PR

  **Crypto layer (the Python surface for the native crate)** — `hathor/crypto/shielded/` (9 files): `__init__` (exposes the ops +
  `SHIELDED_CRYPTO_AVAILABLE`), `_bindings.py` / `_bindings.pyi` (the FFI surface to the future `hathor_ct_crypto`), and the operation wrappers `asset_tag`,
  `balance`, `commitment`, `ecdh`, `range_proof`, `surjection`. It **degrades gracefully**: if the native lib is absent (which it is, until PR 8),
  `SHIELDED_CRYPTO_AVAILABLE` is `False` and the wrappers only raise when actually *called* — importing never fails. _Moved here from PR 4 to ship alongside 
  its first consumer (verification)._

  **Verification semantics** — `hathor/verification/transaction_verifier.py` (the bulk of the change):
  - shielded proof verification — Pedersen **commitment validity**, **Borromean range proofs**, and **asset surjection proofs**;
  - **balance dispatch** — `verify_transparent_balance` (renamed in PR 1) for the clear side, `verify_shielded_balance` for the homomorphic balance equation;
  - **token rules / permissions** and **input handling** made shielded-aware.

  **Verification dispatch + header verifiers** — `hathor/verification/verification_service.py`: feature-gated dispatch, plus basic + full verification of the
   shielded headers.

  **Output sigops & script-size** — `hathor/verification/vertex_verifier.py`: shielded outputs are counted toward the sigops and script-size limits.

  **Consensus activation rule** — `hathor/consensus/consensus.py`: the consensus-level activation hook for shielded transactions.

  **Supporting** — vertex-API + token-accounting additions in `hathor/transaction/{base_transaction,transaction,token_info}.py`; verifier wiring in
  `hathor/builder/builder.py`; one nanocontract test adaptation.

  **`hathorlib`: no change** — this PR uses the data model / serialization published with PR 4.

  | Deferred | Lands in |
  |----------|----------|
  | Wallet (ECDH UTXO discovery + decryption) | PR 6 |
  | Indexes, sync, API, events, DAG builder | PR 7 |
  | Native Rust crate (`hathor_ct_crypto`) + build/CI wiring | PR 8 |
  | Proof / balance / round-trip test vectors (need the native lib) | PR 8 |
  | Mint/melt verification (`0x14`/`0x15`) | post-plan extension (marked with `NOTE` comments) |

  ### 🔒 Safety: fully gated by `ENABLE_SHIELDED_TRANSACTIONS` (default `DISABLED`)

  With the flag disabled (the default — and the only state this PR ships), **no shielded verification is reachable**:

  - **The PR-4 parser gate still holds.** A flag-off node cannot even *deserialize* a transaction carrying header `0x12`/`0x13`, so the new verification
  paths are never entered.
  - **Defense-in-depth in dispatch.** `verification_service` routes to the shielded verifiers only when `features.shielded_transactions` is true — `False`
  when the flag is off.
  - **The crypto layer is dormant.** `SHIELDED_CRYPTO_AVAILABLE` is `False` (no native lib until PR 8); every proof wrapper raises if called without it, so
  nothing can *silently* pass verification.
  - **No production caller reaches the shielded checks** with the flag off — no shielded data can exist.

  Net effect: a flag-off node behaves exactly as before this PR.

  ### Testing

  - `make check` (custom checks + yamllint + ruff + mypy) passes.
  - Existing suite is green; the shielded **gating tests** assert the flag-off invariant (a flag-off node rejects `0x12`/`0x13`).
  - The proof / balance / round-trip tests that exercise *real* crypto require the native crate and therefore land alongside it in PR 8 (or behind a
  skip-guard once it's a dev dependency).


hathor/verification/transaction_verifier.py
```diff
@@ -46,6 +52,7 @@
     TooManySigOps,
     TooManyTokens,
     TooManyWithinConflicts,
+    TrivialCommitmentError,
     UnusedTokensError,
```
LFRezende, line 51: Is `TrivialCommitmentError` to the balance of commitments or to a particular one?

hathor/verification/transaction_verifier.py
```diff
@@ -150,7 +157,19 @@ def _verify_inputs(
...
+                # Standard transparent output
+                pass
+            elif spent_tx.shielded_outputs:
+                shielded_idx = input_tx.index - len(spent_tx.outputs)
+                if shielded_idx >= len(spent_tx.shielded_outputs):
+                    raise InexistentInput('Output spent by this input does not exist: {} index {}'.format(
+                        input_tx.tx_id.hex(), input_tx.index))
+            else:
+                raise InexistentInput('Output spent by this input does not exist: {} index {}'.format(
+                    input_tx.tx_id.hex(), input_tx.index))
```
LFRezende, line 170: If not transparent, then see if shielded. If above length and not shielded, error.

hathor/verification/transaction_verifier.py
```diff
@@ -294,26 +310,58 @@ def verify_transparent_balance(
...
                 case TokenVersion.DEPOSIT:
-                    if token_info.has_been_melted():
-                        withdraw += get_deposit_token_withdraw_amount(settings, token_info.amount)
-                    if token_info.has_been_minted():
-                        deposit += get_deposit_token_deposit_amount(settings, token_info.amount)
+                    if not is_shielded:
+                        if token_info.has_been_melted():
+                            withdraw += get_deposit_token_withdraw_amount(settings, token_info.amount)
+                        if token_info.has_been_minted():
+                            deposit += get_deposit_token_deposit_amount(settings, token_info.amount)
```
LFRezende, line 315: Gated OFF by is shielded - OK.

hathor/verification/transaction_verifier.py
```diff
@@ -294,26 +310,58 @@ def verify_transparent_balance(
...
                     continue
 
                 case _:
                     assert_never(token_info.version)
 
-        # check whether the deposit/withdraw amount is correct
         htr_expected_amount = withdraw - deposit
         htr_info = token_dict[settings.HATHOR_TOKEN_UID]
-        if htr_info.amount > htr_expected_amount:
+        if not is_shielded and htr_info.amount > htr_expected_amount:
```
LFRezende, line 325: If shielded --> amount unbalances due to cryptographic hiddance.

hathor/verification/transaction_verifier.py
```diff
@@ -335,16 +384,58 @@ def verify_transparent_balance(
...
             assert token_uid == HATHOR_TOKEN_UID
             assert not token_info.can_mint
             assert not token_info.can_melt
             return
         assert token_uid != HATHOR_TOKEN_UID
+        if is_shielded:
+            # In shielded txs, transparent deficit/surplus is expected — balance is
+            # verified cryptographically. Authority-based mint/melt is separately
+            # blocked by verify_no_mint_melt.
+            return
```
LFRezende, line 440: [Note]: In a fully shielded transaction, this function should not be called, as there is no way to know the token uid of a blinded asset tag without either being the sender or the receiver after a rewind. 

Only in amount-hidden transactions this makes sense.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+    def calculate_shielded_fee(settings: HathorSettings, tx: Transaction) -> int:
+        """Calculate the total fee required for shielded outputs."""
+        from hathorlib.transaction.shielded_tx_output import AmountShieldedOutput, FullShieldedOutput
+        fee = 0
+        for output in tx.shielded_outputs:
+            if isinstance(output, AmountShieldedOutput):
+                fee += settings.FEE_PER_AMOUNT_SHIELDED_OUTPUT
+            elif isinstance(output, FullShieldedOutput):
+                fee += settings.FEE_PER_FULL_SHIELDED_OUTPUT
+        return fee
```
LFRezende: [Note]: We may mix hence both amount and fully shielded outputs in a single shielded transaction. In other words, we tag the shielded level at output scope, not transaction scope.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+
+    def verify_shielded_fee(self, tx: Transaction) -> None:
+        """Verify the transaction declares sufficient fees for its shielded outputs."""
+        if not tx.has_fees():
+            raise InvalidShieldedOutputError('shielded transactions require a fee header')
```
LFRezende: [Note]: All Shielded Outputs require a FeeHeader, and it makes sense - there is no shielded transaction with null fee.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+                f'minimum shielded fee is {expected_shielded_fee}'
+            )
+
+    def verify_shielded_outputs(self, tx: Transaction) -> None:
+        """Top-level: calls all basic shielded checks."""
+        self.verify_commitments_valid(tx)
+        self.verify_authority_restriction(tx)
+        self.verify_range_proofs(tx)
+        self.verify_trivial_commitment_protection(tx)
+        self.verify_shielded_fee(tx)
```
LFRezende: [Note]: Checks done -> If commitment is 33 bytes, authority on mint/melt (for the future), range proof validity, protection on the commitment and if there is a valid fee implemented.

LFRezende: Does not check Asset Surjection --> SHOULD NOT be called whenever checking a FullShielded Transaction

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+
+        spent_txs: dict[bytes, BaseTransaction] = {}
+        for tx_input in tx.inputs:
+            if tx_input.tx_id not in spent_txs:
+                spent_txs[tx_input.tx_id] = tx.storage.get_transaction(tx_input.tx_id)
+
+        asset_tag_cache: dict[bytes, bytes] = {}
+
+        self.verify_surjection_proofs(tx, spent_txs=spent_txs, asset_tag_cache=asset_tag_cache)
+
```
LFRezende: Why wrap `self.verify_surjection_proofs` like this? To get the dict on spent_txs? It could be done inside the function by only passing tx.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+        for i, output in enumerate(tx.shielded_outputs):
+            if len(output.commitment) != COMMITMENT_SIZE:
+                raise InvalidShieldedOutputError(
+                    f'shielded output {i}: commitment must be {COMMITMENT_SIZE} bytes, '
+                    f'got {len(output.commitment)}'
+                )
+            if not validate_commitment(output.commitment):
+                raise InvalidShieldedOutputError(
+                    f'shielded output {i}: invalid commitment (not a valid curve point)'
+                )
```
LFRezende: [Note]: Checks if the point is valid in the `secp256k1` curve -> good.

`C == v*H + r*G`, hence either C was random/noisily built and/or the token tag (hash) was not generated by a hash-to-curve function.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+            if isinstance(output, FullShieldedOutput):
+                if len(output.asset_commitment) != ASSET_COMMITMENT_SIZE:
+                    raise InvalidShieldedOutputError(
+                        f'shielded output {i}: asset_commitment must be {ASSET_COMMITMENT_SIZE} bytes, '
+                        f'got {len(output.asset_commitment)}'
+                    )
+                if not validate_generator(output.asset_commitment):
+                    raise InvalidShieldedOutputError(
+                        f'shielded output {i}: invalid asset_commitment (not a valid curve point)'
+                    )
```
LFRezende: [Note]: Same comment as before.

`A == H + r*G`, hence either A is random or H is not generated by a Hash-to-Curve.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+                )
+            if isinstance(output, FullShieldedOutput):
+                if len(output.asset_commitment) != ASSET_COMMITMENT_SIZE:
+                    raise InvalidShieldedOutputError(
+                        f'shielded output {i}: asset_commitment must be {ASSET_COMMITMENT_SIZE} bytes, '
+                        f'got {len(output.asset_commitment)}'
+                    )
+                if not validate_generator(output.asset_commitment):
+                    raise InvalidShieldedOutputError(
+                        f'shielded output {i}: invalid asset_commitment (not a valid curve point)'
```
LFRezende: Both `Pedersen Commitments` and Blinded Asset Tags are generators points on the secp256k1 curve. Why use different functions to check their validity and yet the same error message with the same argument for invalidity?

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+                    )
+                try:
+                    get_public_key_from_bytes_compressed(output.ephemeral_pubkey)
+                except (ValueError, TypeError):
+                    raise InvalidShieldedOutputError(
+                        f'shielded output {i}: invalid ephemeral_pubkey (not a valid secp256k1 point)'
+                    )
+
+    def verify_range_proofs(self, tx: Transaction) -> None:
+        """Every shielded output must have valid Bulletproof range proof."""
```
LFRezende: **We are NOT using Bulletproof range proofs.** Our current rust crypto stack is based on secp256k1-zkp package used by Liquid, which leverages Borromean range proofs. Our goal is to thread towards Bulletproofs, yes, but not now. 

All mentions on range proofs in this PR need to change the implementation description, from Bulletproof to Borromean.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+                raise InvalidShieldedOutputError(f'shielded output {i}: unknown type')
+
+            try:
+                if not verify_range_proof(output.range_proof, output.commitment, generator):
+                    raise InvalidRangeProofError(
+                        f'shielded output {i}: range proof verification failed'
+                    )
+            except ValueError as e:
+                raise InvalidRangeProofError(f'shielded output {i}: {e}') from e
+
```
LFRezende: Does `tx.shielded_outputs` also contemplate **the shielded INPUTS** of the transaction? 

**[IMPORTANT QUESTION + NOTE TO SELF]**
Do we need to check the range proofs of all shielded UTXOs and STXOs within a transaction, or just the shielded outputs for each transaction and with the pedersen commitments we lock the values safe, reducing the proof size requirements?

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+    def verify_range_proofs(self, tx: Transaction) -> None:
+        """Every shielded output must have valid Bulletproof range proof."""
+        from hathor.crypto.shielded import verify_range_proof
+        from hathorlib.transaction.shielded_tx_output import AmountShieldedOutput, FullShieldedOutput
+
+        asset_tag_cache: dict[bytes, bytes] = {}
+        for i, output in enumerate(tx.shielded_outputs):
+            if isinstance(output, AmountShieldedOutput):
+                token_index = output.token_data & 0x7F
```
LFRezende: [NOTE]: Does the check for range proof for all outputs in the tx.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+    ) -> bytes:
+        """Derive an asset tag, using the cache if available."""
+        from hathor.crypto.shielded import derive_asset_tag
+        normalized = self._normalize_token_uid(token_uid)
+        if asset_tag_cache is not None:
+            if normalized not in asset_tag_cache:
+                asset_tag_cache[normalized] = derive_asset_tag(normalized)
+            return asset_tag_cache[normalized]
+        return derive_asset_tag(normalized)
+
```
LFRezende: [NOTE]: Look up later the `derive_asset_tag ` implementation and how it differs from simply normalizing the token uid.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+        transparent and shielded inputs; each FullShieldedOutput must prove its
+        asset is one of them.
+        """
+        from hathor.crypto.shielded import verify_surjection_proof
+        from hathorlib.transaction.shielded_tx_output import AmountShieldedOutput, FullShieldedOutput
+
+        assert tx.storage is not None
+        domain_generators: list[bytes] = []
+        for tx_input in tx.inputs:
+            spent_tx = spent_txs[tx_input.tx_id] if spent_txs else tx.storage.get_transaction(tx_input.tx_id)
```
LFRezende: Why pass as an argument `spent_txs` in the first place, if we can get directly the id from get_transaction? Is it just to pass the dict?

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+            spent_index = tx_input.index
+            if spent_index < len(spent_tx.outputs):
+                spent_output = spent_tx.outputs[spent_index]
+                if spent_output.is_token_authority():
+                    continue
+                token_uid = spent_tx.get_token_uid(spent_output.get_token_index())
+                domain_generators.append(self._get_or_derive_asset_tag(token_uid, asset_tag_cache))
```
LFRezende: [NOTE]: Transparent input.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+                elif isinstance(shielded_out, AmountShieldedOutput):
+                    token_index = shielded_out.token_data & 0x7F
+                    try:
+                        token_uid = spent_tx.get_token_uid(token_index)
+                    except (IndexError, NotImplementedError) as e:
+                        raise InvalidShieldedOutputError(
+                            f'spent shielded output token_data index {token_index} '
+                            f'is invalid for the spent transaction'
+                        ) from e
+                    domain_generators.append(self._get_or_derive_asset_tag(token_uid, asset_tag_cache))
```
LFRezende: [NOTE]: Shielded Input. Get (like the transparent) the generator of it so to put in the domain for later verification

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+
+        has_full_shielded = any(isinstance(o, FullShieldedOutput) for o in tx.shielded_outputs)
+        if has_full_shielded and not domain_generators:
+            raise InvalidSurjectionProofError(
+                'FullShieldedOutput requires at least one input to form a surjection proof domain'
+            )
+
+        for i, output in enumerate(tx.shielded_outputs):
+            if isinstance(output, FullShieldedOutput):
```
LFRezende: [Note]: Now, check the outputs (codomain check).

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+                        f'shielded output {i}: FullShieldedOutput requires surjection proof'
+                    )
+                try:
+                    if not verify_surjection_proof(
+                        output.surjection_proof,
+                        output.asset_commitment,
+                        domain_generators,
+                    ):
+                        raise InvalidSurjectionProofError(
+                            f'shielded output {i}: surjection proof verification failed'
```
LFRezende: [NOTE]: For each shielded output, check the proof for the domain generators. Need to check `verify_surjection_proof` and the implementations.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+    def verify_no_mint_melt(self, token_dict: TokenInfoDict) -> None:
+        """Reject mint/melt operations in transactions with shielded outputs."""
+        for token_uid, token_info in token_dict.items():
+            if token_info.version == TokenVersion.NATIVE:
+                continue
+            if token_info.can_mint and token_info.has_been_minted():
+                raise ShieldedMintMeltForbiddenError(
+                    f'token {token_uid.hex()}: minting is not allowed in transactions '
+                    f'with shielded outputs (transparent surplus: {token_info.amount})'
+                )
```
LFRezende: This function just assumes the tx to be shielded from the get-go.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+            if token_info.can_melt and token_info.has_been_melted():
+                raise ShieldedMintMeltForbiddenError(
+                    f'token {token_uid.hex()}: melting is not allowed in transactions '
+                    f'with shielded outputs (transparent deficit: {token_info.amount})'
+                )
+
+    def verify_shielded_balance(
```
LFRezende: [Note]: Pedersen commitment verification.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+        assert tx.storage is not None
+        transparent_inputs: list[tuple[int, bytes]] = []
+        shielded_inputs: list[bytes] = []
+
+        for tx_input in tx.inputs:
+            spent_tx = spent_txs[tx_input.tx_id] if spent_txs else tx.storage.get_transaction(tx_input.tx_id)
+            spent_index = tx_input.index
+            if spent_index < len(spent_tx.outputs):
+                spent_output = spent_tx.outputs[spent_index]
+                if not spent_output.is_token_authority():
```
LFRezende: This structure keeps repeating. We could add a helper function to sort this out.

hathor/verification/transaction_verifier.py
```diff
@@ -425,3 +516,372 @@ def verify_conflict(self, tx: Transaction, params: VerificationParams) -> None:
...
+            )
+        if has_shielded_inputs_ and not has_shielded_outputs_ and not has_excess:
+            raise ShieldedBalanceMismatchError(
+                'a full-unshield tx (shielded inputs, no shielded outputs) must carry an '
+                'unshield balance header'
+            )
+        if has_excess and not has_shielded_inputs_:
+            raise ShieldedBalanceMismatchError(
+                'unshield balance header requires at least one shielded input'
+            )
```
LFRezende: [NOtE]: Logic table checked, seems alright.

hathor/crypto/shielded/_bindings.py
```diff
@@ -0,0 +1,15 @@
...
+
+_lib: Any = None
+AVAILABLE: bool = False
+
+try:
+    import hathor_ct_crypto
+    _lib = hathor_ct_crypto
+    AVAILABLE = True
+except ImportError:
+    pass
```
glevco: We can remove this from this PR, right?

hathor/verification/verification_service.py
```diff
@@ -139,6 +148,22 @@ def verify_basic(
             assert self._settings.ENABLE_NANO_CONTRACTS
             # nothing to do
 
+        if vertex.has_shielded_outputs():
+            # Feature gate already checked above (before match-case dispatch).
+            assert isinstance(vertex, Transaction)
+            self._verify_basic_shielded_header(vertex)
+
```
LFRezende, line 155: Why not join this with the previous `if vertex.has_shielded_outputs():` clause?

hathor/verification/transaction_verifier.py
```diff
@@ -335,16 +389,60 @@ def verify_transparent_balance(
...
 
+    @classmethod
+    def verify_token_rules(
+        cls,
+        settings: HathorSettings,
+        token_dict: TokenInfoDict,
+        *,
+        shielded_fee: int = 0,
+    ) -> None:
+        """Verify token authority permissions, deposit requirements, and fee correctness.
```
LFRezende, line 395: This function should not be called by a FullyShielded transaction. It requires information on the token. Check if this is called by any function where the context is possibly attachable to a fully shielded transaction, of if I am not correct in my assumption.

hathor/verification/transaction_verifier.py
```diff
@@ -256,36 +280,33 @@ def verify_output_token_indexes(self, tx: Transaction) -> None:
...
         settings: HathorSettings,
-        tx: Transaction,
         token_dict: TokenInfoDict,
+        *,
         allow_nonexistent_tokens: bool = False,
-    ) -> None:
-        """Verify that the sum of outputs is equal of the sum of inputs, for each token. If sum of inputs
-        and outputs is not 0, make sure inputs have mint/melt authority.
+        is_shielded: bool = False,
+    ) -> tuple[int, bool]:
```
LFRezende, line 283: Again: Does it make sense to let this function be solely open for shielded transactions? Shouldn't it have a safeguard for AmountHidden only, since only this kind of shielded transactions would be with its token cleanly visible?

hathor/verification/transaction_verifier.py
LFRezende: `verify_commitments_valid`: check if 33 bytes, valid curve points, count is within limits ok

`verify_authority_restriction` : shielded outputs CANNOT be authority. Checks if AmountOnly the token data field, and if FullShielded has no token data to hold authority mask - ok

`verify_range_proofs`: For each output, check the range proof.  **NOTE: Why pass the asset commitment (asset surjection) to verify the range proof? - because it needs to assess bit by bit the pedersen commitment, and it needs the token tag. in fully shielded, the token tag generator is the blinded asset tag**

`verify_trivial_commitment_protection`: Checks at least two shielded outputs.


DOES NOT CHECK: 
1. ASSET SURJECTION 
2. BALANCE ON PEDERSEN COMMITMENTS

hathor/verification/vertex_verifier.py
```diff
@@ -163,6 +163,13 @@ def verify_outputs(self, vertex: BaseTransaction) -> None:
                     len(output.script), self._settings.MAX_OUTPUT_SCRIPT_SIZE
                 ))
 
+        if hasattr(vertex, 'shielded_outputs'):
+            for shielded_output in vertex.shielded_outputs:
+                if len(shielded_output.script) > self._settings.MAX_OUTPUT_SCRIPT_SIZE:
+                    raise InvalidOutputScriptSize('shielded output script size: {} and max-size: {}'.format(
+                        len(shielded_output.script), self._settings.MAX_OUTPUT_SCRIPT_SIZE
+                    ))
+
```
LFRezende: Jan mentioned previously to avoid `hasattr` declarations.

hathor/verification/transaction_verifier.py
LFRezende: **Answer: `_check_token_permissions_and_deposits` receives `token_dict`, which is built such that only transparent inputs are present, so token and amount are visible. It skips `tx.shielded_outputs`**.

hathor/verification/transaction_verifier.py
LFRezende: **Answer: `_check_token_permissions_and_deposits` receives `token_dict`, which is built such that only transparent inputs are present, so token and amount are visible. It skips `tx.shielded_outputs`**.

hathor/verification/vertex_verifier.py
LFRezende: Answer: We'll change the check, but keep the safeguard.

hathor/verification/verification_service.py
LFRezende: Indeed it can be done and make it simpler - but it is not a blocker. I'll keep it for this merge, but in future work this could be done.

hathor/verification/transaction_verifier.py
LFRezende: While useful, not needed. I think this could break consensus.

hathor/verification/transaction_verifier.py
LFRezende: Yes, it does. Token_dict comes solely from transparent inputs, but if even if just one of them can mint/melt, we block it - for we do not allow shielded transactions to mint/melt for now.

hathor/verification/transaction_verifier.py
LFRezende: This check will need to be delegated for PR 8, when the Rust lib is implemented (low-level implementation).

hathor/verification/transaction_verifier.py
LFRezende: Indeed this is correct - while I get the idea of cache and memoing for better performance, the spent_txs dict created here is not the same consumed by other functions, so we are just creating more work.

This is consensus-sensitive and not a blocker, so I'll resolve for now.

hathor/verification/transaction_verifier.py
LFRezende: Two different functions: deriviing the asset tag is looking at the curve and getting a generator for the normalized token uid.

hathor/verification/transaction_verifier.py
LFRezende: Yes, just outputs. Inputs are from spent outputs, and these have either been checked previously if shielded or are transparent.

hathor/verification/transaction_verifier.py
LFRezende: No, just outputs. Inputs need not verify. They have already been so previously, and the balance check in pedersen commitments already does the job.

hathor/verification/transaction_verifier.py
LFRezende: Fixed all mentions of bulletproof in the comments of the diff.

hathor/verification/transaction_verifier.py
LFRezende: It will need to await for the implementaiton at rust-crate level in PR 8 for this to be clear.

hathor/verification/transaction_verifier.py
LFRezende: Done for memoing, although it is not doing it correctly here - `spent_txs` are different  caches for each function who does so, it is not being shared between functions who consume it.

Hence, it could be revamped, yet it is consensus-sensitive and not critical - just an optimization. I'll leave it like that for now.

hathor/verification/transaction_verifier.py
LFRezende: Asset and Balance verifications done in other functions.

hathor/verification/transaction_verifier.py
LFRezende: I retract this: the is_shielded flag skips this for both.

hathor/verification/transaction_verifier.py
LFRezende: No, it's to check if there is at least two shielded outputs.

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.4.0 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.3.2 and should not be changed by hand.
```
glevco: Please use a Poetry version >= the current one (2.4.0)

hathor/crypto/shielded/balance.py
```diff
@@ -0,0 +1,40 @@
...
+        shielded_outputs: List of 33B commitment bytes for each shielded output.
+        excess_blinding_factor: Optional 32B scalar used on full-unshield txs to
+            carry `sum(r_in) - sum(r_out)` to the output side. Mutually exclusive
+            with `shielded_outputs`: must be None whenever there is at least one
+            shielded output, and must be set whenever there are shielded inputs
+            but no shielded outputs.
+    """
+    if excess_blinding_factor is not None:
+        if shielded_outputs:
+            raise ValueError('excess_blinding_factor must be None when shielded_outputs is non-empty')
```
glevco: The function's signature doesn't accept `shielded_outputs` as `None`

hathor/crypto/shielded/balance.py
```diff
@@ -0,0 +1,40 @@
...
+        if not shielded_inputs:
+            raise ValueError('excess_blinding_factor requires at least one shielded input')
+        if len(excess_blinding_factor) != 32:
+            raise ValueError('excess_blinding_factor must be 32 bytes')
```
glevco: Why not move those checks do the rust side and drop the wrapper entirely? I mean, they would be invalid calls on the rust side, too. Also, the len == 32 check can be enforced statically in rust.

hathor/crypto/shielded/asset_tag.py
```diff
@@ -0,0 +1,18 @@
...
+def normalize_token_uid(token_uid: bytes) -> bytes:
+    """Normalize a token UID to 32 bytes for the crypto library.
+
+    Hathor uses b'\\x00' (1 byte) for HTR and 32-byte hashes for custom tokens.
+    The crypto library always expects 32-byte token UIDs.
+    """
+    if len(token_uid) == _CRYPTO_TOKEN_UID_SIZE:
+        return token_uid
+    if len(token_uid) == 1:
+        return token_uid.ljust(_CRYPTO_TOKEN_UID_SIZE, b'\x00')
```
glevco:
```suggestion
    if token_uid == HTR_TOKEN_UID:
        token_uid = b'\x00' * _CRYPTO_TOKEN_UID_SIZE
    assert len(token_uid) == _CRYPTO_TOKEN_UID_SIZE
    return token_uid
```

glevco (changes requested):

hathor/transaction/base_transaction.py
```diff
@@ -1021,6 +1026,12 @@ def get_token_index(self) -> int:
         """The token uid index in the list"""
         return self.token_data & self.TOKEN_INDEX_MASK
 
+    @staticmethod
+    def mode() -> OutputMode:
+        """Return the output mode (TRANSPARENT for standard TxOutput)."""
+        from hathorlib.transaction.shielded_tx_output import OutputMode as _OutputMode
+        return _OutputMode.TRANSPARENT
```
glevco, line 1033: Why static? Doesn't it depend on the transaction?

hathor/verification/verification_service.py
```diff
@@ -139,6 +148,22 @@ def verify_basic(
...
+    def _verify_basic_shielded_header(self, tx: Transaction) -> None:
+        """Shielded verifications that don't need storage."""
+        from hathor.transaction.exceptions import InvalidShieldedOutputError, TxValidationError
+        try:
+            self.verifiers.tx.verify_shielded_outputs(tx)
+        except TxValidationError:
+            self.verifiers.tx.log.debug('shielded basic verification failed', tx=tx.hash_hex)
+            raise
+        except RuntimeError as e:
+            raise InvalidShieldedOutputError(f'shielded crypto library error: {e}') from e
```
glevco: Why not just raise the correct exception from the lib? I don't see how these exception conversions are useful.

glevco: Why log on TxValidationError but not on RuntimeError?

hathor/verification/verification_service.py
```diff
@@ -206,6 +231,22 @@ def verify(self, vertex: BaseTransaction, params: VerificationParams) -> None:
             self.verifiers.nano_header.verify_method_call(vertex, params)
             self.verifiers.nano_header.verify_seqnum(vertex, params)
 
+        if vertex.has_shielded_outputs():
+            # Feature gate is already enforced in verify_basic, which is always called first.
```
glevco: Add an `assert params.features.shielded_transactions` here instead.

hathor/verification/verification_service.py
```diff
@@ -206,6 +231,22 @@ def verify(self, vertex: BaseTransaction, params: VerificationParams) -> None:
...
+    def _verify_shielded_header(self, tx: Transaction, params: VerificationParams) -> None:
+        """Shielded verifications that need storage (balance, surjection)."""
+        from hathor.transaction.exceptions import InvalidShieldedOutputError, TxValidationError
+        try:
+            self.verifiers.tx.verify_shielded_outputs_with_storage(tx)
+        except TxValidationError:
+            self.verifiers.tx.log.debug('shielded full verification failed', tx=tx.hash_hex)
+            raise
+        except RuntimeError as e:
+            raise InvalidShieldedOutputError(f'shielded crypto library error: {e}') from e
```
glevco: Same here about exceptions.

hathor/verification/transaction_verifier.py
glevco, line 1: The nano header has its own verifier, so I think we should move the shielded-only verifications to a new shielded headers verifier too.

glevco (changes requested):

hathor/verification/transaction_verifier.py
LFRezende: I think this level of refactor and creation of a new class would need to be better discussed with Marcelo - it is a significant refactor, and future PRs (not only from the plan but also other ones like #1670 do not work with it like this). Could be done in its separate PR perhaps though, in the future.

hathor/crypto/shielded/_bindings.py
LFRezende: I disagree - by leaving the rust crate empty the try/except fails and keeps the AVAILABLE = False, gating shielded transactions off. 

For now, it works, but in future PRS maybe we could delete this, or revamp it.

hathor/crypto/shielded/asset_tag.py
LFRezende: Better - I'll push with it.

hathor/crypto/shielded/balance.py
LFRezende: Indeed - I'll allow the input to have typing "... | None" as well, as the blinding factor being in excess means not to be allowed any shielded output.

hathor/crypto/shielded/balance.py
LFRezende: I'm leaving all the Rust work to PR 8, where indeed many of these checks are done as well. 

The Python wrappers duplicate the work, yes (which may be good in a sense, although unnecesary) but until PR 8 lands, we need to see it being implemented first before we apply it. 


Deleting them prematurelly now seems risky to me, perhaps deleting them if we see them as not only unnecessary by duplicated AFTER we implement their counterparts in PR 8 seems better for me.

hathor/transaction/base_transaction.py
LFRezende: `TxOutput` is always a transparent class, so returning TRANSPARENT is a constant.  Even if the transaction has a ShieldedHeader, FeeHeader and many other outputs with Amount/FullShielded modes, the TxOutput instance is used solely for the transparent output present in the Tx, so the mode it returns is constant.

hathor/verification/verification_service.py
LFRezende: I'm minimizing Rust alterations now - this is PR 8 work. I created the Rust crate only as an empty stub for the future.

LFRezende: Fair enough, I'll change it.

hathor/verification/transaction_verifier.py
LFRezende: I think it is best to leave this to a future PR.

hathor/crypto/shielded/_bindings.py
glevco: Per slack chat, this is not doing anything since the lib is defined.

hathor/crypto/shielded/balance.py
glevco: Removed

hathor/crypto/shielded/balance.py
glevco: Removed

hathor/verification/verification_service.py
glevco: Thanks

hathor/verification/verification_service.py
glevco: Thanks

hathor/verification/verification_service.py
glevco: Thanks

hathor/verification/transaction_verifier.py
glevco: Done

hathor/crypto/shielded/_bindings.py
```diff
@@ -0,0 +1,15 @@
+"""Import the native Rust module with graceful fallback."""
```
jansegre: What does "graceful fallback" mean in this context? Normally without further context if I read that I'll assume that there is a Python implementation and a Rust implementation, and this module tries to import the Rust implementation and when that is not available it "gracefully fallsback" to the Python implementation. But clearly this is not happening as this module doesn't seem to expose anything usable other than `AVAILABLE: bool`.

hathor/crypto/shielded/_bindings.py
jansegre: I haven't checked the Slack conversation, but I'm leaning towards removing this. What do we gain with the `AVAILABLE: bool` approach? I think I'd rather have `hathor_ct_crypto` be a normal "hard" dependency and if it's not available the node doesn't even start. Does that make sense?

hathor/verification/transaction_verifier.py
```diff
@@ -335,16 +382,62 @@ def verify_transparent_balance(
...
             assert token_uid == HATHOR_TOKEN_UID
             assert not token_info.can_mint
             assert not token_info.can_melt
             return
         assert token_uid != HATHOR_TOKEN_UID
+        if is_shielded:
+            # In shielded txs, transparent deficit/surplus is expected — balance is
+            # verified cryptographically. Authority-based mint/melt is separately
+            # blocked by verify_no_mint_melt.
+            # NOTE (mint/melt — postponed): the extension swaps this backstop to
```
jansegre: I'd say we should mark these comments with `TODO` instead of `NOTE`

LFRezende: Done

LFRezende: Also deleted the bindings file.

hathor/crypto/shielded/_bindings.py
LFRezende: Delete the file - the fallback is problematic, it leads to a soft error. It is best to leave the file as a hard dep and let the node break.

hathor/crypto/shielded/_bindings.py
LFRezende: Just deleted the file.

glevco: Thanks!

hathor/verification/transaction_verifier.py
```diff
@@ -256,36 +273,33 @@ def verify_output_token_indexes(self, tx: Transaction) -> None:
...
         withdraw = 0
         has_nonexistent_tokens = False
 
         for token_uid, token_info in token_dict.items():
-            cls._check_token_permissions(token_uid, token_info)
+            cls._check_token_permissions(token_uid, token_info, is_shielded=is_shielded)
             match token_info.version:
                 case None:
-                    # When a token is not found, we can't assert the HTR value since we don't know the token version.
-                    # This is only possible for nanos, because they may create the missing token in execution-time.
```
glevco, line 288: Why did you remove this comment?

hathor/verification/transaction_verifier.py
```diff
@@ -335,16 +382,62 @@ def verify_transparent_balance(
 
         assert htr_info.amount == htr_expected_amount
 
+    # NOTE: this and the sibling _check_token_permissions* / verify_transparent_balance stay here —
+    # they operate on `token_dict`, a transparent-flow concept shared by both paths.
```
glevco, line 386: Seems like a development-time comment that should be removed.

glevco (changes requested):

hathor/verification/transaction_verifier.py
```diff
@@ -294,26 +308,58 @@ def verify_transparent_balance(
...
+
+    @classmethod
+    def verify_transparent_balance(
+        cls,
+        settings: HathorSettings,
+        tx: Transaction,
+        token_dict: TokenInfoDict,
+        allow_nonexistent_tokens: bool = False,
+        *,
+        shielded_fee: int = 0,
```
glevco, line 341: It looks like this param is never used. Is this correct?

hathor/verification/transaction_verifier.py
```diff
@@ -256,36 +273,33 @@ def verify_output_token_indexes(self, tx: Transaction) -> None:
                 raise InvalidToken('token uid index not available: index {}'.format(output.get_token_index()))
 
     @classmethod
-    def verify_transparent_balance(
+    def _check_token_permissions_and_deposits(
```
glevco, line 276: It seems this method only does one check when `is_shielded=True`: the `if not allow_nonexistent_tokens`.

Then, I guess it would be better to move this check outside this method, and the shielded path wouldn't even need to call it.

hathor/verification/transaction_verifier.py
glevco, line 1: I'm not really confident about the changes in this file, they're too convoluted and touch very important checks. It's hard to follow the paths and make sure there are no uncovered edge cases. We have normal txs, nano txs, shielded txs, and shielded nano txs, all sharing the same path with conditionals for each case. Then we'll have the new decimals places project touching the same code.

Maybe this warrants a refactor to make things more streamlined. I imagine that, instead of a sequence of method calls that encode the required checks, we need to make a "rule" a first-class entity. Then, verification can be a list of rules. Then we can easily define different rule sets for each combination of tx (normal x nano x shielded), and it will be easy to spot what skips what. This should be explored as a separate PR pointing to master and merged before this one.

glevco (changes requested):

hathor/verification/transaction_verifier.py
glevco: Initially we can do this only for the balance verification, to unblock this PR. In the future, we may do this for the whole verification process.

hathor/verification/shielded_header_verifier.py
```diff
@@ -0,0 +1,423 @@
...
+          sum(C_in) == sum(C_out) + fee*H_HTR
+        For full-unshield txs (shielded inputs, no shielded outputs): the tx
+        carries an UnshieldBalanceHeader with `excess = sum(r_in) − sum(r_out)`
+        and verification checks:
+          sum(C_in) == sum(C_out) + excess*G + fee*H_HTR
+
+        Mutual-exclusion invariant: a shielded tx must carry either a
+        ShieldedOutputsHeader or an UnshieldBalanceHeader, not both and not
+        neither.
+        """
```
LFRezende, line 340: Blocks `ShieldedOutputsHeader ` and `UnshieldBalanceHeader` simultaneously.

hathor/verification/shielded_header_verifier.py
```diff
@@ -0,0 +1,423 @@
...
+        # whether the header happens to carry a non-empty commitment list, so
+        # a malformed empty-list header can't evade the check.
+        has_shielded_outputs_ = tx.has_shielded_outputs()
+        has_shielded_inputs_ = bool(shielded_inputs)
+        excess_bf = tx.excess_blinding_factor
+        has_excess = excess_bf is not None
+        if has_shielded_outputs_ and has_excess:
+            raise ShieldedBalanceMismatchError(
+                'a shielded tx cannot carry both shielded outputs and an unshield balance header'
+            )
```
glevco, line 400: We should make this check by simply checking the existence of both headers in the beginning of the verification, and then this would just become an assertion.

glevco (changes requested):

PR #1711, opened 2026-05-27
refactor(decimal): integer math in deposit calculations [part 6]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1708

### Motivation

HTR deposit/withdraw calculations for deposit-based tokens currently use float math to calculate percentages. This is problematic for integers with a high amount of decimal places, because float math starts losing precision for the percentage multiplication.

This PR refactors those calculations so they use integer math instead, with no change in behavior in relation to current code. The new representation of the percentage uses ppb, parts per billion, so we're able to represent the `10^-7` value used in the Ekvilibro network.

### Acceptance Criteria

- Refactor `HathorSettings.TOKEN_DEPOSIT_PERCENTAGE` to `HathorSettings.TOKEN_DEPOSIT_PERCENTAGE_NUMERATOR` and `TOKEN_DEPOSIT_PERCENTAGE_DENOMINATOR`, representing the percentage in parts per billion instead of a float.
- Refactor `get_deposit_token_deposit_amount` and `get_deposit_token_withdraw_amount` to use integer math, that is, ceil div and floor div, instead of multiplication by a float.
- Update `/version` API to include new fields `token_deposit_percentage_numerator` and `token_deposit_percentage_denominator`, and deprecate `token_deposit_percentage`.
- Remove unused `get_deposit_token_amount_from_htr`.
- No change in behavior.

### Checklist

### DIFF
--- a/hathor/version_resource.py
+++ b/hathor/version_resource.py
@@ -45,7 +45,8 @@ class VersionResponse(ResponseModel):
     min_tx_weight: int = Field(description="Minimum transaction weight")
     min_tx_weight_coefficient: float = Field(description="Minimum transaction weight coefficient")
     min_tx_weight_k: float = Field(description="Minimum transaction weight k constant")
-    token_deposit_percentage: float = Field(description="Token deposit percentage")
+    token_deposit_percentage: float = Field(description="Token deposit percentage. DEPRECATED")
+    token_deposit_percentage_ppb: int = Field(description="Token deposit percentage in ppb (parts per billion)")
     reward_spend_min_blocks: int = Field(description="Minimum blocks before reward can be spent")
     max_number_inputs: int = Field(description="Maximum number of inputs per transaction")
     max_number_outputs: int = Field(description="Maximum number of outputs per transaction")
@@ -74,6 +75,7 @@ VersionResponse.openapi_examples = {
             min_tx_weight_coefficient=1.6,
             min_tx_weight_k=100,
             token_deposit_percentage=0.01,
+            token_deposit_percentage_ppb=10**7,
             reward_spend_min_blocks=300,
             max_number_inputs=256,
             max_number_outputs=256,
@@ -135,7 +137,8 @@ class VersionResource(Resource):
             min_tx_weight=self._settings.MIN_TX_WEIGHT,
             min_tx_weight_coefficient=self._settings.MIN_TX_WEIGHT_COEFFICIENT,
             min_tx_weight_k=self._settings.MIN_TX_WEIGHT_K,
-            token_deposit_percentage=self._settings.TOKEN_DEPOSIT_PERCENTAGE,
+            token_deposit_percentage=self._settings.TOKEN_DEPOSIT_PERCENTAGE_PPB / 10**9,
+            token_deposit_percentage_ppb=self._settings.TOKEN_DEPOSIT_PERCENTAGE_PPB,
             reward_spend_min_blocks=self._settings.REWARD_SPEND_MIN_BLOCKS,
             max_number_inputs=self._settings.MAX_NUM_INPUTS,
             max_number_outputs=self._settings.MAX_NUM_OUTPUTS,
--- a/hathor/version_resource.py
+++ b/hathor/version_resource.py
@@ -46,7 +46,8 @@ class VersionResponse(ResponseModel):
     min_tx_weight_coefficient: float = Field(description="Minimum transaction weight coefficient")
     min_tx_weight_k: float = Field(description="Minimum transaction weight k constant")
     token_deposit_percentage: float = Field(description="Token deposit percentage. DEPRECATED")
-    token_deposit_percentage_ppb: int = Field(description="Token deposit percentage in ppb (parts per billion)")
+    token_deposit_percentage_numerator: int = Field(description="Token deposit percentage numerator")
+    token_deposit_percentage_denominator: int = Field(description="Token deposit percentage denominator")
     reward_spend_min_blocks: int = Field(description="Minimum blocks before reward can be spent")
     max_number_inputs: int = Field(description="Maximum number of inputs per transaction")
     max_number_outputs: int = Field(description="Maximum number of outputs per transaction")
@@ -75,7 +76,8 @@ VersionResponse.openapi_examples = {
             min_tx_weight_coefficient=1.6,
             min_tx_weight_k=100,
             token_deposit_percentage=0.01,
-            token_deposit_percentage_ppb=10**7,
+            token_deposit_percentage_numerator=10**7,
+            token_deposit_percentage_denominator=10**9,
             reward_spend_min_blocks=300,
             max_number_inputs=256,
             max_number_outputs=256,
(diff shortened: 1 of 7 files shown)


hathor/version_resource.py
```diff
@@ -45,7 +45,8 @@ class VersionResponse(ResponseModel):
     min_tx_weight: int = Field(description="Minimum transaction weight")
     min_tx_weight_coefficient: float = Field(description="Minimum transaction weight coefficient")
     min_tx_weight_k: float = Field(description="Minimum transaction weight k constant")
-    token_deposit_percentage: float = Field(description="Token deposit percentage")
+    token_deposit_percentage: float = Field(description="Token deposit percentage. DEPRECATED")
+    token_deposit_percentage_ppb: int = Field(description="Token deposit percentage in ppb (parts per billion)")
```
msbrogli: Why parts per billion?

hathorlib/hathorlib/nanocontracts/runner/token_fees.py
```diff
@@ -83,4 +83,6 @@ def _calculate_unit_fee_token_fee(settings: HathorSettings, fee_payment_token: T
     """Calculate the fee for handling a fee-based token"""
     if fee_payment_token.token_id == HATHOR_TOKEN_UID:
         return settings.FEE_PER_OUTPUT
-    return int(settings.FEE_PER_OUTPUT / settings.TOKEN_DEPOSIT_PERCENTAGE)
+    numerator = settings.FEE_PER_OUTPUT * 10**9
```
msbrogli: Why not update the settings (instead of adjusting it here)? The output v1 deserializer can do this multiplication when needed.

msbrogli (changes requested):

hathor/version_resource.py
glevco: Because the Ekvilibro network uses `0.0000001`, which is 100 ppb. If I had used ppm or anything less precise, we wouldn't be able to represent it as int.

hathorlib/hathorlib/nanocontracts/runner/token_fees.py
glevco: What would you update it to? This multiplication has nothing to do with the token version scaling, it's just a percentage calculation in integer math. Before it was `FEE_PER_OUTPUT / 0.01` and now it's `FEE_PER_OUTPUT * 10**9 // 10**7` (the `ppb` way to calculate percentages).

Token version scaling lands in another PR.

jansegre (dismissed):

hathorlib/hathorlib/conf/settings.py
```diff
@@ -134,9 +134,8 @@ def GENESIS_TOKEN_ATOMIC_UNITS(self) -> int:
     @property
     def FEE_DIVISOR(self) -> int:
         """Divisor used for evaluating fee amounts"""
-        result = 1 / self.TOKEN_DEPOSIT_PERCENTAGE
-        assert result.is_integer()
-        return int(result)
+        assert 10**9 % self.TOKEN_DEPOSIT_PERCENTAGE_PPB == 0
+        return 10**9 // self.TOKEN_DEPOSIT_PERCENTAGE_PPB
```
msbrogli: I don't like magic constants like `10**9`. Should we use two settings, one for the numerator and another for the denominator?

hathorlib/hathorlib/utils/__init__.py
```diff
@@ -81,8 +81,14 @@ def not_none(optional: Optional[_T], message: str = 'Unexpected `None`') -> _T:
...
+    return ceil_div(settings.TOKEN_DEPOSIT_PERCENTAGE_PPB * abs(mint_amount), 10**9)
 
 
 def get_deposit_token_withdraw_amount(settings: 'HathorSettings', melt_amount: int) -> int:
-    return floor(abs(settings.TOKEN_DEPOSIT_PERCENTAGE * melt_amount))
+    return settings.TOKEN_DEPOSIT_PERCENTAGE_PPB * abs(melt_amount) // 10**9
+
+
+def ceil_div(a: int, b: int) -> int:
+    """Calculate ceil division using integer math for non-negative operands."""
```
msbrogli: Just make it explicit that it calculates `ceil(a / b)` for integers.

hathorlib/hathorlib/utils/__init__.py
```diff
@@ -81,8 +81,14 @@ def not_none(optional: Optional[_T], message: str = 'Unexpected `None`') -> _T:
...
+    assert a >= 0 and b >= 0
```
msbrogli, line 99: `b > 0`

msbrogli (dismissed):

hathorlib/hathorlib/nanocontracts/runner/token_fees.py
msbrogli: Got it. Thanks!

hathor/version_resource.py
msbrogli: Got it. Thanks!

hathorlib/hathorlib/conf/settings.py
glevco: Refactored to numerator and denominator in https://github.com/HathorNetwork/hathor-core/pull/1711/commits/5e2852ba909fb4267d0d945822fe0604f676b3ba

hathorlib/hathorlib/utils/__init__.py
glevco: Updated docstring in https://github.com/HathorNetwork/hathor-core/pull/1711/commits/5e2852ba909fb4267d0d945822fe0604f676b3ba

hathorlib/hathorlib/utils/__init__.py
glevco: I kept it accepting 0 to keep previous behavior, which accepted 0 and raised `ZeroDivisionError` instead of `AssertionError`. I think it's more useful to keep it like this than converting it to `AssertionError`, what do you think?

msbrogli: I agree. That's ok. Thanks!

hathor/version_resource.py
```diff
@@ -135,7 +137,10 @@ def render_GET(self, request):
             min_tx_weight=self._settings.MIN_TX_WEIGHT,
             min_tx_weight_coefficient=self._settings.MIN_TX_WEIGHT_COEFFICIENT,
             min_tx_weight_k=self._settings.MIN_TX_WEIGHT_K,
-            token_deposit_percentage=self._settings.TOKEN_DEPOSIT_PERCENTAGE,
+            token_deposit_percentage=(
+                self._settings.TOKEN_DEPOSIT_PERCENTAGE_NUMERATOR / self._settings.TOKEN_DEPOSIT_PERCENTAGE_DENOMINATOR
+            ),
+            token_deposit_percentage_ppb=self._settings.TOKEN_DEPOSIT_PERCENTAGE_NUMERATOR,
```
msbrogli: Why not provided both numerator and denominator?

msbrogli (dismissed):

hathorlib/hathorlib/utils/__init__.py
msbrogli: Thanks!

hathorlib/hathorlib/conf/settings.py
msbrogli: Thanks!

hathor/version_resource.py
glevco: Makes sense, updated in [6e8d089](https://github.com/HathorNetwork/hathor-core/pull/1711/commits/6e8d089c45bbf4a4b6f9b0d598d3b715333c0090)

msbrogli: Thanks!

msbrogli (approved):

jansegre (approved):

PR #1712, opened 2026-05-27
feat(shielded): wallet support for shielded outputs (gated OFF)

LFRezende:
### Motivation

In shielded transactions, with all the data being hidden from external observers, one needs to be able to recover them - this is to be the receiver. 

The wallet of the receiver needs to be able to 1. Detect incoming payments with shielded UTXOs in it and 2. Unravel the proofs by rewind, extracting back the hidden amount being transferred and the token, if encrypted as well.

By usage of the ECDH secret-sharing technique, a wallet is able to rewind the present proofs and accomplish 2. Item 1, on the other hand, is dealt solely by identifying the shielded output proper APIs. 

In accordance to the original 8-step-plan on implementing shielded outputs in Hathor, this is step 6 **[6/8]**

Original plan: https://github.com/HathorNetwork/hathor-core/blob/feat/shielded-outputs/docs/plans/shielded-pr-split.md

Current stack: Confidential Transactions, modeled on Liquid's Confidential Assets, with Pedersen Commitments, Borromean Range Proofs (t.b.d.), Asset Surjection Proofs, ECDH rewind.

This PR's implementation or not does not change consensus rules, as it just affects the wallet's perception of shielded outputs. Invalid transactions are still invalid, but shielded outputs are non-visible if PR-6 is not implemented.

### Acceptance Criteria:

1. Recovery of hidden amount in AmountHidden and FullyShielded transactions must be attainable.
2. Recovery of encrypted token tags in FullyShielded transactions.
3. All interactions with shielded outputs must be gated by an **ENABLE_SHIELDED_TRANSACTIONS/SHIELDED_CRYPTO_AVAILABLE = False**. Only toggle on after PR-8 lands.
4. If toggled True, wallet's current version after this PR skips shielded outputs in balance (doesn't compute/recognize them as the user's.). 
5. For push-to-merge purposes, this PR is intended now to be pushed after PR-5. Henceforth,**some files have been duplicated - they come from PR-5, and are added so to allow imports of PR-6 to pass. They must and will be excluded when PR-5 is merged**.
6. Many changes originally proposed to be handled by the genesis-plan have been deferred to PRs 7 and/or 8. Look below for a more thorough table analysis.
7. Aside from the changes on item 4, this must stay **hathor-core scoped only** - no changes in hathorlib, which was already de-mirrored in implementation in previous PRs (duplicated code was erased and all references are not in hathorlib anymore).

__________________________________________________________________________

  #### What the plan asked for

  | Plan item | Status | Notes |
  |---|---|---|
  | `base_wallet.py` (~228 lines, ECDH discovery + decryption) | ✅ Implemented | Landed at ~280 lines; delta covers the wallet adaptations below. |
  | `thin_wallet/address_balance.py` | ✅ Implemented | Skip shielded outputs; add advisory `has_shielded` flag to JSON. |
  | `thin_wallet/address_search.py` | ✅ Implemented | Same pattern. |
  | `test_shielded_wallet.py` | ❌ Deferred to PR 8 | Needs native Rust crate for ECDH fixtures. |
  | hathorlib: no change | ✅ Honored | Zero hathorlib changes. |

  #### What was brought forward / adapted

  Consequences of decisions in PRs 4 and 5 — not deviations from the *intent* of PR 6:

  | # | Adaptation | Why |
  |---|---|---|
  | W1 | `is_shielded_output(index) -> bool` on `GenericVertex` | Forwarded from PR 7's Option-B bundle. Wallet + both thin-wallet APIs are its first consumers. The rippling `resolve_spent_output` union return stays in PR 7. |
  | W2 | `recover_shielded_secrets()` in NEW `hathor/crypto/shielded/recovery.py` | cleanUp had it in `hathor/transaction/shielded_tx_output.py`, which PR 4
  de-mirrored to hathorlib. The function needs hathor-core crypto, so it gets a new home that keeps "hathorlib: no change" intact. |
  | W3a | `extract_key_bytes()` helper added to `base_wallet.py` | cleanUp had it in `hathor/crypto/shielded/ecdh.py`, which PR 5's `f4b38f1c` deleted. It's a pure-Python wallet utility (no crypto primitives), so the right home is `base_wallet.py`. |
  | W3b | Temporary `_resolve_spent_output(self, ...)` helper on `BaseWallet` | Local duplicate of what PR 7's `resolve_spent_output` union will do. Marked
  TEMPORARY; whoever picks up the PR-7 union must remove it. |

  | Deferred | Lands in | Reason |
  |---|---|---|
  | `test_shielded_wallet.py` | PR 8 | Module-level `import hathor_ct_crypto` for real ECDH fixtures. |
  | `resolve_spent_output` union return + ~10 callsite migration | PR 7 | Ripples into scripts, nano, indexes, resources/transaction API. Wallet uses the
  temporary helper today. |
  | `get_related_addresses` shielded iteration + `to_json_extended` shielded tokens block | PR 7 | Part of the indexes/sync/API/events surface. |
  | Native Rust crate (`hathor_ct_crypto`) | PR 8 | Until then the crate is an empty stub. |


htr-rs/crates/hathor-ct-crypto/hathor_ct_crypto.pyi
glevco, line 1: I think we should use kwargs-only params for most functions in this file that have 2 or more params with the same type.

hathor/crypto/shielded/recovery.py
```diff
@@ -0,0 +1,104 @@
...
+
+    if not output.ephemeral_pubkey:
+        raise ValueError('output has no ephemeral_pubkey for ECDH recovery')
+
+    shared_secret = derive_ecdh_shared_secret(private_key_bytes, output.ephemeral_pubkey)
+    nonce = derive_rewind_nonce(shared_secret)
+
+    token_uid: bytes
+    if isinstance(output, AmountShieldedOutput):
+        token_uid = get_token_uid(output.token_data & 0x7F)
```
glevco: Use the constant from `TxOutput` in `base_transaction.py` instead.

hathor/wallet/base_wallet.py
```diff
@@ -589,16 +666,131 @@ def on_new_tx(self, tx: BaseTransaction) -> None:
...
+        """Return the output (transparent or shielded) that an input at `index` spends.
+
+        TEMPORARY (PR 6): this duplicates locally what `BaseTransaction.resolve_spent_output` will do
+        once its shielded-aware union return lands in PR 7. When that lands, delete this helper and
+        switch the call sites back to `output_tx.resolve_spent_output(index)`. Whoever picks up the
+        PR-7 `resolve_spent_output` union should remove this.
+        """
+        if output_tx.is_shielded_output(index):
+            return output_tx.shielded_outputs[index - len(output_tx.outputs)]
+        return output_tx.outputs[index]
```
glevco: Can't we just move that method to this PR?

hathor/crypto/shielded/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+# would AttributeError. With this constant at False, the wallet's existing
+# guard (`if not SHIELDED_CRYPTO_AVAILABLE: return False`) keeps the discovery
+# loop structurally dormant, even if an operator enables
+# ENABLE_SHIELDED_TRANSACTIONS. The constant only flips to True when PR-8
+# ships the real Rust implementation — not when PR-5 merges.
+#
+# PR-5 cleanup (post-Jan's _bindings removal): this constant is gone on master.
+# When PR-6 rebases post-PR-5, drop this line AND remove the `if not
+# SHIELDED_CRYPTO_AVAILABLE: return False` block in hathor/wallet/base_wallet.py.
+SHIELDED_CRYPTO_AVAILABLE: bool = False
```
glevco: I looks like this is unnecessary, why not just reuse `ENABLE_SHIELDED_TRANSACTIONS`?

glevco (changes requested):

hathor/crypto/shielded/__init__.py
LFRezende: Done, duplicated the flag needlessly.

hathor/wallet/base_wallet.py
LFRezende: Migrated the method and its plumbing to this PR now. Better sooner.

hathor/crypto/shielded/recovery.py
LFRezende: Just did so.

htr-rs/crates/hathor-ct-crypto/hathor_ct_crypto.pyi
LFRezende: Fair enough! See if the implementation I did suffices.

glevco: Thanks!

hathor/crypto/shielded/recovery.py
glevco: Thanks!

hathor/wallet/base_wallet.py
glevco: Thanks!

hathor/crypto/shielded/__init__.py
glevco: Thanks!

hathor/transaction/base_transaction.py
```diff
@@ -465,8 +479,12 @@ def add_address_from_output(output: 'TxOutput') -> None:
...
         for txin in self.inputs:
             tx2 = self.storage.get_transaction(txin.tx_id)
-            txout = tx2.resolve_spent_output(txin.index)
-            add_address_from_output(txout)
+            # resolve_spent_output is shielded-aware; both transparent TxOutput and
+            # shielded outputs expose `.script`, so we parse it directly.
+            resolved = tx2.resolve_spent_output(txin.index)
+            script_type_resolved = parse_address_script(resolved.script)
+            if script_type_resolved:
+                addresses.add(script_type_resolved.address)
```
glevco: This is the same as `add_address_from_output` just above.

hathor/transaction/base_transaction.py
```diff
@@ -814,11 +832,24 @@ def serialize_output(tx: BaseTransaction, tx_out: TxOutput) -> dict[str, Any]:
...
-            output['tx_id'] = tx2.hash_hex
-            output['index'] = tx_in.index
-            ret['inputs'].append(output)
+            # Shielded inputs need a different serialization shape than transparent ones,
+            # since shielded outputs have a commitment instead of a value.
+            if tx2.is_shielded_output(tx_in.index):
+                shielded_out = tx2.resolve_spent_output(tx_in.index)
+                output_data: dict[str, Any] = {
+                    'type': 'shielded',
+                    'commitment': shielded_out.commitment.hex(),  # type: ignore[union-attr]
```
glevco: assert the output type instead of ignoring

hathor/transaction/base_transaction.py
```diff
@@ -814,11 +832,24 @@ def serialize_output(tx: BaseTransaction, tx_out: TxOutput) -> dict[str, Any]:
...
+                    'script': shielded_out.script.hex(),
+                    'tx_id': tx2.hash_hex,
+                    'index': tx_in.index,
+                }
+            else:
+                tx2_out = tx2.outputs[tx_in.index]
```
glevco: You can use `resolve_spent_output` here too and move it before the if

hathor/transaction/transaction.py
```diff
@@ -389,8 +389,23 @@ def _get_token_info_from_inputs(self, nc_block_storage: NCBlockStorage) -> Token
 
         for tx_input in self.inputs:
             spent_tx = self.get_spent_tx(tx_input)
-            spent_output = spent_tx.resolve_spent_output(tx_input.index)
 
+            try:
+                resolved = spent_tx.resolve_spent_output(tx_input.index)
+            except IndexError:
+                # Out of bounds — will be caught by _verify_inputs
+                continue
```
glevco, line 397: That's an assertion in `_verify_inputs`. It has to change, right?

hathor/nanocontracts/vertex_data.py
```diff
@@ -51,10 +51,18 @@ def _get_txin_output(vertex: BaseTransaction, txin: TxInput) -> TxOutput | None:
...
+    # Use resolve_spent_output for shielded-aware lookup
+    try:
+        resolved = vertex2.resolve_spent_output(txin.index)
+    except IndexError:
+        return None
+
+    # Only return TxOutput; shielded outputs lack value/token_data for TxOutputData
+    from hathor.transaction import TxOutput as _TxOutput
+    if not isinstance(resolved, _TxOutput):
+        return None
```
glevco: I think we can just keep the assertion at this point. It'll fail as soon as we add a test with nano + shielded, which is better than the None that will (maybe?) fail silently

hathor/transaction/resources/transaction.py
```diff
@@ -93,7 +93,18 @@ def get_tx_extra_data(
     for index, tx_in in enumerate(tx.inputs):
         if tx.storage:
             tx2 = tx.storage.get_transaction(tx_in.tx_id)
-            tx2_out = tx2.resolve_spent_output(tx_in.index)
+
+            if tx2.is_shielded_output(tx_in.index):
+                from hathorlib.transaction.shielded_tx_output import AmountShieldedOutput, FullShieldedOutput
+                shielded_out = tx2.resolve_spent_output(tx_in.index)
+                assert isinstance(shielded_out, (AmountShieldedOutput, FullShieldedOutput))
+                output = _shielded_output_to_json(shielded_out, decode_script=True)
```
glevco, line 101: Why don't we include this in the `to_json_extended` method? We do for normal outputs.

glevco (changes requested):

hathor/transaction/base_transaction.py
LFRezende: Just fixed it.

hathor/transaction/base_transaction.py
LFRezende: Done.

hathor/transaction/base_transaction.py
LFRezende: Did so. Good catch!

hathor/transaction/transaction.py
LFRezende: I did a class check in resolve_outputs in the check in `_verify_inputs`. See if it is decent.

hathor/nanocontracts/vertex_data.py
LFRezende: Good point. We'll do it so.

hathor/transaction/resources/transaction.py
LFRezende: Included.

hathor/transaction/base_transaction.py
glevco: Thanks!

hathor/transaction/base_transaction.py
glevco: Thanks!

hathor/transaction/base_transaction.py
glevco: Thanks!

hathor/nanocontracts/vertex_data.py
glevco: Thanks!

hathor/transaction/resources/transaction.py
glevco: Thanks!

glevco (approved):

hathor/transaction/transaction.py
glevco: Thanks!

jansegre (approved):

PR #1713, opened 2026-05-27
feat: external events simulator scenario

r4mmer:
### Motivation

Each project may want to create custom scenarios to simulate and use the events simulator in integration tests.

### Acceptance Criteria

- Add new option to the event simulator cli run a scenario from an external file
- Load external module and run simulation
- Create tests to cover new use-cases.

### Checklist


hathor_cli/events_simulator/events_simulator.py
```diff
@@ -36,7 +36,16 @@ def create_parser() -> ArgumentParser:
     parser = create_parser()
     possible_scenarios = [scenario.name for scenario in Scenario]
 
-    parser.add_argument('--scenario', help=f'One of {possible_scenarios}', type=str, required=True)
+    scenario_group = parser.add_mutually_exclusive_group(required=True)
+    scenario_group.add_argument('--scenario', help=f'One of {possible_scenarios}', type=str)
+    scenario_group.add_argument('--file', help='external scenario file, e.g. "./my_scenario.py"', type=str)
```
glevco: You could use `type=FileType('r')` (not sure if it will be useful since you need to load the module downstream, otherwise you can use `type=Path` at least.

hathor_cli/events_simulator/external_scenario.py
```diff
@@ -0,0 +1,65 @@
...
+            raise ValueError(f'External scenario file not found: {file_path}')
+        module_name = '_hathor_external_' + hashlib.md5(file_path.encode()).hexdigest()[:8]
+        spec = importlib.util.spec_from_file_location(module_name, file_path)
+        if spec is None:
+            raise ValueError(f'Could not load module spec from: {file_path}')
+        if spec.loader is None:
+            raise ValueError(f'No loader available for: {file_path}')
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        return module
```
glevco: Why not just `importlib.import_module(file_path)`?

hathor/simulator/utils.py
```diff
@@ -177,3 +178,27 @@ def gen_new_double_spending(manager: HathorManager, *, use_same_parents: bool =
 
     manager.cpu_mining_service.resolve(tx2)
     return tx2
+
+
+def create_dag_builder(manager: 'HathorManager') -> 'DAGBuilder':
```
glevco, line 183: This is the same as `_create_dag_builder` from `scenario.py`, why not reuse/move it?

r4mmer: This is exactly the same, i'm just exposing it as an util.

hathor_cli/events_simulator/external_scenario.py
r4mmer: From what I found the `spec_from_file_location` -> `module_from_spec` -> `exec_module` is the correct way to load a module from an arbitrary location, since `import_module` resolves by name the module must be on the `sys.path`.
This approach also does not touch the `sys.modules` so there is no way to use cleverly named files to monkey-patch some internal module.

Since loading from an arbitrary path is the intention this seems like the most stable way to do this.

hathor_cli/events_simulator/events_simulator.py
r4mmer: Changed to use `Path` type.

glevco: Thanks!

hathor/simulator/utils.py
glevco: Then I think we should remove one of them

r4mmer: Done

glevco: Thanks!

glevco (approved):

jansegre (approved):

PR #1714, opened 2026-06-01
feat(decimal): update codebase to use new versioned token amount model [part 17b]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1755

### Motivation

This PR is the main implementation change in the Token Amount V2 project. Its diff is large mostly for tests, the production changes are actually not large, but extremely important, so they should be reviewed with extreme care.

It updates the current hathorlib `int` aliases, `UnsignedAmount` and `SignedAmount`, to point to the actual new types from `htr_lib`. This effectively makes all internal token amount handling in the full node use the new types instead of ints, which requires conversions at all boundaries. That's what causes the large diffs for tests (all fixtures had to be updated), and it can't be done in a separate PR.

To review this PR it's worth revisiting https://github.com/HathorNetwork/hathor-core/pull/1717, that introduced the new models. Here's a cheatsheet:

- `UnsignedAmount` is a versioned new type that holds either V1 or V2 amounts internally. All operations (equality, comparison, arithmetic) is handled by it using V2-normalized values. It's unsigned.
- `SignedAmount` is a signed counterpart, unversioned. It always holds V2 balances and it's used only internally.
- Therefore:
  - `UnsignedAmount.raw()` returns the internal `int` representation in its own version.
  - `UnsignedAmount.normalized()` returns the `int` V2 representation, always.
  - `UnsignedAmount.to_unsigned()` converts a `UnsignedAmount` to a `SignedAmount` for internal usage (which is always V2).
  - `SignedAmount.raw()` returns the internal `int` representation (which is always V2).

### Acceptance Criteria

- Update temporary hathorlib `int` aliases `UnsignedAmount` and `SignedAmount` to use the actual `htr_lib` types.

#### UTXO world

- Update `TxOutput.value` to use `UnsignedAmount` instead of `int`. Same for actions and fee amounts.
- Update `TokenInfo.amount` to use `SignedAmount` instead of `int`.
- Update inputs/outputs balance verification accordingly. Everything is checked in V2-normalized space.
- Handling of token authorities always uses `raw`.

#### Nano world

- Update all internal amount handling in the `Runner` and related classes to use V2.
- Data coming INTO nano runtime is converted using `raw`, that is, the transaction's own token amount version, to keep compatibility with existing Blueprints. Here are them:
  - `TxOutputData.value` in `VertexData`.
  - `NCDepositAction.amount` and `NCWithdrawalAction.amount`.
  - `BlueprintEnv.get_current_balance` and `get_balance_before_current_call`.
  - `ContractAccessor.get_current_balance`.

#### Deposit-based tokens

- 1% mint/melt is now done in V2-normalized space, and rounds to the nearest cent (up for deposit, down for withdrawal).

#### Other

- Set the `UnsignedAmount` decimal places on `run_node` setup, before the node starts.
- Update handling of block rewards to use V1 tokens only.
- Update the DAGBuilder to use V2 tokens internally, all txs are hardcoded as V1.
- Update tokens index and UTXO index to use V2 tokens internally.
- Update APIs: updated relevant `/v2` routes to use decimal strings in requests and responses.
- Add new `HathorSettings.FEE_TOKEN_AMOUNT_PER_OUTPUT` computed property.

### Checklist


hathor/transaction/base_transaction.py
```diff
@@ -969,8 +972,6 @@ def __init__(self, value: TokenAmount, script: TxOutputScript, token_data: int =
         assert isinstance(value, TokenAmount), 'value is %s, type %s' % (str(value), type(value))
         assert isinstance(script, TxOutputScript), 'script is %s, type %s' % (str(script), type(script))
         assert isinstance(token_data, int), 'token_data is %s, type %s' % (str(token_data), type(token_data))
-        if value <= 0 or value > MAX_OUTPUT_VALUE:
-            raise InvalidOutputValue
```
glevco, line 993: **Review aid:** This is safe to remove because negatives are gated statically by the `UnsignedAmount` type, zero is gated by the vertex verifier, and the max is gated by deserialization (so we don't need to take the transaction's token amount version here).

hathorlib/hathorlib/daa.py
```diff
@@ -42,8 +42,8 @@ def minimum_tx_weight(tx: 'Transaction', *, fix_parents: bool = True) -> float:
     # We need to take into consideration the decimal places because it is inside the amount.
     # For instance, if one wants to transfer 20 HTRs, the amount will be 2000.
     # Max below is preventing division by 0 when handling authority methods that have no outputs
-    decimal_places = tx.get_token_amount_version().get_decimal_places(settings)
-    amount = max(1, tx.sum_outputs) / (10 ** decimal_places)
+    # TODO: Update hathorlib with normalization changes.
```
glevco, line 34: Hathorlib will be updated in a separate PR.

hathor/daa/common.py
```diff
@@ -194,8 +194,7 @@ def _minimum_tx_weight(settings: HathorSettings, tx: 'Transaction', test_mode: T
     # We need to take into consideration the decimal places because it is inside the amount.
     # For instance, if one wants to transfer 20 HTRs, the amount will be 2000.
     # Max below is preventing division by 0 when handling authority methods that have no outputs
-    decimal_places = tx.get_token_amount_version().get_decimal_places(settings)
-    amount = max(1, tx.sum_outputs) / (10 ** decimal_places)
+    amount = max(1, tx.sum_outputs.normalized()) / (10 ** settings.TOKEN_AMOUNT_V2_DECIMAL_PLACES)
```
glevco, line 186: **Review aid:** before, the `sum_outputs` returned an `int` and I was dividing by the decimal places of the transaction. Now, since `sum_outputs` returns an `UnsignedAmount`, I just normalize it to V2 and divide it by V2 decimal places, which is always correct.

hathor/daa/daa.py
```diff
@@ -106,7 +106,7 @@ def get_block_dependencies(
     def get_tokens_issued_per_block(self, height: int) -> UnsignedAmount:
         """Return the number of tokens issued (aka reward) per block of a given height."""
         amount = _get_base_tokens_issued_per_block(self._settings, height)
-        return amount // self._config.reward_reduction_factor
+        return UnsignedAmount.from_v1(amount // self._config.reward_reduction_factor)
```
glevco, line 98: **Review aid:** Block rewards are always created as V1. The `amount` that comes from `_get_base_tokens_issued_per_block` is a raw V1 number.

hathor/indexes/rocksdb_tokens_index.py
```diff
@@ -331,7 +331,7 @@ def _create_genesis_info(self) -> None:
             name=self._settings.HATHOR_TOKEN_NAME,
             symbol=self._settings.HATHOR_TOKEN_SYMBOL,
             version=TokenVersion.NATIVE,
-            total=self._settings.GENESIS_TOKEN_ATOMIC_UNITS,
+            total=UnsignedAmount.from_v1(self._settings.GENESIS_TOKEN_ATOMIC_UNITS),
```
glevco, line 323: **Review aid:** Block rewards are always V1.

hathor/indexes/rocksdb_tokens_index.py
```diff
@@ -350,7 +350,7 @@ def _get_value_info(self, token_uid: bytes, *, create_default: bool = True) -> _
     @override
     def add_to_total(self, token_uid: bytes, amount: SignedAmount) -> None:
         dict_info = self._get_value_info(token_uid, create_default=True)
-        dict_info.total += amount
+        dict_info.total += amount.raw()
```
glevco, line 342: **Review aid:** `SignedAmount.raw()` is always V2, and the index uses V2 internally (always sums `UnsignedAmount.normalized()`).

hathor/mining/block_template.py
```diff
@@ -99,7 +100,7 @@ def to_dict(self) -> dict:
         return {
             'data': self.generate_minimally_valid_block().get_struct_without_nonce().hex(),
             'versions': sorted(self.versions),
-            'reward': self.reward,
+            'reward': self.reward.raw(),
```
glevco, line 92: **Review aid:** Block rewards are always V1, so it's safe to use `raw()`.

hathor/nanocontracts/vertex_data.py
```diff
@@ -78,7 +78,7 @@ def create_txoutput_data_from_txout(txout: TxOutput) -> TxOutputData:
     """Create a TxOutputData from a TxOutput."""
     parsed = parse_address_script(txout.script)
     return TxOutputData(
-        value=txout.value,
+        value=txout.value.raw(),
```
glevco, line 70: **Review aid:** this is one of the few places where data is transferred INTO the nano runtime. The `txout.value` is a versioned `UnsignedAmount`, so we use `raw()` to get the un-normalized value.

- Since V1 contracts can only be called by V1 transactions, this raw value will be an int with 2 decimal places. This is compatible with the current state of the network.
- Since V2 contracts can only be called by V2 transactions, this raw value will be an int with 18 decimal places. Ready for new V2 Blueprints.

hathor/transaction/headers/nano_header.py
```diff
@@ -58,19 +58,20 @@ def to_nc_action(self, tx: Transaction) -> NCAction:
...
-                melt = self.amount & TxOutput.TOKEN_MELT_MASK > 0
+                mint = raw_amount & TxOutput.TOKEN_MINT_MASK > 0
+                melt = raw_amount & TxOutput.TOKEN_MELT_MASK > 0
                 self._validate_authorities(token_uid)
                 return NCGrantAuthorityAction(token_uid=token_uid, mint=mint, melt=melt)
             case NCActionType.ACQUIRE_AUTHORITY:
-                mint = self.amount & TxOutput.TOKEN_MINT_MASK > 0
-                melt = self.amount & TxOutput.TOKEN_MELT_MASK > 0
+                mint = raw_amount & TxOutput.TOKEN_MINT_MASK > 0
+                melt = raw_amount & TxOutput.TOKEN_MELT_MASK > 0
```
glevco, line 63: **Review aid:** this is one of the few places where data is transferred INTO the nano runtime. The `NanoHeaderAction.amount` is a versioned `UnsignedAmount`, so we use `raw()` to get the un-normalized value.

For `DEPOSIT`/`WITHDRAWAL`:

- Since V1 contracts can only be called by V1 transactions, this raw value will be an int with 2 decimal places. This is compatible with the current state of the network.
- Since V2 contracts can only be called by V2 transactions, this raw value will be an int with 18 decimal places. Ready for new V2 Blueprints.

For `GRANT_AUTHORITY`/`ACQUIRE_AUTHORITY`:

- Authorities always use the raw value. This allow us to use the same mint/melt masks for both versions.

hathor/transaction/base_transaction.py
```diff
@@ -1046,11 +1047,11 @@ def is_standard_script(self) -> bool:
 
     def can_mint_token(self) -> bool:
         """Whether this utxo can mint tokens"""
-        return self.is_token_authority() and ((self.value & self.TOKEN_MINT_MASK) > 0)
+        return self.is_token_authority() and ((self.value.raw() & self.TOKEN_MINT_MASK) > 0)
 
     def can_melt_token(self) -> bool:
         """Whether this utxo can melt tokens"""
-        return self.is_token_authority() and ((self.value & self.TOKEN_MELT_MASK) > 0)
+        return self.is_token_authority() and ((self.value.raw() & self.TOKEN_MELT_MASK) > 0)
```
glevco, line 1043: **Review aid:** Authorities always use the raw value. This allow us to use the same mint/melt masks for both versions.

hathor/verification/transaction_verifier.py
```diff
@@ -302,9 +304,9 @@ def verify_transparent_balance(
 
                 case TokenVersion.DEPOSIT:
                     if token_info.has_been_melted():
-                        withdraw += get_deposit_token_withdraw_amount(settings, token_info.amount)
+                        withdraw += get_deposit_token_withdraw_amount(settings, (-token_info.amount).to_unsigned())
```
glevco, line 296: **Review aid:** This was a negative value converted inside `get_deposit_token_withdraw_amount` with `abs`. Since `TokenAmount` doesn't support negatives, we removed the `abs` from there and convert here at the call site, instead.

hathorlib/hathorlib/nanocontracts/runner/runner.py
```diff
@@ -1303,9 +1310,16 @@ def syscall_get_nano_settings(self) -> NanoSettings:
...
+                fee_per_output: int
+                assert self._settings.FEE_TOKEN_AMOUNT_PER_OUTPUT.is_v1()
+                match self.token_amount_version:
+                    case TokenAmountVersion.V1:
+                        fee_per_output = self._settings.FEE_TOKEN_AMOUNT_PER_OUTPUT.raw()
+                    case TokenAmountVersion.V2:
+                        fee_per_output = self._settings.FEE_TOKEN_AMOUNT_PER_OUTPUT.normalized()
+                    case _:
+                        assert_never(self.token_amount_version)
+                return NanoSettings(fee_per_output=fee_per_output)
```
glevco, line 1333: **Review aid:** this is one of the few places where data is transferred INTO the nano runtime. The `NanoSettings.fee_per_output` must contain a raw int value scaled respectively to the version in the runner (which comes from the transaction). Since `FEE_TOKEN_AMOUNT_PER_OUTPUT` is V1, we send `raw()` for a V1 runtime and `normalized()` for a V2 runtime.

hathorlib/hathorlib/nanocontracts/blueprint_env.py
glevco, line 1: **Review aid:** this is one of the few places where data is transferred INTO the nano runtime. The balance returned from the runner is an internal `SignedAmount`, so we convert it to the runtime's versioned unsigned value and use `raw()` to get the un-normalized value.

- Since V1 contracts can only be called by V1 transactions, this raw value will be an int with 2 decimal places. This is compatible with the current state of the network.
- Since V2 contracts can only be called by V2 transactions, this raw value will be an int with 18 decimal places. Ready for new V2 Blueprints.

hathorlib/hathorlib/nanocontracts/contract_accessor.py
```diff
@@ -71,7 +71,8 @@ def get_current_balance(self, token_uid: TokenUid | None = None) -> Amount:
         then this method will return 47 HTR.
         """
         balance = self.__runner.get_current_balance(self.__contract_id, token_uid)
-        return Amount(balance.value)
+        token_amount = balance.value.to_unsigned().to_version(self.__runner.token_amount_version)
+        return Amount(token_amount.raw())
```
glevco, line 64: **Review aid:** this is one of the few places where data is transferred INTO the nano runtime. The balance returned from the runner is an internal `SignedAmount`, so we convert it to the runtime's versioned unsigned value and use `raw()` to get the un-normalized value.

- Since V1 contracts can only be called by V1 transactions, this raw value will be an int with 2 decimal places. This is compatible with the current state of the network.
- Since V2 contracts can only be called by V2 transactions, this raw value will be an int with 18 decimal places. Ready for new V2 Blueprints.

hathorlib/hathorlib/utils/__init__.py
```diff
@@ -82,16 +82,21 @@ def not_none(optional: Optional[_T], message: str = 'Unexpected `None`') -> _T:
...
 
 
 def get_deposit_token_withdraw_amount(settings: 'HathorSettings', melt_amount: UnsignedAmount) -> UnsignedAmount:
-    numerator = settings.TOKEN_DEPOSIT_PERCENTAGE_NUMERATOR * abs(melt_amount)
+    numerator = settings.TOKEN_DEPOSIT_PERCENTAGE_NUMERATOR * melt_amount.normalized()
     denominator = settings.TOKEN_DEPOSIT_PERCENTAGE_DENOMINATOR
-    return numerator // denominator
+    rounded_down_to_0 = numerator // denominator
+    rounded_down_to_cent = (rounded_down_to_0 // 10**16) * 10**16
+    return UnsignedAmount.from_v2(rounded_down_to_cent)
```
glevco, line 89: **Review aid:** this includes a new rule that should be backwards compatible with the current state of mainnet/testnet, but will behave differently when V2 is enabled.

It's the same calculation as before, but now it not only rounds floats (values < 0), but also rounds to the nearest cent, up for deposit and down for withdraw.

jansegre (approved):

PR #1716, opened 2026-06-03
refactor(decimal): renames and reorganization [part 8]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1722

### Motivation

In preparation for the main project PR to implement the new decimals place version, this PR does a few refactors, clean ups, and improvements so we can decrease the diff in the main PR. There's no change in any behavior, and the versioning system is just scaffolded — it's completely unused right now.

### Acceptance Criteria

- Rename `DecimalsVersion` to `TokenAmountVersion`.
- Thread the transaction's `TokenAmountVersion` through the `Runner`. It will be used in the future.
- Thread the transaction's `TokenAmountVersion` through the `BalanceRules`. It will be used in the future.
- Add new `get_blueprint_class_and_token_amount_version` method to `BlueprintService`, and use it in the nano state API.
- Remove duplicate functions `validate_token_name_and_symbol` and `validate_fee_amount` from `hathor`, use their counterparts from `hathorlib` instead.
- Remove `HathorManager.get_best_block_nc_runner` (it was only used in one test).
- Rename `NCFee.get_htr_value` to `__get_htr_value__`, moving it out of the user API for Blueprints.
- Refactors in `HathorSettings`:
  - Replace `VERTEX_DECIMAL_PLACES` by `TOKEN_AMOUNT_V1_DECIMAL_PLACES` and `TOKEN_AMOUNT_V2_DECIMAL_PLACES` (and refactor the `/version` API accordingly).
  - Rename `FEE_PER_OUTPUT` to `FEE_PER_OUTPUT_V1`.
- Remove support for Python 3.9 and 3.10 from hathorlib and add `htr_lib` as a dependency (currently unused).
- Update `htr-rs/justfile` to include a `fmt` command.

### Checklist

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -69,6 +69,7 @@ from hathor.verification.verification_service import VerificationService
 from hathor.vertex_handler import VertexHandler
 from hathor.wallet import BaseWallet
 from hathorlib.base_transaction import tx_or_block_from_bytes as lib_tx_or_block_from_bytes
+from hathorlib.token_amount_version import TokenAmountVersion
 
 if TYPE_CHECKING:
     from hathor.websocket.factory import HathorAdminWebsocketFactory
@@ -410,7 +411,7 @@ class HathorManager:
         if save_to:
             self.profiler.dump_stats(save_to)
 
-    def get_nc_runner(self, block: Block) -> Runner:
+    def get_nc_runner(self, block: Block, *, token_amount_version: TokenAmountVersion) -> Runner:
         """Return a contract runner for a given block."""
         nc_storage_factory = self.consensus_algorithm.nc_storage_factory
         block_storage = get_block_storage_from_block(nc_storage_factory, block)
@@ -418,12 +419,8 @@ class HathorManager:
         return self.runner_factory.create(
             runtime_version=features.nano_runtime_version,
             block_storage=block_storage,
+            token_amount_version=token_amount_version
         )
-
-    def get_best_block_nc_runner(self) -> Runner:
-        """Return a contract runner for the best block."""
-        best_block = self.tx_storage.get_best_block()
-        return self.get_nc_runner(best_block)
 
     def get_nc_block_storage(self, block: Block) -> NCBlockStorage:
         """Return the nano block storage for a given block."""
--- a/hathor/nanocontracts/blueprint_service.py
+++ b/hathor/nanocontracts/blueprint_service.py
@@ -30,6 +30,7 @@ from hathor.nanocontracts.exception import (
 from hathor.transaction.storage import TransactionStorage
 from hathor.transaction.storage.exceptions import TransactionDoesNotExist
 from hathorlib.nanocontracts.types import BlueprintId
+from hathorlib.token_amount_version import TokenAmountVersion
 
 if TYPE_CHECKING:
     from hathor import Blueprint
@@ -72,17 +73,28 @@ class BlueprintService:
         # XXX: maybe use N blocks confirmation, like reward-locks
         return blueprint_tx
 
-    def get_blueprint_class(self, blueprint_id: BlueprintId) -> type[Blueprint]:
-        """Returns the blueprint class associated with the given blueprint_id.
+    def get_blueprint_class_and_token_amount_version(
+        self,
+        blueprint_id: BlueprintId,
+    ) -> tuple[type[Blueprint], TokenAmountVersion]:
+        """Returns the blueprint class associated with the given blueprint_id, and its TokenAmountVersion.
 
         The blueprint class could be in the catalog (first search), or it could be the tx_id of an on-chain blueprint.
         """
         from hathor.nanocontracts import OnChainBlueprint
         blueprint = self._get_blueprint(blueprint_id)
         if isinstance(blueprint, OnChainBlueprint):
-            return blueprint.get_blueprint_class()
+            return blueprint.get_blueprint_class(), blueprint.get_token_amount_version()
         else:
-            return blueprint
+            return blueprint, TokenAmountVersion.V2
+
+    def get_blueprint_class(self, blueprint_id: BlueprintId) -> type[Blueprint]:
+        """Returns the blueprint class associated with the given blueprint_id.
+
+        The blueprint class could be in the catalog (first search), or it could be the tx_id of an on-chain blueprint.
+        """
+        blueprint_class, _ = self.get_blueprint_class_and_token_amount_version(blueprint_id)
+        return blueprint_class
 
     def get_blueprint_source(self, blueprint_id: BlueprintId) -> str:
         """Returns the source code associated with the given blueprint_id.
--- a/hathor/nanocontracts/execution/block_executor.py
+++ b/hathor/nanocontracts/execution/block_executor.py
@@ -298,6 +298,7 @@ class NCBlockExecutor:
 
         runner = self._runner_factory.create(
             runtime_version=runtime_version,
+            token_amount_version=tx.get_token_amount_version(),
             block_storage=block_storage,
             seed=rng_seed,
         )
--- a/hathor/nanocontracts/resources/state.py
+++ b/hathor/nanocontracts/resources/state.py
@@ -48,6 +48,7 @@ class NanoContractStateResource(Resource):
     def __init__(self, manager: 'HathorManager') -> None:
         super().__init__()
         self.manager = manager
+        self.nc_storage_factory = manager.consensus_algorithm.nc_storage_factory
 
     def render_GET(self, request: 'Request') -> bytes:
         request.setHeader(b'content-type', b'application/json; charset=utf-8')
@@ -135,9 +136,11 @@ class NanoContractStateResource(Resource):
         else:
             block = self.manager.tx_storage.get_best_block()
 
+        from hathor.nanocontracts.storage import get_block_storage_from_block
+        block_storage = get_block_storage_from_block(self.nc_storage_factory, block)
+
         try:
-            runner = self.manager.get_nc_runner(block)
-            nc_storage = runner.get_storage(nc_id_bytes)
+            nc_storage = block_storage.get_contract_storage(nc_id_bytes)
         except NanoContractDoesNotExist:
             # Nano contract does not exist at this block
             request.setResponseCode(404)
@@ -148,7 +151,10 @@ class NanoContractStateResource(Resource):
             return error_response.json_dumpb()
 
         blueprint_id = nc_storage.get_blueprint_id()
-        blueprint_class = self.manager.blueprint_service.get_blueprint_class(blueprint_id)
+        blueprint_class, token_amount_version = (
+            self.manager.blueprint_service.get_blueprint_class_and_token_amount_version(blueprint_id)
+        )
+        runner = self.manager.get_nc_runner(block, token_amount_version=token_amount_version)
 
         value: Any
         # Get balances.
--- a/hathor/transaction/transaction.py
+++ b/hathor/transaction/transaction.py
@@ -36,7 +36,7 @@ from hathor.transaction.static_metadata import TransactionStaticMetadata
 from hathor.transaction.token_info import TokenInfo, TokenInfoDict, TokenVersion, get_token_version
 from hathor.transaction.util import VerboseCallback
 from hathor.types import TokenUid, VertexId
-from hathorlib.decimal_places import VertexDecimalVersion
+from hathorlib.token_amount_version import TokenAmountVersion
 from hathorlib.transaction.shielded_tx_output import ShieldedOutput
 
 T = TypeVar('T', bound=AnyVertexHeader)
@@ -267,7 +267,7 @@ class Transaction(GenericVertex[TransactionStaticMetadata]):
     def to_json(self, decode_script: bool = False, include_metadata: bool = False) -> dict[str, Any]:
         json = super().to_json(decode_script=decode_script, include_metadata=include_metadata)
         json['tokens'] = [h.hex() for h in self.tokens]
-        json['decimal_version'] = self.get_decimal_version().value
+        json['token_amount_version'] = self.get_token_amount_version().value
 
         if self.is_nano_contract():
             nano_header = self.get_nano_header()
@@ -345,7 +345,7 @@ class Transaction(GenericVertex[TransactionStaticMetadata]):
         nano_header = self.get_nano_header()
 
         for action in nano_header.get_actions():
-            rules = BalanceRules.get_rules(self._settings, action)
+            rules = BalanceRules.get_rules(self._settings, action, self.get_token_amount_version())
             if action.token_uid not in token_dict:
                 # we try to load this token version from storage in case it's not in the inputs
                 token_dict[action.token_uid] = TokenInfo(
@@ -489,9 +489,7 @@ class Transaction(GenericVertex[TransactionStaticMetadata]):
         self.set_static_metadata(static_metadata)
 
     @final
-    def get_decimal_version(self) -> VertexDecimalVersion:
-        """
-        Return the decimal-places version under which this transaction's token amounts are interpreted.
-        """
+    def get_token_amount_version(self) -> TokenAmountVersion:
+        """Return the version under which this transaction's token amounts are interpreted."""
         # Transactions are always V1. This will be updated in the future.
-        return VertexDecimalVersion.V1
+        return TokenAmountVersion.V1
--- a/hathor/version_resource.py
+++ b/hathor/version_resource.py
@@ -26,7 +26,6 @@ from hathor.manager import HathorManager
 from hathor.transaction.token_info import TokenVersion
 from hathor.types import BlockId, TransactionId
 from hathor.utils.pydantic import BaseModel, Hex
-from hathorlib.decimal_places import VertexDecimalVersion
 
 
 class NativeTokenInfo(BaseModel):
@@ -55,9 +54,8 @@ class VersionResponse(ResponseModel):
     display_decimal_places: int = Field(
         description="Number of decimal places to be displayed for all tokens; it is simply cosmetic"
     )
-    decimal_places_by_version: dict[VertexDecimalVersion, int] = Field(
-        description="Number of decimal places for each supported vertex decimal version"
-    )
+    token_amount_v1_decimal_places: int = Field(description="Number of decimal places for V1")
+    token_amount_v2_decimal_places: int = Field(description="Number of decimal places for V2")
     genesis_block_hash: Hex[BlockId] = Field(description="Genesis block hash in hex")
     genesis_tx1_hash: Hex[TransactionId] = Field(description="Genesis transaction 1 hash in hex")
     genesis_tx2_hash: Hex[TransactionId] = Field(description="Genesis transaction 2 hash in hex")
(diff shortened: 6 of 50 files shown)


hathorlib/hathorlib/nanocontracts/types.py
```diff
@@ -559,7 +560,7 @@ class NCFee:
     token_uid: TokenUid
     amount: int
 
-    def get_htr_value(self, settings: HathorSettings) -> int:
+    def __get_htr_value__(self, settings: HathorSettings, token_amount_version: TokenAmountVersion) -> int:
```
glevco, line 563: @jansegre since this is moving the method out of the user API, we must add it to the OCB guide.

jansegre: Done.

hathor/version_resource.py
```diff
@@ -81,7 +79,8 @@ class VersionResponse(ResponseModel):
             max_number_outputs=256,
             decimal_places=2,
             display_decimal_places=2,
-            decimal_places_by_version={VertexDecimalVersion.V1: 2},
+            token_amount_v1_decimal_places=2,
```
msbrogli, line 84: Why two separate fields instead of the previous dict?

hathor/manager.py
```diff
@@ -410,21 +411,17 @@ def stop_profiler(self, save_to: Optional[str] = None) -> None:
...
-    def get_nc_runner(self, block: Block) -> Runner:
+    def get_nc_runner(self, block: Block, *, token_amount_version: TokenAmountVersion) -> Runner:
         """Return a contract runner for a given block."""
         nc_storage_factory = self.consensus_algorithm.nc_storage_factory
         block_storage = get_block_storage_from_block(nc_storage_factory, block)
         features = Features.from_vertex(settings=self._settings, feature_service=self.feature_service, vertex=block)
         return self.runner_factory.create(
             runtime_version=features.nano_runtime_version,
             block_storage=block_storage,
+            token_amount_version=token_amount_version
```
msbrogli, line 422: Why don't we put this flag inside the runtime version?

hathor_tests/nanocontracts/blueprints/unittest.py
```diff
@@ -121,14 +122,19 @@ def _register_blueprint_contents(
 
         return self._register_blueprint_class(blueprint_class, blueprint_id)
 
-    def build_runner(self, runtime_version: NanoRuntimeVersion = NanoRuntimeVersion.V2) -> TestRunner:
+    def build_runner(
+        self,
+        runtime_version: NanoRuntimeVersion = NanoRuntimeVersion.V2,
+        token_amount_version: TokenAmountVersion = TokenAmountVersion.V1,
```
msbrogli, line 128: Same here.

hathor_tests/nanocontracts/on_chain_blueprints/test_bet.py
```diff
@@ -175,7 +176,9 @@ def initialize_contract(self) -> None:
 
         # set expected self objects:
         self.nc_id = ContractId(VertexId(nc_init_tx.hash))
-        self.runner = TestRunner.from_runner(self.manager.get_nc_runner(block))
+        self.runner = TestRunner.from_runner(
+            self.manager.get_nc_runner(block, token_amount_version=TokenAmountVersion.V1)
```
msbrogli, line 180: Isn't the block or the contract enough to say which version will be used? I don't like having to manually provide this configuration.

hathor/transaction/transaction.py
```diff
@@ -267,7 +267,7 @@ def get_related_addresses(self) -> set[str]:
     def to_json(self, decode_script: bool = False, include_metadata: bool = False) -> dict[str, Any]:
         json = super().to_json(decode_script=decode_script, include_metadata=include_metadata)
         json['tokens'] = [h.hex() for h in self.tokens]
-        json['decimal_version'] = self.get_decimal_version().value
+        json['token_amount_version'] = self.get_token_amount_version().value
```
msbrogli, line 270: Why `token_amount_version`? I guess it should be either just `amount_version` or `output_amount_version`.

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -298,6 +298,7 @@ def execute_transaction(
 
         runner = self._runner_factory.create(
             runtime_version=runtime_version,
+            token_amount_version=tx.get_token_amount_version(),
```
msbrogli, line 301: Why not do it within a new runtime version?

hathor/nanocontracts/resources/state.py
```diff
@@ -136,7 +137,10 @@ def render_GET(self, request: 'Request') -> bytes:
             block = self.manager.tx_storage.get_best_block()
 
         try:
-            runner = self.manager.get_nc_runner(block)
+            runner = self.manager.get_nc_runner(
+                block,
+                token_amount_version=TokenAmountVersion.V1,  # Dummy version, we'll set the correct one below.
```
msbrogli: This case shows that it's the wrong place to set the token amount. It should come from the contract (and its underlying blueprint), not as a parameter.

hathor/nanocontracts/blueprint_service.py
```diff
@@ -72,17 +73,28 @@ def get_on_chain_blueprint(self, blueprint_id: BlueprintId) -> OnChainBlueprint:
...
         The blueprint class could be in the catalog (first search), or it could be the tx_id of an on-chain blueprint.
         """
         from hathor.nanocontracts import OnChainBlueprint
         blueprint = self._get_blueprint(blueprint_id)
         if isinstance(blueprint, OnChainBlueprint):
-            return blueprint.get_blueprint_class()
+            return blueprint.get_blueprint_class(), blueprint.get_token_amount_version()
         else:
-            return blueprint
+            return blueprint, TokenAmountVersion.V2
```
msbrogli, line 89: I guess we should place the version inside the `Blueprint` object and let the runner use it.

msbrogli (changes requested):

hathorlib/hathorlib/nanocontracts/types.py
glevco: Thanks!

hathor/version_resource.py
glevco: This change here reflects the same change in settings, which replaced the `VERTEX_DECIMAL_PLACES` dict with two fields, `TOKEN_AMOUNT_V1_DECIMAL_PLACES` and `TOKEN_AMOUNT_V2_DECIMAL_PLACES`.

The reason for this change is that initially, when I introduced `VERTEX_DECIMAL_PLACES` in the first PR of this project, I wanted to make it a dynamic field, which would allow other networks to keep the dict with a single version (for example, the Ekvilibro network would keep only V1).

However, during implementation, I understood that the code itself cannot be dynamic over a variable number of token versions per network. Instead of inheriting this dynamic complexity in the model ([PR part 8b](https://github.com/HathorNetwork/hathor-core/pull/1722)), I decided that the code should explicitly support exactly two versions, instead. Then, hathor will use 2 and 18 decimal places for V1 and V2 respectively, and Ekvilibro will simply define both V1 and V2 to be the same, with 2 decimal places.

By doing that, the underlying model can be specific for the two existing versions, while networks can still individually set whether they're using a different V2 or not.

hathor/manager.py
glevco: They're orthogonal. The nano runtime version is used to introduce new syscalls, for example. It only changes the underlying runtime, regardless of blueprint and transaction versions. The V2 runtime is already activated, and it just added the new `get_settings` syscall. We will have both Token Amount V1 and V2 transactions/blueprints running on Nano Runtime V2, which is the current one. The Nano Runtime V1 simply doesn't exist anymore.

Maybe what you were thinking about is the _Blueprint version V2_, which was another side project, but that was not merged — nor prioritized — yet. There's an [open discussion](https://github.com/HathorNetwork/hathor-core/pull/1625#discussion_r3138647295) in its PR related to how this versioning would work, that would affect whether we could do this or not: whether Blueprints should choose versions explicitly or we would force all new blueprints to be V2.

To summarize, assuming we merge everything as I designed, we would have 3 separate and orthogonal versioning systems:

1. **Nano Runtime Version.** This one is already merged and activated. It simply allows the introduction of _new_ syscalls, and therefore is completely transparent to users. They can already create new Blueprints with the new syscall.
2. **Blueprint Version.** This is in review. It would allow _changing_ the behavior of existing syscalls. As soon as activated, new blueprints would automatically use the new version and their code would behave accordingly. Almost transparent to users, since they don't have to do anything, except be aware of the new behavior when writing new blueprints.
3. **Token Amount Version.** The current project. The only thing that defines this version is a transaction (via the bit flag). Therefore, a Blueprint's Token Amount Version comes from the transaction that created the OCB — transitively, a contract's Token Amount Version comes from its Blueprint. This one is not transparent to users, as interacting with a V1 contract requires a V1 transaction, and a V2 contract requires a V2 transaction.

hathor_tests/nanocontracts/blueprints/unittest.py
glevco: See https://github.com/HathorNetwork/hathor-core/pull/1716#discussion_r3417102166

hathor_tests/nanocontracts/on_chain_blueprints/test_bet.py
glevco: The contract, yes (from its Blueprint), however I need to correlate it with the transaction starting the call chain. It's not manually provided. A V1 contract (one that uses a V1 Blueprint) can only be interacted with from a V1 transaction, and analogous for V2.

hathor/transaction/transaction.py
glevco: I didn't go with `output_amount_version` because it's used in more places than just outputs (nano actions, fees, nano balances, etc — it talks about token amounts anywhere in the codebase, not just in outputs).

I went with `token_amount_version` instead of `amount_version` to couple it with the model ([in part 8b](https://github.com/HathorNetwork/hathor-core/pull/1717)) which I called `TokenAmount` and `TokenAmountVersion`. I didn't just use `Amount` because we already have two types in the codebase with this name, and I wanted to prevent confusion.

hathor/nanocontracts/execution/block_executor.py
glevco: Please see https://github.com/HathorNetwork/hathor-core/pull/1716#discussion_r3417102166

hathor/nanocontracts/resources/state.py
glevco: > It should come from the contract (and its underlying blueprint), not as a parameter.

It _does_ come from the contract, via its Blueprint. The correct version was set just below after the blueprint is retrieved. Please see https://github.com/HathorNetwork/hathor-core/pull/1716#discussion_r3417111742

> This case shows that it's the wrong place to set the token amount.

I agree that the dummy version is not ideal, however I disagree on the reason. It's not because the runner shouldn't receive it, but because the runner shouldn't be initialized here. It's a dependency issue, we need the version to get a runner to get a nc storage to get the blueprint, to then get the correct version and set it on the runner.

I fixed this by removing that dependency chain in https://github.com/HathorNetwork/hathor-core/pull/1716/changes/2f699ba22476d9d5bcb677cefe48d5dd3e7e3812. There's no dummy needed anymore.

hathor/nanocontracts/blueprint_service.py
glevco: I think it would be weird injecting it on the class, we would need a new field hidden behind dunder. What would be the advantage?

hathor/version_resource.py
msbrogli: I don't get the Ekvilibro case since there won't be any feature activation there. So v1 would be the only one, right?

hathor_tests/nanocontracts/on_chain_blueprints/test_bet.py
msbrogli: Since the blueprint version is immutable, it's a verification we can do to forbid v1 transactions from interacting with v2 blueprints and vice versa. The runner can do it as well and fail/skip execution if that happens.

hathor/transaction/transaction.py
msbrogli: Got it. I'm not sure I like `TokenAmount` but I don't dislike it either.

hathor/manager.py
msbrogli: But I guess runtime comprises everything inside the runner, including the storage access. Isn't it a better design to increase the runtime version to v3?

v1 = 2 decimals, no new syscalls
v2 = 2 decimals + new syscalls
v3 = 18 decimals + new syscalls

hathor/nanocontracts/blueprint_service.py
msbrogli: You're right. That would be bad design. But such a map would be needed only for builtin blueprints, right? I mean, OCBs would always have it. So we could add to the catalog or maybe simply remove builtin blueprints.

hathor/version_resource.py
glevco: > I don't get the Ekvilibro case since there won't be any feature activation there. So v1 would be the only one, right?

Exactly. However, it's not possible for the _code itself_ (the model, the functions, etc) to be agnostic to the fact that we want to support 2 versions.

Before this PR, the `VERTEX_DECIMAL_PLACES` dict meant the decimal places AND the number of versions itself were dynamic per network. Then, Ekvilibro would set it as `VERTEX_DECIMAL_PLACES = {V1: 2}`.

Now, the decimal places is still dynamic per network, however the number of versions is not. The fact that we support 2 versions is not dynamic per network, because it's something structural in the code. Then, Ekvilibro will simply set both `TOKEN_AMOUNT_V1_DECIMAL_PLACES = 2` and `TOKEN_AMOUNT_V2_DECIMAL_PLACES = 2`. This means the conversions will work transparently for Ekvilibro: since the decimal places are the same in V1 and V2, nothing is changed.

hathor/manager.py
glevco: We cannot do that because the runtime version doesn't depend on the transaction. There's no way, structurally, to create a transaction that will run in the V1 runtime — it was designed like this because it wouldn't make sense to purposefully run a new transaction on an older runtime with _less_ syscalls.

For the token amount version, it's different. We must support the two versions concurrently. Existing Blueprints (such as the Dozer pool, for example), were implemented with 2 decimal places in mind. We'll use 18 decimals internally (in the Runner, storage, etc), but:

- Syscalls must still respond those existing blueprints with 2 decimals (for example, the `get_current_balance` syscall).
- Field descriptors (`__get__`) must still return numerical values with 2 decimals.
- Method arguments must still be serialized with 2 decimals.

In other words, existing blueprints will remain using 2 decimals. To support that, we'll keep accepting V1 transactions in the network. V1 transactions can only call V1 contracts (with V1 blueprints), and V1 contracts can only call other V1 contracts. Analogously for V2. It's the only safe way to do this, in my view.

>  Isn't it a better design to increase the runtime version to v3?
>
> v1 = 2 decimals, no new syscalls
> v2 = 2 decimals + new syscalls
> v3 = 18 decimals + new syscalls

Concluding, if we do that, we wouldn't be able to continue supporting the V1 transactions with 2 decimals for the existing blueprints.

hathor_tests/nanocontracts/on_chain_blueprints/test_bet.py
glevco: Exactly! That's implemented in a future PR. But we'll still need to compare the Blueprint's version with the transaction's version, that's why the version is a new parameter on the Runner (it comes from the transaction).

This PR only threads this parameter so we can use it in that future PR to enforce that verification.

hathor/transaction/transaction.py
glevco: I propose you review [PR part 8b](https://github.com/HathorNetwork/hathor-core/pull/1717) and suggest naming improvements there, if necessary. It's where the model is defined.

hathor/nanocontracts/blueprint_service.py
glevco: Yes, OCBs just use their transaction's version. The harcoded V2 above is only necessary for builtin blueprints (that `else` branch is exactly that). I hardcoded V2 for builtin blueprints because it doesn't really matter, since we don't have any builtin blueprints.

hathor/version_resource.py
msbrogli: I still don't get why it is not possible for the code to be agnostic. It seems that, after activation, if the deserializer reads `amount = 10**16 * amount if v1 else amount`, the rest of the code would be the same.

hathor/nanocontracts/blueprint_service.py
msbrogli: Got it. Ok. Thanks!

hathor/version_resource.py
glevco: > the deserializer reads `amount = 10**16 * amount if v1 else amount`

The `normalization_factor = 10**16` must be dynamic according to the network config. Its formula is `10 ** (TOKEN_AMOUNT_V2_DECIMAL_PLACES - TOKEN_AMOUNT_V1_DECIMAL_PLACES)`. So there you already need a definition for the V2 number of decimal places. If Ekvilibro does 2 in both versions, the formula collapses to `normalization_factor = 1`.

glevco: If we use the dynamic `VERTEX_DECIMAL_PLACES` dict, we introduce a couple of complexities:

- What's the internal version used in full node calculations? Is it always the last version in the dict? Is it the one with the largest number of decimal places? Is it configurable?
- The normalization factor would be dynamic over the internal version and the vertex version. Instead of a single normalization factor, we need one per version pair.

Doing that is over generalizing for an use case we don't plan on supporting: having some networks with 1 version, others with 2 versions, and others with N versions.

The "hardcoded" approach with exactly two versions in the codebase is the simpler choice for our current needs, and the Ekvilibro special case is easily handled by having the two versions with the same number of decimal places.

jansegre (dismissed):

jansegre (approved):

hathor/version_resource.py
msbrogli: After our review meeting last week, I agree that we can move forward and change later if needed.

hathor/transaction/transaction.py
msbrogli: Ok. Thanks!

hathor/nanocontracts/resources/state.py
msbrogli: I'm not sure this approach would work since a call stack might have contracts with different versions. See this comment: https://github.com/HathorNetwork/rfcs/pull/115/changes#r3453771807

glevco: > a call stack might have contracts with different versions

That is not allowed, it fails. That's implemented in PR https://github.com/HathorNetwork/hathor-core/pull/1736

hathor_tests/nanocontracts/on_chain_blueprints/test_bet.py
msbrogli: Ok. Thanks for clarifying.

hathor/nanocontracts/resources/state.py
msbrogli: Never mind. It can't have different versions in the same stack.

hathor/manager.py
msbrogli: Got it. Thanks!

msbrogli (approved):

PR #1717, opened 2026-06-03
feat(decimals): implement UnsignedAmount and SignedAmount models [part 9]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1716

### Motivation

This PR introduces the model that is going to be used by the new token amount version upgrade in the protocol. It contains two load-bearing types, `UnsignedAmount` and `SignedAmount`, written in Rust for increased type safety and reliability.

The PR only contains changes in `htr-rs`, and does not currently affect any `hathor-core` code. However, it must be reviewed with extreme attention as the invariants encoded in those new types are fundamental for the changes in the subsequent PRs.

#### The `UnsignedAmount` type

This type represents a token amount (from an output value, a Nano action, a fee, and anywhere else a token amount is required). It's statically non-negative by wrapping the `BigUint` type from the widely adopted `num-bigint` Rust crate. It abstracts a versioned amount that can be constructed from both V1 and V2 values (with 2 and 18 decimals places, respectively). All interactions between instances of this type go through the `normalize` method, which scales V1 amounts to V2. This makes it easy and safe to handle these values in the Python side, no matter what are their versions.

#### The `SignedAmount` type

A counterpart to `UnsignedAmount`, but signed (wrapping `BigInt`). It'll be used internally by the full node to handle arithmetic between `UnsignedAmount`s when the value can be negative. Since it's internal only, it's not versioned, and always represents balances in V2.

Note that equality, comparison, and arithmetic between heterogeneous types is not supported. So it's impossible to add a `UnsignedAmount` to a `SignedAmount`, or an `int` to either of them, for example. This is not only intentional, but load-bearing. It's the only way to make sure we're not adding values of incompatible versions in a dynamic language like Python, because an invalid operation will just fail instead of silently succeeding with an incorrect value.

### Acceptance Criteria

- Add myself as a code owner of the `htr-rs` code.
- Create a new crate, `htr-lib`, which is a pure-Rust crate. The existing `htr-lib-py` will contain wrapper bindings for Python.
- Add a group of related dependencies to `htr-rs`: `num-bigint`, `num-derive`, and `num-traits`. Also, add `num-bigint` feature to `pyo3`.
- Create new Rust types `UnsignedAmount` and `SignedAmount` in `htr-lib`.
- Create new Python wrappers `PyUnsignedAmount` and `PySignedAmount` in `htr-lib-py` and update library bindings.
- Update README.md and CLAUDE.md accordingly.

### Checklist


htr-rs/crates/htr-lib/src/token_balance.rs
```diff
@@ -0,0 +1,166 @@
...
+use crate::token_amount::TokenAmount;
+use num_bigint::BigInt;
+use std::cmp::Ordering;
+
+/// Signed token balance, stored as a `BigInt` in the V2-normalized unit.
+// TODO: In the future we may profile and optimize with a variant for small ints as u64.
+#[derive(Debug)]
+pub struct TokenBalance(BigInt);
+
+impl TokenBalance {
```
msbrogli: I'd rather rename them to `SignedTokenAmount` and `UnsignedTokenAmount`. It is way more intuitive, communicates clearly and less error prone.

htr-rs/crates/htr-lib-py/htr_lib.pyi
msbrogli, line 1: Should docstrings be placed in this file?

htr-rs/crates/htr-lib/src/token_amount.rs
```diff
@@ -0,0 +1,596 @@
...
+            TokenAmount::V2 { normalized } => normalized,
+        }
+    }
+
+    pub fn as_bool(&self) -> bool {
+        self.raw() != &BigUint::ZERO
+    }
+
+    /// Lift to a [`TokenBalance`] (which is signed) holding the normalized value.
+    pub fn to_balance(&self) -> TokenBalance {
```
msbrogli: `to_signed()`?

htr-rs/crates/htr-lib/src/token_amount.rs
```diff
@@ -0,0 +1,596 @@
...
+
+/// Non-negative token amount tagged with its decimal-places version.
+///
+/// `V1` keeps both the original `raw` value and its V2-scaled `normalized` form so a V1
+/// vertex round-trips through this type without loss. `V2` only needs `normalized` because
+/// raw and normalized coincide. Ordering, equality, and arithmetic work on `normalized`,
+/// so V1 and V2 can be mixed in those operations; arithmetic always returns a V2 result.
+// TODO: In the future we may profile and optimize with a variant for small ints as u64.
+#[derive(Debug, Clone)]
+pub enum TokenAmount {
```
msbrogli: Why are all these necessary? I guess a cleaner approach is to have only one bigint and just adjust the serializer and deserializer to handle v1 and v2. Internally, everything would be just SignedAmount and UnsignedAmount.

msbrogli (changes requested):

htr-rs/crates/htr-lib/src/token_balance.rs
glevco: Renamed to `SignedAmount` and `UnsignedAmount` in https://github.com/HathorNetwork/hathor-core/pull/1717/commits/532ab467f9021b76d2d52f8931dddc2ed0eeabc0, as discussed in our call

htr-rs/crates/htr-lib/src/token_amount.rs
glevco: Renamed in https://github.com/HathorNetwork/hathor-core/pull/1717/commits/532ab467f9021b76d2d52f8931dddc2ed0eeabc0

htr-rs/crates/htr-lib/src/token_amount.rs
glevco: As discussed in our call, I think we have to keep this.

htr-rs/crates/htr-lib-py/htr_lib.pyi
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1717/commits/3bd9181d2a5ccd0d518ad60878cd076c63ced02e

jansegre: I don't mind, but is this usual or even loaded anywhere?

jansegre (dismissed):

htr-rs/crates/htr-lib-py/htr_lib.pyi
glevco: Yes, IDEs for example extract it so a tooltip shows the docstring when hovering method names in call sites. I'm sure there are other places too.

msbrogli (approved):

htr-rs/crates/htr-lib/src/token_balance.rs
msbrogli: Thanks!

htr-rs/crates/htr-lib-py/htr_lib.pyi
msbrogli: Thanks!

htr-rs/crates/htr-lib/src/token_amount.rs
msbrogli: Thanks!

htr-rs/crates/htr-lib/src/token_amount.rs
msbrogli: Agreed. Thanks!

jansegre (approved):

PR #1718, opened 2026-06-03
refactor(decimals): swap int for TokenAmount/TokenBalance on the static surface [part 10]

glevco:
### Motivation

The previous PR introduced the new token amount model types, `UnsignedAmount` and `SignedAmount`. In preparation to use those types across the codebase, this PR creates types with the same names that are just aliases to `int`. This means we can change all function, method, and attribute type signatures to the new types, without actually changing any logic.

To be clear, this PR **does not** use the types introduced in the previous PR. It's simply a typing trick to allows us to remove some lines from the diff of the main PR, which is coming next. What must be reviewed with attention is the deliberate choice of which type to use in each situation. `UnsignedAmount` is used whenever an value cannot be negative, and `SignedAmount` is used when it can.

### Acceptance Criteria

- Add temporary `int` aliases on `hathorlib`: `UnsignedAmount` and `SignedAmount`.
- Change all usages of `int` that represent token amounts (unsigned) to the `UnsignedAmount` alias.
- Change all usages of `int` that represent token balances (signed) to the `SignedAmount` alias.

### Checklist

### DIFF
--- a/hathor/transaction/exceptions.py
+++ b/hathor/transaction/exceptions.py
@@ -74,7 +74,7 @@ from hathorlib.exceptions import (  # noqa: F401
     VerifyFailed,
     WeightError,
 )
-from hathorlib.token_amount import SignedAmount
+from hathorlib.token_amount import UnsignedAmount
 
 
 class ForbiddenMint(InputOutputMismatch):
@@ -82,8 +82,8 @@ class ForbiddenMint(InputOutputMismatch):
 
     from hathor.types import TokenUid
 
-    def __init__(self, amount: SignedAmount, token_uid: TokenUid) -> None:
-        super().__init__(f'{-amount} {token_uid.hex()} tokens minted, but there is no mint authority input')
+    def __init__(self, amount: UnsignedAmount, token_uid: TokenUid) -> None:
+        super().__init__(f'{amount} {token_uid.hex()} tokens minted, but there is no mint authority input')
 
 
 class ForbiddenMelt(InputOutputMismatch):
@@ -95,8 +95,8 @@ class ForbiddenMelt(InputOutputMismatch):
         super().__init__(msg)
 
     @classmethod
-    def from_token(cls, amount: SignedAmount, token_uid: TokenUid) -> 'ForbiddenMelt':
-        return cls(f'{-amount} {token_uid.hex()} tokens melted, but there is no melt authority input')
+    def from_token(cls, amount: UnsignedAmount, token_uid: TokenUid) -> 'ForbiddenMelt':
+        return cls(f'{amount} {token_uid.hex()} tokens melted, but there is no melt authority input')
 
 
 class InvalidRangeProofError(TxValidationError):
--- a/hathor/transaction/exceptions.py
+++ b/hathor/transaction/exceptions.py
@@ -74,6 +74,7 @@ from hathorlib.exceptions import (  # noqa: F401
     VerifyFailed,
     WeightError,
 )
+from hathorlib.token_amount import SignedAmount
 
 
 class ForbiddenMint(InputOutputMismatch):
@@ -81,9 +82,8 @@ class ForbiddenMint(InputOutputMismatch):
 
     from hathor.types import TokenUid
 
-    def __init__(self, amount: int, token_uid: TokenUid) -> None:
-        super().__init__('{} {} tokens minted, but there is no mint authority input'.format(
-            (-1) * amount, token_uid.hex()))
+    def __init__(self, amount: SignedAmount, token_uid: TokenUid) -> None:
+        super().__init__(f'{-amount} {token_uid.hex()} tokens minted, but there is no mint authority input')
 
 
 class ForbiddenMelt(InputOutputMismatch):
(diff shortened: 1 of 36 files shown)


hathor/transaction/exceptions.py
```diff
@@ -74,16 +74,16 @@
...
 
 class ForbiddenMint(InputOutputMismatch):
     """Tokens were minted without authority inputs"""
 
     from hathor.types import TokenUid
 
-    def __init__(self, amount: int, token_uid: TokenUid) -> None:
-        super().__init__('{} {} tokens minted, but there is no mint authority input'.format(
-            (-1) * amount, token_uid.hex()))
+    def __init__(self, amount: SignedAmount, token_uid: TokenUid) -> None:
```
msbrogli: As it is already explicit a mint error, why not use UnsignedAmount and let the caller adjust the signal when needed?

hathor/transaction/exceptions.py
```diff
@@ -95,9 +95,8 @@ def __init__(self, msg: str) -> None:
         super().__init__(msg)
 
     @classmethod
-    def from_token(cls, amount: int, token_uid: TokenUid) -> 'ForbiddenMelt':
-        return cls('{} {} tokens melted, but there is no melt authority input'.format(
-            (-1) * amount, token_uid.hex()))
+    def from_token(cls, amount: SignedAmount, token_uid: TokenUid) -> 'ForbiddenMelt':
```
msbrogli: Same here.

hathorlib/hathorlib/token_amount.py
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TypeAlias
+
+# Temporary aliases.
+SignedAmount: TypeAlias = int
```
msbrogli, line 18: Why not `NewType`? It would allow mypy to flag incorrect types (and it will be an actual new type soon, I guess).

msbrogli (dismissed):

jansegre (dismissed):

hathor/transaction/exceptions.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1718/commits/c9286246a0c178c8c50803f72fcf859d2b949aa8

hathor/transaction/exceptions.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1718/commits/c9286246a0c178c8c50803f72fcf859d2b949aa8

hathorlib/hathorlib/token_amount.py
glevco: Because operations such as addition would return an `int`, not the `NewType`, so I would need to manually wrap them. That would happen in multiple places that would have to be changed again and increase the diff.

hathor/transaction/exceptions.py
msbrogli: Thanks!

hathor/transaction/exceptions.py
msbrogli: Thanks!

hathorlib/hathorlib/token_amount.py
msbrogli: Make sense. As most operations are just add and sub, a simple class with `__add__` and `__sub__` methods returning wrapping object might be enough to fix most issues, right?

glevco: Probably. In fact I do that in PR https://github.com/HathorNetwork/hathor-core/pull/1742, but _only_ [for tests](https://github.com/HathorNetwork/hathor-core/blob/483fb4c7cb8cba7513adc2dd59c5431b064e47e3/hathor_tests/token_amount.py), in a shim with all `htr_lib` API. That allows me to update all test fixtures in a standalone PR, which before were the bulk of the diff in the main PR.

The reason I did it only for tests and kept the aliases in production is that I did not want to change the runtime type of production to anything that is not the hardened Rust types, since I rely the safety of this update on the static analysis of Rust.

The current plan is the following:

1. `int` aliases (this PR) in the whole codebase (productions + tests). Correctness is guaranteed by the fact that there's no runtime change at all. The review should only check whether something signed was named as unsigned or vice-versa.
2. New type with full `htr_lib` API, but _only_ for tests (PRs https://github.com/HathorNetwork/hathor-core/pull/1742 and https://github.com/HathorNetwork/hathor-core/pull/1743). Correctness is guaranteed by the fact that no production code is touched. The review should only check whether any tests were incorrectly removed, or actual fixture numbers changed, etc. Large diff, but mostly brainless review.
3. Alias type flip, from `int` to the actual `htr_lib` types (main PR, https://github.com/HathorNetwork/hathor-core/pull/1714). Correctness is guaranteed by the rules on the Rust side: operations are only defined for equal types, conversions are necessarily explicit, etc.

The idea of introducing the new type here, not only for tests but for production, has one advantage and one disadvantage, in my opinion:

- The advantage is what you said, we get mypy to type check early in this PR. However this doesn't really buy us anything, because according to the plan this PR doesn't touch runtime, and type checking will be enforced in the last step of the plan anyways, by mypy AND the Rust compiler.
- The disadvantage is that we introduce a new intermediate _runtime_ type in production, without the strong safety guarantees (even though it's a pretty simple type). If for example we end up having to make a new hathor-core release at this point, we would release a version with a runtime type that is unnecessary.

I could do it if you think it's worth it, but I don't think it is.

jansegre: Since the main objective of the PR is to minimize the diff later, and not an actual replacement of the int/amount class. I don't really care if it's an alias or a new-type. We'll replace the actual class later anyway, right?

glevco: Yes

msbrogli: Ok. Agreed. Thanks!

msbrogli (approved):

jansegre (approved):

PR #1725, opened 2026-06-09
feat(decimal): napi binding for htr-lib

r4mmer:
### Motivation

Add JS bindings for the new rust htr-lib

### Acceptance Criteria

- Create JS bindings using napi-rs

### Checklist


htr-rs/crates/htr-lib-napi/src/signed_amount.rs
glevco, line 1: Claude found a bug that apparently exists in all ops in both types, verified by this test:

```js
import test from 'ava'
import { UnsignedAmount, SignedAmount } from '../index.js'

test.before(() => {
  UnsignedAmount.setNormalizationFactor(2, 18)
})

// Passing the WRONG wrapper class (a real napi-wrapped object, not a plain number)
// must throw a catchable error — not silently return a garbage result or corrupt
// memory. napi's bare `&T` recovery only checks that the arg carries *some* native
// wrap, not that it is *this* class.
test('SignedAmount.eq rejects an UnsignedAmount argument (wrong wrapper class)', (t) => {
  const s = new SignedAmount(5n)
  const u = UnsignedAmount.fromV2(5n)
  // @ts-expect-error intentionally passing the wrong wrapper class
  t.throws(() => s.eq(u))
})

test('UnsignedAmount.eq rejects a SignedAmount argument (wrong wrapper class)', (t) => {
  const u = UnsignedAmount.fromV2(5n)
  const s = new SignedAmount(5n)
  // @ts-expect-error intentionally passing the wrong wrapper class
  t.throws(() => u.eq(s))
})
```

The method signatures accept any wrapped runtime object as the correct Rust type without checking, which leads to memory confusion and therefore is UB. The fix would be to add `#[napi(strict)]` on all of them.

htr-rs/crates/htr-lib-napi/apply-operators.cjs
```diff
@@ -0,0 +1,19 @@
...
+// return the underlying value as a bigint so ordering compares with full precision.
+// Note: == / === remain reference identity (use .eq()); + / - coerce to bigint (use .add/.sub).
+module.exports = function applyOperators(binding) {
+  binding.UnsignedAmount.prototype[Symbol.toPrimitive] = function (hint) {
+    return hint === 'string' ? this.toString() : this.normalized()
+  }
+  binding.SignedAmount.prototype[Symbol.toPrimitive] = function (hint) {
+    return hint === 'string' ? this.toString() : this.raw()
+  }
+}
```
glevco, line 22: Why not throw here instead of coerce? This will make errors explicit in JS which will help you convert types where necessary, and force you to use the comparison methods instead of operators. It would be analogous to what we do in the Python bindings.

htr-rs/crates/htr-lib-napi/PUBLISHING.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Versioning
+
+```bash
+npm version <patch|minor|major>   # the `version` script runs `napi version`, which keeps the
+                                  # per-platform npm/<target> package versions in sync
+```
+
+Then commit, and tag as `htr-lib-napi-v<version>` (e.g. `htr-lib-napi-v0.1.0`) to trigger the CI
+publish (see below).
```
glevco: I don't think the CI is configured to trigger it automatically in `.github/workflows/htr-rs-napi-publish.yml`

.github/workflows/htr-rs-napi-publish.yml
```diff
@@ -0,0 +1,168 @@
...
+        type: string
+        required: true
+      dry-run:
+        description: Build and assemble without publishing
+        type: boolean
+        default: true
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
glevco: Should we do that for a publishing CI?

htr-rs/justfile
```diff
@@ -5,8 +5,8 @@ export RUSTFLAGS := "-D warnings"
 help:
   @just --list
 
-# Run the full local check suite: check, fmt-check, clippy, test, sort, audit.
-all: check fmt-check clippy test sort audit
+# Run the full local check suite: check, fmt-check, clippy, test, sort, audit, node-test.
+all: check fmt-check clippy test sort audit node-test
```
glevco: I think we should leave `node-test` out of `just all`, so we don't necessarily need a node environment when developing pure Rust in other crates of this workspace. Maybe create a `just all-js` or similar?

glevco (changes requested):

.github/workflows/htr-rs.yml
```diff
@@ -143,11 +143,47 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: cargo-llvm-cov,cargo-nextest
+      # The napi UnsignedAmount/SignedAmount wrappers are exercised by the ava suite (the
+      # node-test job), not by cargo nextest, so exclude them from the Rust lcov instead of
+      # letting them read as ~0%. convert.rs keeps its real Rust coverage; a future htr-rs-napi
+      # codecov flag can measure these from an ava-driven instrumented run.
```
glevco, line 152: Couldn't we call JS from Rust to exercise those paths instead?

htr-rs/crates/htr-lib-napi/src/unsigned_amount.rs
```diff
@@ -0,0 +1,246 @@
...
+        Ok(UnsignedAmount::from_inner(InnerUnsignedAmount::from_v1(
+            try_napi_to_biguint(&amount)?,
+        )))
+    }
+
+    #[napi]
+    pub fn from_v2(amount: BigInt) -> napi::Result<UnsignedAmount> {
+        Ok(UnsignedAmount::from_inner(InnerUnsignedAmount::from_v2(
+            try_napi_to_biguint(&amount)?,
+        )))
```
glevco: Why no catch_unwind in this case?

htr-rs/crates/htr-lib-napi/src/convert.rs
```diff
@@ -0,0 +1,121 @@
+//! Conversions between napi's `BigInt` and `num-bigint`'s `BigUint`/`BigInt`.
+//!
+//! napi's `BigInt` stores magnitude as little-endian `u64` limbs in `words` plus a
+//! `sign_bit`. `num_bigint::BigUint::to_u64_digits()` yields the same little-endian
+//! `u64` limbs, so the magnitude maps across directly.
+
+use napi::bindgen_prelude::BigInt;
+use num_bigint::{BigInt as RsBigInt, BigUint, Sign};
```
glevco: Nit: it's confusing to use `BigInt` from napi and `BigUint` from num-bigint. Maybe you should alias the napi one, instead. Optional.

htr-rs/crates/htr-lib-napi/src/convert.rs
```diff
@@ -0,0 +1,121 @@
...
+/// napi `BigInt` magnitude -> `BigUint`. Ignores `sign_bit`; callers that must reject
+/// negatives use [`try_napi_to_biguint`].
+pub fn napi_to_biguint(value: &BigInt) -> BigUint {
+    let digits: Vec<u32> = value
+        .words
+        .iter()
+        .flat_map(|w| [*w as u32, (*w >> 32) as u32])
+        .collect();
+    BigUint::from_slice(&digits)
+}
```
glevco: Maybe call it `napi_magnitude_to_biguint` to make it clear that the sign is ignored. Otherwise I think I would expect this would panic on negative, instead of ignoring.

htr-rs/crates/htr-lib-napi/src/convert.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// napi `BigInt` -> `BigUint`, rejecting negative inputs with a JS-throwable error.
+pub fn try_napi_to_biguint(value: &BigInt) -> napi::Result<BigUint> {
+    let is_zero = value.words.iter().all(|w| *w == 0);
```
glevco: Why not just this?

```suggestion
    let is_zero = value == BigInt::from(0u64);
```

glevco: Same in `napi_to_rsbigint`

glevco (changes requested):

htr-rs/crates/htr-lib-napi/src/signed_amount.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn op_neg(&self) -> SignedAmount {
+        SignedAmount::from_inner(-&self.inner)
+    }
+
+    /// Identity, mirroring Python's `SignedAmount.__pos__`. Returns an equal
+    /// `SignedAmount` (value semantics); not guaranteed to be the same reference.
+    #[napi(js_name = "pos")]
+    pub fn op_pos(&self) -> SignedAmount {
+        SignedAmount::from_inner(InnerSignedAmount::new(self.inner.raw().clone()))
+    }
```
glevco: Is it not possible to return a reference to self?

glevco (changes requested):

htr-rs/crates/htr-lib-napi/apply-operators.cjs
r4mmer: This has 2 intentions.
When printing to the stdout via console.log or other string operation it uses `.toString()` so we can use it safely.

When using on any other operator it is converted to bigint, so we can make `value > 10n` instead of `value.op_gt(new SignedAmount(10n))`

If we throw here we cannot use any comparison operators.
Do you think it would be the better option?

.github/workflows/htr-rs.yml
r4mmer: The test works by testing the built bindings (`*.node` file) with JS tests using ava.
Meaning that during the test there is no invocation of rust code that can be tracked to count as coverage.

I am trying to create a coverage mapping so we can run the ava tests against an addon with LLVM instrumentation so we can get the actual rust lines coverage report then merge with the current rust coverage (similar to what we do in the wallet-lib integration tests), but this is not the priority.

We can also test the rust bindings in rust to achieve coverage, but it would not mean much, the test on JS means we are behaving as intended, like an integration test.

htr-rs/crates/htr-lib-napi/src/unsigned_amount.rs
r4mmer: The result unwrapping is always `Ok` because `from_v2` never reads the normalization factor, so it does not panic.
No panic, no need for `catch_unwind`

htr-rs/crates/htr-lib-napi/src/signed_amount.rs
r4mmer: Will update to use strict mode, but I intentionally used an `Either<&SignedAmount, &UnsignedAmount>` so some methods can accept both, do you think we should not?

htr-rs/crates/htr-lib-napi/PUBLISHING.md
r4mmer: updated the file

.github/workflows/htr-rs-napi-publish.yml
r4mmer: this was to prevent multiple publish actions running for the same `ref`.
But I updated to not cancel the in-progress one when dispatching a new one.

htr-rs/justfile
r4mmer: Done

htr-rs/crates/htr-lib-napi/src/convert.rs
r4mmer: I aliased the napi one so BigInt and BigUInt are rust only to avoid confusion.

htr-rs/crates/htr-lib-napi/src/convert.rs
r4mmer: Done

htr-rs/crates/htr-lib-napi/src/convert.rs
r4mmer: Good call on the readability, I extracted this behavior to a `napi_magnitude_is_zero` helper and used it on both.
I kept the helper as a magnitude only check rather than ` == BigInt::from(0u64)` because napi's `BigInt::PartialEq` also compares the sign_bit, making it possible to have positive and negative zero values which return as not equal, but if we compare the magnitude only we can avoid this.

htr-rs/crates/htr-lib-napi/src/signed_amount.rs
r4mmer: Done

.github/workflows/htr-rs-napi-publish.yml
```diff
@@ -20,8 +20,10 @@ on:  # yamllint disable-line rule:truthy
         default: true
 
 concurrency:
+  # Never cancel an in-progress publish: a half-cancelled run could leave some per-platform
+  # packages published and the main package not.
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
glevco: We could just remove this block, right?

htr-rs/crates/htr-lib-napi/src/signed_amount.rs
glevco: I'm fine with `Either`. The problem was the type confusion and UB.

htr-rs/crates/htr-lib-napi/apply-operators.cjs
glevco: Ok, it's fine for now. You can change it back if it bites you during usage in JS.

htr-rs/crates/htr-lib-napi/src/unsigned_amount.rs
glevco: We don't have known explicit panic in this function (like setting the normalization factor), however the code can always panic, right? Same for all other methods — why not simply use `catch_unwind` in all?

.github/workflows/htr-rs-napi-publish.yml
r4mmer: done

htr-rs/crates/htr-lib-napi/src/unsigned_amount.rs
r4mmer: Done

glevco (dismissed):

glevco (dismissed):

htr-rs/crates/htr-lib-napi/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "dependencies": {
+    "@napi-rs/wasm-runtime": "^1"
+  },
+  "devDependencies": {
+    "@napi-rs/cli": "^3",
+    "@oxc-node/core": "^0.1.0",
+    "@types/node": "^25",
+    "ava": "^8",
+    "typescript": "^5"
+  },
```
pedroferreira1, line 49: We should pin the versions

pedroferreira1 (dismissed):

glevco (dismissed):

glevco (approved):

pedroferreira1 (approved):

jansegre (approved):

PR #1728, opened 2026-06-11
fix(metrics): use DAA-version-aware avg_time in hash_rate calculation

luislhl:
## Summary

- `Metrics.calculate_new_hashrate()` used a fixed `avg_time_between_blocks=30s` (from settings at construction time) to back-calculate hash rate from block weight. After `REDUCE_DAA_TARGET` activates, new blocks target 7.5s and carry proportionally lower weights — causing the formula to report 4× less than the actual hash rate.
- Pass `daa_factory` to `Metrics` so `calculate_new_hashrate` uses the correct `avg_time_between_blocks` per block version (V1=30s, V2=7.5s) via `daa_factory.create_from_block(block).avg_time_between_blocks`.
- Apply the same fix to `mining_info.py` which hardcoded `log(30, 2)` regardless of block version.

## Root cause

The formula is `hash_rate = 2^(block.weight - log2(avg_time_between_blocks))`, derived from `weight = log2(avg_time) + log2(hash_rate)`. V2 blocks have weights 2 units lower (log2(30/7.5) = 2), but the metrics code still divided by log2(30), giving `hash_rate_reported = actual / 4`.

## Test plan

- [x] All existing metrics tests pass (`8 passed`)
- [x] All mining_info and reduce_daa tests pass (`44 passed`)
- [x] mypy: no issues on modified files
- [x] Validated on a live mainnet node: fixed node reports `1.807e16` H/s vs `4.518e15` H/s on broken nodes — exactly 4× correction as expected

### DIFF
--- a/hathor/manager.py
+++ b/hathor/manager.py
@@ -219,6 +219,7 @@ class HathorManager:
             tx_storage=self.tx_storage,
             reactor=self.reactor,
             websocket_factory=self.websocket_factory,
+            daa_factory=self.daa_factory,
         )
 
         self.wallet = wallet
--- a/hathor/metrics.py
+++ b/hathor/metrics.py
@@ -28,6 +28,7 @@ from hathor.transaction.block import Block
 from hathor.transaction.storage import TransactionRocksDBStorage, TransactionStorage
 
 if TYPE_CHECKING:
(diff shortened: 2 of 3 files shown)


luislhl:
Replaced by https://github.com/HathorNetwork/hathor-core/pull/1735

PR #1730, opened 2026-06-12
feat(shielded): mint/melt headers (serialization + verification)

msbrogli:
### Motivation

Shielded transactions declare per-token supply changes (mint/melt) publicly so those amounts can enter the augmented homomorphic balance equation as unblinded scalars (RFC `0000-shielded-outputs-mint-melt`). Header IDs `0x14`/`0x15` were reserved by #1702, which left placeholder notes in `hathor/transaction/headers/__init__.py` and `hathorlib/headers/__init__.py` for this follow-up.

This PR adds the `MintHeader` (`0x14`) and `MeltHeader` (`0x15`) data model + wire (de)serialization **and** the full structural verification of those headers. Everything is gated behind `ENABLE_SHIELDED_TRANSACTIONS` (DISABLED by default), so no existing tx path changes.

### Acceptance Criteria

**Data model & serialization**
- `MintHeader`/`MeltHeader`/`MintMeltEntry` in hathorlib, driven through the `Serializer`/`Deserializer` framework (mirroring `ShieldedOutputsHeader`). Wire format: `header_id(1) | num_entries(1) | (token_index(1) | amount(8 BE))…`.
- Entry bounds (≤16 entries, `token_index ∈ [1,16]`, `amount ∈ [1, 2**64)`, unique within a header) enforced at construction and at deserialize-time; malformed input raises `InvalidMintMeltHeaderError`.
- hathor-core parses/serializes via dedicated `(de)serialize_mint_header`/`_melt_header` functions with explicit `case` arms, gated by `ENABLE_SHIELDED_TRANSACTIONS`.

**Verification (Rules M1–M4)**
- **M1** `verify_mint_melt_requires_shielded` — a Mint/Melt header is valid only with a `ShieldedOutputsHeader` or `UnshieldBalanceHeader`.
- **M3 + bounds** `verify_mint_melt_headers_well_formed` — `token_index ≤ len(tx.tokens)` and a token cannot appear in both headers.
- nano compat `verify_mint_melt_nano_compatibility` — a token cannot be supply-changed via both a NanoHeader action and a Mint/Melt header.
- **M2** `verify_mint_melt_authority_inputs` — each minted/melted token needs a matching mint/melt **authority input** (TCT exempt for its new token).
- **M4** `verify_no_undeclared_mint_melt` — transparent surplus/deficit on a non-native token in a shielded tx must be declared in a header.
- TCT `verify_minted_tokens` — a shielded token-creation tx declares initial supply via a single MintHeader entry for `token_index=1`, and cannot melt the new token.
- `VertexVerifier.get_allowed_headers`, when `shielded_transactions` is active, admits Mint **and** Melt headers (plus shielded) on `REGULAR_TRANSACTION`, and on `TOKEN_CREATION_TRANSACTION` admits the `MintHeader` (plus shielded) but **not** a `MeltHeader` — a creation tx mints, it does not melt.
- tx accessors `has/get_mint_header`, `has/get_melt_header`.

Tests cover wire-format round-trips, malformed input, the feature gate (parser + verifier), and rules M1–M4 + the TCT rule. ruff/isort/flake8/mypy clean; hathorlib suite green.

### Out of scope (shielded-balance PR)

Crypto-bound pieces that cannot run without the shielded crypto layer: `verify_shielded_balance` and the augmented balance equation, the full `is_shielded()` balance dispatch, and the consensus `_shielded_activation_rule` reorg rule.

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged (gated OFF by default; no consensus behavior change while the feature is disabled).


hathor/transaction/vertex_parser/_headers.py
```diff
@@ -115,13 +131,15 @@ def serialize_header(
                 serialize_unshield_balance_header,
             )
             serialize_unshield_balance_header(serializer, header)
+        case MintHeader() | MeltHeader():
+            serializer.write_bytes(header.serialize())
```
msbrogli: Why not follow the same pattern as the other headers?

hathor/transaction/vertex_parser/_headers.py
```diff
@@ -134,5 +152,7 @@ def get_header_sighash_bytes(header: AnyVertexHeader, *, decimal_version: Vertex
             serializer = Serializer.build_bytes_serializer()
             serialize_fee_header(serializer, header, decimal_version=decimal_version)
             return bytes(serializer.finalize())
+        case MintHeader() | MeltHeader():
+            return header.get_sighash_bytes()
```
msbrogli: Same here.

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def serialize_entries(entries: list[MintMeltEntry]) -> bytes:
+    """Serialize: num_entries(1) | entries..."""
+    parts: list[bytes] = []
+    parts.append(int_to_bytes(len(entries), 1))
+    for entry in entries:
+        parts.append(int_to_bytes(entry.token_index, 1))
+        parts.append(struct.pack('!Q', entry.amount))
+    return b''.join(parts)
```
msbrogli: It should use the newest serialization structure. See how other stuff is serialized/deserialized.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli, line 1: This PR is missing the verifier that a MintHeader or MeltHeader can only exist when a transaction has shielded inputs or outputs and also has a mint or melt authority. Other than that, the same token id cannot exist in both a MintHeader and a MeltHeader when both are present.

hathor/transaction/vertex_parser/_headers.py
msbrogli: Done in c9004759 — added `serialize_mint_header`/`serialize_melt_header` in `_mint_melt_header.py` and replaced the generic fallback with explicit `case MintHeader()` / `case MeltHeader()` arms calling them, matching `_unshield_balance_header.py`.

hathor/transaction/vertex_parser/_headers.py
msbrogli: Done — same pattern applied to `get_header_sighash_bytes`: explicit `case MintHeader()`/`case MeltHeader()` arms that build a serializer and call `serialize_mint_header`/`serialize_melt_header`, like the nano/fee arms.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Done — rewrote the hathorlib (de)serialization on the `Serializer`/`Deserializer` framework, mirroring `ShieldedOutputsHeader` (`build_bytes_serializer`, `write_bytes(int_to_bytes(...))`, `read_byte`/`read_bytes`/`read_all`). No more raw `struct.pack`/offset indexing. Wire format is unchanged (`int_to_bytes` is big-endian).

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Added the full mint/melt verification in c9004759, no crypto required:
- **M1** `verify_mint_melt_requires_shielded` — header only valid with a ShieldedOutputs/UnshieldBalance header.
- **M2** `verify_mint_melt_authority_inputs` — each minted/melted token needs a matching mint/melt authority input (TCT exempt for its new token).
- **M3** `verify_mint_melt_headers_well_formed` — token_index bounds + a token cannot appear in both headers.
- **M4** `verify_no_undeclared_mint_melt` — undeclared transparent surplus/deficit rejected.
- plus nano-compat and the TCT initial-supply rule.

Wired into `verify_without_storage` (M1/M3/nano) and `_verify_tx` (M2/M4). The crypto-bound `verify_shielded_balance` and the consensus reorg rule stay with the shielded-balance PR (they can't run without the crypto layer, which isn't on master yet). See the updated PR description.

hathor/verification/transaction_verifier.py
```diff
@@ -350,6 +352,196 @@ def _check_token_permissions(token_uid: TokenUid, token_info: TokenInfo) -> None
...
+        if not tx.has_mint_header() and not tx.has_melt_header():
+            return
+        if tx.has_shielded_outputs() or tx.has_unshield_balance_header():
+            return
+        raise ShieldedMintMeltForbiddenError(
+            'MintHeader/MeltHeader requires the transaction to carry a '
+            'ShieldedOutputsHeader or UnshieldBalanceHeader (Rule M1)'
+        )
+
+    def verify_mint_melt_nano_compatibility(self, tx: Transaction) -> None:
```
msbrogli: Not sure we need this verification.

hathor/verification/vertex_verifier.py
```diff
@@ -219,6 +221,13 @@ def get_allowed_headers(self, vertex: BaseTransaction, params: VerificationParam
                     allowed_headers.add(NanoHeader)
                 if params.features.fee_tokens:
                     allowed_headers.add(FeeHeader)
+                # A shielded TCT carries shielded outputs of the new token plus a
+                # MintHeader declaring its initial supply.
+                if params.features.shielded_transactions:
+                    allowed_headers.add(ShieldedOutputsHeader)
+                    allowed_headers.add(UnshieldBalanceHeader)
+                    allowed_headers.add(MintHeader)
+                    allowed_headers.add(MeltHeader)
```
msbrogli: Should we remove `MeltHeader` from this TCT?

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+    uniqueness check local to the header.
+    """
+
+    _HEADER_ID: ClassVar[bytes] = b''
+    _HEADER_NAME: ClassVar[str] = ''
+
+    entries: list[MintMeltEntry] = field(default_factory=list)
+
+    @classmethod
+    def deserialize(cls, tx: BaseTransaction, buf: bytes) -> tuple[_MintMeltHeaderBase, bytes]:
```
msbrogli, line 132: Why this method? Can't we just remove it?

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+            raise
+        except (IndexError, struct.error) as e:
+            # OutOfDataError (truncation) is a struct.error subclass, caught here.
+            raise ValueError(f'malformed {cls._HEADER_NAME}: {e}') from e
+
+        # Whatever follows this header (subsequent headers) is the unconsumed leftover.
+        remaining = bytes(deserializer.read_all())
+        return cls(entries=entries), remaining
+
+    def serialize(self) -> bytes:
```
msbrogli, line 135: Why this method? Can't we just remove it?

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+        """Serialize: header_id(1) | num_entries(1) | entries..."""
+        from hathorlib.serialization import Serializer
+
+        serializer = Serializer.build_bytes_serializer()
+        serializer.write_bytes(self._HEADER_ID)
+        serialize_entries(serializer, self.entries)
+        return bytes(serializer.finalize())
+
+    def get_sighash_bytes(self) -> bytes:
```
msbrogli, line 138: Why this method? Can't we just remove it?

hathor/verification/vertex_verifier.py
msbrogli: Good call — removed in d474e752. A creation tx mints its new token and never melts, so a TCT no longer admits a MeltHeader (it's rejected at the header gate). Added a test asserting MintHeader is admitted on a TCT but MeltHeader is not.

hathor/verification/transaction_verifier.py
msbrogli: This is the guard you asked me to add alongside the other mint/melt checks. Its purpose: a single token must not have its supply changed through **two** channels in the same tx — a NanoHeader action **and** a Mint/Melt header. If both declared the same token, the public scalar fed into the augmented balance equation would be ambiguous/double-counted. Cross-token combinations are still allowed.

That said, if you'd rather this interaction be handled in the nano/balance PR (where the augmented equation actually lives) I'm happy to drop `verify_mint_melt_nano_compatibility` from this PR — just say the word and I'll remove it + its test.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: These three implement the abstract `VertexBaseHeader` interface — `deserialize`, `serialize` and `get_sighash_bytes` are all `@abstractmethod` on the base, so a concrete header *must* provide them (otherwise the dataclass can't be instantiated). They're the client-side (de)serialization path used by hathorlib's `VertexParser` (e.g. `VertexParser.get_header_parser(0x14).deserialize(tx, buf)`, exercised in `test_basic.py::test_tx_with_mint_melt_headers`), exactly like `UnshieldBalanceHeader` / `ShieldedOutputsHeader`. So they can't just be removed without breaking client parsing.

They live once on the shared `_MintMeltHeaderBase` (not duplicated per header). If your concern is the broader dual path — hathorlib class methods *plus* hathor-core's free functions in `_mint_melt_header.py` — that's the established pattern for every shielded header, not specific to mint/melt; happy to discuss reworking it across all of them if you'd prefer.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Same as the `deserialize` thread above: `serialize` is an `@abstractmethod` of `VertexBaseHeader` and is used by hathorlib's `VertexParser` to re-serialize the header client-side. Removing it makes the dataclass abstract (non-instantiable). Mirrors `UnshieldBalanceHeader.serialize`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Same here: `get_sighash_bytes` is an `@abstractmethod` of `VertexBaseHeader`. It returns the full serialization so any mutation to the declared mint/melt amounts invalidates the tx signatures. Required and used by the signing/verification path, like every other header.

hathor/transaction/vertex_parser/_headers.py
```diff
@@ -134,5 +156,15 @@ def get_header_sighash_bytes(header: AnyVertexHeader, *, decimal_version: Vertex
...
+        case MintHeader():
+            from hathor.transaction.vertex_parser._mint_melt_header import serialize_mint_header
+            serializer = Serializer.build_bytes_serializer()
+            serialize_mint_header(serializer, header)
+            return bytes(serializer.finalize())
+        case MeltHeader():
+            from hathor.transaction.vertex_parser._mint_melt_header import serialize_melt_header
+            serializer = Serializer.build_bytes_serializer()
+            serialize_melt_header(serializer, header)
+            return bytes(serializer.finalize())
```
glevco, line 181: `ShieldedOutputsHeader` and `UnshieldBalanceHeader` are not here in `get_header_sighash_bytes`. Why is that?

hathor/verification/transaction_verifier.py
```diff
@@ -350,6 +353,206 @@ def _check_token_permissions(token_uid: TokenUid, token_info: TokenInfo) -> None
         if token_info.has_been_minted() and not token_info.can_mint:
             raise ForbiddenMint(token_info.amount, token_uid)
 
+    def verify_mint_melt_basic(self, tx: Transaction) -> None:
```
glevco: I think those should be moved to their own verifier, analogous to the `NanoHeaderVerifier`. I asked Luis Felipe the same for the other new headers.

hathor/verification/token_creation_transaction_verifier.py
```diff
@@ -32,10 +32,37 @@ def verify_minted_tokens(self, tx: TokenCreationTransaction, token_dict: dict[To
...
+                    'shielded token creation transaction must have exactly one MintHeader entry '
+                    f'for the new token (token_index=1); got {len(new_token_entries)}'
+                )
+            # MintMeltEntry enforces 1 <= amount < 2**64, so amount <= 0 cannot occur.
+            # A TCT cannot melt the token it is creating — there is nothing to destroy yet.
+            if tx.has_melt_header():
+                if any(e.token_index == 1 for e in tx.get_melt_header().entries):
+                    raise InvalidToken(
+                        'token creation transaction cannot melt the new token (token_index=1)'
+                    )
```
glevco, line 62: On the `VertexVerifier` you're forbidding a TCT of containing a `MeltHeader`.

hathor/verification/transaction_verifier.py
```diff
@@ -350,6 +353,206 @@ def _check_token_permissions(token_uid: TokenUid, token_info: TokenInfo) -> None
...
+        upstream and downstream.
+        """
+        if not tx.has_mint_header() and not tx.has_melt_header():
+            return
+        if tx.has_shielded_outputs() or tx.has_unshield_balance_header():
+            return
+        raise ShieldedMintMeltForbiddenError(
+            'MintHeader/MeltHeader requires the transaction to carry a '
+            'ShieldedOutputsHeader or UnshieldBalanceHeader (Rule M1)'
+        )
```
glevco: If mint/melt always requires a shielded header, why have separate headers at all? It's up to two extra bytes that are useless. The mint/melt data could just be part of the shielded header model, and we would just route them through serialization depending on the header version.

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+    from hathorlib.base_transaction import BaseTransaction
+    from hathorlib.serialization import Deserializer, Serializer
+
+
+# Per RFC 0000-shielded-outputs-mint-melt §4.1.
+MAX_MINT_MELT_ENTRIES = 16
+AMOUNT_SIZE = 8  # amount is a u64, big-endian
+
+
+@dataclass(frozen=True)
```
glevco: Add `slots=True, kw_only=True`

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+    sneak a negative value into the Pedersen-augmented balance equation.
+    Validated at construction so programmatic builders fail fast at the call
+    site, and again at deserialize-time via this same constructor.
+    """
+
+    token_index: int
+    amount: int
+
+    def __post_init__(self) -> None:
+        if not 1 <= self.token_index <= MAX_MINT_MELT_ENTRIES:
```
glevco: For clarity:

```suggestion
        if not (1 <= self.token_index <= MAX_MINT_MELT_ENTRIES):
```

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+            raise ValueError(
+                f'token_index must be in [1, {MAX_MINT_MELT_ENTRIES}]; got {self.token_index}'
+            )
+        if not 1 <= self.amount < 2 ** 64:
```
glevco: For clarity:

```suggestion
        if not (1 <= self.amount < 2 ** 64):
```

glevco: Move `2**64` to a constant

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+    The strict positive amount bound (1 <= amount < 2**64) is what lets us
+    skip a range proof on these headers: by construction the declared scalar
+    cannot encode a negative or zero supply change, so it cannot be used to
+    sneak a negative value into the Pedersen-augmented balance equation.
+    Validated at construction so programmatic builders fail fast at the call
+    site, and again at deserialize-time via this same constructor.
+    """
+
+    token_index: int
+    amount: int
```
glevco, line 55: This would be a great candidate for a Rust-native model. We could enforce the min/max limits on the type level. I would say it's not required for this PR, so if you don't want to change this now, we can proceed, but it would be nice.

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+AMOUNT_SIZE = 8  # amount is a u64, big-endian
+
+
+@dataclass(frozen=True)
+class MintMeltEntry:
+    """A single (token_index, amount) entry in a MintHeader or MeltHeader.
+
+    Wire format: token_index(1B) | amount(8B BE).
+
+    The strict positive amount bound (1 <= amount < 2**64) is what lets us
```
glevco, line 46: This will have to change when we introduce the 18 decimal places, right?

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+        if token_index in seen_indexes:
+            raise ValueError(f'{header_name}: duplicate token_index {token_index}')
+        seen_indexes.add(token_index)
+        # MintMeltEntry enforces the token_index/amount bounds.
+        entries.append(MintMeltEntry(token_index=token_index, amount=amount))
+
+    return entries
+
+
+@dataclass(frozen=True)
```
glevco: Add `slots=True`

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+        serialize_entries(serializer, self.entries)
+        return bytes(serializer.finalize())
+
+    def get_sighash_bytes(self) -> bytes:
+        # Full serialization is bound to the signature: any mutation to the
+        # declared mint/melt amounts invalidates all signatures over the tx.
+        return self.serialize()
+
+
+@dataclass(frozen=True)
```
glevco: Add `slots=True`

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+class MintHeader(_MintMeltHeaderBase):
+    """Publicly declares per-token supply created by this shielded transaction."""
+
+    _HEADER_ID: ClassVar[bytes] = VertexHeaderId.MINT_HEADER.value
+    _HEADER_NAME: ClassVar[str] = 'MintHeader'
+
+
+@dataclass(frozen=True)
```
glevco: Add `slots=True`

hathorlib/hathorlib/headers/mint_melt_header.py
glevco, line 132: These are not used so you can just `raise NotImplementedError` in them, instead of duplicating the implementation. In a future PR we should remove them.

glevco: See https://github.com/HathorNetwork/hathor-core/pull/1730/changes#r3421932850

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+    Design note: a single header carries a list of (token_index, amount)
+    entries rather than spreading entries across multiple headers. The codebase
+    enforces one-instance-per-header-type (see verify_headers in hathor-core's
+    vertex_verifier), so allowing repeated MintHeader/MeltHeader instances would
+    require relaxing that invariant. Bundling the entries keeps the per-entry
+    uniqueness check local to the header.
+    """
+
+    _HEADER_ID: ClassVar[bytes] = b''
+    _HEADER_NAME: ClassVar[str] = ''
```
glevco, line 120: I think we don't need these.

hathor/transaction/vertex_parser/_mint_melt_header.py
```diff
@@ -0,0 +1,107 @@
...
+        header_id = bytes(deserializer.read_bytes(1))
+        if verbose:
+            verbose('header_id', header_id)
+        if header_id != header_id_value:
+            raise InvalidMintMeltHeaderError(
+                f'{header_name}: unexpected header id: expected {header_id_value!r}, got {header_id!r}'
+            )
+        entries = deserialize_entries(deserializer, header_name=header_name)
+    except InvalidMintMeltHeaderError:
+        raise
```
glevco: Why not just remove this clause?

glevco (changes requested):

hathor/transaction/vertex_parser/_headers.py
msbrogli: They are not being added in this PR. I guess they should be there from previous PRs.

hathor/verification/transaction_verifier.py
msbrogli: Ok. I'll do it.

hathor/verification/token_creation_transaction_verifier.py
msbrogli: I know. This is a second layer of protection. I guess it is better to "fail the transaction" instead of failing an assertion.

hathor/verification/transaction_verifier.py
msbrogli: Mint/melt with shielded outputs require the headers. If only inputs are shielded, there will be no shielded outputs at all in that transaction, for example. I'm not sure I really understood your suggestion.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Very likely. We still haven't mapped all conflicts between 18 decimal places and shielded outputs but it is one of them.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Good idea.

hathor/transaction/vertex_parser/_headers.py
glevco: @LFRezende

hathor/verification/token_creation_transaction_verifier.py
glevco: But the assertion would never be triggered on a correct implementation. This check is a business rule that's never enforced, in fact you can't even test it through normal tests, so I think it gives the false impression that this is accepted.

hathor/verification/transaction_verifier.py
glevco: Sorry. What I meant was, if `MintHeader/MeltHeader` require either a `ShieldedOutputsHeader` or a `UnshieldBalanceHeader`, why do `MintHeader/MeltHeader` exist as a separate entity?

Instead, you could simply move `mint_entries` and `melt_entries` to `ShieldedOutputsHeader` and `UnshieldBalanceHeader`, and route them through serialization, for example:

- `b'\x12'`: constructs a `ShieldedOutputsHeader` without mint/melt entries.
- `b'\x13'`: constructs a `UnshieldBalanceHeader` without mint/melt entries.
- `b'\x14'`: constructs a `ShieldedOutputsHeader` with mint entries.
- `b'\x15'`: constructs a `UnshieldBalanceHeader` with mint entries.
- Etc

Granted, we would have 8 header versions for all combinations. But we would be saving header version bytes as we would use a single one instead of 3 in the worst case (1 for `ShieldedOutputsHeader` + 1 for `MintHeader` + 1 for `MeltHeader`).

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+
+if TYPE_CHECKING:
+    from hathorlib.base_transaction import BaseTransaction
+    from hathorlib.serialization import Deserializer, Serializer
+
+
+# Per RFC 0000-shielded-outputs-mint-melt §4.1.
+MAX_MINT_MELT_ENTRIES = 16
+AMOUNT_SIZE = 8  # amount is a u64, big-endian
+
```
LFRezende, line 38: The amount in any shielded output must be tied to a range proof, which ties the value to `0<= amount <= 2^k`, where `k` is the number of bits (`k ~(log2(amount))`). 

We should, therefore, not only treat `2**64` as a constant `MAX_AMOUNT = 2**64`, like Gabriel mentioned below, but also have it dependent on the decided `RANGE_PROOF_BITS` constant.

Essentially: `MAX_AMOUNT = 2** RANGE_PROOF_BITS - 1`

There is no reason to allow mint/melt amounts be "free" up to a constant of `2^64` if the hidden value is always capped by whichever number of bits the range proof allows/constraints.

hathorlib/hathorlib/headers/mint_melt_header.py
LFRezende: Agreed, also see my previous comment:  https://github.com/HathorNetwork/hathor-core/pull/1730/changes#r3422503129

hathor/transaction/vertex_parser/_headers.py
LFRezende: @msbrogli @glevco just looked it up - it is more problematic than that. 

In the original `feat/shielded-outputs` branch, onto which our pr-split steps were based on,  and already on master after pr-4 (https://github.com/HathorNetwork/hathor-core/pull/1702) , we do not add these as matches at all - in any step of the pr-split plan. 

We do the following (PR-4 onward/master)

```python
    match header:
        case NanoHeader():
             # ...
            return bytes(serializer.finalize())
        case FeeHeader():
             # ...
            return bytes(serializer.finalize())
        case _:
            return header.get_sighash_bytes()
```

And in PR-5 onward (https://github.com/HathorNetwork/hathor-core/pull/1709),  we have the `ShieldedOutputsHeader ` and `UnshieldBalanceHeader` with the method `get_sighash_bytes` defined. 

```python
    # For ShieldedOutputsHeader, in its respective module (PR-5)

    def get_sighash_bytes(self) -> bytes:
        """Include in sighash: header_id + count + per-output sighash bytes."""
        parts: list[bytes] = []
        parts.append(VertexHeaderId.SHIELDED_OUTPUTS_HEADER.value)
        parts.append(int_to_bytes(len(self.shielded_outputs), 1))

        for output in self.shielded_outputs:
            parts.append(output_sighash_bytes(output))

        return b''.join(parts)

    # For UnshieldBalanceHeader, in its respective module (PR-5)

    def get_sighash_bytes(self) -> bytes:
        # The full serialization is bound to the signature, so any mutation of
        # the scalar invalidates signatures over the tx.
        return self.serialize()

```

So we neved added these classes to `get_header_sighash_bytes`, in master or in any step of the original plan. I think Claude didn't catch that due to the fallback still enabling the headers to work. 

But now, we have the "never case", where it raises an error. 
If we're to add these classes now, one possibility:

```python

        # ...
        # ----- Shielded Cases------ #
        case ShieldedOutputsHeader():
             return header.get_sighash_bytes()

        case UnshieldBalanceHeader():
             return header.get_sighash_bytes() // or header.serialize() too, it's possible.
 
        # ----- Mint/Melt ------ #
        case MintHeader():
             # ...
            return bytes(serializer.finalize())

        case MeltHeader():
             # ...
            return bytes(serializer.finalize())

        # ----- Exception ----- #
        case _:
            raise SomeError

```

The respective modules are added in PR-5.

hathor/transaction/vertex_parser/_mint_melt_header.py
LFRezende: Agreed. Feels unnecessary.

hathor/verification/transaction_verifier.py
LFRezende: The `ShieldedVerifier`, in case you'd like a reference, is in PR-5. 

https://github.com/HathorNetwork/hathor-core/pull/1709

hathorlib/hathorlib/headers/mint_melt_header.py
LFRezende, line 55: I see many thrown constants that aren't that explained and that could potentially be meddled with in the future.

Adding to Gabriel's comment, we could centralize all constants related to Mint/Melt in a class or similar.

Not truly necessary, but something like the following would be tidier:

```python

from ...  import RANGE_PROOF_BITS
from .... import DECIMALS

...

@dataclass
class MintMeltHeaderConstants():
      
        MAX_NUMBER_ENTRIES = MAX_MINT_MELT_ENTRIES
        MIN_AMOUNT=1
        MAX_AMOUNT= 2**RANGE_PROOF_BITS
        DECIMALS= DECIMALS
        ...
```

Bundling it all in the same class to make it clearer and referenced.

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,177 @@
...
+        try:
+            header_id = bytes(deserializer.read_bytes(1))
+            if header_id != cls._HEADER_ID:
+                raise ValueError(
+                    f'{cls._HEADER_NAME}: unexpected header id: '
+                    f'expected {cls._HEADER_ID!r}, got {header_id!r}'
+                )
+            entries = deserialize_entries(deserializer, header_name=cls._HEADER_NAME)
+        except ValueError:
+            raise
```
LFRezende: Do we need this too?

hathorlib/tests/test_basic.py
```diff
@@ -294,6 +294,62 @@ def test_tx_with_unshield_balance_header(self):
...
+            (MeltHeader, VertexHeaderId.MELT_HEADER.value, b'\x15'),
+        ]
+        for header_cls, id_value, id_byte in cases:
+            entries = [MintMeltEntry(token_index=1, amount=1000), MintMeltEntry(token_index=3, amount=2 ** 63)]
+            header = header_cls(entries=entries)
+
+            # Wire format: header_id(1) | num_entries(1) | (token_index(1) | amount(8 BE))...
+            wire = header.serialize()
+            self.assertEqual(wire[:1], id_value)
+            self.assertEqual(wire[:1], id_byte)
```
LFRezende: Won't `wire[:1]` yield the same value?

hathor/transaction/vertex_parser/_headers.py
msbrogli: Fixed in 86be6e12. Added an explicit `case ShieldedOutputsHeader() | UnshieldBalanceHeader(): return header.get_sighash_bytes()` arm before the Mint/Melt arms, so these no longer fall through to the `raise AssertionError('unreachable')` guard (which would have crashed sighash computation for those headers). Thanks @LFRezende for tracing that it got dropped when the generic fallback was replaced.

hathor/verification/transaction_verifier.py
msbrogli: Done in 86be6e12. Extracted a dedicated `MintMeltVerifier` (`hathor/verification/mint_melt_verifier.py`) holding all six `verify_mint_melt_*` methods, wired into `VertexVerifiers` as `mint_melt` (modeled on `TokenCreationTransactionVerifier`), and updated the call sites in `verification_service.py`. The methods are removed from `TransactionVerifier`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Done in 86be6e12 — `@dataclass(frozen=True, slots=True, kw_only=True)` on `MintMeltEntry`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Applied in 86be6e12: `if not (1 <= self.token_index <= MAX_MINT_MELT_ENTRIES):`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Applied in 86be6e12: `if not (1 <= self.amount < MAX_AMOUNT):`.

msbrogli: Done in 86be6e12 — added `MAX_AMOUNT = 2 ** 64` and use it in both the bound check and the error message.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Done in 86be6e12 — added `slots=True`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Done in 86be6e12 — added `slots=True`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Done in 86be6e12 — added `slots=True`.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Done in 86be6e12 — `deserialize`/`serialize`/`get_sighash_bytes` on `_MintMeltHeaderBase` now `raise NotImplementedError`, with a comment that the real path is the framework free functions in `_mint_melt_header.py`. Note this means hathorlib `base_transaction` cannot parse a tx carrying these headers yet — deliberate until a follow-up drops them from the `VertexBaseHeader` interface. Tests updated to exercise the free-function path.

msbrogli: Replaced with a `raise NotImplementedError` stub in 86be6e12 (see the NotImplementedError thread). It remains only to satisfy the `VertexBaseHeader` ABC; the real (de)serialization is the free functions.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Stubbed with `NotImplementedError` in 86be6e12; slated for full removal from the base interface in a follow-up.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Same treatment — stubbed with `NotImplementedError` in 86be6e12.

hathor/transaction/vertex_parser/_mint_melt_header.py
msbrogli: Removed in 86be6e12. `InvalidMintMeltHeaderError` subclasses `TxValidationError` (not `ValueError`/`SerializationError`), so it was never caught by the `except (SerializationError, ValueError)` below — the re-raise clause was dead code. Dropped it.

hathorlib/tests/test_basic.py
msbrogli: Right — collapsed in 86be6e12: assert `id_value == id_byte` once, then check `wire[:1]` once.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Partly addressed in 86be6e12: hoisted `MAX_AMOUNT = 2 ** 64` (the u64 wire bound) with a comment that it should tighten to `2 ** RANGE_PROOF_BITS` once that constant lands. Kept it at u64 for now since `RANGE_PROOF_BITS` is not defined on this branch — it comes with the crypto layer.

hathor/verification/transaction_verifier.py
msbrogli: Keeping `verify_mint_melt_nano_compatibility` in this PR (86be6e12 moved it to the new `MintMeltVerifier` unchanged). It guards the one genuinely ambiguous interaction: the same token having its supply changed through both a NanoHeader action and a Mint/Melt header in one tx would feed a double-counted public scalar into the augmented balance equation. Cross-token combinations stay allowed. Happy to relocate it next to the equation in the nano/balance PR if you prefer, but I would rather not drop the guard entirely.

hathor/verification/token_creation_transaction_verifier.py
msbrogli: Keeping this as defense-in-depth. You are right that `VertexVerifier` already rejects a `MeltHeader` on a TCT at the header gate, so in a correct implementation this branch is unreachable. I would rather fail the transaction explicitly than rely solely on the gate, in case the allowed-header set is ever changed. I take the point that an unreachable business rule can read as if the case were supported — I can add a comment marking it a redundant safety net, or drop it if you feel strongly. Leaving it for now.

hathor/verification/transaction_verifier.py
msbrogli: This is a larger wire-format decision, so I would rather not fold it into this PR. The tradeoff for whoever picks it up: separate MintHeader/MeltHeader keep the supply-change declaration orthogonal to the shielded-output payload and avoid a combinatorial set of header versions; your proposal collapses the 3-headers-worst-case into one version byte at the cost of ~8 versions encoding the present/absent combinations. Both are viable. I will keep the current 3-header layout here and we can settle the encoding separately before it ships — I do not want to lock the wire format without consensus.

hathorlib/hathorlib/headers/mint_melt_header.py
msbrogli: Agreed it is a good fit, and as you noted it is not required for this PR. Leaving the Python model here; we can move it to a Rust-native type that enforces the min/max bounds at the type level in a follow-up.

msbrogli: Partially done in 86be6e12: hoisted `MAX_AMOUNT` next to the existing `MAX_MINT_MELT_ENTRIES` / `AMOUNT_SIZE` module constants. I held off on bundling them into a dedicated constants class to keep this diff focused; that (together with tying `MAX_AMOUNT` to `RANGE_PROOF_BITS`) is a good follow-up once the range-proof constant exists.

jansegre (approved):

hathorlib/hathorlib/headers/mint_melt_header.py
LFRezende: Ok!

hathorlib/hathorlib/headers/mint_melt_header.py
LFRezende: Fair enough!

hathorlib/hathorlib/headers/mint_melt_header.py
LFRezende: Perfect.

hathorlib/tests/test_basic.py
LFRezende: Great!

hathor/transaction/vertex_parser/_headers.py
LFRezende: Great!

glevco: Thanks!

hathor/verification/transaction_verifier.py
glevco: Thanks!

hathor/verification/token_creation_transaction_verifier.py
glevco: The point is that this is not defense-in-depth, it's the opposite:

> in case the allowed-header set is ever changed

If `MeltHeader` is accidentally added to the allowed-header set, it would be incorrectly accepted here. The correct defense-in-depth/safety-net _is_ the assertion.

hathor/verification/transaction_verifier.py
glevco: I think this is an AI response and it needs your input, @msbrogli.

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Thanks!

hathorlib/hathorlib/headers/mint_melt_header.py
glevco: Both are unused.

hathor/transaction/vertex_parser/_mint_melt_header.py
glevco: Thanks!

hathor/verification/mint_melt_verifier.py
```diff
@@ -0,0 +1,245 @@
...
+                # Shielded inputs cannot be authority outputs (parent Rule 7).
+                continue
+            spent_output = spent_tx.outputs[spent_index]
+            if not spent_output.is_token_authority():
+                continue
+            token_uid = spent_tx.get_token_uid(spent_output.get_token_index())
+            if spent_output.can_mint_token():
+                mint_authorities.add(token_uid)
+            if spent_output.can_melt_token():
+                melt_authorities.add(token_uid)
```
glevco, line 194: Why not use the pre-calculated `TokenInfoDict` for this?

hathor/verification/mint_melt_verifier.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def verify_mint_melt_nano_compatibility(self, tx: Transaction) -> None:
+        """Reject same-token mint/melt declared via both NanoHeader actions and Mint/Melt headers.
+
+        Per the user's choice on RFC unresolved Q3, a NanoHeader may coexist with
+        Mint/Melt headers in the same tx. Cross-token combinations are fine, but
+        a single token cannot be minted (or melted) through both channels at once
+        because the amount would be ambiguous and the augmented balance equation
+        would double-count.
+        """
```
glevco, line 127: Why is this rule necessary? The amount would not be ambiguous because the nano actions have explicit amounts.

hathor/verification/verification_service.py
```diff
@@ -322,6 +328,12 @@ def verify_without_storage(self, vertex: BaseTransaction, params: VerificationPa
             assert self._settings.ENABLE_NANO_CONTRACTS
             self._verify_without_storage_nano_header(vertex, params)
 
+        # Storage-free MintHeader/MeltHeader checks (Rules M1, M3, well-formedness).
+        # Header acceptance per tx version is already gated by verify_headers; this
+        # is a no-op unless a Mint/Melt header is present.
+        if isinstance(vertex, Transaction):
+            self.verifiers.mint_melt.verify_mint_melt_basic(vertex)
```
glevco, line 335: Move to `_verify_without_storage_tx`

glevco (changes requested):

hathor/verification/mint_melt_verifier.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def verify_mint_melt_basic(self, tx: Transaction) -> None:
+        """Top-level: basic (no-storage) verification for MintHeader/MeltHeader.
+
+        Fires whenever either header is present. Performs Rules M1, M3, the
+        well-formedness checks against tx.tokens length, and the NanoHeader
+        same-token guard.
+        """
+        if not tx.has_mint_header() and not tx.has_melt_header():
+            return
```
LFRezende, line 55: If it fires whenever either is present, an assert might be more pertinent here - it should not fire if none of the headers are present , hence it should be an assert.

```python
assert tx.has_mint_header() or tx.has_melt_header(), 'At least one header must be present.'
```

LFRezende, line 55: Not necessary, but an assert on "has shielded outputs" could be interesting too, as this function should not be called for mint/melt headers without a shielded header.

```python

assert tx.is_shielded(), 'Mint/Melt headers must occur solely in shielded txs'

```

hathor/verification/mint_melt_verifier.py
```diff
@@ -0,0 +1,245 @@
...
+        self.verify_mint_melt_headers_well_formed(tx)
+        self.verify_mint_melt_requires_shielded(tx)
```
LFRezende, line 57: You can put this function first and drop the mint/melt headers check (and the shielded suggestion as well).  You need not pass through the whole computation of `verify_mint_melt_headers_well_formed` just to see later via a quick check that the headers are not really there.

hathorlib/hathorlib/headers/mint_melt_header.py
```diff
@@ -0,0 +1,155 @@
...
+    """A single (token_index, amount) entry in a MintHeader or MeltHeader.
+
+    Wire format: token_index(1B) | amount(8B BE).
+
+    The strict positive amount bound (1 <= amount < 2**64) is what lets us
+    skip a range proof on these headers: by construction the declared scalar
+    cannot encode a negative or zero supply change, so it cannot be used to
+    sneak a negative value into the Pedersen-augmented balance equation.
+    Validated at construction so programmatic builders fail fast at the call
+    site, and again at deserialize-time via this same constructor.
```
LFRezende, line 51: How do we guarantee the value between 1 and 2^64 is the same as the blinded value in the pedersen commitment? It could declare something in the amount and encode something different in the range proof. 

The range proofs only ensure 0 and 2^64 - 1.  So someone could sneak in v as zero and the blinding factor would hide it - and the range-proof would still confirm it.

We could perhaps do a range proof for the first 63 bits and force the last to be 1. Not sure if this trick has any flaws but it is a possibility.

LFRezende (commented):
All comments but one are small scope and non-critical.

I just have the final question as an impeditive - to make sure we're onboard with the range proofs|pedersen commitments vs declared amount.  Being answered/rebutted/corrected, I can approve the rest without correction.

PR #1732, opened 2026-06-15
feat(storage): move the RocksDB connection to Rust

msbrogli:
## What

Rust now owns the primary RocksDB handle, exposed to Python as `htr_lib.RocksDb`. This is the first phase of a larger migration; it moves **only** the RocksDB connection to Rust with semantics unchanged (no serde in the swap).

## How

- **Rust (`htr-rs/crates/htr-lib/storage/`)** — new self-contained module exposing `RocksDb`/`RocksDbWriteBatch`/`RocksDbIterator` pyclasses with the exact op surface hathor-core uses through python-rocksdb: `get`/`multi_get`/`put`/`delete`, atomic multi-CF write batches, chunked iterators (one FFI call per chunk; modes `first`/`last`/`seek`/`seek_for_prev`, reverse), `key_may_exist`, `get_property`, and column-family management. Every call releases the GIL; `RocksDb::native()` shares the same `Arc<DB>` with future Rust-native readers.
- **Python** — `RocksDBStorage` opens the database through this handle. A python-rocksdb-shaped facade (`hathor/storage/rocksdb_compat.py`) preserves every call shape the ~15 consumers use, so they only swap `import rocksdb` for the compat module. Iterators fetch 256 pairs per FFI call.
- `python-rocksdb` stays a dependency **only** for the differential test.

## Compatibility

Cross-binding file compatibility is deliberately not maintained — the bundled librocksdb 10.x writes SST footers python-rocksdb cannot read. The migration path is a fresh database + re-sync.

## Testing

- 20 Rust unit tests (op surface, atomicity, chunk/seek semantics, reopen persistence, close behavior).
- Python differential suite applying the same seeded op stream to both backends and comparing every observable (scan order both directions across chunk sizes, point reads, multi_get, seeks at every boundary).
- Storage / index / cache suites green; `mypy`, `ruff`, `cargo fmt`/`clippy`/`sort` clean.


hathor/storage/rocksdb_storage.py
```diff
@@ -39,34 +48,8 @@ def __init__(
...
-            # this means the db doesn't exist, a repair will create one
-            rocksdb.repair_db(db_path, options)
-            cf_names = []
-
-        # we need to open all column families
-        column_families = {cf: rocksdb.ColumnFamilyOptions() for cf in cf_names}
-
-        # finally, open the database
-        self._db = rocksdb.DB(db_path, options, column_families=column_families)
+        inner = htr_lib.RocksDb(db_path, cache_capacity)
```
msbrogli, line 51: Why this? Shouldn't `rocksdb_compat.DB()` be directly inject by the builder?

plans/rust-rocksdb-storage.md
msbrogli, line 1: We can remove this file.

jansegre (dismissed):

PR #1733, opened 2026-06-15
feat(benchmarking): tracked UTXO transaction-processing TPS benchmark

msbrogli:
### Motivation

We're phasing the large optimization branch (`feat/rust-script-executor`) into small PRs, and we want a **tracked transactions-per-second number for every PR**, compared against the master baseline — so each phase (RocksDB-in-Rust, verification, …) shows whether it actually moved TPS. This adds that benchmark, built **on top of** the existing `sync_v2` benchmark and tracked in Bencher with an automatic PR-vs-master comparison.

It measures plain UTXO transactions only (no nano contracts), on a real reactor with real (temp) RocksDB. A synthetic DAG (spend-chains of txs, optionally confirmed by blocks) is built once with the DAG Builder and fed to a fresh manager, with **transaction processing and block processing timed separately**:

- transactions enter through the realtime relay path (`vertex_handler.on_new_relayed_vertex`) — parse + full verification + consensus;
- blocks are then connected (`vertex_handler.on_new_block`), confirming the txs.

Each phase ends with `tx_storage.flush()` so the **actual RocksDB write is inside the measured window** (the tx-storage cache otherwise flushes asynchronously and would hide storage cost — exactly what the RocksDB PR optimizes). Only the payload is timed; the reward-lock base chain and funding tx are connected first, untimed.

It is written against entrypoints that exist on master, so it runs identically on master and on every future PR. The original branch's `bench_real_tps.py` could not be reused — it needs `verify_bytes` / `ScriptVerificationMode.RUST`, which don't exist on master.

Example output (master baseline, `--txs 2000 --blocks 1`, one run):

```
real reactor, 2,000 txs, 1 block(s), median of 1 run(s)

| inputs/outputs per tx | tx/s | UTXOs/s (tx/s x N) | txs (s) | blocks (s) | total (s) | tx/s vs 1-in |
|                     1 |  685 |                685 |   2.920 |      0.090 |     3.010 |            — |
|                     2 |  437 |                875 |   4.572 |      0.096 |     4.668 |       -36.1% |
|                     4 |  262 |              1,049 |   7.624 |      0.108 |     7.733 |       -61.7% |
|                     8 |  146 |              1,170 |  13.681 |      0.136 |    13.817 |       -78.7% |
|                    16 |   78 |              1,246 |  25.675 |      0.180 |    25.855 |       -88.6% |
```

`tx/s` is transaction-phase throughput; the steep drop with heavier txs is master's baseline weakness (per-input signature verification, no Rust batching) that later PRs should flatten. In UTXO-throughput terms it still rises (685 → 1,246 UTXOs/s).

### Acceptance Criteria

- `extras/benchmarking/tps/bench_tps.py` measures UTXO-only, processing-time TPS on a real reactor with real RocksDB, sweeping 1/2/4/8/16 inputs/outputs per tx (median of N runs).
- Transaction processing and block processing are timed **separately**, each including a storage flush so the RocksDB write is measured; the report shows tx/s plus transactions / blocks / total seconds.
- `--blocks 0` measures transactions only (no confirming block).
- `--output` emits Bencher Metric Format JSON (per input-count: `throughput` tx/s + `txs-seconds` / `blocks-seconds` / `total-seconds`); `--baseline` adds a vs-master column.
- CI tracks it in Bencher: `base_tps_benchmarks.yml` (push to master) and `pr_tps_benchmarks.yml` (PRs, posts the PR-vs-master comparison), using the `json` adapter with no benchmark-data download.
- Runs on master and every future PR using only master-stable entrypoints; `ruff`, `mypy`, and `yamllint` clean.

### Checklist


extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+
+# IMPORTANT: the settings must be selected *before* importing any hathor module, because
+# most of them read the global settings at import time. We use the same unittests config
+# the test suite uses, so that the genesis output script matches the DAG Builder genesis
+# wallet and the reward-lock window is small.
+from hathorlib.conf import UNITTESTS_SETTINGS_FILEPATH  # noqa: E402
+
+os.environ.setdefault(
+    'HATHOR_CONFIG_YAML',
+    os.environ.get('HATHOR_TEST_CONFIG_YAML', UNITTESTS_SETTINGS_FILEPATH),
```
glevco, line 82: Shouldn't we use mainnet or at least testnet? Unittests could disable some specific verification, for example.

glevco (dismissed):

extras/benchmarking/tps/bench_tps.py
msbrogli: Yes, we can. I guess it wouldn't make a difference for the current setup but I'm ok changing it.

extras/benchmarking/tps/.env
```diff
@@ -0,0 +1,6 @@
+N_TXS=2000
+N_BLOCKS=1
+N_RUNS=3
+INPUTS=1,2,4,8,16
+BENCH_FILE=tps_results.json
+EXPECTED_NPROC=4
```
LFRezende, line 9: No variable for the number of outputs? 

`N_OUTPUTS=2` for example (good default)

For now they will not matter much as outputs seem not to muddle TPS as inputs do. But when we bring shielded outputs to the stack, they'll give us a run for our money.

We'd have to have something like:

`N_OUTPUTS=1, 2, 4, 8, 16` as well.

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+  - transactions are connected through the realtime relay path
+    (``vertex_handler.on_new_relayed_vertex``) — parse + full verification + consensus;
+  - blocks are then connected (``vertex_handler.on_new_block``), confirming the txs.
+
+Each phase ends with ``tx_storage.flush()`` so the actual RocksDB write it produced is
+inside the measured window (the tx-storage cache otherwise flushes asynchronously). The
+real reactor means storage is real (temp) RocksDB, so the number is honest wall-clock
+processing time. Only the payload is timed — the reward-lock base chain and the funding
+tx are connected first, untimed.
+
```
LFRezende, line 39: Discussion: If we're willing to know the "best average TPS" of the full-node verification, counting the `tx_storage.flush()` time for when the rocksDB offloads the written SST files to storage will amortize that timing - it'd drop the average from, say, 600 TPS to maybe 580 TPS (example) within a flush timeframe, but until the flush, it was higher. 

Not a bug or a problem - we can keep it if we're willing to count the flushing as part of our "best average TPS" definition.

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+    num_blocks: int,
+    num_inputs: int,
+) -> Generator[Any, Any, tuple[float, float]]:
+    """Connect the payload, timing transaction-processing and block-processing separately.
+
+    Returns ``(txs_seconds, blocks_seconds)``. Transactions enter through the realtime relay
+    path (``on_new_relayed_vertex``); blocks are then connected with no deps, confirming the
+    already-present txs. Each phase ends with ``tx_storage.flush()`` so the RocksDB write it
+    produced is inside its measured window (the cache otherwise flushes asynchronously). With
+    ``num_blocks == 0`` there is no block phase, so ``blocks_seconds`` is 0."""
```
LFRezende: Again, my personal take is to remove `tx_storage.flush()` from time computing. But it is a matter of opinion - true TPS fetch vs stable TPS fetch.

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+        names = ([f'g{r}_s'] if num_inputs > 1 else []) + [f'g{r}_t{i}' for i in range(num_txs)]
+
+        # transaction phase: parse + verify + consensus + storage flush
+        t0 = time.perf_counter()
+        for name in names:
+            ok = manager.vertex_handler.on_new_relayed_vertex(parse(name))
+            assert ok, f'tx {name} rejected'
+        storage.flush()
+        txs_seconds += time.perf_counter() - t0
+
```
LFRezende, line 306: Put a `tf = time.perf_counter()` above `txs_seconds` and then just do `txs_seconds += tf - t0` - cleaner.

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+        # the per-group source (num_inputs > 1) is the chain root and must connect first
+        names = ([f'g{r}_s'] if num_inputs > 1 else []) + [f'g{r}_t{i}' for i in range(num_txs)]
+
+        # transaction phase: parse + verify + consensus + storage flush
+        t0 = time.perf_counter()
+        for name in names:
+            ok = manager.vertex_handler.on_new_relayed_vertex(parse(name))
+            assert ok, f'tx {name} rejected'
+        storage.flush()
+        txs_seconds += time.perf_counter() - t0
```
LFRezende, line 305: Linking to my personal take: we can flush every single time, but fetch the time flag before the storage flush, like this:

```python

# Start the counting
t0 = time.perf_counter()
        for name in names:
            # verify groups
            ok = manager.vertex_handler.on_new_relayed_vertex(parse(name))
            assert ok, f'tx {name} rejected'

# Finish verification time
tf = time.perf_counter()

storage.flush()    # flush for consistency, but without measuring the time it took

# Actual, non-flush-amortized time of verification:
txs_seconds += tf - t0
```

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+        if has_blocks:
+            block = parse(f'x{r + 1}')
+            assert isinstance(block, Block)
+            t1 = time.perf_counter()
+            ok = yield manager.vertex_handler.on_new_block(block, deps=[])
+            assert ok, f'payload block x{r + 1} rejected'
+            storage.flush()
+            blocks_seconds += time.perf_counter() - t1
+    return txs_seconds, blocks_seconds
+
```
LFRezende, line 317: Idem here:

```python
# Start
t1 = time.perf_counter()
ok = yield manager.vertex_handler.on_new_block(block, deps=[])
assert ok, f'payload block x{r + 1} rejected'
# End
t2 =  time.perf_counter()  

# Flush - put it here or above, if we wish to measure it with the TPS time (I personally am not a partisan of this).
storage.flush()

blocks_seconds += t2 - t1 # Block total time
```

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+    """Per input-count, reduce the per-run (txs_s, blocks_s) to medians and a tx/s figure.
+
+    tx/s uses only the transaction phase — the headline 'transaction processing' throughput.
+    min/max are kept so the report and Bencher can show run-to-run spread."""
+    stats: dict[int, dict[str, float]] = {}
+    for n, run_list in samples.items():
+        txs_t = [t for t, _ in run_list]
+        blk_t = [b for _, b in run_list]
+        tot_t = [t + b for t, b in run_list]
+        tps = [total_txs / t for t in txs_t]
```
LFRezende, line 376: This does not seem right.  

The total number of transactions `total_txs` is for a given batch. 
Each `t` is the total time measured of the flow of a single transaction `tx`. 

The tps value should be  `tps = [1/t for t in txs_t if t > 0 else raise Exception]` and we should remove `total_txs` as a function parameter.

Example: Say we send `total_txs == 2000` for run `n == 2`. This run gets the following times:

`txs_t = [ 0.1 0.05 0.2 ... 1]` seconds. So, if one `tx` takes these seconds, then we have `tps = [10 20 5 ... 1]` transactions per second (reverse). 

Doing `tps = [ total_txs/t for t in txs_t]` multiplies these values by 2000.

extras/benchmarking/tps/bench_tps.py
```diff
@@ -0,0 +1,516 @@
...
+    *,
+    total_txs: int,
+) -> dict[int, dict[str, float]]:
+    """Per input-count, reduce the per-run (txs_s, blocks_s) to medians and a tx/s figure.
+
+    tx/s uses only the transaction phase — the headline 'transaction processing' throughput.
+    min/max are kept so the report and Bencher can show run-to-run spread."""
+    stats: dict[int, dict[str, float]] = {}
+    for n, run_list in samples.items():
+        txs_t = [t for t, _ in run_list]
```
LFRezende, line 373: Isn't `total_txs == len(txs_t)`, or at least `total_txs += len(txs_t)`? If so, we could delete it since it is an unnecessary input and make it a local variable.

extras/benchmarking/tps/.env
msbrogli: Good catch. I'll do it now.

LFRezende: Thanks!

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 After a closer look I'm keeping the unittests config, but it's not a drop-in switch — here's the constraint. The bench builds its DAG with the DAG Builder, whose genesis wallet/output script has to match the active settings' genesis, and it relies on the small reward-lock window (`REWARD_SPEND_MIN_BLOCKS`) to spend the funding reward quickly. mainnet/testnet genesis wouldn't match the builder's funding wallet, so the synthetic DAG wouldn't be spendable without reworking the whole funding path.

The pipeline under test (parse → full verify → consensus) is the same code regardless of network, and the settings that actually affect verification here (reward lock, weights) are already realistic. If we later want to catch a network-specific verification toggle I'm happy to revisit, but it'd be a separate funding rework rather than a config flip.

extras/benchmarking/tps/.env
msbrogli: 🤖 Done in 6b3a41bd. Added an independent `--outputs` / `OUTPUTS` knob: left empty it matches inputs (the previous diagonal sweep); set to a list it sweeps the full inputs × outputs matrix. A quick local check (inputs=2 fixed) shows outputs do cost a little, as you predicted: 437 → 413 → 359 tx/s for 1 → 4 → 16 outputs. BMF keys are now `tps/NNin-NNout` and the table gained an `outputs` column.

This needed dropping the spend-chain (it forced inputs == outputs): each tx now spends `num_inputs` UTXOs minted by per-group feeder txs (untimed scaffolding) and creates `num_outputs` terminal outputs, balanced exactly so the filler adds nothing that skews the counts. Payload txs are parent-chained instead of spend-chained, so a single block still confirms the whole group.

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 Keeping the flush inside the measured window — by design, and documented in the module/`feed` docstrings. The goal is honest end-to-end wall-clock: the tx-storage cache flushes to RocksDB asynchronously, so stopping the timer before `flush()` would charge the bench for verification but let the actual write land outside the window, reporting a TPS the node can't sustain once the cache drains.

You're right that it amortizes the per-SST write across the batch — that's exactly the "sustainable TPS" we want to track rather than the pre-flush burst. I see the true-vs-stable framing and I'm choosing stable on purpose. If we ever want the burst figure too, a `--no-flush-timing` mode would be the clean way to add it.

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 Same call as the discussion thread: flush stays in the window intentionally (sustainable TPS vs pre-flush burst). Noted as a deliberate definition, not an oversight.

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 This is equivalent to the current `txs_seconds += time.perf_counter() - t0` — one clock read, then subtract. The separate `tf` snapshot only buys something if we want to read the clock *before* the flush, which we don't (flush stays measured — see the flush discussion thread). Kept the inline form to avoid the extra local.

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 Rationale on the flush discussion thread above: flush is kept inside the timed window on purpose (sustainable vs burst TPS), so I kept the clock read after the flush rather than snapshotting before it.

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 Same as the tx phase — flush stays inside the block-phase window by design (see the flush discussion thread).

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 The formula is correct — `txs_t` isn't a list of per-transaction times, it's one total per run. `feed()` returns a single `txs_seconds` for the whole transaction phase of a run, so with `runs=3` `txs_t` has 3 entries, each the total seconds to process all `total_txs` payload txs in that run. So `total_txs / t` is the txs-per-second for that run, and we take the median across runs.

`[1/t for t in txs_t]` would treat each entry as a single tx's latency, which it isn't (it'd give ~0.4 tx/s instead of hundreds). `total_txs` = num_txs × groups, computed in `main`. I added a comment to `aggregate()` spelling this out so it doesn't read as per-tx.

extras/benchmarking/tps/bench_tps.py
msbrogli: 🤖 Not equal: `len(txs_t)` is the number of *runs* (e.g. 3), while `total_txs` is the number of *payload txs per run* (num_txs × groups, e.g. 2000) — different quantities, so `total_txs` does need to come from `main`. See the sibling thread on the tx/s formula for the full breakdown.

extras/benchmarking/tps/bench_tps.py
glevco: What about the overridden `MIN_TX_WEIGHT/MIN_SHARE_WEIGHT/MAX_TX_WEIGHT_DIFF/MAX_TX_WEIGHT_DIFF_ACTIVATION`, they all could incorrectly lead to lower transaction verification times. `SLOW_ASSERTS=True` is also enabled.

You could simply create a new `benchmark.yml` setting that `extends` mainnet and overrides the genesis with the same from unittests.

PR #1736, opened 2026-06-18
feat(decimals): implement runner verifications [part 12]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1719

### Motivation

Prevent Nano calls across different `TokenAmountVersion`s.

### Acceptance Criteria

- Add `TokenAmountVersion` to the blueprint catalog.
- Add verification to forbid the runner from instantiating a Blueprint with a `TokenAmountVersion` different from the one in the Runner (which comes from the tx). This covers creating contracts, public/view calls, and public/view proxy calls.
- Add verification to forbid a contract to upgrade to a Blueprint with an older `TokenAmountVersion`.
- Updating a blueprint and calling another contract in the same call follows the initial blueprint version.

### Checklist

- [x] If you are requesting a merge into `master`, confirm this code is production-ready and can be included in future releases as soon as it gets merged

PR #1740, opened 2026-06-22
feat(api): serve v1a/v2 APIs with version-aware nginx config and docs [part 13]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1736

### Motivation

The v2 APIs work needs the node to serve both the existing `/v1a` API and the new `/v2` API side by side, and our tooling has to follow. The `hathor-cli generate_nginx_config` generator was pinned to a single prefix, and the OpenAPI docs described one path per unversioned route with no notion of versions. This PR makes the node serve both prefixes and makes both the nginx config and the OpenAPI docs version-aware, driven by a per-route `x-api-versions` declaration in the OpenAPI spec, so v2 can be gated to only the routes that support it. Existing routes keep working unchanged by defaulting to v1a.

#### Serving

- Add an `APIVersion` enum (`v1a`, `v2`) carried on `Resource`, and drop the single `API_VERSION_PREFIX` setting.
- Mount a parallel `/v2` resource tree and serve both `/v1a` and `/v2` from the root, dual-registering the v2-capable resources (`create_tx`, `thin_wallet/address_balance`, `thin_wallet/token`, `utxo_search`, `nano_contract/state`, `wallet/balance`, `wallet/send_tokens`).

#### nginx config

- Generate one `location` block per declared API version, including the websocket and `403` fallback locations, driven by `x-api-versions`; the catch-all `404` is emitted once.
- Routes without `x-api-versions` default to `v1a` only, so v2 is gated to the routes that opt in.

#### OpenAPI docs

- Expand the unversioned merged spec into version-prefixed paths (`/v1a/...`, `/v2/...`). The expansion runs only in the docs path, leaving the unversioned dict that nginx consumes unchanged.
- Support an optional `x-api-version-overrides` block, deep-merged per version, so a version can document a different request/response schema.
- Extend `@api_endpoint` with per-version request/response models that lower into the same override representation, so a single expander handles both legacy `.openapi` dicts and decorated endpoints.
- Annotate the five public v2-mounted routes with `x-api-versions: ['v1a', 'v2']` plus a no-op v2 override scaffold, to be filled in once the decimal schema shapes are finalized.

---

#### nginx config

To check the diff on the generated nginx config after this PR, run this:

```bash
git switch master && git pull
BEFORE=$(poetry run hathor-cli generate_nginx_config)
git switch feat/decimal/v2-apis && git pull
AFTER=$(poetry run hathor-cli generate_nginx_config)
diff -y <(echo "$BEFORE") <(echo "$AFTER")
```

Inspect the output. The only diff should be the addition of `/v2` configs for the 5 updated public API routes, plus the 403 for `/v2` itself.

#### OpenAPI docs

To check the diff on the generated open api docs after this PR, run this:

```bash
git switch master && git pull
poetry run hathor-cli generate_openapi_json before.json
git switch feat/decimal/v2-apis && git pull
poetry run hathor-cli generate_openapi_json after.json
diff <(jq -S . before.json) <(jq -S . after.json)
```

Inspect the output. The only diff should be the addition of `/v2` docs for the 7 updated API routes, and the path update to include `/v1a` in all existing routes.

### Checklist


hathor/builder/resources_builder.py
```diff
@@ -263,15 +274,19 @@ def create_resources(self) -> server.Site:
                 NCDryRunResource,
             )
             nc_resource = Resource()
```
msbrogli: Maybe rename to v1?

hathor/nanocontracts/resources/state.py
```diff
@@ -317,6 +317,12 @@ class NCStateResponse(Response):
 NanoContractStateResource.openapi = {
     '/nano_contract/state': {
         'x-visibility': 'public',
+        'x-api-versions': ['v1a', 'v2'],
+        'x-api-version-overrides': {
+            # TODO(decimals): v2 mirrors v1a here. Add the v2 request/response schema
+            # delta (decimal token amounts) when the v2 shape is finalized.
+            'v2': {},
+        },
```
msbrogli: Why not add two entries, one for each version?

hathor/api/openapi/generator.py
```diff
@@ -79,27 +80,27 @@ def _build_parameters(self, metadata: EndpointMetadata) -> list[dict[str, Any]]:
...
-                    'schema': self._get_schema_ref(metadata.request_model),
+                    'schema': self._get_schema_ref(request_model),
                 },
             },
         }
 
-    def _get_response_models(self, metadata: EndpointMetadata) -> list[type[BaseModel]]:
-        """Extract individual response models from response_model (handles Union types)."""
-        if metadata.response_model is None:
+    def _get_response_models(self, response_model: Any) -> list[type[BaseModel]]:
```
msbrogli: Why `Any`?###

hathor/api/openapi/generator.py
```diff
@@ -111,15 +112,15 @@ def _get_response_models(self, metadata: EndpointMetadata) -> list[type[BaseMode
             return models
 
         # Single model
-        return [metadata.response_model]
+        return [response_model]
 
-    def _build_responses(self, metadata: EndpointMetadata) -> dict[str, Any]:
+    def _build_responses(self, response_model: Any) -> dict[str, Any]:
```
msbrogli: Same here. Why `Any`?

hathor/transaction/resources/create_tx.py
```diff
@@ -131,6 +131,12 @@ def _verify_unsigned_skip_pow(self, tx: Transaction) -> None:
 CreateTxResource.openapi = {
     '/create_tx': {
         'x-visibility': 'public',
+        'x-api-versions': ['v1a', 'v2'],
+        'x-api-version-overrides': {
+            # TODO(decimals): v2 mirrors v1a here. Add the v2 request/response schema
+            # delta (decimal token amounts) when the v2 shape is finalized.
+            'v2': {},
+        },
```
msbrogli: Why don't we have two separate entries? Something like `CreateTxResourceV1` and `CreateTxResourceV2`? They can share the same implementation but they are separate APIs with different OpenAPI documentations. Same for the other endpoints.

msbrogli (changes requested):

hathor/builder/resources_builder.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1740/changes/5157fc7e4998944f8f856a066d9231d19d568dfa

hathor/nanocontracts/resources/state.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1740/changes/5157fc7e4998944f8f856a066d9231d19d568dfa

hathor/api/openapi/generator.py
glevco: Reverted in https://github.com/HathorNetwork/hathor-core/pull/1740/changes/5157fc7e4998944f8f856a066d9231d19d568dfa

hathor/api/openapi/generator.py
glevco: Reverted in https://github.com/HathorNetwork/hathor-core/pull/1740/changes/5157fc7e4998944f8f856a066d9231d19d568dfa

hathor/transaction/resources/create_tx.py
glevco: Done in https://github.com/HathorNetwork/hathor-core/pull/1740/changes/5157fc7e4998944f8f856a066d9231d19d568dfa

hathor/builder/resources_builder.py
msbrogli: Thanks!

hathor/nanocontracts/resources/state.py
msbrogli: Thanks!

hathor/api/openapi/generator.py
msbrogli: Thanks!

hathor/api/openapi/generator.py
msbrogli: Thanks!

hathor/transaction/resources/create_tx.py
msbrogli: Thanks!

msbrogli (dismissed):

jansegre (dismissed):

msbrogli (approved):

jansegre (approved):

PR #1741, opened 2026-06-23
feat(decimals): add __str__ and parse to token amounts [part 14]

glevco:
Depends on #1740

### Motivation

Token amounts are stored as integers, but API requests and responses read and enter them as decimal strings (e.g. `1.5`), with the decimal point placed according to the amount's version (2 fractional digits for V1, 18 for V2). This PR adds that human-facing decimal rendering — exposed to Python as `__str__` on both `UnsignedAmount` and `SignedAmount` — and its inverse, `UnsignedAmount.parse`, which reads a decimal string into a V2 amount.

### Acceptance Criteria

- Add a `decimal` module to `htr-lib` housing the shared `format_decimal` / `parse_fixed_point` helpers and the `ParseDecimalError` type.
- Rename `set_normalization_factor` to `set_decimal_places`; the global `OnceLock` now stores the `(v1_decimal_places, v2_decimal_places)` pair instead of the derived factor.
- Add a `Display` impl to `UnsignedAmount` and `SignedAmount` in `htr-lib`, exposed to Python as `__str__`.
- Implement `UnsignedAmount.parse`, the inverse of the V2 `__str__`: it reads a decimal string carrying an explicit decimal point and builds a V2 amount.
- Update `htr_lib.pyi`.

### Checklist

### DIFF
--- a/htr-rs/crates/htr-lib-py/src/signed_amount.rs
+++ b/htr-rs/crates/htr-lib-py/src/signed_amount.rs
@@ -56,6 +56,10 @@ impl PySignedAmount {
         format!("{:?}", self.0)
     }
 
+    fn __str__(&self) -> String {
+        self.0.to_string()
+    }
+
     fn __bool__(&self) -> bool {
         self.0.as_bool()
     }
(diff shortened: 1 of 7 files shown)


htr-rs/crates/htr-lib-py/src/signed_amount.rs
```diff
@@ -143,6 +147,45 @@ mod tests {
...
+    #[test]
+    fn py_str_renders_fixed_point_decimal_with_sign() {
+        UnsignedAmount::set_decimal_places(2, 18);
+        Python::initialize();
+        Python::attach(|py| {
+            for (input, expected) in [
+                (0_i64, "0.000000000000000000"),
+                (5, "0.000000000000000005"),
+                (-5, "-0.000000000000000005"),
+                (1_500_000_000_000_000_000, "1.500000000000000000"),
```
msbrogli: Should it be `"1.5"` instead?

htr-rs/crates/htr-lib/src/decimal.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+impl std::error::Error for ParseDecimalError {}
+
+/// Parse a decimal string into the integer it denotes scaled by `10^decimal_places`, the inverse
+/// of [`format_fixed_point`]. Accepts a bare integer, or an integer and fractional part separated
+/// by a single `.` with digits on both sides; the fractional part may carry at most
+/// `decimal_places` digits.
+pub(crate) fn parse_fixed_point(
```
msbrogli, line 74: We should validate that there's only one `.`.

msbrogli (approved):

htr-rs/crates/htr-lib-py/src/signed_amount.rs
glevco: This is outdated.

htr-rs/crates/htr-lib/src/decimal.rs
glevco: This is done here:

```rs
    let Some((integer_part, fractional_part)) = input.split_once('.') else {
        return Err(ParseDecimalError::MissingDecimalPoint);
    };
    if fractional_part.contains('.') {
        return Err(ParseDecimalError::MultipleDecimalPoints);
    }
```

htr-rs/crates/htr-lib-py/src/signed_amount.rs
msbrogli: Thanks!

htr-rs/crates/htr-lib/src/decimal.rs
msbrogli: Thanks!

msbrogli (dismissed):

jansegre (dismissed):

jansegre (approved):

msbrogli (approved):

PR #1742, opened 2026-06-24
test(decimals): pre-wrap test fixtures behind a transitional shim [part 15]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1741

### Motivation

#1714 updates the codebase to the new versioned token amount model, flipping the `hathorlib.token_amount` `int` aliases to the real `htr_lib` types. That flip forces every test fixture to wrap its literal amounts, which dominates the diff. This PR lands the value-preserving fixture wrapping ahead of the flip, so the implementation PR is left with the production changes and only the behavioural test updates.

The wrapping stays green against the current `int`-based production code because the shim is itself an `int` subclass: every wrapped value is a plain integer at runtime. A single indirection module lets the implementation PR repoint the fixtures at `htr_lib` in one place, without touching each module again.

### Acceptance Criteria

- Add `hathor_tests/token_amount.py`, an `int`-subclass shim mirroring the `htr_lib` `UnsignedAmount`/`SignedAmount` API with V1 and V2 collapsed to an identity unit.
- Route the cleanly-mechanical test modules through the shim, wrapping their fixtures with `UnsignedAmount.from_v1(x)`, `.to_signed()`, and `UnsignedAmount.zero()`.
- Make no production changes and keep the `hathorlib.token_amount` aliases as `int`.

### Checklist

### DIFF
--- a/hathor_tests/nanocontracts/test_call_other_contract.py
+++ b/hathor_tests/nanocontracts/test_call_other_contract.py
@@ -28,6 +28,7 @@ from hathor.nanocontracts.types import (
 )
 from hathor.transaction.token_info import TokenVersion
 from hathor_tests.nanocontracts.blueprints.unittest import BlueprintTestCase
+from hathor_tests.token_amount import SignedAmount, UnsignedAmount
 
 COUNTER_NC_TYPE = make_nc_type(int)
 CONTRACT_NC_TYPE: NCType[ContractId | None] = make_nc_type(ContractId | None)  # type: ignore[arg-type]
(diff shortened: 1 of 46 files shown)


hathor_tests/nanocontracts/test_call_other_contract.py
```diff
@@ -323,33 +342,51 @@ def test_getting_funds_from_another_contract(self) -> None:
...
-            Balance(value=0, can_mint=False, can_melt=False), self.runner.get_current_balance(self.nc1_id, token3_uid)
+            Balance(
+               value=SignedAmount(), can_mint=False, can_melt=False
+           ), self.runner.get_current_balance(self.nc1_id, token3_uid)
         )
 
         self.assertEqual(
-            Balance(value=21, can_mint=False, can_melt=False), self.runner.get_current_balance(self.nc2_id, token1_uid)
+            Balance(
+               value=UnsignedAmount.from_v1(21).to_signed(), can_mint=False, can_melt=False
```
msbrogli, line 365: Why not `SignedAmount(21)`?

msbrogli (dismissed):

hathor_tests/nanocontracts/test_call_other_contract.py
glevco: Because `SignedAmount(21)` would be equivalent to `UnsignedAmount.from_v2(21).to_signed()`, and I want to keep the same test fixture values.

msbrogli: I don't get it but ok. We can keep it as it is now.

jansegre: I'd even prefer if we don't even allow `SignedAmount(21)`/`UnsignedAmount(21)`, it's ambiguous whether we mean `21.000000000000000000` (integer amount), `0.210000000000000000` (v1) or `0.000000000000000021` (v2). Ideally I think the integer amount makes most sense but a `UnsignedAmount.from_integer` should be enough for that. But I guess this isn't the PR for this change.

jansegre (dismissed):

msbrogli (approved):

hathor_tests/nanocontracts/test_call_other_contract.py
glevco: `UnsignedAmount(21)` doesn't exist because it's versioned, you have to use either `UnsignedAmount.from_v1()` or `UnsignedAmount.from_v2()`. `SignedAmount(21)` does exist because it's not versioned, but we could remove it in favor of only `SignedAmount.from_v2()`, for clarity. I would prefer to do it on another PR.

msbrogli: Great! Thanks!

jansegre (approved):

PR #1749, opened 2026-07-03
fix(nc-types): type `Amount` should not be `int` in runtime

jansegre:
### Motivation

Nano contract values typed as `Amount` were being serialized and deserialized through `VarUint32NCType`, whose deserialization returns a plain `int`. As a result, reading an `Amount` field, calling a method that returns `Amount`, or receiving an `Amount` argument would yield a plain `int` rather than an `Amount` new-type instance, silently dropping the semantic type on the static surface. This PR makes `Amount` round-trip as an actual `Amount`.

### Acceptance Criteria

- Add an `AmountNCType` (subclass of `VarUint32NCType`) that serializes identically but deserializes into an `Amount` new-type instance, and export it from the `nc_types` packages in both `hathor` and `hathorlib`.
- Map the `Amount` field type to `AmountNCType` (instead of `VarUint32NCType`) so contract fields, arguments, and return values typed as `Amount` deserialize back as `Amount`, while plain `int` fields remain unaffected.
- Preserve wire compatibility: the serialized bytes are unchanged, so only the deserialized Python type differs and existing stored data continues to decode.
- Cover the behavior with tests verifying that field/arg/return type maps resolve `Amount` to `AmountNCType`, that deserialization (including from plain-int input) yields an `Amount`, that byte round-trips hold across a range of values, and that a blueprint's `Amount` field getter returns an `Amount` instance.

### Checklist

### DIFF
--- a/hathor/nanocontracts/nc_types/__init__.py
+++ b/hathor/nanocontracts/nc_types/__init__.py
@@ -10,6 +10,7 @@ from hathorlib.nanocontracts.nc_types import (  # noqa: F401
     FIELD_TYPE_TO_NC_TYPE_MAP,
     RETURN_TYPE_TO_NC_TYPE_MAP,
     AddressNCType,
+    AmountNCType,
     BoolNCType,
     Bytes32NCType,
     BytesLikeNCType,
--- a/hathor_tests/nanocontracts/test_blueprint.py
+++ b/hathor_tests/nanocontracts/test_blueprint.py
@@ -6,7 +6,7 @@ from hathor.nanocontracts.context import Context
 from hathor.nanocontracts.exception import BlueprintSyntaxError, NCFail, NCInsufficientFunds, NCViewMethodError
 from hathor.nanocontracts.nc_types import make_nc_type_for_arg_type as make_nc_type
 from hathor.nanocontracts.storage.contract_storage import Balance, BalanceKey
-from hathor.nanocontracts.types import Address, NCDepositAction, NCWithdrawalAction, TokenUid, public, view
+from hathor.nanocontracts.types import Address, Amount, NCDepositAction, NCWithdrawalAction, TokenUid, public, view
 from hathor_tests.nanocontracts.blueprints.unittest import BlueprintTestCase
 
 STR_NC_TYPE = make_nc_type(str)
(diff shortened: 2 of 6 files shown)


msbrogli (approved):

hathor/nanocontracts/nc_types/__init__.py
glevco, line 1: Don't we have to do the same for other wrapper types, too?

```python
    Address: AddressNCType,
    Amount: AmountNCType,
    BlueprintId: Bytes32NCType,  # this
    ContractId: Bytes32NCType,   # this
    Timestamp: Uint32NCType,  # this
    TokenUid: TokenUidNCType,
    TxOutputScript: BytesLikeNCType[TxOutputScript],
    VertexId: Bytes32NCType,  # this
    SignedData: SignedDataNCType,
    (Address, ContractId): CallerIdNCType,
```

hathor_tests/nanocontracts/test_blueprint.py
```diff
@@ -87,12 +87,28 @@ def my_private_method_nop(self) -> int:
...
+    @public
+    def initialize(self, ctx: Context, y: Amount) -> None:
+        self.y = y
+
+    @view
+    def get_y(self) -> Amount:
+        # Reading `self.y` must yield an `Amount` instance, not a plain `int`. This assert would
+        # fire if the field were mapped to a plain integer NCType (e.g. `Amount: VarUint32NCType`).
+        assert isinstance(self.y, Amount)
+        return self.y
```
glevco, line 102: We have to check that this change in behavior will not break any contracts in mainnet/testnet, although it's unlikely.

glevco (changes requested):

hathor_tests/nanocontracts/test_blueprint.py
jansegre: I did check on every mainnet blueprint and there is no code that is affected. I did not check on testnet.

hathor/nanocontracts/nc_types/__init__.py
jansegre: `Bytes32NCType` implements `_actual_type` to store and instantiate the correct type so they don't have this problem, so out of those only `Timestamp` would have this problem.

jansegre: Implemented it in #1759 which should replace this PR.

jansegre:
Replaced by #1759

PR #1759, opened 2026-07-13
fix(nc-types): preserve actual types when deserializing NC values

jansegre:
### Motivation

Nano contract values typed as `Amount` or `Timestamp` were being deserialized through `VarUint32NCType` and returning a value of plain `int`. As a result, reading an that field, or calling a method that returns `Amount`/`Timestamp`, or receiving an `Amount`/`Timestamp` argument would yield a plain `int` rather than the expected new-type instance.

### Acceptance Criteria

- Values deserialized through `NCType.deserialize` and `NCType.json_to_value` are wrapped back into the declared actual class (e.g. `Amount`, `Timestamp`) when it is a newtype-style subclass of the produced value's type, instead of degrading to the base type.
- `NCType.from_type` records the declared (post-alias) type in a new `_actual_type` slot on the `NCType` base class; `BytesLikeNCType` and `NamedTupleNCType` reuse the inherited slot instead of declaring their own, and directly constructed `NCType` instances remain unaffected (no wrapping).
- Plain base types (e.g. a field declared as `int`) are not wrapped, and generic aliases (`tuple[X, ...]`, unions, containers) keep working, with their inner types wrapped through recursion.
- Add `hathorlib` tests covering bytes and JSON roundtrips for `Amount`/`Timestamp`, container composition, mapping-independent wrapping, and no-regression for bytes-like types (`Address`, `TokenUid`, `ContractId`, etc.).
- Add a blueprint test verifying that field getters return instances of the declared classes at runtime.
- All existing mainnet OCBs have been verified against this change and there is no impact (none of the Blueprints are affected by the change in type for `Amount` or `Timestamp`)

### Checklist

### DIFF
--- a/hathorlib/hathorlib/nanocontracts/nc_types/nc_type.py
+++ b/hathorlib/hathorlib/nanocontracts/nc_types/nc_type.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Generic, NamedTuple, TypeAlias, TypeVar, final
+from typing import Any, Generic, NamedTuple, TypeAlias, TypeVar, final
 
 from typing_extensions import Self
 
@@ -15,6 +15,7 @@ from hathorlib.nanocontracts.nc_types.utils import (
     get_usable_origin_type,
 )
 from hathorlib.serialization import Deserializer, Serializer
+from hathorlib.utils.typing import is_subclass
 
 T = TypeVar('T')
 
@@ -39,10 +40,17 @@ class NCType(ABC, Generic[T]):
         nc_types_map: TypeToNCTypeMap
 
     # XXX: subclasses must override this if they need any properties
-    __slots__ = ()
+    __slots__ = ('_actual_type',)
 
     # XXX: subclasses must initialize this property
     _is_hashable: bool
+
+    # XXX: set by `NCType.from_type` to the declared (post-alias) type, and used to wrap deserialized values back
+    #      into newtype-like classes (e.g. `Amount`, `Timestamp`). It is unset when an NCType is constructed
+    #      directly, so it must be read with `getattr(self, '_actual_type', None)`. Some subclass constructors
+    #      also assign it, possibly to a non-class (e.g. `NodeNCType` uses the `NodeId` NewType), which
+    #      `_to_actual_type` must skip.
+    _actual_type: type[Any]
 
     @final
     @staticmethod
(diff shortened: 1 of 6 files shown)


msbrogli (approved):

hathorlib/hathorlib/nanocontracts/nc_types/nc_type.py
```diff
@@ -85,6 +98,24 @@ def _from_type(cls, type_: type[T], /, *, type_map: TypeMap) -> Self:
...
+
+        This is what makes a field/arg/return declared as a newtype-like class (e.g. `Amount`, `Timestamp`)
+        yield instances of that class instead of the underlying base type (e.g. `int`). Only newtype-style
+        upcasts are made: the declared type must be a plain class and a subclass of the value's type.
+        """
+        actual: type[Any] | None = getattr(self, '_actual_type', None)
+        if actual is None or not isinstance(actual, type):
+            return value
+        if isinstance(value, actual) or not is_subclass(actual, type(value)):
+            return value
```
glevco, line 113: In which cases would those be reached?

glevco (approved):

hathorlib/hathorlib/nanocontracts/nc_types/nc_type.py
jansegre: Let me take a look, those `or not ...` do look weird.

jansegre: - First `if`:
  - **`actual is None`** — NCTypes constructed directly instead of via `NCType.from_type`, so there's no declared type to wrap into (e.g. the `DictNCType(...)` built inside `NodeNCType`, or `VarUint32NCType()` in tests).
  - **`not isinstance(actual, type)`** — some constructors assign `_actual_type` a non-class: `NodeNCType` uses `BytesLikeNCType(NodeId)`, and `NodeId` is a `typing.NewType`, for which `isinstance(value, actual)` would raise `TypeError`.
- Second `if`:
  - **`isinstance(value, actual)`** — the common path: most `_deserialize` impls already construct the actual class (`AddressNCType`, `Bytes32NCType`, namedtuples, builtins...), so this skips a useless re-construction — and for `NamedTupleNCType`, `actual(value)` would actually crash.
  - **`not is_subclass(actual, type(value))`** — the reverse direction of the `isinstance` above, so not redundant: it requires `actual` to derive from `type(value)`, i.e. the wrap is a base→newtype conversion (`int` → `Amount`). If they're unrelated (possible with a custom `TypeMap`), we leave the value for `_check_value` to reject properly instead of calling an incompatible constructor or silently coercing (a stray `bool` would otherwise become `Amount(True) == Amount(1)`).

Only when all four fall through is `actual(value)` called.

PR #1761, opened 2026-07-14
refactor(fees): introduce the fee policy configuration model [part 2/3]

glevco:
Depends on https://github.com/HathorNetwork/hathor-core/pull/1763

### Motivation

Fee amounts are configured as a single scalar setting, `FEE_PER_OUTPUT_V1`, which can only express a fee charged in HTR, and only with V1 decimal precision. The upcoming stable-fee work needs fees to be charged in tokens other than HTR (stablecoins) and with sub-cent precision, so that scalar cannot carry it.

This PR replaces it with the fee policy configuration model, and threads it through every fee computation, without introducing any new fee behavior. Only fee policy `V1` is defined, and its HTR policy charges the same `0.01` HTR per unit as before, so the change is behavior-preserving.

### Acceptance Criteria

- Add `FeePolicyVersion`, `FeePolicy`, and `FeePolicyPerToken`.
- Replace `HathorSettings.FEE_PER_OUTPUT_V1` and `FEE_TOKEN_AMOUNT_PER_OUTPUT` with the `FEE_POLICIES` setting.
- Replace `FeeHeader.total_fee_amount()` and `NCFee.__get_htr_value__()` with `get_fee_charge()`, which validates and aggregates fee payments in a single pass into a `FeeCharge`.
- Make fee payment validation policy-aware in the mint/melt syscalls.
- Reject duplicate fee entries for the same token in the nano runner.

### Checklist

### DIFF
--- a/hathorlib/hathorlib/nanocontracts/nano_settings.py
+++ b/hathorlib/hathorlib/nanocontracts/nano_settings.py
@@ -1,7 +1,39 @@
 # SPDX-FileCopyrightText: Hathor Labs
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
+
 from dataclasses import dataclass
+from typing import TYPE_CHECKING, Self
+
+from htr_lib import UnsignedAmount
+from typing_extensions import assert_never
+
+from hathorlib.conf.fee_policy import FeePolicy as SettingsFeePolicy
+from hathorlib.conf.settings import HathorSettings
+from hathorlib.token_amount_version import TokenAmountVersion
+
+if TYPE_CHECKING:
+    from hathorlib.nanocontracts import NanoRuntimeVersion
+    from hathorlib.nanocontracts.types import TokenUid
+
+
+@dataclass(slots=True, frozen=True, kw_only=True)
+class FeePolicy:
+    fee_based_tokens: int
+    amount_shielded: int
+    full_shielded: int
+
+    @classmethod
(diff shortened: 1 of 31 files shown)


hathorlib/hathorlib/nanocontracts/nano_settings.py
```diff
@@ -10,4 +39,30 @@ class NanoSettings:
     This dataclass contains information about the settings used by the current Nano runtime.
     It is returned by the `get_settings` syscall. Note that settings are not constant, they may change over time.
     """
-    fee_per_output: int
+    fee_policies: dict[TokenUid, FeePolicy]
```
glevco, line 45: TODO: to change this, we have to check that no OCBs are using it.

hathorlib/hathorlib/nanocontracts/runner/runner.py
```diff
@@ -450,7 +449,11 @@ def _unsafe_call_another_contract_public_method(
                 raise NCInvalidContext('action amount must be positive')
 
         # Validate fees
+        seen_fee_tokens: set[TokenUid] = set()
         for fee in fees:
+            if fee.token_uid in seen_fee_tokens:
+                raise NCInvalidFee(f'duplicate fees for token {fee.token_uid.hex()}')
+            seen_fee_tokens.add(fee.token_uid)
```
glevco, line 462: TODO: this is a new rule. We must check whether mainnet breaks it.

PR #1769, opened 2026-07-17
feat(decimals): update nano serialization [part 20]

glevco:
Depends on #1746

### Motivation

Nano-related (de)serialization must support token amount V2. Serialization happens in three places: method call args, method call returns, and contract fields persisted in the global state trie. Args and returns must follow the calling transaction's token amount version — a V1 tx only calls V1 blueprints and a V2 tx only V2 — while field storage must be version-independent, since the state trie must be identical across all nodes, so fields always use the V2 encodings. This PR threads the version through those paths, implements the V2 length-prefix encodings for `int` and `Amount`, and fixes the version gating that prevented V2 transactions from validating.

### Acceptance Criteria

- Thread `TokenAmountVersion` through nano (de)serialization: `Method`/`ArgsOnly`/`ReturnOnly` parsing follows the `Runner`'s version.
- Serialize fields with the global V2 encodings regardless of any contract's version; `FORCE_LEGACY_FIELDS` toggles the whole storage layer to V1, for tests only.
- Add signed support to `encode_length_prefix_varint`/`decode_length_prefix_varint`: minimal-length signed payloads, with canonicality checks rejecting redundant leading `0x00`/`0xff` bytes.
- Add `VarInt32V2NCType` (signed, for `int`) and `VarUint32V2NCType` (unsigned, for `Amount`) using LPE, mapped in the V2 branch of `update_type_map`; fix the signed LEB128 bound so the eager value check matches the codec's encodable range.
- Model the serialization version of `SignedData` and `NCRawArgs` as class identity: version-abstract bases with concrete `SignedDataV1`/`V2` and `NCRawArgsV1`/`V2` subclasses, routed to the plain names blueprint code sees by each blueprint's own version; the version tag is a dunder so the sandbox cannot reassign it, and `SignedData` equality is version-aware.
- Accept the token amount version bit in `verify_signal_bits` via a shared `TOKEN_AMOUNT_VERSION_MASK`, so V2 transactions are gated by `verify_token_amount_version` instead of rejected as unknown signal bits.
- Default the simulator's blueprint registrations to `TokenAmountVersion.V2`, matching its runner.
- Memoize the per-version arg/return/field type maps instead of rebuilding them per call; encode the DAG builder's token amount version onto on-chain blueprint vertices.
- Update tests: V1/V2 nano contract coverage, LPE byte-level storage expectations, varint bounds, versioned factory coverage in hathorlib, and signal-bits coverage; make hathorlib pass mypy from both environments.

### Checklist


hathorlib/pyproject.toml
```diff
@@ -90,6 +90,14 @@ module = [
...
 disallow_untyped_defs = false
 
+# `hathor` is only imported under TYPE_CHECKING and is not a hathorlib dependency: it is absent in
+# hathorlib's own environment and installed without a py.typed marker in hathor-core's environment.
+[[tool.mypy.overrides]]
+module = [
+    "hathor.*",
+]
+ignore_missing_imports = true
+
```
glevco: Why this?

hathorlib/hathorlib/nanocontracts/nc_types/__init__.py
```diff
@@ -137,34 +147,40 @@
...
     If you need to customize the mapping use `NCType.from_type` instead.
     """
-    return NCType.from_type(type_, type_map=_ARG_TYPE_MAP)
+    from hathorlib.nanocontracts.fields import update_type_map
+    type_map = NCType.TypeMap(ESSENTIAL_TYPE_ALIAS_MAP.copy(), ARG_TYPE_TO_NC_TYPE_MAP.copy())
+    update_type_map(type_map.nc_types_map, token_amount_version)
+    return NCType.from_type(type_, type_map=type_map)
 
 
 _RETURN_TYPE_MAP = NCType.TypeMap(ESSENTIAL_TYPE_ALIAS_MAP, RETURN_TYPE_TO_NC_TYPE_MAP)
```
glevco: Can we remove this line?

hathor/nanocontracts/execution/block_executor.py
```diff
@@ -325,7 +325,7 @@ def execute_transaction(
             context = nano_header.get_context()
             assert context.block.hash == vertex_metadata.first_block
 
-            nc_args = NCRawArgs(nano_header.nc_args_bytes)
+            nc_args = get_nc_raw_args_class(runner.token_amount_version)(nano_header.nc_args_bytes)
```
glevco: Bind to var instead of inline

hathor/__init__.py
```diff
@@ -20,9 +23,9 @@
     NCFee,
     NCGrantAuthorityAction,
     NCParsedArgs,
-    NCRawArgs,
+    NCRawArgsV2 as NCRawArgs,
     NCWithdrawalAction,
-    SignedData,
+    SignedDataV2 as SignedData,
```
glevco, line 28: Remove these, they're error-prone. Be explicit in the blueprints that need it.

PR #1772, opened 2026-07-22
fix(resources): return success false on short hex_tx and unknown token in /decode_tx

### LINKED ISSUE #1770: [bug] Harden /decode_tx: catch decode-time exceptions (short hex_tx → IndexError, unknown token → KeyError) and return success:false
## Summary

`GET /decode_tx` is designed to answer malformed input with a graceful `{'success': false, ...}` JSON body — it already catches `ValueError` (bad hex) and `struct.error` (malformed body). But **two other decode-time exceptions escape uncaught**, each turning arbitrary internet input on a public endpoint into an **HTTP 500 + a `critical` log line** (and thus alert noise):

1. **`IndexError` on a too-short `hex_tx`** — a valid-hex `hex_tx` shorter than 2 bytes (e.g. empty, or `00`) passes `bytes.fromhex()`, then `vertex_parser.deserialize()` reads `version = data[1]` *before* checking length. Observed on **mainnet, 2026-07-16**.
2. **`KeyError('unknown token')`** — a decodable tx that references a `token_uid` not in the node's tokens index reaches `get_tx_extra_data()` → `tokens_index.get_token_info()` → `raise KeyError('unknown token')`. Long-standing; already reported in **#378**.

Both are the **same class** the `/thin_wallet` endpoints were just hardened against in **#1758 / #1760** (user input reaching an uncaught exception on a public API resource). #1758 was explicitly scoped to `/thin_wallet`; `/decode_tx` is a `transaction` resource and was **not** covered, so the gap persists here. This issue tracks closing it for `/decode_tx` and absorbs #378.

## Traceback 1 — `IndexError` (mainnet, 2026-07-16)

```
File ".../hathor/transaction/resources/decode_tx.py", line 55, in render_GET
    tx = self.manager.vertex_parser.deserialize(tx_bytes)
File ".../hathor/transaction/vertex_parser/_vertex_parser.py", line 59, in deserialize
    version = data[1]
              ~~~~^^^
IndexError: index out of range
```

## Traceback 2 — `KeyError` (from #378, still reproducible in current code)

```
File ".../hathor/transaction/resources/decode_tx.py", line 58, in render_GET
    data = get_tx_extra_data(tx)
File ".../hathor/transaction/resources/transaction.py", line 131, in get_tx_extra_data
    token_info = tokens_index.get_token_info(bytes.fromhex(token_uid_hex))
File ".../hathor/indexes/rocksdb_tokens_index.py", line 491, in get_token_info
    raise KeyError('unknown token')
KeyError: 'unknown token'
```

## Root cause

`hathor/transaction/resources/decode_tx.py` `render_GET`:

```python
ARGS = ['hex_tx']
...
try:
    tx_bytes = bytes.fromhex(parsed['args']['hex_tx'])
    tx = self.manager.vertex_parser.deserialize(tx_bytes)   # <-- IndexError (short input)
    ...
    data = get_tx_extra_data(tx)                            # <-- KeyError('unknown token')
except ValueError:                     # bytes.fromhex on bad hex
    data = {'success': False, 'message': 'Invalid hexadecimal data'}
except struct.error:                   # struct.unpack on malformed body
    data = {'success': False, 'message': 'Could not decode transaction'}
```

- `bytes.fromhex('')`/`bytes.fromhex('00')` succeed, then `deserialize` reads the 2nd byte (`version = data[1]`) with no length guard → `IndexError`.
- `get_tx_extra_data` loops over the tx's referenced tokens and calls `get_token_info`, which raises `KeyError('unknown token')` for a token the node doesn't index — no guard at the call site (`transaction.py:128-131`).
- Neither `IndexError` nor `KeyError` is `ValueError`/`struct.error`, so both escape `render_GET` → Twisted 500 + `critical` traceback → alert.

## Expected behavior

The endpoint's contract is "**never 500 on user input**" — it already returns `{'success': false, 'message': ...}` for undecodable input. A too-short payload and an unknown-token reference are both just "input I couldn't fully decode," so both should return the same graceful `success: false` response — no traceback, no `critical` log line, no alert.

## Suggested fix

- **Minimal:** broaden the `except` in `decode_tx.render_GET` to also catch `IndexError` and `KeyError` (or catch `Exception` at this API boundary) and map to the existing `{'success': False, 'message': 'Could not decode transaction'}`.
- **Better (robust):**
  - `vertex_parser.deserialize` should length-check and raise a typed decode error (e.g. `struct.error` / a domain error) on truncated input instead of a bare `IndexError`.
  - unknown-token lookups in `get_tx_extra_data` should be a handled, typed condition rather than a bare `KeyError` bubbling up.
- **Systemic:** extend the #1758 hardening audit **beyond `/thin_wallet`** to the public `transaction` resources (`/decode_tx`, `/push_tx`, …) so no public endpoint can 500 on malformed input. Add regression tests mirroring #1760's (a sub-2-byte `hex_tx`; a tx referencing an unknown token).

## References

- Absorbs **#378** ("Decode TX API: KeyError: 'unknown token'") — same endpoint, same class, still reproducible.
- Sibling hardening (the `/thin_wallet` slice of this same class): **#1758** (issue), **#1760** (PR).

## Impact

Low per request — a single HTTP 500, the node keeps running. But these are assertions/exceptions reachable from **arbitrary internet input on a public endpoint**, producing recurring `critical` log lines / alerts, and a latent robustness gap on a public API. Found during on-call triage of a mainnet occurrence.

SAY-5:
### Motivation

Fixes #1770. A valid-hex `hex_tx` shorter than 2 bytes escapes the existing handlers with an `IndexError` (deserialize reads the version byte before checking length), and a decodable tx referencing a token uid missing from the tokens index escapes with `KeyError('unknown token')`. Both turn arbitrary input on this public endpoint into an HTTP 500 plus a critical log line instead of the documented `success: false` response.

### Acceptance Criteria

- `/decode_tx` returns `{'success': false, ...}` for a too-short `hex_tx` instead of an HTTP 500
- `/decode_tx` returns `{'success': false, ...}` when the tx references an unknown token
- Regression test for the short `hex_tx` case in `test_decodetx.py`

### Checklist

### DIFF
--- a/hathor/transaction/resources/decode_tx.py
+++ b/hathor/transaction/resources/decode_tx.py
@@ -47,8 +47,10 @@ class DecodeTxResource(Resource):
             data = get_tx_extra_data(tx)
         except ValueError:
             data = {'success': False, 'message': 'Invalid hexadecimal data'}
-        except struct.error:
+        except (IndexError, struct.error):
             data = {'success': False, 'message': 'Could not decode transaction'}
+        except KeyError:
+            data = {'success': False, 'message': 'Unknown token'}
 
         return json_dumpb(data)
 
--- a/hathor_tests/resources/transaction/test_decodetx.py
+++ b/hathor_tests/resources/transaction/test_decodetx.py
@@ -46,6 +46,12 @@ class DecodeTxTest(_BaseResourceTest._ResourceTest):
 
         self.assertFalse(data_error2['success'])
 
+        # Valid hex but too short to hold a version byte
+        response_error3 = yield self.web.get("decode_tx", {b'hex_tx': b'00'})
+        data_error3 = response_error3.json_value()
+
+        self.assertFalse(data_error3['success'])
+
         # Token creation tx
         script_type_out = parse_address_script(genesis[0].outputs[0].script)
         address = script_type_out.address


jansegre (approved):
Good change.

msbrogli (approved):

